PR #37, opened 2016-10-30
add instructions to clients for running unittest

sviswana:
fix minor bug for tagging
rename unittest .py files to conventional unittest format


mpvartak (approved):
Looks good.

PR #54, opened 2016-11-01
Frontend

weienlee:
adding frontend directory


mpvartak (changes requested):
Looks good overall! 

Few high level comments:
- Please remove any mongo stuff from the code
- Related to above, there are classes specific to mongo that you can remove now
- Pleas put all files generated by thrift into .gitignore
- Fix indentation to 2 or 4 spaces (just be consistent)

mpvartak (changes requested):
The only thing remaining seems to be indentation -- could you please fix?

mpvartak (approved):

PR #57, opened 2016-11-01
e2e test cases for Linear Model - and fix to syncing on other sample …

sviswana:
…files


mpvartak (changes requested):
Can you also create a separate folder for E2E?

PR #63, opened 2016-11-02
add read_csv_sync so pandas dataframe automatically tracked

sviswana:


mpvartak (approved):
One small comment. Looks good otherwise.

PR #195, opened 2017-02-17
Minor pom changes to fix server build

flikka:
See https://github.com/mitdbg/modeldb/issues/194


mpvartak:
Thanks for the fixes! Look good to me, merging.

PR #197, opened 2017-02-17
Added RedHat provisioning script - tested on RHEL 7.1

ghost:


mpvartak:
Looks good. Thanks for adding instructions for RHEL!

Issue #191, opened 2017-02-12
[Installation] Make ModelDB compatible with thrift 0.10.0


mpvartak:
Currently this is blocked on a version issue related to the thrift-maven-plugin. Will be resolved when that is unblocked.

mpvartak:
Fixed by #192

mpvartak:
Breaks build for 0.9.3. Need two build solutions for the two thrift versions.

PR #198, opened 2017-02-18
fix for supporting thrift 0.10.0 along with thrift 0.9.3

mpvartak:
See #191


mpvartak:
Tested with 0.10.0 and 0.9.3

PR #216, opened 2017-03-06
Frontend

weienlee:
- fix url issues
 - rename files 
 - deprecate old models page
 - support filtering on json metadata


frontend/public/js/util.js
```diff
@@ -10,4 +10,31 @@
...
+
+
+JSON.flatten = function(data) {
+    var result = {};
+    function recurse (cur, prop) {
+        if (Object(cur) !== cur) {
+            result[prop] = cur;
+        } else if (Array.isArray(cur)) {
+          /*
+             for(var i=0, l=cur.length; i<l; i++)
```
mpvartak, line 23: What's going on here?

weienlee: i found code online to flatten JSON objects, which is how I decided to deal with nested objects in the metadata

mpvartak: do you plan to use the commented out code? if not, let's remove it!

re: flattening: currently filtering works for top-level keys right. are you planning to add in nested filtering? i think it would make sense because a lot of the metrics are nested

mpvartak:
Can we add a way to dismiss the modal dialog on pressing ESC?

mpvartak:
looks good otherwise

frontend/public/js/util.js
weienlee: hmm not sure.
so right now the flattening works with nested objects

for example,
{
    config: {
        l1: 20
    }
}

will get flattened to 
{
    config.l1: 20
}

The commented code was for lists inside an object, an it would just use the index when flattening. Wasn't sure how to support flattening lists properly, since we probably don't want the order to matter. So I commented it out for now, and will revisit how to deal lists in the json later.

mpvartak:
LGTM.

PR #226, opened 2017-03-20
[MRG] Unify configs

mpvartak:


client/syncer.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "experiments": [],
+  "experimentRun": {
+    "description": "Default Run Description"
+  },
+  "shouldCountRows": false,
+  "shouldStoreGSCVE": false,
+  "shouldStoreSpecificModels": false,
+  "git": {
+    "versionCode": "yes",
```
mpvartak: make this no for now

client/shell/mdb_code_versioning.py
```diff
@@ -17,10 +17,10 @@ def version(version_config):
     session = requests.Session()
     session.headers['Authorization'] = 'token %s' \
         % version_config[constants.ACCESS_TOKEN_KEY]
-    
+
     # check if the repo exists, if not create it
     repo_exists_url = 'https://api.github.com/repos/%s/%s' % \
-        (version_config[constants.GIT_USERNAME_KEY], 
+        (version_config[constants.GIT_USERNAME_KEY],
```
mpvartak, line 23: These lines will fail now

client/python/modeldb/utils/ConfigUtils.py
```diff
@@ -1,68 +1,71 @@
...
+        self.config =  json.load(file(filename, 'r'))
         self.validate_config()
 
     def validate_config(self):
         # TODO: what all do we expect to be here?
-        if constants.PROJECT_KEY not in self.config:
+        if 'project' not in self.config:
             print "Project must be defined in config file. Cannot proceed."
             sys.exit(-1)
-        if constants.GIT_USERNAME_KEY not in self.config:
```
mpvartak: I don't quite get the reason for using the strings in this file but using constants.ABC in other files. I see that this is what you were asking about before -- the reason I had pulled out the constants was in case we wanted to change one of them to another string. Changing them in one place is a lot easier than changing them in multiple files. I'd actually recommend sticking with one or another. 

Not sure if I'm missing something here to please let me know if I am!

client/shell/mdb_code_versioning.py
RogerTangos: This is just a whitespace change from my linter. How does it fail?

client/python/modeldb/utils/ConfigUtils.py
RogerTangos: We can keep constants, but will need to define new ones so that they
- match `syncer.json`'s keys
- allow for nesting, which syncer.json currently uses.

Basically, if we use constants, we need to introduce a separation of _input_ and _output_ constants. This will allow the python code to continue to ingest the same `syncer.json` file that Scala uses.

Would introducing a bunch of new input constants be ok with you?

client/shell/mdb_code_versioning.py
mpvartak: If the constants are removed -- I'm unclear on what we are doing there. Is constants going to be around?

RogerTangos:
Responding to other email questions inline: Reading from the new config does work.

I'm also updating the scala code and documentation to point to the new syncer file.

client/shell/mdb_code_versioning.py
RogerTangos: `ConfigUtils.ConfigReader`'s methods return exactly what they did before, so version_config is identical to the previous version, and this continues to function. 

(ConfigUtils.ConfigReader ConfigConstants to define _outputs_. I didn't want to touch that, since I wasn't sure what parts of the program, like this one, used them, or how.)

client/python/modeldb/utils/ConfigUtils.py
```diff
@@ -1,68 +1,71 @@
...
+
+        if 'experiments' not in self.config or len(self.config['experiments']) < 1:
             print "Experiment '%s' not defined in config file. Using default." \
                 % expt_name
             return None
-        
+
         experiment = None
-        experiments = self.config[constants.EXPT_KEY]
+        experiments = self.config.get('experiments', [])
```
mpvartak: let's remove the experiments section of the config. It's confusing to most users and not used a lot at the moment,

client/python/modeldb/utils/ConfigUtils.py
mpvartak: Ok got it. How about the following?
- let's remove constants.ABC; we can use the actual strings when reading in the config file
- when creating objects from the config (e.g. for project etc), let's define structs s.t. we can reference their fields directly. Then we don't need to use the keys in some places and not use them in others.

So we can have a similar interface for all accesses to the config.

RogerTangos: I think I've got it. Objects from the config sounds good. Structs s.t. = structure tags?

client/python/modeldb/utils/ConfigUtils.py
RogerTangos: Updated so that this just returns whatever is in the experiment key

RogerTangos:
This should be ready to merge.

I'd like to have some other config methods return structs (that aren't yet defined), but first want to clean up the python imports to help lessen confusion when refactoring.

mpvartak:
Looks good. Could you add a comment regarding testing? OK to merge after
that.

RogerTangos:
Tests are passing (or in the case of testPipelineEvent.py, sometimes failing) exactly as they did before.

RogerTangos:
My own notation here:
```
OK: /modeldb/client/python/modeldb/tests/basic/ModelDbSyncerBaseTest.py 
OK: /modeldb/client/python/modeldb/tests/sklearn/testFitEvent.py 
OK: /modeldb/client/python/modeldb/tests/sklearn/testGridSearchEvent.py 
OK: /modeldb/client/python/modeldb/tests/sklearn/testMetricEvent.py
2Failures: /modeldb/client/python/modeldb/tests/sklearn/testPipelineEvent.py 
======================================================================
FAIL: test_overall_pipeline_fit_event (__main__.TestPipelineEvent)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "testPipelineEvent.py", line 90, in test_overall_pipeline_fit_event
    utils.is_equal_transformer_spec(spec, expected_spec, self)
  File "/Users/arcarter/code/modeldb/client/python/modeldb/tests/utils.py", line 134, in is_equal_transformer_spec
    tester.assertEqual(len(spec1.hyperparameters), len(spec2.hyperparameters))
AssertionError: 14 != 10

======================================================================
FAIL: test_pipeline_first_fit_stage (__main__.TestPipelineEvent)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "testPipelineEvent.py", line 131, in test_pipeline_first_fit_stage
    utils.is_equal_transformer_spec(spec, expected_spec, self)
  File "/Users/arcarter/code/modeldb/client/python/modeldb/tests/utils.py", line 134, in is_equal_transformer_spec
    tester.assertEqual(len(spec1.hyperparameters), len(spec2.hyperparameters))
AssertionError: 7 != 3

OK: /modeldb/client/python/modeldb/tests/sklearn/testProjectCreation.py 
OK: /modeldb/client/python/modeldb/tests/sklearn/testRandomSplitEvent.py 
OK: /modeldb/client/python/modeldb/tests/sklearn/testTransformEvent.py
```

mpvartak:
LGTM.

PR #238, opened 2017-04-05
[MRG] Reduce Docker image size and split into one image per service

### LINKED ISSUE #213: [Improvement] Reduce size of docker image
Current image is several GBs. Can we reduce?

justinanderson:
When finished, this will resolve #213.

I still need to:

- [x] Stop hardcoding hostnames and ports in Docker images
    - [x] Java backend server
    - [x] Node frontend
- [x] Create a docker-compose.yml for the server
- [x] Make docker-compose.yml able to build the images from scratch
- [x] Update readmes
- [x] Upload prebuilt images to Docker Hub.

The backend is 1.35 GB and the frontend is 1.16 GB on my computer. Official mongo is 360 MB.

Simplest usage, if you have [Docker Compose](https://docs.docker.com/compose/install/), is:

```bash
cd /repo/root/folder
docker-compose up
# Logs should be visible and the site should be reachable at http://localhost.
```

If you don't have Docker Compose, you can do the same manually with:

```bash
# Build images
docker build -t backend -f Dockerfile-backend .
docker build -t frontend -f Dockerfile-frontend .

# Create a network for them to see each other over
docker network create modeldb

# Mongo server
docker run -d \
    --name mongo \
    --net modeldb \
    -p 27017:27017 \
    mongo:latest

# ModelDB backend server
docker run -d \
    --name backend \
    --net modeldb \
    -p 6543:6543 \
    backend

# ModelDB frontend server (reachable at http://localhost)
docker run -d \
    --name frontend \
    --net modeldb \
    -p 80:3000 \
    frontend
```


justinanderson:
@arubisov Is this along the lines of what you want at Scotiabank?

justinanderson:
Going to skip making Docker images for the client libraries in this PR.

justinanderson:
Ready for a look. I'll build and push images to Docker Hub once this has been merged into master.

README.md
```diff
@@ -44,11 +44,29 @@ Alternatively, ModelDB native clients for  ```spark.ml``` and ```scikit-learn```
...
+### Automated Setup
+
+If you have [Docker Compose](https://docs.docker.com/compose/install/) installed, you can bring up a ModelDB server with just a couple commands.
+
+2. **Clone the repo**
+
+    ```bash
+    git clone https://github.com/mitdbg/modeldb
+    ```
+
```
mpvartak, line 59: Please add the folder from where to run this cmd

.gitignore
```diff
@@ -1,8 +1,4 @@
-# DS_Store
 .DS_Store
-*/.DS_Store
```
mpvartak, line 3: I remember that these lines were there b/c there were D_Store files at various places

README.md
```diff
@@ -44,11 +44,29 @@ Alternatively, ModelDB native clients for  ```spark.ml``` and ```scikit-learn```
 
 ## Setup and Installation
 
+### Automated Setup
```
mpvartak: Call this Docker setup?

dockerbuild/README.md
```diff
@@ -1,36 +1,69 @@
...
+1. **Install Docker Compose**
 
-You should replace *ID* with the ID that is printed after building.
+    [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
 
-Once you have started the built container, please run `interactive_installer.sh` to install the dependencies that require
-accepting license agreements. It will then also build ModelDB. You should commit this as a new tag so that you do not need to wait for
-it to install again. Future runs can then `start_modeldb.sh` to build modelDB (You can also build modelDB manually, just use `add_dependencies_to_path.sh`
-to add the dependencies to your PATH first).
+2. **Run ModelDB server**
```
mpvartak: This will start frontend too, so maybe "Run ModelDB"

dockerbuild/all-in-one/README.md
```diff
@@ -0,0 +1,38 @@
+This Dockerfile and its sibling scripts are from an earlier configuration of ModelDB. They are preserved here but not required by other Dockerfiles.
```
mpvartak: Maybe rename this folder to archive? b/c the other docker compose stuff is also all-in-one in some sense.

mpvartak (commented):
Some comments mainly about the README/docs.

I will test this out later today and update here.

.gitignore
justinanderson: I've seen it happen in other projects where there are redundant gitignore rules, especially for annoying files like .DS_Store. One `.DS_Store` matches all instances of .DS_Store in all subfolders, per the [gitignore docs](https://git-scm.com/docs/gitignore):

> A leading `**` followed by a slash means match in all directories. For example, `**/foo` matches file or directory `foo` anywhere, the same as pattern `foo`.

justinanderson:
@mpvartak All your comments should be addressed by these new commits.

.gitignore
mpvartak: Cool, SG

dockerbuild/Dockerfile-backend
```diff
@@ -0,0 +1,67 @@
...
+    apt-get clean && \
+    update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
+
+WORKDIR /root
+
+# Install Thrift 0.10.0
+RUN wget -q http://archive.apache.org/dist/thrift/0.10.0/thrift-0.10.0.tar.gz && \
+    tar -xzf thrift-0.10.0.tar.gz && \
+    cd thrift-0.10.0 && \
+    ./configure --without-python && \
```
mpvartak, line 31: we only need java for the backend so maybe we can also specify other languages if that helps with faster config too.

dockerbuild/README.md
```diff
@@ -1,36 +1,69 @@
...
-`add_dependencies_to_path.sh`: A script that adds the dependencies to the PATH.
+    You can build ModelDB's Docker images locally if you prefer, but prebuilt images are available on [MIT DBg's Docker Hub organization](https://hub.docker.com/r/mitdbg/) and will be downloaded automatically by Docker if you skip this step.
 
-`Dockerfile`: A script that tells Docker how to build the image.
+    ```bash
+    docker build -t mitdbg/modeldb-backend -f dockerbuild/Dockerfile-backend .
+    docker build -t mitdbg/modeldb-frontend -f dockerbuild/Dockerfile-frontend .
+    ```
+
+2. **Set up ModelDB containers**
```
mpvartak: This setup is the "demo" setup and doesn't quite mirror how users would actually set up modeldb using docker. There may not be a docker network step involved at all. For example, they'll spin up a mongodb container (or reuse an existing mongo server) and then spin up the modeldb backend and frontend containers. 

I tried that right now and the backend loops with "MongoDB is unavailable - sleeping" when mongo is alive and can even see those connections. I think this has to do with either the way the communication between the backend docker container and mongo is happening or the timeouts.Can you look into this?

Let's rewrite this part as if the user is running docker run on each of the individual images.

mpvartak: I image that the frontend startup will also need to change a little because it waits on the backend to start

mpvartak (changes requested):
I tested out the docker functionality.
- The docker compose for the demo works flawlessly! Nice work
- For the non-demo use case, I'd imagine that users spin up mongo, backend and frontend containers independently. The setup is failing in this case right now due to the wait_for_X code (more in the detailed comments). Let's fix this -- it is fair to assume that if the user is running docker run on individual components, they have the dependencies (e.g. backend on mongo) taken care of.

mpvartak:
Looks great otherwise!

docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+    build:
+      context: .
+      dockerfile: dockerbuild/Dockerfile-frontend
+    image: mitdbg/modeldb-frontend:latest
+    networks:
+      default:
+        aliases:
+          - frontend
+    ports:
+      - "80:3000"
```
mpvartak: Also, unless we plan to start the frontend on 80 for the non-docker setup as well, can we stick with 3000 for consistency?

dockerbuild/Dockerfile-backend
justinanderson: Java is the only one that ends up getting built. It would build Python if it wasn't explicitly excluded, but none of the others would be built either way.

```
thrift 0.10.0

Building Plugin Support ...... : no
Building C++ Library ......... : no
Building C (GLib) Library .... : no
Building Java Library ........ : yes
Building C# Library .......... : no
Building Python Library ...... : no
Building Ruby Library ........ : no
Building Haxe Library ........ : no
Building Haskell Library ..... : no
Building Perl Library ........ : no
Building PHP Library ......... : no
Building Dart Library ........ : no
Building Erlang Library ...... : no
Building Go Library .......... : no
Building D Library ........... : no
Building NodeJS Library ...... : no
Building Lua Library ......... : no
```

dockerbuild/README.md
```diff
@@ -66,4 +66,4 @@ It is also possiblet to run ModelDB in Docker without Docker Compose.
     docker start mongo backend frontend
     ```
 
-    Shortly after running this command, ModelDB should be reachable at [http://localhost/](http://localhost/).
+    Shortly after running this command, ModelDB frontend will be reachable at [http://localhost/3000](http://localhost/3000) and you will be able to log data into ModelDB via ModelDB clients.
```
justinanderson: @mpvartak Should we also mention that the ModelDB server to log data into will be at localhost:6543?

dockerbuild/Dockerfile-backend
mpvartak: Sounds good!

dockerbuild/README.md
mpvartak: fixed

dockerbuild/README.md
justinanderson: @mpvartak The docker network step is required for the containers to find each other by hostname (container name). They could always find each other by IP, but a container's IP may be different each time it runs. There's a `--net=host` option which puts the containers directly on the host, so everything could find each other at localhost, but `--net=host` doesn't work on Docker for Mac.

I can simplify it down to just 3 `docker run` commands instead of `create` and `start`, but I'd recommend we still suggest users create a docker network. I can also split out the mongo part and give instructions for people who already have a MongoDB server.

justinanderson:
@mpvartak Ready for another look.

mpvartak (approved):

mpvartak:
Looks good. There's a conflict in start_server.sh though that needs to be fixed.

ghost:
@justinanderson this looks fantastic, exactly what I had in mind. I'll be working on getting this running early next week.

PR #239, opened 2017-04-08
[MRG] Add more functions to metadata API

mcslao:
* Add functions that can get project and model ids, and can update other fields
* Distinguish between scalar and vector fields for model metadata
* Create sample to show new methods in action

To see the new functions in action:
* Load a sample project by running `[modeldb_folder]/client/python/samples/basic/BasicSyncAll.py`
* View the model metadata of Sample Project in the frontend.
* Run `[modeldb_folder]/client/python/samples/basic/LoggingSample.py`
* See the changes to the model metadata in the frontend.

Current additions are:
```java
  public List<Integer> getProjectIds(Map<String, String> keyValuePairs) throws TException

  public List<Integer> getModelIds(Map<String, String> keyValuePairs) throws TException

  public boolean updateProject(int projectId, String key, String value) throws TException

  public boolean createOrUpdateScalarField(int modelId, String key, String value, String valueType) throws TException

  public boolean createVectorField(int modelId, String vectorName, Map<String, String> vectorConfig) throws TException

  public boolean updateVectorField(int modelId, String key, int valueIndex, String value, String valueType) throws TException

  public boolean appendToVectorField(int modelId, String vectorName, String value, String valueType) throws TException
```

PR #241, opened 2017-04-19
[MRG] run server tests individually

RogerTangos:
Something travis was doing makes the tests run in parallel, so this runs them in sequence. It necessitates that new mvn tests are added to the `.travis.yml` file. Unfortunately, I'm not able to find a way around it, so this is what works for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,20 +68,31 @@ before_script:
 script:
 
   # scala sbt tests
-  # - echo =================== SEVER TESTS ===================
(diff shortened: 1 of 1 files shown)


RogerTangos:
It seems like this is continuing to get nondeterministic failures. The push and pr have identical codebases. 😖

screen shot 2017-04-19 at 4 43 16 pm

mcslao:
@RogerTangos I fixed what I think is a nondeterministic test in #239 so hopefully it also fixes the problem.

Issue #244, opened 2017-04-21
[FEATURE] PyPi Package for Modeldb


RogerTangos:
There should be a pip installable package for modeldb's python client. This will allow people to more easily deploy modeldb.

A PyPi package may require moving syncer.json, or duplicating its hardcoded values into `ConfigUtils.py`

This is currently being worked on at the [PyPI test site](https://testpypi.python.org/pypi/modeldb/0.0.0.dev1)

RogerTangos:
Closed by #245 . `pip install modeldb`

PR #245, opened 2017-04-21
[MRG] Create PyPI package for modeldb

RogerTangos:
Relates to #244. Pip installable modeldb python package.


RogerTangos:
This is up and working! @justinanderson has confirmed that he's able to pip install and use the package.

https://pypi.python.org/pypi/modeldb

RogerTangos:
Users are able to `pip install modeldb`, and use a wizard (`python -m modeldb`) to create a customized syncer setup. 

Pip install information is available in the top level README.md. custom syncer creation is documented in `client/python/README.rst` and on [PyPI/modeldb](https://pypi.python.org/pypi/modeldb).

To try it out, `pip install modeldb`

client/python/modeldb/__main__.py
```diff
@@ -0,0 +1,182 @@
...
+    new_config['thrift']['host'] = thrift_host
+
+    prompt = 'port (default=%s): ' % (config['thrift']['port'])
+    thrift_port = raw_input(prompt)
+    if thrift_port == '':
+        thrift_port = config['thrift']['port']
+    thrift_port = int(thrift_port)
+    new_config['thrift']['port'] = thrift_port
+
+    print('\n== Syncing Strategy =='
```
mpvartak: Some of these are not going to mean much to the user, so perhaps we just expose the relevant ones? Here are the ones that are relevant:

== Output Filename ==
output filename (default=custom_syncer.json): 

== Thrift ==
host (default=localhost): 
port (default=6543): 

== Syncing Strategy ==eager/lazy
syncing strategy (default=eager): 

== Project ==
name (default=Demo): 
author (default=modeldb_users): 
description (default=Project to hold all models from the demo): 

== Experiment ==
name (default=Default Experiment): 
description (default=Default Experiment): 

== Experiment Run Description ==
experiment run description (default=Default Run Description):

client/python/README.md
```diff
@@ -1,3 +0,0 @@
-This folder contains ModelDB's
```
mpvartak, line 1: Did you mean to delete these? Those links might have relevant info, please double check.

mpvartak: And why was the README.md --> rst?

client/python/modeldb/__main__.py
```diff
@@ -0,0 +1,182 @@
...
+        project_name = config['project']['name']
+    new_config['project']['name'] = project_name
+
+    prompt = 'author (default=%s): ' % (config['project']['author'])
+    project_author = raw_input(prompt)
+    if project_author == '':
+        project_author = config['project']['author']
+    new_config['project']['author'] = project_author
+
+    prompt = 'description (default=%s): ' % (config['project']['description'])
```
mpvartak: A bunch of this seems repeated, can it be rewritten to iterate over a list of keys or something?

client/python/modeldb/__main__.py
```diff
@@ -0,0 +1,182 @@
...
+        raise ValueError('invalid literal for boolean. Not a string.')
+
+    lower_value = value.lower()
+    if lower_value in valid:
+        return valid[lower_value]
+    else:
+        raise ValueError('invalid literal for boolean: "%s"' % value)
+
+
+if __name__ == '__main__':
```
mpvartak: Nit: Can we have this helper script be called modeldb_config or something? python -m modeldb makes it seem that we are running modeldb.

client/python/modeldb/basic/ModelDbSyncerBase.py
```diff
@@ -203,9 +204,7 @@ def clear_buffer(self):
     def initialize_thrift_client(self, thrift_config):
         # use defaults if thrift_config values are empty
         if not (thrift_config.port and thrift_config.host):
-            syncer_location = os.path.abspath(
-                os.path.join(os.path.dirname(__file__), self.config_file))
-            config_reader = ConfigReader(syncer_location)
+            config_reader = ConfigReader(self.config_path)
```
mpvartak: self.config_path is not being set anywhere

mpvartak (dismissed):
Tested this out. Works as expected! Tested syncer.json creation too.

Few fixes suggested.

client/python/modeldb/__main__.py
RogerTangos: np. Would `python -m modeldb setup_config` work?

client/python/modeldb/__main__.py
RogerTangos: yep. I'll hide the others.

client/python/README.md
RogerTangos: I migrated this to the .rst. PyPI can render .rst files, but not .md, and it seemed redundant to have both. Github will display either .rst or .md in the file folder, but not both.

client/python/modeldb/basic/ModelDbSyncerBase.py
RogerTangos: This works - config_path is set as a class variable, which is passed to the instance. It'd be good to set that explicitly though, so will update.

client/python/modeldb/__main__.py
justinanderson: I'd lean toward `create_config` since that's what it's doing.

client/python/modeldb/__main__.py
RogerTangos: If I replaced the slightly nicer formatting, it'd be possible to use the keys already supplied in syncer.json. This wouldn't look quite as good, but would reduce the number of lines of code. Would that be useful?

client/python/README.md
mpvartak: SG

client/python/modeldb/__main__.py
mpvartak: sure, create_config is great

client/python/modeldb/__main__.py
mpvartak: You can probably define a function that given a <key (or nested key), type> gets the user input for that key and updates the config. And then you can iterate over the keys we care about.

client/python/modeldb/basic/ModelDbSyncerBase.py
mpvartak: Cool

client/python/modeldb/__main__.py
RogerTangos: Updated so that prompts are generated programmatically based on whatever's in syncer.json. I think it's worth keeping the user keys we don't care about in there too.

If we manually exclude them, we're effectively duplicating keys from `syncer.json` in `__main__.py`, and lose out on the benefit of doing it programmatically. LMK. It's still very easy to throw in if we want.

client/python/modeldb/syncer.json
```diff
@@ -0,0 +1 @@
+../../syncer.json
```
justinanderson, line 1: ModelDB doesn't currently support Windows, but if that's on the roadmap then we'll need to not rely on symlinks. Git on Windows has iffy support for turning POSIX symlinks into Windows ones, and this may not work reliably.

client/python/modeldb/__main__.py
```diff
@@ -0,0 +1,94 @@
...
+    try:
+        int(s)
+        return True
+    except ValueError:
+        return False
+
+
+if __name__ == '__main__':
+    if sys.argv[1] == 'create_config':
+        create_config()
```
justinanderson: If you wrap this in a try-except block, users can exit gracefully with ctrl-c:

```py
try:
    create_config()
except KeyboardInterrupt:
    pass
```

client/python/modeldb/__main__.py
```diff
@@ -0,0 +1,94 @@
...
+import json
+import sys
+
+def create_config():
+    file_dir = os.path.dirname(__file__)
+    filename = file_dir + '/syncer.json'
+    old_config = json.load(file(filename, 'r'))
+    new_config = json.load(file(filename, 'r'))
+
+    raw_input(
```
justinanderson: I recommend changing this to a `print()` call. Pausing for input here gives the impression that there's a choice at this point and that pressing enter now will choose all defaults.

justinanderson (dismissed):

PR #248, opened 2017-04-26
[WIP] Pull out thrift functions to expose

mcslao:


RogerTangos (commented):
Hey @mcslao - is it be possible to include this in the ModelDB file, or merge them in another way, so that existing build scripts work? It seems like this is dependent on at least `LinearModel`, which is defined in ModelDB.thrift

mcslao:
@RogerTangos yeah, you can include other files in a thrift file, but you'd still have to add `../scripts/gen_thrift_file.py java '../thrift/ModelDB_API.thrift' './src/main/thrift/'` to `start_server.sh`

See http://stackoverflow.com/questions/21721532/is-there-any-import-statement-like-thing-in-apache-thrift

PR #276, opened 2017-11-27
[MRG] Moving server.metadata package to server.storage.metadata

### LINKED ISSUE #275: Fix IDE warnings for server.storage.metadata package
Class and Interfaces for edu.mit.csail.db.ml.server.storage.metadata has the right `package` construct but are available under a different directory. IntelliJ IDEA was complaining about missing imports.

ashwanthkumar:
While the package statement in the source files are correct the directory structure was a bit off according to java conventions. Fixes #275


vidit-bhatia:
@ashwanthkumar  The folder for these files are correct . as they are not storage classes and must be in metadata folder the correct thing to do would be making the changes in the package of these files and change the references in all other files.

I am working on a fix and will open a PR soon which have these fixes but feel free to commit if you can keep the package same and change the references

ashwanthkumar:
@vidit-bhatia I've made the changes as requested, please let me know if this is what you had in mind.

ravishetye:

PR #297, opened 2018-06-27
add root path

YujiOshima:
Running the ModelDB under a proxy, which does map a subdirectory to the ModelDB' root.
You can add a path prefix by setting the path to `ROOT_PATH` environment value.

The same concept as `path_prefix` argument of TensorBoard. https://github.com/tensorflow/tensorboard/blob/master/tensorboard/program.py#L226

Signed-off-by: YujiOshima <yuji.oshima0x3fd@gmail.com>

### DIFF
--- a/frontend/app.js
+++ b/frontend/app.js
@@ -11,6 +11,11 @@ var projects = require('./routes/projects');
 
 var app = express();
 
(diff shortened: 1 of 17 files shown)


ravishetye:

PR #310, opened 2018-10-30
Modeldb pv

mpvartak:


server/src/main/resources/reference.conf
```diff
@@ -2,12 +2,12 @@ modeldb: {
...
     "password": "",
-    "jdbcUrl": "jdbc:sqlite:modeldb.db",
-    "jdbcTestUrl": "jdbc:sqlite:modeldb_test.db",
+    "jdbcUrl": "jdbc:sqlite:/db/modeldb.db",
+    "jdbcTestUrl": "jdbc:sqlite:/db/modeldb_test.db",
     "databaseType": "sqlite"
   },
   metadataDb: {
-    "host" : "localhost",
+    "host" : "MONGODB_HOST",
```
jlewi, line 10: How do we set MONGODB_HOST?  Is that an environment variable?

mpvartak: Yes, exactly. This param gets set indirectly but is specified as an ENV variable (see Dockerfile-backend)

PR #327, opened 2020-03-14
Add proposal to push status closer to final elements in diff

conradoverta:
1. Every time we refer to `a` or `b`, there is a `status` at the same level. The messages associated with `a` and `b` should not be nested (that is, they are composed only of basic types). So the comparison for modified should be easier.
2. Comment out some unused imports that were causing issue in my local machine.
3. Add example of how this would change the dataset path case. The logic isn't much more complicated, it just goes deeper.


zlv (approved):

evgeniilezhnin (approved):

conradoverta:
Kirill approved via Slack messages.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diffFactory/DatasetBlobDiffFactory.java
```diff
@@ -2,14 +2,8 @@
...
 
-import ai.verta.modeldb.versioning.BlobDiff;
-import ai.verta.modeldb.versioning.BlobExpanded;
-import ai.verta.modeldb.versioning.DatasetBlob;
-import ai.verta.modeldb.versioning.DatasetDiff;
-import ai.verta.modeldb.versioning.PathDatasetComponentBlob;
-import ai.verta.modeldb.versioning.PathDatasetDiff;
-import ai.verta.modeldb.versioning.S3DatasetComponentBlob;
-import ai.verta.modeldb.versioning.S3DatasetDiff;
+import ai.verta.modeldb.versioning.*;
```
ravishetye, line 5: @conradoverta : are these changes expected?

conradoverta: Yes. I was just adding an example and the autoformatting did this I think.

PR #337, opened 2020-03-16
Refactor diff computation

conradoverta:
The goal of this generation script is to facilitate the manipulation of blobs and diffs, and its expansion. It contains:

1. A script to create Java classes for the blobs and diffs defined in protobuf. These classes have `toProto` and `fromProto` methods to handle conversion. The main use is to have `setX` methods in order to avoid the builder complexity until it's necessary.
2. A visitor system (see `src/main/java/ai/verta/modeldb/versioning/blob/visitors`) that can navigate through the blobs and diffs. An example is the validation of input, which is given in `Validator.java` and is similar to the current validation for dataset.
3. A simplification of the diff process with the new API, which is present in `src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java`. By leveraging the compositional nature of the blobs, it can very easily compute the diff for all objects (and with a few adjustments of the API, the creation of the diff algorithm can actually be fully automated).

I would seriously recommend considering the use of this code for the following reasons:

a. less code to test and maintain, since the wrappers are auto-generated.
b. easier to extend, since adding a new diff is a few lines of code at most
c. easier to maintain, as the diff logic is simpler (we go from the "root" through the trees instead of doing multiple passes and conversions)


conradoverta:
Closed in favor of https://github.com/VertaAI/modeldb/pull/366.

PR #341, opened 2020-03-16
models associated with commit

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/entities/ExperimentRunEntity.java
```diff
@@ -180,7 +180,7 @@ public ExperimentRunEntity(ExperimentRun experimentRun) throws InvalidProtocolBu
       mappedBy = "experimentRunEntity",
       cascade = CascadeType.ALL)
   @LazyCollection(LazyCollectionOption.FALSE)
-  private List<VersioningModeldbEntityMapping> versioningModeldbEntityMappings = new ArrayList<>();
+  private List<VersioningModeldbEntityMapping> versioned_inputs = new ArrayList<>();
```
ravishetye, line 183: Why the snake case? `versionedInputs`

backend/src/main/java/ai/verta/modeldb/entities/ExperimentRunEntity.java
```diff
@@ -412,13 +412,12 @@ public void setAttributeMapping(List<AttributeEntity> attributeMapping) {
     this.attributeMapping.addAll(attributeMapping);
   }
 
-  public List<VersioningModeldbEntityMapping> getVersioningModeldbEntityMappings() {
-    return versioningModeldbEntityMappings;
+  public List<VersioningModeldbEntityMapping> getVersioned_inputs() {
```
ravishetye, line 415: here as well

backend/src/main/java/ai/verta/modeldb/entities/ExperimentRunEntity.java
```diff
@@ -412,13 +412,12 @@ public void setAttributeMapping(List<AttributeEntity> attributeMapping) {
...
+    return versioned_inputs;
   }
 
-  public void setVersioningModeldbEntityMappings(
-      List<VersioningModeldbEntityMapping> versioningModeldbEntityMappings) {
-    this.versioningModeldbEntityMappings = versioningModeldbEntityMappings;
+  public void setVersioned_inputs(List<VersioningModeldbEntityMapping> versioned_inputs) {
```
ravishetye, line 419: here

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
```diff
@@ -341,6 +341,58 @@ public void listCommitBlobs(
...
+  public void listBlobExperimentRuns(
+      ListBlobExperimentRunsRequest request,
+      StreamObserver<ListBlobExperimentRunsRequest.Response> responseObserver) {
+    QPSCountResource.inc();
+    try (RequestLatencyResource latencyResource =
+        new RequestLatencyResource(modelDBAuthInterceptor.getMethodName())) {
+      if (request.getCommitSha().isEmpty()) {
+        throw new ModelDBException("Commit SHA should not be empty", Code.INVALID_ARGUMENT);
+      }
+
```
ravishetye, line 380: we should add check that there is a blob at the location

ravishetye (changes requested):

backend/src/main/java/ai/verta/modeldb/entities/ExperimentRunEntity.java
anandjakhaniya: we have used other fields as snack cases also. the reason is we have used the same name as the proto name so when we called FindExperimentRun with KeyValueQuery we just pass `hyperparameter.loss` & `versioned_inputs.repository_id` as a key and used that key directly in the Criteria query.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
anandjakhaniya: but if blob location not found in DB these will be returned empty response. still, we need to add a check for location?

ravishetye (approved):

PR #342, opened 2020-03-17
compute environment sha unique

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -208,7 +208,8 @@ private void validateEnvironmentVariableName(String name) throws ModelDBExceptio
   private String computeSHA(EnvironmentBlob blob) throws NoSuchAlgorithmException {
     StringBuilder sb = new StringBuilder();
     sb.append("env:");
-    for (EnvironmentVariablesBlob environmentVariablesBlob : blob.getEnvironmentVariablesList()) {
+    for (EnvironmentVariablesBlob environmentVariablesBlob :
+        new HashSet<>(blob.getEnvironmentVariablesList())) {
```
anandjakhaniya: Why this change required? it just for loop you can use whatever I mean List or Set I think it doesn't affect to hash also.

ravishetye:
@evgeniilezhnin  this can be closed? or still, needs review?

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: HashSet ignores order of elements. That's why I have used it. But it doesn't guarantee order, so I better change it to sort.

evgeniilezhnin:
> @evgeniilezhnin this can be closed? or still, needs review?

it needs review

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -227,23 +222,9 @@ private String computeSHA(PythonEnvironmentBlob blob) throws NoSuchAlgorithmExce
...
     sb.append(":requirements:");
-    for (PythonRequirementEnvironmentBlob pythonRequirementEnvironmentBlob :
-        blob.getRequirementsList()) {
-      sb.append(":library:")
-          .append(pythonRequirementEnvironmentBlob.getLibrary())
-          .append(":constraint:")
-          .append(pythonRequirementEnvironmentBlob.getConstraint());
-      appendVersion(sb, pythonRequirementEnvironmentBlob.getVersion());
-    }
+    sb.append(new HashSet<>(blob.getRequirementsList()).hashCode());
```
ravishetye: would hahsCode be same across calls/ across restarts?
we would want the same content of requirementList to be create identical sha

conradoverta:
The autogenerated code has hash computation, which should be stable (it checks for new fields so that we can add them without changing the SHA). If we want to change how SHA is computed, should we use that instead?

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: changed base and used generated files to compute hash

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -208,12 +208,15 @@ private void validateEnvironmentVariableName(String name) throws ModelDBExceptio
...
-    }
+    List<
+            ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model
+                .EnvironmentVariablesBlob>
+        environmentVariables =
+            ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model
+                .EnvironmentBlob.fromProto(blob)
+                .EnvironmentVariables;
+    sb.append(
+        environmentVariables == null ? "null" : new HashSet<>(environmentVariables).hashCode());
```
anandjakhaniya: why you are set to null as a string in StringBuilder? here something wrong. you should just leave it if it is null.

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -226,24 +229,15 @@ private String computeSHA(PythonEnvironmentBlob blob) throws NoSuchAlgorithmExce
...
-    for (PythonRequirementEnvironmentBlob pythonRequirementEnvironmentBlob :
-        blob.getRequirementsList()) {
-      sb.append(":library:")
-          .append(pythonRequirementEnvironmentBlob.getLibrary())
-          .append(":constraint:")
-          .append(pythonRequirementEnvironmentBlob.getConstraint());
-      appendVersion(sb, pythonRequirementEnvironmentBlob.getVersion());
-    }
+    sb.append(
+        blobNew.Requirements == null ? "null" : new HashSet<>(blobNew.Requirements).hashCode());
```
anandjakhaniya: Same comment like above

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -226,24 +229,15 @@ private String computeSHA(PythonEnvironmentBlob blob) throws NoSuchAlgorithmExce
...
     sb.append(":constraints:");
-    for (PythonRequirementEnvironmentBlob pythonConstraintEnvironmentBlob :
-        blob.getConstraintsList()) {
-      sb.append(":library:")
-          .append(pythonConstraintEnvironmentBlob.getLibrary())
-          .append(":constraint:")
-          .append(pythonConstraintEnvironmentBlob.getConstraint());
-      appendVersion(sb, pythonConstraintEnvironmentBlob.getVersion());
-    }
+    sb.append(blobNew.Constraints == null ? "null" : new HashSet<>(blobNew.Constraints).hashCode());
```
anandjakhaniya: same as above comments

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -220,12 +220,16 @@ private void validateEnvironmentVariableName(String name) throws ModelDBExceptio
...
+    List<
+            ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model
+                .EnvironmentVariablesBlob>
+        environmentVariables =
+            ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model
+                .EnvironmentBlob.fromProto(blob)
+                .getEnvironmentVariables();
+    if (environmentVariables != null) {
+      sb.append("env:");
+      sb.append(new HashSet<>(environmentVariables).hashCode());
```
ravishetye: Are you sure Hash set will guarantee you the same order across restarts? We want the same set of envVars to map to same SHA.

Why not use environmentVariables itself? It is getting sorted in the auto gen code.

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -239,23 +243,21 @@ private String computeSHA(PythonEnvironmentBlob blob) throws NoSuchAlgorithmExce
...
-      appendVersion(sb, pythonRequirementEnvironmentBlob.getVersion());
+    ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model.PythonEnvironmentBlob
+        blobNew =
+            ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model
+                .PythonEnvironmentBlob.fromProto(blob);
+    List<ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model.PythonRequirementEnvironmentBlob> requirements = blobNew
+        .getRequirements();
+    if (requirements != null) {
+      sb.append(":requirements:");
+      sb.append(new HashSet<>(requirements).hashCode());
```
ravishetye: similar to above.

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
```diff
@@ -239,23 +243,21 @@ private String computeSHA(PythonEnvironmentBlob blob) throws NoSuchAlgorithmExce
...
-      sb.append(":library:")
-          .append(pythonConstraintEnvironmentBlob.getLibrary())
-          .append(":constraint:")
-          .append(pythonConstraintEnvironmentBlob.getConstraint());
-      appendVersion(sb, pythonConstraintEnvironmentBlob.getVersion());
+    List<ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model.PythonRequirementEnvironmentBlob> constraints = blobNew
+        .getConstraints();
+    if (constraints != null) {
+      sb.append(":constraints:");
+      sb.append(new HashSet<>(constraints).hashCode());
```
ravishetye: same.

ravishetye (commented):
I do not get the requirement of hashSet here

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: oh sorry. Forgot about that. Fixed it. Also added a small change into template to use setters instead of set operator to force sort in fromProto.

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/EnvironmentContainer.java
ravishetye: shouldn't this be removed?

backend/src/main/java/ai/verta/modeldb/versioning/blob/visitors/Visitor.java
```diff
@@ -3,101 +3,103 @@
...
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class Visitor {
-  public void preVisitListOfAutogenBlob(List<AutogenBlob> lst) throws ModelDBException {
+  public void preVisitListOfAutogenBlobDiff(List<AutogenBlobDiff> lst) throws ModelDBException {
     if (lst == null) return;
-    for (AutogenBlob val : lst) {
-      preVisitAutogenBlob(val);
+    for (AutogenBlobDiff val : lst) {
```
ravishetye: I do not understand why this file changed

backend/src/test/java/ai/verta/modeldb/blobs/BlobEquality.java
```diff
@@ -1,22 +1,29 @@
 package ai.verta.modeldb.blobs;
 
-import static org.junit.Assert.*;
-import static org.junit.Assume.*;
-
+import ai.verta.modeldb.ModelDBException;
```
ravishetye: i do not understand why this file changed

backend/src/main/java/ai/verta/modeldb/versioning/blob/visitors/Visitor.java
conradoverta: I believe this file is auto-generated (it might be missing the header).

evgeniilezhnin: for some reason I have different result than in master when I run ./swagger_codegen.sh and reformat step in maven. Did I miss something here?

evgeniilezhnin: It mostly imports and method order.

I did just a small change in generation template. But when I don't do any changes it's still different

ravishetye: @conradoverta : could you review the autogen files and the process. the code merged from #425 makes sense to me otherwise.

ravishetye (approved):

PR #343, opened 2020-03-17
temporal hack

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/versioning/blob/diffFactory/BlobDiffFactory.java
```diff
@@ -39,6 +42,18 @@ public BlobExpanded getBlobExpanded() {
       final Builder builder =
           BlobDiff.newBuilder().addAllLocation(Arrays.asList(location.split("#")));
       builder.setStatus(DiffStatus.MODIFIED);
+      //hack
```
ravishetye:
```suggestion
      //FIXME : remove hack
```

so that it pops up in IDE list

conradoverta:
This worked great for the demo! Should we close it now? I'm not sure we want to have it in.

PR #360, opened 2020-03-18
Feature/vr 3656 blob diff refactoring with tests

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
```diff
@@ -457,32 +456,40 @@ boolean childContains(Set<?> list, Set<?> sublist) {
     }
   }
 
-  List<BlobDiff> getAddedBlobDiff(
+  List<ai.verta.modeldb.versioning.BlobDiff> getAddedBlobDiff(
```
conradoverta, line 459: FYI: we can change the name of the autogenerated things if it makes life simpler.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
```diff
@@ -76,11 +47,14 @@
...
                   // Key collision and one element, process it
-                  //                  if (elA != null && elB != null && elA.size() == 1 &&
-                  // elB.size() == 1) {
-                  //                    return Stream.of(computer.apply(elA.iterator().next(),
-                  // elB.iterator().next()));
-                  //                  }
+                  if (elA != null && elB != null && elA.size() == 1 && elB.size() == 1) {
+                    final F nextA = elA.iterator().next();
+                    final F nextB = elB.iterator().next();
+                    if (nextA.equals(nextB)) {
```
conradoverta, line 53: Nice catch. Let's add a property test that says "diff(A, A) must be null" to make sure this is covered.

conradoverta (approved):
LGTM. Suggested adding one autogenerated test. I'll merge this to my branch so that I can do small changes on top of it based on Kirill's feedback.

backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
evgeniilezhnin: that's a good idea. Now when I have tried to change filenames (I thought it will be easier) it appears that I should make some changes in model.java, but also in the generation script. I will spend some time to understand how to do it, but if it will be too hard, I'll postpone it and ask for your help.

evgeniilezhnin: So I can't just change --file-suffix from java to Model.java in swagger_codegen.sh ? It always expects just java?
I have tried to make corresponding changes in swagger_codegen.sh api.java, model.java, but it always generates the wrong files. Did I forget anything?

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
evgeniilezhnin: So when we generate random data for it then there can be empty keys. And then this test fails. Should we just check a similar case in a unit test where we set data manually or postpone it?

conradoverta: I don't understand the case you're talking about. Could you give a concrete example?

evgeniilezhnin: concrete example: https://github.com/VertaAI/modeldb/pull/399#discussion_r396412051

conradoverta: Sorry I missed this. I'll take a look over the weekend.

PR #361, opened 2020-03-18
commit test cases & delete commit bug

anandjakhaniya:
- Fix delete commit bug (**Reason:** throw an exception when delete commit due to branch & commit mapping does not delete from Branch table)
- enable `location_prefix` support for ListCommitBlobsRequest with test cases


conradoverta (commented):
Let's focus on getting the functionality tested before adding new functionality. More specifically, we don't need `location_prefix` right now. No system is using that parameter, so implementing it right now is not a good use of time. It's ok now that it's done, but let's avoid for the other APIs.

ravishetye (approved):

PR #367, opened 2020-03-18
VR-3659: Don't get SpaCy models from pip env

convoliution:
A fix for [this behavior](https://vertaai.slack.com/archives/CV3KW9TT6/p1584463452001500) we saw during the demo.

SpaCy models have a consistent naming scheme (defined [here](https://spacy.io/models#conventions), and verified [here](https://github.com/explosion/spacy-models/blob/master/compatibility.json)), so they can be reliably filtered out of `pip freeze`. It's particular enough that I don't expect any false positives.

### DIFF
--- a/client/verta/requirements.txt
+++ b/client/verta/requirements.txt
@@ -20,6 +20,7 @@ pandas
 pillow<7.0
 scikit-learn<0.21
 scipy<1.3
+spacy
 tensorflow<2.0
 torch
 xgboost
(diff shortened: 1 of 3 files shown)


conradoverta (commented):
Can we add that to the tests, since we already have tests installing spacy?

convoliution:
> Can we add that to the tests, since we already have tests installing spacy?

Sure, will do!

convoliution:
@conradoverta Test added; depends on https://github.com/VertaAI/cluster-setup/pull/1035

conradoverta (approved):

PR #369, opened 2020-03-18
Enforce that lists inside autogenerated are sorted

conradoverta:
Use getters/setters and move the sorting from the tests to inside the classes themselves. This will ensure internal diffs and hashes are consistent.


backend/src/main/java/ai/verta/modeldb/versioning/autogenerated/_public/modeldb/versioning/model/CommandLineEnvironmentDiff.java
```diff
@@ -141,19 +141,37 @@ public int hashCode() {
 
   public CommandLineEnvironmentDiff setA(List<String> value) {
     this.A = Utils.removeEmpty(value);
+    if (this.A != null) {
+      this.A.sort(Comparator.comparingInt(String::hashCode));
```
evgeniilezhnin: I don't think that sorting is applicable for command line because in our blob API it is the only collection that is really list and we should maintain an order of elements in it

evgeniilezhnin: I think that we should just ignore command line when adding sort in autogenerated

evgeniilezhnin: Or don't enforce it and make us able to get sorted or not sorted list

conradoverta: Oh, wow! That is correct. Let me figure out how to fix it. Nice catch.

conradoverta: Oh wow! Nice catch. Let me figure out how to do this.

conradoverta: I did a small fix on this with a solution that makes sense to me so far. Please take a look. I also added comments so that we don't forget what is the reasoning.

evgeniilezhnin (approved):

PR #376, opened 2020-03-19
Vr 3616 fix diff and add commit from diff bugs

evgeniilezhnin:
- fixes for diff
- fixes for commit with diff


backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
```diff
@@ -278,8 +286,230 @@ public void createCommit(
...
+                  pythonEnvironmentDiffConstraintsList) {
+                validate(pythonConstraintEnvironmentDiff, "constraint");
+                if (pythonConstraintEnvironmentDiff.hasB()) {
+                  requirements.add(pythonConstraintEnvironmentDiff.getB());
+                  requirementsSet.add(pythonConstraintEnvironmentDiff.getB());
+                }
+              }
+              if (requirementsSet.size() != requirements.size()) {
+                throw new ModelDBException(
+                    "There are recurring requirements or constraints", Code.INVALID_ARGUMENT);
```
conradoverta: Wouldn't make more sense to check this in the commit created after applying the diff instead of the diff itself? It's possible the duplicated entries come from the diff and original commit.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
```diff
@@ -278,8 +286,230 @@ public void createCommit(
     }
   }
 
-  // TODO: add validation VR-3587
-  private void validateBlobDiffs(CreateCommitRequest request) throws ModelDBException {}
+  private void validateBlobDiffs(CreateCommitRequest request) throws ModelDBException {
```
conradoverta: This can be replaced with the visitor pattern on the generated code, if you'd prefer. It ensures the validation is called for everything. I personally got lost trying to review on what was logic (which requires more attention) and what was boilerplate. But I'm ok leaving as is if we think it's best.

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/BlobContainer.java
```diff
@@ -68,8 +70,30 @@ protected void validate(PathDatasetBlob path) throws ModelDBException {
   }
 
   protected void validate(PathDatasetComponentBlob path) throws ModelDBException {
+    Optional<String> result = validateReturnMessage(path);
```
conradoverta: I'm not sure I get why the validation was separated from the message creation.

conradoverta (commented):
I don't have enough context on what were the issues that this is supposed to fix and a lot of code shifted around, so just leaving a couple of notes for now.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
evgeniilezhnin: yes, you are right. It checks just B list too. Removed it and enabled validation after blobs creation.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
evgeniilezhnin: I'm not sure how to add different validation logic in generated code for each blob type with it. Maybe it will be easier if we'll ignore A blobs validation. And for diffs we'll just validate statuses. But I think there should be validation for A blobs too.

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/BlobContainer.java
evgeniilezhnin: just to use lambdas

conradoverta: Doesn't sound like a good enough reason to me.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
conradoverta: We don't need to add validation in the generated code. The visitor is done outside and overrides what methods it wants. See https://github.com/VertaAI/modeldb/blob/feature/diffapi/backend/src/main/java/ai/verta/modeldb/versioning/blob/visitors/Validator.java for instance.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -633,7 +633,7 @@ public void deleteBranchByCommit(Long repoId, String commitHash) {
       while (!childCommitSHAs.isEmpty()) {
         String childCommit = childCommitSHAs.remove(0);
         commitSHAs.add(childCommit);
-        Query sqlQuery = session.createSQLQuery(getParentCommitsQuery + childCommit + "\'");
+        Query sqlQuery = session.createSQLQuery(getParentCommitsQuery + childCommit + "\"");
```
ravishetye: single quote for Postgres . Also slightly surprized that doing in just here and not the prefix worked.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
evgeniilezhnin: moved validation to other branch, will follow up on that

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/BlobContainer.java
evgeniilezhnin: moved validation to other branch, will follow up on that

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
evgeniilezhnin: removed the change

backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
```diff
@@ -417,19 +417,19 @@ boolean childContains(Set<?> list, Set<?> sublist) {
...
-              locationBlobsMapCommitB.values().stream()
-                  .map(Entry::getValue)
-                  .collect(Collectors.toSet()));
-      Map<String, BlobExpanded> locationBlobsCommon =
-          getLocationWiseBlobExpandedMapFromCollection(commonBlobs.values());
-      modifiedLocations.removeAll(locationBlobsCommon.keySet());
+      Map<String, Set<BlobExpanded>> blobsB = getCollectToMap(locationBlobsMapCommitB);
+      Map<String, Set<BlobExpanded>> blobsA = getCollectToMap(locationBlobsMapCommitA);
+      for (Map.Entry<String, Set<BlobExpanded>> entry : blobsA.entrySet()) {
+        Set<BlobExpanded> ent = blobsB.get(entry.getKey());
```
ravishetye, line 425: Commenting on how we compute the modified set will be helpful.

evgeniilezhnin: fixed

ravishetye (approved):
LGTM . lets wait few minutes if @ad-47  wanted to review this as well.

backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
```diff
@@ -416,20 +416,24 @@ boolean childContains(Set<?> list, Set<?> sublist) {
...
-      modifiedLocations.removeAll(locationBlobsCommon.keySet());
+      // get B sha -> blobs
+      Map<String, Set<BlobExpanded>> blobsB = getCollectToMap(locationBlobsMapCommitB);
+      // get A sha -> blobs
+      Map<String, Set<BlobExpanded>> blobsA = getCollectToMap(locationBlobsMapCommitA);
+      // delete blobs same with A
+      for (Map.Entry<String, Set<BlobExpanded>> entry : blobsA.entrySet()) {
+        Set<BlobExpanded> ent = blobsB.get(entry.getKey());
+        if (ent != null) {
+          ent.removeAll(entry.getValue());
```
anandjakhaniya, line 427: please make sure that call by reference updates the value in `blobsB` map.

backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
```diff
@@ -450,6 +454,20 @@ boolean childContains(Set<?> list, Set<?> sublist) {
...
 
+  private Map<String, Set<BlobExpanded>> getCollectToMap(
+      Map<String, Entry<BlobExpanded, String>> locationBlobsMapCommit) {
+    return locationBlobsMapCommit.values().stream()
+        .collect(
+            Collectors.toMap(
+                Entry::getValue,
+                entry -> new HashSet<>(Collections.singletonList(entry.getKey())),
+                (m1, m2) -> {
+                  HashSet<BlobExpanded> newHash = new HashSet<>(m1);
```
anandjakhaniya: Instead of HashSet please use the LinkedHashSet because we need to maintain the sequence order of record.

anandjakhaniya (changes requested):
Drop some comments otherwise looks good to me.

backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
evgeniilezhnin: it does

backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
evgeniilezhnin: fixed

PR #392, opened 2020-03-22
fix index issue in oss the right way

ravishetye:
#374 creates more problems than it fixes.
@ad-47 Never edit tags, unless you have a really good reason for doing so.


anandjakhaniya (approved):
Sorry about that. Really I don't know it was already applied to the database.

PR #396, opened 2020-03-23
Feature/vr 3688 commit in multiple locations fix

evgeniilezhnin:


backend/src/main/resources/liquibase/create-versioning-tables-changelog2.xml
```diff
@@ -296,4 +296,17 @@
 				constraintName="ck_versioning_modeldb_entity_mapping"
 				tableName="versioning_modeldb_entity_mapping"/>
 	</changeSet>
+	<changeSet id="changeKeysOfFolderElement" author="lezhevg">
+		<dropTable tableName="folder_element"/>
```
conradoverta, line 317: I’m not sure I follow why we need to drop the table

conradoverta (changes requested):

backend/src/main/resources/liquibase/create-versioning-tables-changelog2.xml
evgeniilezhnin: fixed

ravishetye:
@evgeniilezhnin this needs to be edited as per the recommendation in slack discussion, right?

evgeniilezhnin:
> @evgeniilezhnin this needs to be edited as per the recommendation in slack discussion, right?

had already did that

backend/src/main/resources/liquibase/create-versioning-tables-changelog2.xml
ravishetye: break them into changeset as mentioned in the message. A single changeset if interrupted won't rollback create table.

ravishetye:
> > @evgeniilezhnin this needs to be edited as per the recommendation in slack discussion, right?
> 
> had already did that

break them into changeset as mentioned in the message. A single changeset if interrupted won't rollback create table.

evgeniilezhnin:
> > > @evgeniilezhnin this needs to be edited as per the recommendation in slack discussion, right?
> > 
> > 
> > had already did that
> 
> break them into changeset as mentioned in the message. A single changeset if interrupted won't rollback create table.

done

ravishetye (approved):

PR #393, opened 2020-03-23
Fix test cases for ListCommitsLog endpoint

anandjakhaniya:

PR #397, opened 2020-03-23
added index

anandjakhaniya:
- test case fixes also added in other PR so if that PR merge then this test case changes need be to resolve.  PR: https://github.com/VertaAI/modeldb/pull/393


ravishetye (changes requested):
precondition should be reveresed

ravishetye (approved):

PR #409, opened 2020-03-24
Change name of autogenerated entities

conradoverta:


evgeniilezhnin (approved):
Works all right.
But also need renames in utils: https://github.com/VertaAI/modeldb/pull/410

PR #412, opened 2020-03-24
Feature/vr 3616 validation refactoring

evgeniilezhnin:
1. Change blob validation to use visitors from autogen.
2. Blob diff validation implementation.


backend/src/main/java/ai/verta/modeldb/versioning/blob/visitors/Validator.java
```diff
@@ -1,20 +1,2310 @@
...
+        break;
+      case MODIFIED:
+        modified.validate();
+        break;
+      default:
+        throw new ModelDBException("Unknown type", Status.Code.INVALID_ARGUMENT);
+    }
+  }
+
+  @Override
```
conradoverta: You don't need to override if you're just going to use the default behavior. It also makes things much more difficult to review. Also, you don't have to implement the "deep" version of the methods unless you want to. They will call their equivalent non-deep and then call for each subfield, which is usually the desired behavior.

See the original for an example of just implementing the cases that will have extra logic.

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/PathDatasetComponentContainer.java
```diff
@@ -0,0 +1,24 @@
+package ai.verta.modeldb.versioning.blob.container;
+
+import ai.verta.modeldb.ModelDBException;
+import ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model.AutogenGitCodeBlob;
+import ai.verta.modeldb.versioning.autogenerated._public.modeldb.versioning.model.AutogenPathDatasetComponentBlob;
+import io.grpc.Status.Code;
+
+public class PathDatasetComponentContainer implements BlobContainerBase {
```
conradoverta: Why do we need a container wrapping around the autogen functionality? Seems like a good amount of boilerplate that the autogen was created to remove.

backend/src/main/java/ai/verta/modeldb/versioning/blob/visitors/Validator.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/PathDatasetComponentContainer.java
evgeniilezhnin: because I thought that we need to check a and b only if it should present in diff. But now I think that it is not so important. Fixed.

ravishetye:
The base need s to be changed to master.
A summary of changes or motivation in the PR summary will be helpful

evgeniilezhnin:
> The base need s to be changed to master.
> A summary of changes or motivation in the PR summary will be helpful

changed.

1. Change blob validation to use visitors from autogen.
2. Blob diff validation implementation.

ravishetye (approved):

backend/src/main/java/ai/verta/modeldb/versioning/blob/visitors/Validator.java
```diff
@@ -1,20 +1,391 @@
...
+    if (autogenBlob.getCode() == null
+        && autogenBlob.getDataset() == null
+        && autogenBlob.getEnvironment() == null
+        && autogenBlob.getConfig() == null) {
+      throw new ModelDBException("Unknown blob type specified", Status.Code.INVALID_ARGUMENT);
+    }
+  }
+
+  private void preVisitDeep(ProtoType proto) throws ModelDBException {
+    if (isNull(proto)) {
```
conradoverta: Should we just fix this in the base class? Seems like an important fix that will save a lot of duplicated code too.

evgeniilezhnin: I see that postVisit has already all necessary null checks. Can I use post visit instead of pre visit like in the last commit?

conradoverta: Sure. It does sound like a bug on the pre, so that needs to be fixed anyway.

PR #417, opened 2020-03-25
Allow slash in the branch name

conradoverta:


convoliution:
gRPC never ceases to amaze me.

PR #421, opened 2020-03-25
Delete child before deleting parent in test

convoliution:
The backend now returns a `Bad Request` when deleting a parent commit while it has children:
```python
{'thread': 'grpc-default-executor-56',
 'level': 'WARN',
 'loggerName': 'ai.verta.modeldb.utils.ModelDBUtils',
 'message': 'Exception occured: Commit has the child, please delete child commit first',
 'endOfBatch': False,
 'loggerFqcn': 'org.apache.logging.log4j.spi.AbstractLogger',
 'instant': {'epochSecond': 1585176229, 'nanoOfSecond': 357000000},
 'threadId': 268,
 'threadPriority': 5}
```

This PR fixes a test.

### DIFF
--- a/client/verta/tests/conftest.py
+++ b/client/verta/tests/conftest.py
@@ -285,7 +285,10 @@ def commit(repository):
     yield commit
 
     if commit.id is not None:
-        utils.delete_commit(repository.id, commit.id, repository._conn)
+        try:
+            utils.delete_commit(repository.id, commit.id, repository._conn)
+        except:
+            pass  # may have already been deleted in test
 
 
 @pytest.fixture
--- a/client/verta/tests/conftest.py
+++ b/client/verta/tests/conftest.py
@@ -7,6 +7,8 @@ import random
 import shutil
 import string
 
+import requests
+
 import verta
 from verta import Client
 from verta._internal_utils import _utils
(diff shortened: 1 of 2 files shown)


client/verta/tests/conftest.py
```diff
@@ -267,14 +269,9 @@ def experiment_run(client):
...
     name = _utils.generate_default_name()
     repo = client.get_or_create_repository(name)
-    root_id = repo.get_commit()
 
     yield repo
 
-    try:
-        utils.delete_commit(repository.id, root_id, repository._conn)
-    except:
-        pass  # may have already been deleted in test
```
convoliution, line 277: The back end also returns an error when deleting the root (or the `master` branch, didn't investigate which), so we're just not doing that anymore.

(fun fact! because of the blanket `except:` here, there's actually a error that's gone uncaught ever since this fixture was written!)

conradoverta: Never catch all =)

client/verta/tests/conftest.py
```diff
@@ -285,7 +282,14 @@ def commit(repository):
...
 
     if commit.id is not None:
-        utils.delete_commit(repository.id, commit.id, repository._conn)
+        try:
+            utils.delete_commit(repository.id, commit.id, repository._conn)
+        except requests.HTTPError as e:
+            try:
+                if e.response.status_code == 404 and e.response.json()['code'] == 5:
+                    pass  # already deleted in test
+            except:
```
conradoverta, line 291: Catch all!

conradoverta (approved):
LGTM

client/verta/tests/conftest.py
convoliution: 🔥

PR #422, opened 2020-03-26
entities owner deletion part 2

anandjakhaniya:
- added logic when deleting the entity
- added test cases for the type of condition

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
@@ -107,7 +107,7 @@ public class DatasetVersionServiceImpl extends DatasetVersionServiceImplBase {
     Set<String> datasetIdSet = new HashSet<>(datasetIdDatasetVersionIdMap.values());
 
     List<String> accessibleDatasetVersionIds = new ArrayList<>();
-    List<String> allowedDatasetIds;
+    List<String> allowedDatasetVersionIds;
     // Validate if current user has access to the entity or not
     if (datasetIdSet.size() == 1) {
       roleService.isSelfAllowed(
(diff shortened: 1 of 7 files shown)


backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
```diff
@@ -116,16 +116,39 @@ public DatasetVersionServiceImpl(
...
       accessibleDatasetVersionIds.addAll(requestedDatasetVersionIds);
     } else {
-      allowedDatasetIds =
+      allowedDatasetVersionIds =
           roleService.getSelfAllowedResources(
               ModelDBServiceResourceTypes.DATASET, modelDBServiceActions);
       // Validate if current user has access to the entity or not
-      allowedDatasetIds.retainAll(requestedDatasetVersionIds);
-      accessibleDatasetVersionIds.addAll(allowedDatasetIds);
+      allowedDatasetVersionIds.retainAll(datasetIdSet);
```
ravishetye: datasetIdSet ? is this right? if so was there a bug to begin with?

anandjakhaniya: it was a typo. fixed

ravishetye: I do not see this changed

backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
```diff
@@ -116,16 +116,39 @@ public DatasetVersionServiceImpl(
...
+      }
     }
     return accessibleDatasetVersionIds;
   }
 
+  public List<String> getAccessibleDatasetVersionIDsByAction(
+      List<String> requestedDatasetVersionIds, ModelDBServiceActions modelDBServiceActions) {
+    // Validate if current user has access to the entity or not
+    if (requestedDatasetVersionIds.size() == 1) {
+      roleService.isSelfAllowed(
```
ravishetye, line 137: let us say a user has delete privilages on Dataset and deos not have a role binding at dataset version . how will this call to isSelfAllowed  working?

ravishetye (changes requested):
we can follow up on zoom but this requires followup

ravishetye (approved):

backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
anandjakhaniya: resolved conversation over zoom

PR #423, opened 2020-03-26
VR-3618 blobs and diffs negative unit tests

evgeniilezhnin:


backend/src/test/java/ai/verta/modeldb/PublicTestSequenceSuite.java
```diff
@@ -22,6 +27,11 @@
   RepositoryTest.class,
   CommitTest.class,
   DiffTest.class,
-  MetadataTest.class
+  MetadataTest.class,
```
ravishetye, line 30: Please verify these tests work without uacservice

ravishetye (approved):
left a comment to check in OSS setup.
For future and depending on the time required for this PR, please add a summary comment of how an entity look so that  others can reuse the entity for their  tests easily. In this case the blob difs.
adding a json representation of a toString helps the team

backend/src/test/java/ai/verta/modeldb/PublicTestSequenceSuite.java
evgeniilezhnin: All 5 tests are unit tests, they should work without dependencies and without uac

PR #424, opened 2020-03-26
VR-3681: Add autocapture flag to applicable blobs

convoliution:


client/verta/verta/environment/_python.py
```diff
@@ -41,10 +43,12 @@ class Python(_environment._Environment):
...
-        super(Python, self).__init__(env_vars=env_vars)
-        self._capture_python_version()
-        self._capture_requirements(requirements)
+    def __init__(self, requirements=None, constraints=None, env_vars=None, _autocapture=True):
+        super(Python, self).__init__(env_vars, _autocapture)
+
+        if _autocapture:
+            self._capture_python_version()
+        self._capture_requirements(requirements, _autocapture)
         self._capture_constraints(constraints)
```
conradoverta: Why not just skip these directly if not auto capturing?

convoliution: I thought it would make sense to do the check inside the methods because they have related logic (if something, then autocapture, else don't), but your comment has me considering it'll be easier to understand to have the check in `__init__()`.

conradoverta: Yeah. I mean, the method name literally has "capture" in it, so I wouldn't expect it to do any non-capturing =)

conradoverta (approved):

PR #432, opened 2020-03-27
fix diff error on older blobs

ravishetye:
the sha computation code extracted from #342  (mainly because it is off feature branch, and will require unnecessary merge resolutions, and secondly it had a small chqnage yet to be made). Thanks @evgeniilezhnin  for the original code.


conradoverta (approved):

ravishetye:
@evgeniilezhnin : please review and have a follow up if there is a better way to handle this.

evgeniilezhnin (commented):
LGTM

PR #438, opened 2020-03-30
Feature/vr 3750 batch process owner binding

evgeniilezhnin:
have used function from aj/entity-owner-deletion. But also needs code from master branch


ravishetye:
@evgeniilezhnin the migration code does not run in background if you are online, it will help if you can fix that, If @ad-47 has to do it , then you will atleast need to ensure the fix is propogated here.

evgeniilezhnin:
> @evgeniilezhnin the migration code does not run in background if you are online, it will help if you can fix that, If @ad-47 has to do it , then you will atleast need to ensure the fix is propogated here.

have changed a base to the fix and validated that it is working

backend/src/main/java/ai/verta/modeldb/advancedService/AdvancedServiceImpl.java
```diff
@@ -441,18 +440,30 @@ public void getHydratedExperimentRunsInProject(
...
 
-    Map<String, Actions> actions = new HashMap<>();
+    Map<String, Actions> projectActions = new HashMap<>();
     if (projectIdSet.size() > 0) {
-      actions =
+      projectActions =
           roleService.getSelfAllowedActionsBatch(
               new ArrayList<>(projectIdSet), ModelDBServiceResourceTypes.PROJECT);
     }
+    List<Action> projectActionList =
```
ravishetye: thoughts on why this change shows up?

backend/src/main/java/ai/verta/modeldb/batchProcess/OwnerRoleBindingUtils.java
```diff
@@ -55,6 +57,8 @@ public static void execute() {
     LOGGER.info("ExperimentRuns done migration");
     migrateDatasetVersions();
     LOGGER.info("DatasetVersions done migration");
+    migrateRepositories();
```
ravishetye, line 60: can we extract this out of its own migration util with its own flag.

ravishetye (approved):
this looks good , but pasted some question may be we resolve them here or on slack and get this in.

backend/src/test/java/ai/verta/modeldb/versioning/blob/visitors/ValidatorBlobDiffTest.java
```diff
@@ -34,58 +34,58 @@
 import org.junit.runners.Parameterized.Parameters;
 
 /*
- * Examples:
```
evgeniilezhnin, line 37: autoformat

backend/src/main/java/ai/verta/modeldb/advancedService/AdvancedServiceImpl.java
evgeniilezhnin: I think it was a part of some other branch that I was basing my changes on before, but not in master now. Sorry for that

backend/src/main/java/ai/verta/modeldb/batchProcess/OwnerRoleBindingUtils.java
evgeniilezhnin: done

PR #442, opened 2020-03-31
support branches in diff and merge calls

ravishetye:
we should error out if both the commit and branch name are present in the request.
we should use the branch name in merge when available
if could be posssible to have a commit diffed/merged to a branch


backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
```diff
@@ -477,30 +509,86 @@ boolean childContains(Set<?> list, Set<?> sublist) {
...
+      invalidParam = true;
+    } else if (!branchB.isEmpty() && !commitB.isEmpty()) {
+      invalidParam = true;
+    } else if (branchB.isEmpty() && commitB.isEmpty()) {
+      invalidParam = true;
+    }
+
+    if (invalidParam) {
+      throw new ModelDBException(
+          "Branches and Commits both are not allowed in the request", Status.Code.INVALID_ARGUMENT);
```
ravishetye, line 529:
```suggestion
          "Exactly one of Branches and Commits are required in the request", Status.Code.INVALID_ARGUMENT);
```

PR #445, opened 2020-03-31
Feature/vr 3752 collaborator api change

evgeniilezhnin:
small proto change and generated files


ravishetye (approved):
LGTM , the implementation though need to be in uac and modeldb right?

evgeniilezhnin:
> LGTM , the implementation though need to be in uac and modeldb right?

yes, it's moved to uac

PR #448, opened 2020-03-31
Feature/vr 3754 delete should delete bindings

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -296,6 +301,71 @@ public RepositoryEntity getRepositoryById(Session session, RepositoryIdentificat
...
+          WorkspaceType.forNumber(repositoryEntity.getWorkspace_type()),
+          String.valueOf(repositoryEntity.getId()));
+    }
+  }
+
+  private void deleteWorkspaceRoleBindings(
+      String workspaceId,
+      WorkspaceType workspaceType,
+      String resourceId) {
+    if (workspaceId != null && !workspaceId.isEmpty()) {
```
ravishetye: could we refactor the common logic to a common function which takes in the resourceType to delete as  parameters.
we have the deleteWorkspaceRoleBindings at project, dataset and repository now.

This function should exist and just invoke the common function.

ravishetye (changes requested):
looks fine otherwise

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
evgeniilezhnin: fixed

PR #457, opened 2020-04-01
Fix indexing (added exists check before creating index)

anandjakhaniya:


ravishetye:
@conrado this should fix the issue you were seeing.

ravishetye (approved):

PR #461, opened 2020-04-01
fix pushd issue

evgeniilezhnin:


backend/proxy/Makefile
```diff
@@ -1,2 +1,2 @@
 all:
-	pushd ../.. && docker build -t vertaaiofficial/modeldb-proxy -f backend/proxy/Dockerfile . && popd
+	cd ../..; docker build -t vertaaiofficial/modeldb-proxy -f backend/proxy/Dockerfile .
```
ravishetye: there should be a `cd backend/proxy` at the end to have the same effect as popd

ravishetye (changes requested):

backend/proxy/Makefile
evgeniilezhnin: fixed

ravishetye (approved):

PR #435, opened 2020-03-30
Prevent deletion of commit if referred to by branch or tag

anandjakhaniya:

PR #465, opened 2020-04-01
Don't delete commits in tests

convoliution:
This update is needed because of https://github.com/VertaAI/modeldb/pull/435


conradoverta:
I assume that the repository gets deleted, right?

convoliution:
Yep, all the tests still make a `DeleteRepositoryRequest` during teardown.

conradoverta (approved):

PR #475, opened 2020-04-05
Fix failing autogen tests

conradoverta:
They started to fail when Evgenii changed the `fromProto` to use `setFooBar` instead of direct equality, which is the right way. This highlighted another bug: the hash code used for sorting was not consistent. The same fields would generate a different hash. A little bit of digging pointed out that this was because it was hashing the object, including its address.

This PR uses the `toString()` to compute the hashCode for sorting, which should always be consistent. The tests triggered by `mvn -Dtest=ai.verta.modeldb.blobs.* test` now pass.


ravishetye (commented):
Changes makes sense to me ,  but i  would like @evgeniilezhnin to  see it this causes any issues else where by running all the tests

evgeniilezhnin (approved):

PR #476, opened 2020-04-06
Fix bug for Repository listing returns only number of records

anandjakhaniya:


evgeniilezhnin (approved):

conradoverta (commented):
This probably also happens with other things like projects and datasets. Please verify.

ravishetye (approved):

PR #479, opened 2020-04-06
VR-3805 visibility implementation

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
```diff
@@ -448,7 +448,12 @@ public static Throwable findRootCause(Throwable throwable) {
                 .addDetails(Any.pack(defaultInstance))
                 .build();
       } else {
-        LOGGER.error("Exception occurred:", e);
+        LOGGER.error("Exception occurred: {}", e.getMessage());
+        StackTraceElement[] stack = e.getStackTrace();
```
ravishetye, line 452: what motivated this?

ravishetye (approved):

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
zlv: When there is some bug, I can't see why this bug happened in Kibana without printing stack trace. I can just remove this code from here and do this stacktrace print in separate PR or use some log4j tools for that

evgeniilezhnin: When there is some bug, I can't see why this bug happened in Kibana without printing stack trace. I can just remove this code from here and do this stacktrace print in separate PR or use some log4j tools for that

ravishetye: no this is fine for now.
in future either bubble this up in PR description or adda PR comment as a self review

PR #484, opened 2020-04-08
address test related to merge

ravishetye:


backend/src/test/java/ai/verta/modeldb/blobs/DiffAndMerge.java
```diff
@@ -24,8 +24,13 @@ public void diffAndMerge(AutogenBlob a, AutogenBlob b) throws ModelDBException {
...
 
     // Reapplying the diff should not change the result
-    diffedB = DiffMerger.mergeBlob(diffedB, d);
-    assertEquals(newB, diffedB);
+    HashSet<String> conflictSet = new HashSet<String>();
+    diffedB = DiffMerger.mergeBlob(diffedB, d, conflictSet);
+    if (newA == null || newB == null) {
+      assertEquals(newB, diffedB);
+    } else {
+      assertFalse(conflictSet.isEmpty());
```
conradoverta: Is there anything we can say about the content of this? Or do you think just testing that it has anything is enough? The goal should be to assert the strongest statement we can with as few assumptions as we can. But maybe building the assert would be too complex for this case?

backend/swagger_codegen.sh
```diff
@@ -381,12 +388,17 @@ do
         ${type}Diff d = DiffComputer.compute$(echo $type | sed 's,Autogen,,')Diff(newA, newB);
 
         // Applying the diff on top of the original A should get original B
-        ${type}Blob diffedB = DiffMerger.merge$(echo $type | sed 's,Autogen,,')(newA, d);
+        ${type}Blob diffedB = DiffMerger.merge$(echo $type | sed 's,Autogen,,')(newA, d, new HashSet<String>());
         assertEquals(newB, diffedB);
 
+		HashSet<String> conflictSet = new HashSet<String>();
```
conradoverta: Nit: spaces

backend/src/test/java/ai/verta/modeldb/blobs/DiffAndMerge.java
ravishetye: the assert is difficult to generate due to the randomized nature of the inputs.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
```diff
@@ -305,14 +305,14 @@ public static AutogenPythonEnvironmentDiff computePythonEnvironmentDiff(
                     a,
                     b,
                     AutogenPythonEnvironmentBlob::getConstraints,
-                    AutogenPythonRequirementEnvironmentBlob::getLibrary,
+                    e -> e.getLibrary() + e.getConstraint(),
```
ravishetye, line 308: @conradoverta : do the changes make sense to you? If Ihave different hashing in differ and merger then we have issues during merge around modified  being detected as added

conradoverta: Yeah, it seems reasonable to me. I don't know what you mean by "different hashing" though.

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
ravishetye: generates different key for the map

PR #493, opened 2020-04-08
VR-3885: Expand ~ to user directory

convoliution:
I kind of wish `os.path` had a way to enable this behavior across the module, but it doesn't so I manually looked for places where we take a filepath from the user.

I'm certain I got all the spots in the versioning ingredient submodules. I'm not certain I got everything in the rest of the client 😵


convoliution:
Running tests to make sure nothing has broken from this change

conradoverta (approved):

PR #496, opened 2020-04-09
VR-3315 convert config integration test to unit test

evgeniilezhnin:


client/verta/verta/client.py
```diff
@@ -158,22 +159,23 @@ def __init__(self, host=None, port=None, email=None, dev_key=None,
...
+                             None)
+
+          def is_unauthorized(response): return response.status_code == 401
+
+          if is_unauthorized(response):
+              auth_error_msg = "authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`"
+              six.raise_from(requests.HTTPError(auth_error_msg), None)
+
+          _utils.raise_for_http_error(response)
+          print("connection successfully established")
```
convoliution: This should indented by four spaces instead of two spaces
```suggestion
            try:
                response = _utils.make_request("GET",
                                               "{}://{}/api/v1/modeldb/project/verifyConnection".format(conn.scheme, conn.socket),
                                               conn)
            except requests.ConnectionError:
                six.raise_from(requests.ConnectionError("connection failed; please check `host` and `port`"),
                               None)

            def is_unauthorized(response): return response.status_code == 401

            if is_unauthorized(response):
                auth_error_msg = "authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`"
                six.raise_from(requests.HTTPError(auth_error_msg), None)

            _utils.raise_for_http_error(response)
            print("connection successfully established")
```

client/verta/verta/client.py
```diff
@@ -112,10 +112,11 @@ class Client(object):
         Currently active Project.
     expt : :class:`Experiment` or None
         Currently active Experiment.
+    connect : connect to server (False for unit tests)
 
     """
     def __init__(self, host=None, port=None, email=None, dev_key=None,
-                 max_retries=5, ignore_conn_err=False, use_git=True, debug=False):
+                 max_retries=5, ignore_conn_err=False, use_git=True, debug=False, connect=True):
```
convoliution: This lets users know it's not generally for public use.
```suggestion
                 max_retries=5, ignore_conn_err=False, use_git=True, debug=False, _connect=True):
```

client/verta/verta/client.py
```diff
@@ -112,10 +112,11 @@ class Client(object):
         Currently active Project.
     expt : :class:`Experiment` or None
         Currently active Experiment.
+    connect : connect to server (False for unit tests)
```
convoliution: **Also**, could you move this to the `Parameters` section (line 99)?

```suggestion
    _connect : str, default True
        Whether to connect to server (``False`` for unit tests).
```

convoliution (changes requested):
@evgeniilezhnin I realized that this integration test is actually very useful, because it tests that this is possible:
```
client = verta.Client(connect=False)
client.set_experiment_run()
```
without needing `set_project()`/`set_experiment()`.

Converting this to a unit test would lose this coverage. Maybe this can be a separate test instead.

evgeniilezhnin:
> @evgeniilezhnin I realized that this integration test is actually very useful, because it tests that this is possible:
> 
> ```
> client = verta.Client(connect=False)
> client.set_experiment_run()
> ```
> 
> without needing `set_project()`/`set_experiment()`.
> 
> Converting this to a unit test would lose this coverage. Maybe this can be a separate test instead.

restored it

client/verta/verta/client.py
```diff
@@ -112,10 +112,12 @@ class Client(object):
         Currently active Project.
     expt : :class:`Experiment` or None
         Currently active Experiment.
+    _connect : str, default True
+        Whether to connect to server (``False`` for unit tests).
```
convoliution: Could you move this to the `Parameters` section? (line 99)

client/verta/tests/test_entities.py
```diff
@@ -91,6 +91,14 @@ def test_else_http(self):
...
     @pytest.mark.skipif('VERTA_EMAIL' not in os.environ or 'VERTA_DEV_KEY' not in os.environ, reason="insufficient Verta credentials")
     def test_config_file(self):
+        self.test_config_file_with_type(connect = False)
+
+    @pytest.mark.skipif('VERTA_EMAIL' not in os.environ or 'VERTA_DEV_KEY' not in os.environ, reason="insufficient Verta credentials")
+    def test_config_file_connect(self):
+        self.test_config_file_with_type(connect = True)
+        
+    @pytest.mark.skip("util function")
+    def test_config_file_with_type(self, connect):
```
convoliution: Instead of `mark.skip`, I suggest simply renaming this function to something that does _not_ start with `test_` (for example: `def config_file_test(self, connect)`)

Pytest will then ignore it, because Pytest only runs functions that start with `test_`!

convoliution (changes requested):
See comments

client/verta/verta/client.py
evgeniilezhnin: fixed

client/verta/tests/test_entities.py
evgeniilezhnin: fixed

evgeniilezhnin:
> See comments

fixed

convoliution (approved):

PR #499, opened 2020-04-09
Skip model API tests until pandas is dealt with

convoliution:


conradoverta (approved):
That was not what I meant =) I meant to skip the specific tests that fail, not the whole thing and parts that are passing right now.

convoliution:
Pssshh of course, that would be the rational thing to do here ha ha

PR #505, opened 2020-04-11
enhancement for tracking s3 versioning

ravishetye:


protos/protos/public/modeldb/versioning/Dataset.proto
```diff
@@ -21,6 +22,9 @@ message PathDatasetComponentBlob {
 
     string sha256 = 4;
     string md5 = 5;
+
+    string version_id = 6;
+    VersionIDTypeEnum.VersionIDType version_type = 7;
```
conradoverta: It seems very weird to me to have S3 component -> path component -> flag saying it's S3. Seems like a reversal of boundaries and easy to lose the semantics. I would recommend adding `version_id` to the S3 component itself, as that's the only version information we have at the moment and it's very meaningful (when we see the field, we know we're talking about S3 version specifically). If we figure out some other system that we want to generalize, then we can extract that.

protos/protos/public/modeldb/DatasetVersionService.proto
```diff
@@ -77,6 +77,8 @@ message DatasetPartInfo {
     uint64 size = 2;
     string checksum = 3;
     uint64 last_modified_at_source =4;
+    string version_id = 5;
+    VersionIDTypeEnum.VersionIDType version_type = 6;
```
conradoverta: I wouldn't bother adding to the previous system since that might change significantly. We can add later if it turns out it won't.

protos/protos/public/modeldb/versioning/Dataset.proto
ravishetye: it is saying the version-id is provided by s3. I think if s3 versioning is not enabled or the dataset is not s3 and there is an identifier that can act as version id then we can store it in the `version_id` field. This was the thought behind having `CUSTOM` in the typeEnum as well, where the interpretation of the version id is left to the consumer.

protos/protos/public/modeldb/DatasetVersionService.proto
ravishetye: may be we have it in protos and deprioritize the implementation?

conradoverta: I'd prefer not have something in protos until we actually plan to implement it. Same reason that we commented out some fields in the versioning protos.

protos/protos/public/modeldb/versioning/Dataset.proto
conradoverta: I generally favor strong meaning and de-favor arbitrary catch all. I put more weight on how the data will be manipulated, interpreted and reasoned about than how it will be stored, since the latter is much easier to solve.

This approach opens doors for things like setting S3 flag on a non-S3 component. Or putting something meaningless inside the version as a catch all. It shifts the burden from the data structure that will be re-used by multiple systems (backend, client, UI) into application logic that needs to be kept consistent and correct across all 3 (4+ if you consider other clients we decide to built). Furthermore, it makes it easy to pass `PathDatasetComponentBlob` and start doing S3-specialized operations, which is really bad practice. This generates functionality by convention instead of design, which is less favorable as it depends on humans always complying with conventions.

There are good design reasons why overloading functionality in programming languages is preferred on the object type instead of a flag inside the object itself, or using interfaces instead of generic objects and parsing their types. Using strong types is better when the type will matter (an S3-versioning operation should only accept an S3-compatible type as argument). There are very few cases I'd recommend using a flag-type instead of a strong-type.

Note that we had the same discussion when I proposed having a separate type for S3 component vs just a path component with a type flag inside. I'm happy to revisit that discussion to get on the same page.

ravishetye: > Furthermore, it makes it easy to pass PathDatasetComponentBlob and start doing S3-specialized operations, which is really bad practice. This generates functionality by convention instead of design, which is less favorable as it depends on humans always complying with conventions.

Versiond id is similar metadata like path. Why do we not worry about a Path data set having s3urls in its path? 

If there is a usecase where user has a string called `banana` which can be used by the **user** to retrieve a version of their data then there should be a way for them to store and retrieve the version id from modeldb.

I do not have issues moving the version up to s3Component but think it is adding an unnecessary restriction to the design, and we will end up revisiting and restructuring when we support versioning beyond the one provided by s3.

conradoverta: I don't think it's imposing an unnecessary restriction and I don't think it's "metadata like path", since today a path doesn't have a specific meaning for a specific application. If we have some API that receives a PathComponent only and does S3-specific operations based on the path, rest assured that I'll have exactly the same concerns.

If there is a use case like the one you described, we should consider with more care and design with that in mind. Designing for arbitrary hypotheticals is not a good approach. What if the version is a number? Will it be converted into a string and force every system using the API to do conversion and type checks? What if the versioning is defined by a part of string and integer? Will we store them in a single string following some convention that needs to be parsed and validated? What if the version is more complex? There are too many possible open-ended scenarios for us to try to cover all of them based on guess. It's really hard to remove the arbitrary nature later instead of starting with something concrete and generalize later when we actually know what we need.

Considering that the current data versioning API has this same approach of putting a flag and, to my understand, this has caused enough issues on the client that we're considering re-designing the API, I'd also prefer to be a little bit more cautious before using the same solution.

convoliution (approved):

conradoverta (approved):

PR #509, opened 2020-04-13
Have describe() print instead of returning str

convoliution:
Probably makes more sense than the current
Screen Shot 2020-04-13 at 2 52 38 PM


conradoverta:
Why? No big preference either way. I like to return string so that people can customize. Printing can be a bit invasive.

convoliution:
I find this easier, and it makes sense to me for a `describe()` method to simply print.
But if it's not needed, then it's fine. I can shelve this until users bring it up.

PR #497, opened 2020-04-09
Revert commit implementation

anandjakhaniya:

PR #510, opened 2020-04-13
VR-3956: Change Client to use backend revert

convoliution:


client/verta/verta/_repository/commit.py
```diff
@@ -548,10 +548,32 @@ def revert(self, other=None, message=None):
...
+
+        commit_id = response_msg.commit.commit_sha
+        if self.branch_name is not None:
+            # update branch to revert commit
+            set_branch(self._conn, self._repo.id, commit_id, self.branch_name)
+            new_commit = self._repo.get_commit(branch=self.branch_name)
+        else:
+            new_commit = self._repo.get_commit(id=commit_id)
+
+        self.__dict__ = new_commit.__dict__
```
convoliution: This logic may look familiar; it's delegated to a function in #511.

client/verta/verta/_repository/commit.py
```diff
@@ -548,10 +548,32 @@ def revert(self, other=None, message=None):
...
+        if message is not None:
+            msg.content.message = message
 
-        self.apply_diff(other.get_revert_diff(), message)
+        data = _utils.proto_to_json(msg)
+        endpoint = "{}://{}/api/v1/modeldb/versioning/repositories/{}/commits/{}/revert".format(
+            self._conn.scheme,
+            self._conn.socket,
+            self._repo.id,
+            msg.commit_to_revert_sha,
```
conradoverta, line 569: This will always revert the commit onto itself, which is different from what we had before.

convoliution: Not entirely sure what you mean by "revert the commit onto itself" (is the revert commit a child of the correct commit?),
but I'm building the URL according to the protos: https://github.com/VertaAI/modeldb/blob/master/protos/protos/public/modeldb/versioning/VersioningService.proto#L630

conradoverta: The protos have option to use two different commits. You are only using one, from what I can see. You can't revert `C1` and apply the changes to `C2`.

convoliution: Oh, I can't? Then how exactly is this supposed to be used?

cc @ravishetye if he also has anything to add

ravishetye: the base_commit commit on which the revert is to be applied is coming from the body

ravishetye: https://github.com/VertaAI/modeldb/blob/master/protos/protos/public/modeldb/versioning/VersioningService.proto#L416

convoliution: >You are only using one

Is this accounting for lines 552 and 553, where I'm using the two commit IDs?

>You can't revert `C1` and apply the changes to `C2`.

Do you mean "you can't" as in "it's impossible", or as in "this is being done incorrectly"? Because based on the [backend PR](https://github.com/VertaAI/modeldb/pull/497#pullrequestreview-391257314), I think this should be possible.

conradoverta: Oh, I missed that you are setting the two fields, as that's unexpected to me =) I thought one would be set only by the path and completely missed that you set another in the other field. Ignore my comment.

conradoverta (approved):

PR #513, opened 2020-04-14
VR-3946 validation and exception message fix

evgeniilezhnin:
According to logs it is not in notebook, but in environment. Looks like version was empty and it wasn’t checked, so validation needs fix. Also fixed the exception message


ravishetye (approved):

ravishetye (changes requested):
when I said notebook I meant this can be reproduce in the standard notebook: https://github.com/VertaAI/modeldb/blob/master/demos/webinar-2020-4-1/census-s3-oss-versioning.ipynb

Based on your fix we will throw an invalid argument error, I am not sure if making the python and requirement version compulsory is the right fix for this.

tagging @conradoverta  and @convoliution since they will have more context on python requirements

convoliution:
I do not think the backend should raise an error. I believe that fields should be optional from the backend's perspective if things are simply being logged. But I could be wrong about that, so I'd defer to @conradoverta.

@ravishetye This error is surprising to me, since I would have expected the Client to always capture the versions (especially looking at the webinar notebook). Do you have any other details about how you were running it?

evgeniilezhnin:
removed validation, added null checked

ravishetye (approved):

PR #517, opened 2020-04-14
Add Blob type as a column in versioning_modeldb_entity_mapping

anandjakhaniya:


anandjakhaniya:
@ravishetye what about the response I mean we should add blob_type in VersionedInput response for the client? or column added just for the DB.

ravishetye:
> @ravishetye what about the response I mean we should add blob_type in VersionedInput response for the client? or column added just for the DB.

good point lets add it in response as well, I think. it will be good to show that on UI as well

backend/src/main/java/ai/verta/modeldb/entities/ExperimentRunEntity.java
```diff
@@ -75,12 +75,6 @@ public ExperimentRunEntity(ExperimentRun experimentRun) throws InvalidProtocolBu
           RdbmsUtils.generateCodeVersionEntity(
               ModelDBConstants.CODE_VERSION, experimentRun.getCodeVersionSnapshot()));
     }
-
```
ravishetye, line 78: was this not required?

anandjakhaniya: No, it is not required at that point. this removed code snippet was moved on DAO level https://github.com/VertaAI/modeldb/pull/517/files#diff-1a3287003b54e5e5b064b79266c3d82bR301-R305 because we have to populate the blob_type which is we can not fetch from ExperimentRunEntity OR we can not pass it as the parameter.

anandjakhaniya:
https://github.com/VertaAI/modeldb/pull/517#issuecomment-613730183 if we add it in response then it will also populate in request because Request and response both are used same `message`. 
```
message VersioningEntry {
    uint64 repository_id = 1;
    string commit = 2;
    map<string, Location> key_location_map = 3;
}
```

ravishetye (approved):

PR #525, opened 2020-04-15
Update lineage service to use versioning functionality

conradoverta:
This adds small changes to our existing lineage service to drop the existing data versioning solution in favor of the commit-based one.

### DIFF
--- a/protos/protos/public/modeldb/Lineage.proto
+++ b/protos/protos/public/modeldb/Lineage.proto
@@ -28,11 +28,6 @@ message LineageEntry {
     };
 }
 
-// Represent a set of lineage items, essentially an array so that it can be embedded on other proto messages.
-message LineageEntryBatch {
-    repeated LineageEntry items = 1;
-}
-
 // Add every item of "input" as a parent of every item on "output".
 message AddLineage {
     // If not provided, a new ID will be created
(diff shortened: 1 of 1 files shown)


evgeniilezhnin (approved):

conradoverta:
@evgeniilezhnin holding off on this change to avoid breaking the build. You can place your code on top of it.

evgeniilezhnin:
why this new id is required? How can we know this id when we want to delete something? 

I mean it is returned only when we add lineage, but not when we get existing lineage.

protos/protos/public/modeldb/Lineage.proto
```diff
@@ -39,18 +35,27 @@ message LineageEntryBatch {
...
-        bool status = 1;
+        uint64 id = 1;
     }
 }
 
 // Delete every item of "input" as a parent of every item on "output".
 message DeleteLineage {
-    repeated LineageEntry input = 1;
-    repeated LineageEntry output = 2;
+    uint64 id = 1; // Required
```
ravishetye, line 51: is the design that we can partially delete a Lineage entry? if not then just the id should be sufficient, right?

ravishetye (approved):

protos/protos/public/modeldb/Lineage.proto
conradoverta: Yes, it should cover partial delete.

PR #526, opened 2020-04-16
VR-3505: Capture versioning information from S3

convoliution:
Still needs to be able to accept user-specified `version_id`s.
Figuring out the API is complicated because `verta.dataset.S3()` accepts either

- a single string
- a list of strings

that can be either

- a bucket
- a specific key!

I could separate out the multiple argument types into `S3.read_bucket(bucket)` and `S3.read_object(key)` utils [similar to `Python`](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/environment/_python.py#L198), but that would break the existing API and also make it more verbose to use.

### DIFF
--- a/client/verta/tests/test_versioning/test_dataset.py
+++ b/client/verta/tests/test_versioning/test_dataset.py
@@ -95,7 +95,7 @@ class TestS3:
         dataset = verta.dataset.S3("s3://{}/{}".format(bucket, key))
 
         assert len(dataset._msg.s3.components) == 1
-        assert dataset._msg.s3.components[0].version_id == latest_version_id
+        assert dataset._msg.s3.components[0].s3_version_id == latest_version_id
 
     @pytest.mark.skip("Not yet implemented.")
     def test_versioned_object_by_id(self):
--- a/client/verta/tests/test_versioning/test_dataset.py
+++ b/client/verta/tests/test_versioning/test_dataset.py
@@ -60,6 +60,7 @@ class TestS3:
 
         assert len(multiple_dataset._msg.s3.components) == len(bucket_dataset._msg.s3.components)
 
+    @pytest.mark.skip("Not yet implemented.")
     def test_versioned_bucket(self):
         pass
 
@@ -79,6 +80,7 @@ class TestS3:
 
         assert dataset._msg.s3.components[0].version_id == latest_version_id
 
+    @pytest.mark.skip("Not yet implemented.")
     def test_versioned_object_by_id(self):
         pass
 
--- a/client/verta/tests/test_versioning/test_dataset.py
+++ b/client/verta/tests/test_versioning/test_dataset.py
@@ -102,6 +102,7 @@ class TestS3:
 
         bucket = "verta-versioned-bucket"
         key = "data/census-train.csv"
+        s3_url = "s3://{}/{}".format(bucket, key)
 
         # pick a version that's not the latest
         version_ids = [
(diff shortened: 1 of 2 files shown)


conradoverta:
The concern of adopting something like `Python` style isn't breaking or verbosity, but the fact that they represent very different styles. `Python` is divided because you are talking about multiple sources of information that would be confusing if they overlapped. On the other hand, giving a path to a bucket, or to an item, or to a list of them falls naturally inside the same class of "set of locations to register". So having an API similar to the one we have makes things simpler from user POV. We can expand what a `location` means besides just a string though.

convoliution:
Ahhh, I see. I'd been trying to figure out what the conceptual distinction was!

client/verta/verta/dataset/_s3.py
```diff
@@ -93,22 +111,66 @@ def _get_s3_metadata(cls, bucket_name, key=None):
...
+        msg.path.size = obj.get('Size') or obj.get('ContentLength') or 0
         msg.path.last_modified_at_source = _utils.timestamp_to_ms(_utils.ensure_timestamp(obj['LastModified']))
         msg.path.md5 = obj['ETag'].strip('"')
+        if obj.get('VersionId', 'null') != 'null':
+            msg.s3_version_id = obj['VersionId']
 
         return msg
+
+    @staticmethod
+    def location(bucket, key=None, version_id=None):
```
conradoverta: I would recommend parsing S3 url here (or in the constructor) too. I don't see a reason to 1) force different interfaces, 2) force the user to split.

client/verta/verta/dataset/_s3.py
```diff
@@ -93,22 +111,66 @@ def _get_s3_metadata(cls, bucket_name, key=None):
...
     @classmethod
     def _get_s3_obj_metadata(cls, obj, bucket_name, key):
         # pylint: disable=no-member
         msg = _DatasetService.S3DatasetComponentBlob()
         msg.path.path = cls._S3_PATH.format(bucket_name, key)
-        msg.path.size = obj.get('Size') or obj['ContentLength']
+        msg.path.size = obj.get('Size') or obj.get('ContentLength') or 0
         msg.path.last_modified_at_source = _utils.timestamp_to_ms(_utils.ensure_timestamp(obj['LastModified']))
         msg.path.md5 = obj['ETag'].strip('"')
+        if obj.get('VersionId', 'null') != 'null':
```
conradoverta: None? Not sure why we should use a special string like `null`. If there is a reason, add a comment to it?

convoliution: `'null'` is what AWS's S3 API returns when there's no version ID! I'll make a comment

client/verta/tests/test_versioning/test_dataset.py
```diff
@@ -60,6 +60,66 @@ def test_s3_no_duplicates(self):
 
         assert len(multiple_dataset._msg.s3.components) == len(bucket_dataset._msg.s3.components)
 
+    def test_versioned_bucket(self):
+        s3 = pytest.importorskip("boto3").client('s3')
+        S3_PATH = verta.dataset.S3._S3_PATH
+
+        bucket = "verta-versioned-bucket"
```
conradoverta, line 67: I'm assuming you already put some data there.

conradoverta (approved):

client/verta/tests/test_versioning/test_dataset.py
convoliution: There is data there! I almost forgot:
@ravishetye fyi, you had created this bucket to test versioning with. I'm now using this bucket for our automated Client tests. If this would pose any problem for you, I can create a separate bucket.

ravishetye: go for it. just a related datapoint when we version stuff and delete the object, the object is not actually deleted. We need to delete the version to keep the s3 cost in control.

ravishetye (approved):

PR #528, opened 2020-04-16
Add graphql backend

conradoverta:
Screenshot 2020-04-15 19 33 57

@ravishetye please review the build part and integration with our setup. I have verified with docker compose locally as per screenshot above.


backend/graphql/internal/schema/README.txt
```diff
@@ -0,0 +1,4 @@
+1) Re-run "go run github.com/99designs/gqlgen" when you update the schema.
```
ravishetye: would it possible to document the structure of the folder, when would one end up updating which file?

backend/graphql/internal/schema/dataloaders/README.md
```diff
@@ -0,0 +1,5 @@
+1) Get the generator `go get github.com/vektah/dataloaden`
```
ravishetye, line 1: dataloader caches the data, or just batches and optimizes the calls to backend/uac?
we only do it for uac because we anticipate only those calls to have the volume which makes sense to batch?

backend/graphql/internal/schema/definition/actions.graphql
```diff
@@ -0,0 +1,6 @@
+type AllowedActions {
```
ravishetye, line 1: are these files auto generated?

ravishetye (commented):
makes sense to me , added questions for my clarification.

backend/graphql/internal/schema/README.txt
conradoverta: ack

backend/graphql/internal/schema/dataloaders/README.md
conradoverta: Just batches and optimizes if the same call happens over and over, like what happens on collaborators (teams and users).

We don't cache anything between calls as permissions and state might have changed. Caching is a nightmare.

backend/graphql/internal/schema/definition/actions.graphql
conradoverta: No.

backend/graphql/internal/schema/README.txt
conradoverta: Done

ravishetye (approved):

PR #536, opened 2020-04-17
VR-3895: Only undeploy models in successful tests

convoliution:
Note that the model's ExperimentRun and Project will still be deleted, because I don't have a clean way to selectively skip test teardowns.


conradoverta (approved):

conradoverta:
Thanks =)

PR #540, opened 2020-04-18
VR-3794: verta remote add & verta remote use

convoliution:


client/verta/verta/_cli/remote.py
```diff
@@ -14,11 +16,26 @@ def remote():
...
 @click.argument('name')
 def use(name):
     """Set a remote repository as the active one."""
-    click.echo("now using repo {}".format(name))
+    with _config_utils.read_config() as config:
+        if name not in config.get('remotes', {}):
+            raise click.BadParameter("no such remote: '{}'".format(name))
+
+    with _config_utils.write_config() as config:
+        config['current-remote'] = name
```
conradoverta, line 48: We should set the remote added with `add` if it's the only one.

client/verta/tests/test_cli.py
```diff
@@ -19,3 +19,62 @@ def test_init(self):
             result = runner.invoke(cli, ['init'])
             assert not result.exception
             assert result.output.startswith("found existing config file ")
+
+
+class TestRemote:
+    def test_add(self):
+        # TODO: create a new repo and use it
+        name = "banana"
```
conradoverta: Avoid fruit names. It will be good muscle memory.

client/verta/verta/_cli/remote.py
convoliution: Great thinking! 0382b7c

conradoverta (approved):

PR #541, opened 2020-04-18
Get host from env for test

convoliution:
Because of #537, `Client` checks the environment for a `$VERTA_HOST` before checking the config, so this test needs to be updated to account for that.

This also indicates that since #537, anyone who

- has started using the config file since it was released (exactly!) a month ago
- is relying on the config file to set their host
- has `$VERTA_HOST` set in their environment, and it's different from the confiig file

now has to update their setup. But I don't think anyone is.


convoliution:
Merging this now so that tests aren't failing throughout the weekend.

PR #542, opened 2020-04-18
s3VersionID support in diff and  merge

ravishetye:
as indicated by the branch name I thought I was just fixing broken test cases here.
However soon I realized that our support for s3 version in diff was missing.
This PR adds support for diff
the main changes are 
1.  protos change to bubble up diff status to s3ComponentBlobs (previously was at pathCpmponnentBloblevel) : https://github.com/VertaAI/modeldb/pull/542/files?file-filters%5B%5D=.proto#diff-f4aee5c87aa637da2326853d8bddc58bR58
1. the changes are trivial changes to absorb the protos change. 
    1. we are now picking the path and version as a key for S3ComponenBlob in most of the maps
1. added null filtering in asserts in test to match what differ is doing in src, I think this change was always required but am surprized what did it not occur earlier? since the property test are probabilistic we might have just gotten lucky?


protos/protos/public/modeldb/versioning/Dataset.proto
```diff
@@ -56,6 +56,15 @@ message PathDatasetDiff {
 
 message S3DatasetComponentDiff {
     PathDatasetComponentDiff path = 1;
+    S3VersionIdDiff s3_version_id = 2;
+    DiffStatusEnum.DiffStatus status = 3;
```
ravishetye, line 58: @conradoverta : i am finding it difficult to merge S3DatasetComponent without the status here. Does it make sense to have a DiffStatus here?

conradoverta: The recommendation is to add a `S3DatasetComponentVersionDiff` that has status, A, B and C, like we do for everything else like command line in the environment. This should automatically fall into place with the autogen code and the UI.

Why using a status here would help? We can add if it will make things for you, but it will not be consumed by other pieces that have a well defined structure already (unless we decide to change).

conradoverta:
>we are now picking the path and version as a key for S3ComponenBlob in most of the maps

I'd suggest keeping the path alone. I don't see a use case for having the same path but multiple versions in the same dataset. On the other hand, I do see a merge conflict being hidden by keeping 2 versions of the same path.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
```diff
@@ -225,20 +225,16 @@ public static AutogenS3DatasetDiff computeS3DatasetDiff(
                     a,
                     b,
                     AutogenS3DatasetBlob::getComponents,
-                    x -> x.getPath().getPath(),
+                    x -> x.getPath().getPath() + x.getS3VersionId(),
```
conradoverta: Be careful about just concatenating strings, as they can randomly match with another concatenation. Using some unlikely chars like `[s1, s2]` makes it less likely.

conradoverta:
>added null filtering in asserts in test to match what differ is doing in src, I think this change was always required but am surprized what did it not occur earlier? since the property test are probabilistic we might have just gotten lucky?

Very unlikely, since we've run hundreds of times before and it also is random but biased to cover corner cases. The original properties also state exactly what is the expected behavior and were passing because we had the expected behavior.. I'd say this probably indicates a bug in the code itself, since now the property we're trying to assert is more specific than what it should be. The objects should already be handling null filtering by themselves, but maybe the merge function isn't? Generally if something breaks the tests, I think changing the test is not the right solution unless we know what caused the break and can affirm that the new behavior is the correct. It seems here that isn't the case, at least with the information we have currently. We should not have to duplicate the null filtering in the test and in every code that uses the object. Seems like something the object and whatever operates on it (e.g. differ, merge) should handle.

ravishetye:
> > we are now picking the path and version as a key for S3ComponenBlob in most of the maps
> 
> I'd suggest keeping the path alone. I don't see a use case for having the same path but multiple versions in the same dataset. On the other hand, I do see a merge conflict being hidden by keeping 2 versions of the same path.

May be I do not understand "I'd suggest keeping the path alone." correctly. But I think it would be common to dump up just the version of a S3file in a component blob and keeping its path same.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
ravishetye: will do

conradoverta:
> > > we are now picking the path and version as a key for S3ComponenBlob in most of the maps
> > 
> > 
> > I'd suggest keeping the path alone. I don't see a use case for having the same path but multiple versions in the same dataset. On the other hand, I do see a merge conflict being hidden by keeping 2 versions of the same path.
> 
> May be I do not understand "I'd suggest keeping the path alone." correctly. But I think it would be common to dump up just the version of a S3file in a component blob and keeping its path same.

Imagine the following scenario: there is a dataset with component with path P on master with no version (V0 == empty). A user creates a branch B1 that updates that dataset so that P is the same, but the version is V1. Another user creates a branch B2 that updates that dataset so that P is the same, but the version is V2. Both B1 and B2 are merged into master. What is the expected result? For me, I'd expect the second merge to have a conflict since the version field was modified. If we key on both path and version, then I think they'd be considered different components and I'd end up with (P, V0), (P, V1) and (P, V2), all inside the same dataset. Maybe I'm missing something on the merge.

ravishetye:
> > > > we are now picking the path and version as a key for S3ComponenBlob in most of the maps
> > > 
> > > 
> > > I'd suggest keeping the path alone. I don't see a use case for having the same path but multiple versions in the same dataset. On the other hand, I do see a merge conflict being hidden by keeping 2 versions of the same path.
> > 
> > 
> > May be I do not understand "I'd suggest keeping the path alone." correctly. But I think it would be common to dump up just the version of a S3file in a component blob and keeping its path same.
> 
> Imagine the following scenario: there is a dataset with component with path P on master with no version (V0 == empty). A user creates a branch B1 that updates that dataset so that P is the same, but the version is V1. Another user creates a branch B2 that updates that dataset so that P is the same, but the version is V2. Both B1 and B2 are merged into master. What is the expected result? For me, I'd expect the second merge to have a conflict since the version field was modified. If we key on both path and version, then I think they'd be considered different components and I'd end up with (P, V0), (P, V1) and (P, V2), all inside the same dataset. Maybe I'm missing something on the merge.

makes sense , thanks for the example, now I can not remember why I wanted the s3version in there.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
ravishetye: agreed on the other comment of not requiring versionid for generating the map key. So this comment in no longer valid

ravishetye:
> > added null filtering in asserts in test to match what differ is doing in src, I think this change was always required but am surprized what did it not occur earlier? since the property test are probabilistic we might have just gotten lucky?
> 
> Very unlikely, since we've run hundreds of times before and it also is random but biased to cover corner cases. The original properties also state exactly what is the expected behavior and were passing because we had the expected behavior.. I'd say this probably indicates a bug in the code itself, since now the property we're trying to assert is more specific than what it should be. The objects should already be handling null filtering by themselves, but maybe the merge function isn't? Generally if something breaks the tests, I think changing the test is not the right solution unless we know what caused the break and can affirm that the new behavior is the correct. It seems here that isn't the case, at least with the information we have currently. We should not have to duplicate the null filtering in the test and in every code that uses the object. Seems like something the object and whatever operates on it (e.g. differ, merge) should handle.

1. agreed on don't change the test philosophy, hence highlighted the change in PR description
1.  based on the code currently there I can logically argue why this change is required probably on slack/screen share tomorrow.

ravishetye:
> > > added null filtering in asserts in test to match what differ is doing in src, I think this change was always required but am surprized what did it not occur earlier? since the property test are probabilistic we might have just gotten lucky?
> > 
> > 
> > Very unlikely, since we've run hundreds of times before and it also is random but biased to cover corner cases. The original properties also state exactly what is the expected behavior and were passing because we had the expected behavior.. I'd say this probably indicates a bug in the code itself, since now the property we're trying to assert is more specific than what it should be. The objects should already be handling null filtering by themselves, but maybe the merge function isn't? Generally if something breaks the tests, I think changing the test is not the right solution unless we know what caused the break and can affirm that the new behavior is the correct. It seems here that isn't the case, at least with the information we have currently. We should not have to duplicate the null filtering in the test and in every code that uses the object. Seems like something the object and whatever operates on it (e.g. differ, merge) should handle.
> 
> 1. agreed on don't change the test philosophy, hence highlighted the change in PR description
> 2. based on the code currently there I can logically argue why this change is required probably on slack/screen share tomorrow.

fixed this in sanitization instead of the test. Merging this in

ravishetye:
resolved all comments so merging things in.

PR #545, opened 2020-04-20
FindRepositories endpoint

anandjakhaniya:
Added endpoint called FindRepositories for the filter on repository based on Labels, Name, Owner

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
+++ b/backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
@@ -1,0 +1,377 @@
+package ai.verta.modeldb.versioning;
+
+import ai.verta.modeldb.KeyValueQuery;
+import ai.verta.modeldb.ModelDBConstants;
+import ai.verta.modeldb.OperatorEnum;
+import ai.verta.modeldb.dto.WorkspaceDTO;
+import ai.verta.modeldb.entities.metadata.LabelsMappingEntity;
+import ai.verta.modeldb.entities.versioning.RepositoryEntity;
+import ai.verta.modeldb.metadata.IDTypeEnum;
+import ai.verta.modeldb.versioning.RepositoryVisibilityEnum.RepositoryVisibility;
+import com.google.protobuf.Value;
+import com.google.rpc.Code;
+import com.google.rpc.Status;
+import io.grpc.protobuf.StatusProto;
+import java.util.Calendar;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.hibernate.Session;
+import org.hibernate.query.Query;
+
+public class FindRepositoriesQuery {
+  private FindRepositoriesQuery() {}
+
+  private Query findRepositoriesHQLQuery;
+  private Query findRepositoriesCountHQLQuery;
+
+  public Query getFindRepositoriesHQLQuery() {
+    return findRepositoriesHQLQuery;
+  }
+
+  public Query getFindRepositoriesCountHQLQuery() {
+    return findRepositoriesCountHQLQuery;
+  }
+
+  private FindRepositoriesQuery(FindRepositoriesHQLQueryBuilder builder) {
+    this.findRepositoriesHQLQuery = builder.buildQuery();
+    this.findRepositoriesCountHQLQuery = builder.buildCountQuery();
+  }
+
+  // Builder Class
+  public static class FindRepositoriesHQLQueryBuilder {
+    private static final Logger LOGGER =
+        LogManager.getLogger(FindRepositoriesHQLQueryBuilder.class);
+
+    boolean isNeedToWhereCause = false;
+    final Session session;
+    final WorkspaceDTO workspaceDTO;
+    String countQueryString;
+    Map<String, Object> parametersMap = new HashMap<>();
+
+    // optional parameters
+    private List<Long> repoIds;
+    private List<String> repoNames;
+    private List<KeyValueQuery> predicates;
+    private Integer pageNumber = 0;
+    private Integer pageLimit = 0;
+
+    public FindRepositoriesHQLQueryBuilder(Session session, WorkspaceDTO workspaceDTO) {
+      this.session = session;
+      this.workspaceDTO = workspaceDTO;
+    }
+
+    public FindRepositoriesHQLQueryBuilder setRepoIds(List<Long> repoIds) {
+      if (repoIds != null && !repoIds.isEmpty()) {
+        this.repoIds = repoIds;
+        this.isNeedToWhereCause = true;
+      }
+      return this;
+    }
+
+    public FindRepositoriesHQLQueryBuilder setRepoNames(List<String> repoNames) {
+      if (repoNames != null && !repoNames.isEmpty()) {
+        this.repoNames = repoNames;
+        this.isNeedToWhereCause = true;
+      }
+      return this;
+    }
+
+    public FindRepositoriesHQLQueryBuilder setPredicates(List<KeyValueQuery> predicates) {
+      if (predicates != null && !predicates.isEmpty()) {
+        this.predicates = predicates;
+        this.isNeedToWhereCause = true;
+      }
+      return this;
+    }
+
+    public FindRepositoriesHQLQueryBuilder setPageNumber(Integer pageNumber) {
+      this.pageNumber = pageNumber;
+      return this;
+    }
+
+    public FindRepositoriesHQLQueryBuilder setPageLimit(Integer pageLimit) {
+      this.pageLimit = pageLimit;
+      return this;
+    }
+
+    public FindRepositoriesQuery build() {
+      return new FindRepositoriesQuery(this);
+    }
+
+    public Query buildQuery() {
+      Query query = session.createQuery(getHQLQueryString());
+      setParameterInQuery(query);
+
+      LOGGER.debug("Final find repository query : {}", query.getQueryString());
+      if (this.pageNumber != null
+          && this.pageLimit != null
+          && this.pageNumber != 0
+          && this.pageLimit != 0) {
+        // Calculate number of documents to skip
+        int skips = this.pageLimit * (this.pageNumber - 1);
+        query.setFirstResult(skips);
+        query.setMaxResults(this.pageLimit);
+      }
+
+      LOGGER.trace("Final query generated");
+      return query;
+    }
+
+    private void setParameterInQuery(Query query) {
+      if (parametersMap.size() > 0) {
+        parametersMap.forEach(
+            (key, value) -> {
+              if (value instanceof List) {
+                List<Object> objectList = (List<Object>) value;
+                query.setParameterList(key, objectList);
+              } else {
+                query.setParameter(key, value);
+              }
+            });
+      }
+    }
+
+    public Query buildCountQuery() {
+      Query query = session.createQuery(this.countQueryString);
+      setParameterInQuery(query);
+      return query;
+    }
+
+    private String getHQLQueryString() {
+      boolean isNeedOperator = false;
+      String alias = " repo";
+      StringBuilder queryBuilder =
+          new StringBuilder(" FROM ").append(RepositoryEntity.class.getSimpleName()).append(alias);
+
+      StringBuilder joinClause = new StringBuilder();
+      if (this.predicates != null && !this.predicates.isEmpty()) {
+        this.predicates.forEach(
+            keyValueQuery -> {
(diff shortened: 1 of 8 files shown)


backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+    this.findRepositoriesHQLQuery = builder.buildQuery();
+    this.findRepositoriesCountHQLQuery = builder.buildCountQuery();
+  }
+
+  // Builder Class
+  public static class FindRepositoriesHQLQueryBuilder {
+    private static final Logger LOGGER =
+        LogManager.getLogger(FindRepositoriesHQLQueryBuilder.class);
+
+    boolean isNeedToWhereCause = false;
```
ravishetye:
```suggestion
    boolean addWhereClause = false;
```

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -835,4 +835,34 @@ public void deleteBranchByCommit(Long repoId, String commitHash) {
...
+  @Override
+  public FindRepositories.Response findRepositories(FindRepositories request)
+      throws ModelDBException {
+    try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
+      session.beginTransaction();
+      UserInfo userInfo = authService.getCurrentLoginUserInfo();
+      WorkspaceDTO workspaceDTO =
+          roleService.getWorkspaceDTOByWorkspaceName(userInfo, request.getWorkspaceName());
+      FindRepositoriesQuery findRepositoriesQuery =
+          new FindRepositoriesQuery.FindRepositoriesHQLQueryBuilder(session, workspaceDTO)
```
ravishetye, line 848: can we have a constructor instead? Is there a benefit of creating the Builder and having separate Repos

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+      return this;
+    }
+
+    public FindRepositoriesQuery build() {
+      return new FindRepositoriesQuery(this);
+    }
+
+    public Query buildQuery() {
+      Query query = session.createQuery(getHQLQueryString());
+      setParameterInQuery(query);
```
ravishetye, line 96: add short comment on what is happening at each line. specially setParameterInQuery is confusing , since I thought it was about the SQL in operator.
How About setQueryParameters(query, parametersMap) ?

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+
+      LOGGER.debug("Final find repository query : {}", query.getQueryString());
```
ravishetye: just move this log to 111, two logs is unnecessary

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    public Query buildCountQuery() {
+      Query query = session.createQuery(this.countQueryString);
+      setParameterInQuery(query);
+      return query;
+    }
+
+    private String getHQLQueryString() {
+      boolean isNeedOperator = false;
```
ravishetye:
```suggestion
      boolean appendAND = false;
```

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
ravishetye: notice the missing l in Clause

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+                joinClause
+                    .append(ModelDBConstants.LABELS)
+                    .append(".id.entity_type = ")
+                    .append(IDTypeEnum.IDType.VERSIONING_REPOSITORY.getNumber());
+              }
+            });
+      }
+
+      StringBuilder whereClause = new StringBuilder();
+      if (workspaceDTO != null
```
ravishetye, line 172: in verta system this should not happen if not present we default to user's personal workspace

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+      boolean isNeedOperator = false;
+      String alias = " repo";
+      StringBuilder queryBuilder =
+          new StringBuilder(" FROM ").append(RepositoryEntity.class.getSimpleName()).append(alias);
+
+      StringBuilder joinClause = new StringBuilder();
+      if (this.predicates != null && !this.predicates.isEmpty()) {
+        this.predicates.forEach(
+            keyValueQuery -> {
+              if (keyValueQuery.getKey().contains(ModelDBConstants.LABEL)) {
```
ravishetye, line 143: if there are two labels filtered this will create join with two tables? Would the tables need aliases?

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+      }
+
+      if (this.predicates != null && !this.predicates.isEmpty()) {
+        if (isNeedOperator) {
+          whereClause.append(" AND ");
+        }
+        for (int index = 0; index < this.predicates.size(); index++) {
+          KeyValueQuery keyValueQuery = this.predicates.get(index);
+          if (keyValueQuery.getKey().contains(ModelDBConstants.LABEL)) {
+            whereClause.append(ModelDBConstants.LABELS).append(".id.label ");
```
ravishetye: these predicates can not happen on single table. they need to be on different aliases
this is saying `Labels.id.label = "foo" and Labels,id.label = "bar"` which will never be true.
instead, it should be `L1.id.label = "foo" and L2.id.label = "bar"`
unless there is some hibernate magic happening here

if this is unclear create a repo with two tags and construct a  find with keyValuepredicates on both the tags

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+        parametersMap.put(
+            ModelDBConstants.WORKSPACE_TYPE, workspaceDTO.getWorkspaceType().getNumber());
+        isNeedOperator = true;
+      } else if (this.isNeedToWhereCause) {
+        whereClause.append(" WHERE ");
+      }
+
+      if (this.predicates != null && !this.predicates.isEmpty()) {
+        if (isNeedOperator) {
+          whereClause.append(" AND ");
```
ravishetye: Rather than this AND sprinkled at multiple places can we collect predicate strings in an array and then have a function to append them with AND?
it should take the join string as parameter so that we can use the same to construct an or condition.

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -0,0 +1,355 @@
...
+      // Order by clause
+      StringBuilder orderClause =
+          new StringBuilder(" ORDER BY")
+              .append(alias)
+              .append(".")
+              .append(ModelDBConstants.DATE_UPDATED)
+              .append(" DESC");
+
+      StringBuilder finalQueryBuilder = new StringBuilder();
+      if (!joinClause.toString().isEmpty()) {
```
ravishetye, line 221: what happens if there is no join and we add a select? is it not recommended or is outright incorrect and throw s error?

ravishetye: if it still works then i would suggest removing the if

ravishetye (changes requested):

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: this is happening with OSS.

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: - If there is a join then hibernate returns the two entities. I mean root & join entity both so at that time we should have to pass select clause like `finalQueryBuilder.append("SELECT ").append(alias).append(" ");` that means root entity so it will return only a single root entity.
- If there is not a join then in hibernate not need to add Select clause so we build a query from `new StringBuilder(" FROM ").append(RepositoryEntity.class.getSimpleName()).append(alias)` FROM clause so it will return root entity.
So above this is needed in code.

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
ravishetye: makes sense

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: If we are creating a constructor and remove the builder then the builder pattern type set parameter not works. and need to change the entire class.

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: changed

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: changed

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: Fixed logic with the aliases

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: Fixed logic

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: Added

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: Done

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
ravishetye: ok

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
```diff
@@ -157,77 +154,60 @@ private String getHQLQueryString() {
...
-                    .append(".id.entity_hash = ")
-                    .append(alias)
-                    .append(".")
-                    .append(ModelDBConstants.ID)
-                    .append(" AND ");
-                joinClause
-                    .append(joinAlias)
-                    .append(".id.entity_type = ")
-                    .append(IDTypeEnum.IDType.VERSIONING_REPOSITORY.getNumber());
+                String[] joinCauses = new String[2];
```
ravishetye: Clause

ravishetye (approved):
typo in var name but LGTM otherwise

backend/src/main/java/ai/verta/modeldb/versioning/FindRepositoriesQuery.java
anandjakhaniya: Fixed

PR #553, opened 2020-04-22
Print more information in tests

convoliution:


client/verta/tests/conftest.py
```diff
@@ -262,7 +262,10 @@ def client(host, port, email, dev_key):
 def experiment_run(client):
     client.set_project()
     client.set_experiment()
-    return client.set_experiment_run()
+    
+    run = client.set_experiment_run()
+    print("[TEST LOG] Run ID is {}".format(run.id))
```
conradoverta, line 270: Can we spend like 30min and review that we print all the basic information we'd want? I'd prefer for us to avoid having to revisit this every time we identify we don't print something that would clearly help debugging.

convoliution: Could I take that on tomorrow?

convoliution:
@convoliution It would be useful to print the time when a test starts & stops, probably in the `client` fixture.

client/verta/tests/conftest.py
conradoverta: Of course!

conradoverta (approved):

PR #562, opened 2020-04-23
fix liquebase bug

anandjakhaniya:


ravishetye (approved):

ravishetye:
did the change in precondition affect the changeset sha? I assume not and hence approved this. Let me know if it does then we might have to handle it differently

anandjakhaniya:
No, it will not affect to changeset SHA. I have verified it.

PR #568, opened 2020-04-24
find experimentruns without experiment and project

anandjakhaniya:
Proto **PR:** https://github.com/VertaAI/modeldb/pull/569


backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -1237,10 +1361,29 @@ public ExperimentRunPaginationDTO findExperimentRuns(FindExperimentRuns queryPar
...
+              queryParameters.getWorkspaceName());
+          ExperimentRunPaginationDTO experimentRunPaginationDTO = new ExperimentRunPaginationDTO();
+          experimentRunPaginationDTO.setExperimentRuns(Collections.emptyList());
+          experimentRunPaginationDTO.setTotalRecords(0L);
+          return experimentRunPaginationDTO;
+        }
+        projectIds.addAll(workspaceProjectIDs);
+      }
+
+      if (!projectIds.isEmpty()) {
```
ravishetye, line 1388: if projectIds isEmpty what does it mean? can we return early? here it seems if there is no project accessible in workspace we search across all the projects, which is wrong.

ravishetye: Also if you think this is causing https://vertaai.atlassian.net/browse/VR-4091 please assign it to yourself.

ravishetye (changes requested):
there seems to be a bug

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: @ravishetye this if is only for the safety cause because if the workspace has no project then it will return from https://github.com/VertaAI/modeldb/pull/568/files#diff-1a3287003b54e5e5b064b79266c3d82bR1371

PR #573, opened 2020-04-25
Sync protos

conradoverta:
Synchronize the protos content and generation with our private protos. This basically means that we cannot use `protos/public/...` directly and have to use `...` only. This creates a few challenges in the build, but nothing major. Changes and validations:

- python client protos changed from `from ...public.module` to `from ..module`, which is functional equivalent. Verified by importing and running a few commands.
- scala client code was broken, so I fixed. Verified we can build everything.
- backend required some changes in the proto consumption, so I aligned with UAC and fixed what was necessary. Also regenerated the autogen files and fixed changes from private protos. Everything compiles.
- go proxy and graphql compile.


backend/pom.xml
```diff
@@ -222,12 +227,12 @@
...
 				<artifactId>protobuf-maven-plugin</artifactId>
-				<version>0.5.0</version>
+				<version>0.6.1</version>
 				<configuration>
-					<protocArtifact>com.google.protobuf:protoc:3.4.0:exe:${os.detected.classifier}</protocArtifact>
+					<protocArtifact>com.google.protobuf:protoc:3.8.0:exe:${os.detected.classifier}</protocArtifact>
 					<pluginId>grpc-java</pluginId>
 					<pluginArtifact>io.grpc:protoc-gen-grpc-java:1.16.1:exe:${os.detected.classifier}</pluginArtifact>
-					<!--<protoSourceRoot>../protos</protoSourceRoot>-->
+					<protoSourceRoot>src/main/proto/protos/public</protoSourceRoot>
```
ravishetye, line 235: @ad-47 : let us know if this works on windows . we can add steps to reslve it inline or in readme. if required.

protos/scripts/fix_imports.py
```diff
@@ -13,6 +13,11 @@
 
 
 def fix_imports(python_output_dir):
+    known_dirs = set()
```
ravishetye, line 16: @convoliution : should review/be aware of changes to this file.

ravishetye (approved):

backend/pom.xml
anandjakhaniya: @conradoverta @ravishetye In the windows machine above path is not working. it should be like below for supporting on the windows machine.
```suggestion
<protoSourceRoot>../protos/protos/public</protoSourceRoot>
```

PR #576, opened 2020-04-27
VR-3974: don't upload artifact in log_training_data()

convoliution:


client/verta/verta/_internal_utils/_histogram_utils.py
```diff
@@ -0,0 +1,209 @@
...
+
+    # reject non-numbers
+    try:
+        values = list(map(float, values))
+    except ValueError:
+        raise HistogramError(
+            "values must be castable to numbers"
+        )
+
+    # reject non-integral floats
```
conradoverta, line 153: Why?

client/verta/verta/_internal_utils/_histogram_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    histogram : dict
+
+    Raises
+    ------
+    HistogramError
+        If a discrete histogram cannot be calculated from `data`.
+
+    """
+    value_counts = data.value_counts().sort_index()
+    values = value_counts.index.tolist()
```
conradoverta, line 142: I have no idea if the values will actually become the index here.

client/verta/verta/_internal_utils/_histogram_utils.py
```diff
@@ -0,0 +1,209 @@
...
+        values = list(map(float, values))
+    except ValueError:
+        raise TypeError(
+            "unable to generate histogram from non-numeric column {}".format(data.name)
+        )
+
+    # calculate bin boundaries
+    start, stop = min(values), max(values)
+    space = (stop - start)/num_bins
+    bin_boundaries = [start + space*i for i in range(num_bins+1)]
```
conradoverta: Please make sure that the last bin boundary is equal to the stop value. It will help avoid numerical issues.

client/verta/verta/_internal_utils/_histogram_utils.py
```diff
@@ -0,0 +1,209 @@
...
+
+    # fit `data` into bins
+    reference_counts = []
+    for l, r in zip(bin_boundaries[:-1], bin_boundaries[1:]):
+        count = len([value for value in values if l <= value < r])
```
conradoverta, line 219: Make sure that the last bucket includes the last value. Otherwise the total count here will be wrong.

client/verta/tests/test_deployment.py
```diff
@@ -578,6 +578,46 @@ def test_dataframe(self, experiment_run, model_for_deployment):
...
+
+        histograms = {
+            'features': {
+                'binary col': {'histogram': {'binary': {'count': [20, 10]}}, 'type': 'binary'},
+                'discrete col': {'histogram': {'discrete': {
+                    'bucket_values': [0, 1, 2],
+                    'count': [5, 10, 15]}}, 'type': 'discrete'},
+                'continuous col': {'histogram': {'float': {
+                    'bucket_limits': [0, 2.9, 5.8, 8.7, 11.6, 14.5, 17.4, 20.3, 23.2, 26.099999999999998, 29],
+                    'count': [3, 3, 3, 3, 3, 3, 3, 3, 3, 2]}}, 'type': 'float'},
```
conradoverta: `2` is really telling me that this code is wrong, since there were 30 values sent in. Don't just write tests to pass the code you wrote. Write tests to implement the functionality you want. If the code doesn't pass the tests, then it has a bug.

client/verta/tests/test_deployment.py
```diff
@@ -578,6 +578,46 @@ def test_dataframe(self, experiment_run, model_for_deployment):
...
+        endpoint = "{}://{}/api/v1/monitoring/data/references/{}".format(
+            experiment_run._conn.scheme,
+            experiment_run._conn.socket,
+            experiment_run.id,
+        )
+        response = _utils.make_request("GET", endpoint, experiment_run._conn)
+        _utils.raise_for_http_error(response)
+        generated_histograms = response.json()
+
+        assert generated_histograms == histograms
```
conradoverta: Avoid making such strong assertion when you have floats.

client/verta/tests/test_deployment.py
```diff
@@ -578,6 +578,46 @@ def test_dataframe(self, experiment_run, model_for_deployment):
...
+                    'count': [5, 10, 15]}}, 'type': 'discrete'},
+                'continuous col': {'histogram': {'float': {
+                    'bucket_limits': [0, 2.9, 5.8, 8.7, 11.6, 14.5, 17.4, 20.3, 23.2, 26.099999999999998, 29],
+                    'count': [3, 3, 3, 3, 3, 3, 3, 3, 3, 2]}}, 'type': 'float'},
+            },
+            'total_count': 30,
+        }
+
+        experiment_run.log_training_data(X, y)
+        endpoint = "{}://{}/api/v1/monitoring/data/references/{}".format(
```
conradoverta, line 740: Can you make this an unit test too? Just relying on the external system could indicate an issue in the service and not the code.

conradoverta (changes requested):

client/verta/verta/_internal_utils/_histogram_utils.py
convoliution: Because the Historical Data Processor also rejects non-integer floats. I can remove this if it's unnecessary.

client/verta/verta/_internal_utils/_histogram_utils.py
convoliution: That's what `pd.Series.value_counts()`'s return is: `.index` is the unique values from the original data, and `.values` are the counts of each unique value.

I can rename the variable `values` to `unique_values` if that would clarify this.

conradoverta: It's fine, I just didn't know.

client/verta/verta/_internal_utils/_histogram_utils.py
conradoverta: Sounds weird but sure.

client/verta/tests/test_deployment.py
convoliution: Ugh I hadn't even noticed the `2`. That's super careless of me. I've figured out better tests, now.

client/verta/tests/test_deployment.py
```diff
@@ -579,6 +580,175 @@ def test_dataframe(self, experiment_run, model_for_deployment):
...
+
+                # data within buckets
+                assert all(buckets[0] <= series)
+                assert all(series <= buckets[-1])
+
+                # counts correct
+                bin_windows = list(zip(buckets[:-1], buckets[1:]))
+                for i, (l, r) in enumerate(bin_windows[:-1]):
+                    assert sum((l <= series) & (series < r)) == counts[i]
+                assert sum(buckets[-2] <= series) == counts[-1]
```
conradoverta: This test will pass if `buckets[-1]` is always `1e9` as long as all data is smaller than that. Same for `buckets[0]`. Let's assert we found the right boundaries.

client/verta/tests/test_deployment.py
```diff
@@ -579,6 +580,175 @@ def test_dataframe(self, experiment_run, model_for_deployment):
...
+                counts = histogram_data['count']
+
+                # buckets in ascending order
+                assert buckets == list(sorted(buckets))
+
+                # data within buckets
+                assert all(buckets[0] <= series)
+                assert all(series <= buckets[-1])
+
+                # counts correct
```
conradoverta, line 650: Can we add a simple check that the total count from the buckets matches the total too? That's simpler to check than the last assert and a corner case easy to get wrong.

conradoverta (approved):

conradoverta (changes requested):
Clicked the wrong button.

client/verta/tests/test_deployment.py
convoliution: Is line 598 what you have in mind?
```python
# all data points counted
assert sum(histogram_data['count']) == len(series)
```

client/verta/tests/test_deployment.py
convoliution: I've added checks for the leftmost and rightmost boundaries.

Would it make sense to check the other buckets as well? I would just be copy&pasting the logic from the util, which I think defeats the purpose of testing functionality.

client/verta/tests/test_deployment.py
conradoverta: Yes!

client/verta/tests/test_deployment.py
conradoverta: No need to check the other ones. Just the cases that are easy to get wrong.

On a side note, you can always verify the intent of the functionality instead of copying the code. For example in this case, we could check 1) are the bucket sizes about the same? 2) do we get the number of buckets we expected? These two are the properties we want. The code in utils is just a way to get to it.

conradoverta (approved):

PR #588, opened 2020-04-29
getUsersFuzzy endpoint

anandjakhaniya:
**NOTE:** Copy protos from UAC master branch

### DIFF
--- a/protos/protos/public/uac/UACService.proto
+++ b/protos/protos/public/uac/UACService.proto
@@ -26,6 +26,23 @@ message GetUsers {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ravishetye:
@conradoverta : low priority , but what should be the flow here? should the person updating protos-all be responsible for updating oss protos as well?

ravishetye (approved):

conradoverta:
> @conradoverta : low priority , but what should be the flow here? should the person updating protos-all be responsible for updating oss protos as well?

Yes. It's better not to shift work to others.

PR #595, opened 2020-04-30
fix fuzzy owner search bug

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
```diff
@@ -146,6 +146,7 @@
   String INTERNAL_ERROR = "Internal server error";
   String NON_EQ_ID_PRED_ERROR_MESSAGE =
       "Only equality predicates supported on ids. Use EQ Operator.";
+  String USERS_NOT_FOUND_FUZZY_SEARCH = "Users not found for the fuzzy search";
```
conradoverta: We should not include implementation details in user-facing messages.

ravishetye: this is consumed internally.

conradoverta: Are you sure this will never reach the user via a status response? I'm asking because I've seen the others.

ravishetye: Currently, it will not reach the caller of MDB, But I can see if someone accidentally reuses this in some error message. we can change it to 
`String USERS_NOT_FOUND_FUZZY_SEARCH = "MDB: Users not found"; `
And this should still be the same. Would that address your concern?

conradoverta: "Users not found" alone seems reasonable.

ravishetye: we are matching on the error within MDB , so do not want to have a message similar to something with uac could throw.

ravishetye: Example https://github.com/VertaAI/modeldb/pull/595/files#diff-4d9b11fa8f1e2f3791950e90b6555560R343

conradoverta: Then maybe re-using the same class we consume to display user errors is not the right solution to the problem. Maybe we need to define specific error types (e.g. internal, external, user-friendly). Checking if a string matches the description is already something that looks weird, so maybe that is a yellow flag on the solution?

ravishetye (approved):

backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
ravishetye: noted, will address this in a follow up PR.

ravishetye: https://vertaai.atlassian.net/browse/VR-4142

PR #598, opened 2020-05-04
Fix bug: Top-level blob locations are getting duplicated into a subfolder

anandjakhaniya:


conradoverta (commented):
Please add a test for coverage

ravishetye:
> Please add a test for coverage

@ad-47 : filed https://vertaai.atlassian.net/browse/VR-4163 for the tests. I am merging this now since it is causing  client test to fail.

ravishetye (approved):

PR #607, opened 2020-05-05
RemoveResources endpoint proto

anandjakhaniya:


protos/protos/public/uac/Collaborator.proto
```diff
@@ -69,6 +69,15 @@ message GetCollaborator {
     }
 }
 
+message RemoveAllCollaborators {
```
conradoverta: Can we:
1. Move to the role service, since it's possible there are other associations that are not collaborators
2. Change the endpoint to "RemoveResource"? For example, it's possible a role is attached to the resource. This API would remove all references to the resource, which include collaborators but also other potential connections (e.g. owner, org admin).

ravishetye (approved):

protos/protos/public/uac/RoleService.proto
```diff
@@ -266,4 +275,10 @@ service RoleService {
             body : "*"
         };
     };
+
+    rpc removeResource (RemoveResource) returns (RemoveResource.Response){
+        option (google.api.http) = {
+            delete: "/v1/collaborator/removeResource"
```
ravishetye:
```suggestion
            delete: "/v1/collaborator/removeResources"
```

ravishetye (changes requested):
name change to indicate that it can support batch

protos/protos/public/uac/RoleService.proto
anandjakhaniya: Renamed

ravishetye (approved):

PR #610, opened 2020-05-05
merge conflict bug fixes

ravishetye:


backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -135,6 +135,7 @@ private static AutogenVersionEnvironmentDiff getVersionDiff(
       AutogenVersionEnvironmentDiff a,
       AutogenVersionEnvironmentDiff b,
       AutogenVersionEnvironmentBlob c) {
+    if ((a == null && b == null) || (a != null && a.equals(b))) return null;
```
ravishetye, line 138: addition like these  are resolving the actual issue reported in VR-4156

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -397,23 +400,28 @@ private static AutogenCodeDiff getCodeConflictBlob(
...
-            .setPath(
-                new AutogenPathDatasetComponentDiff()
-                    .setA(a == null || a.getPath() == null ? null : a.getPath().getB())
-                    .setB(b == null || b.getPath() == null ? null : b.getPath().getB())
-                    .setC(c == null ? null : c.getPath())
-                    .setStatus(AutogenDiffStatusEnumDiffStatus.fromProto(DiffStatus.CONFLICTED))));
+                    c == null ? null : c.getGitRepo()));
+    if (((a == null || a.getPath() == null) && (b == null || b.getPath() == null))
+        || (a != null && a.equals(b))) return Utils.removeEmpty(diff);
+    diff.setPath(
```
ravishetye, line 413: there was a bug where PathDataset would appear as a conflict when not expected

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffMerger.java
```diff
@@ -70,7 +70,10 @@ else if (!elA.contains(diffB)) {
                         // TODO: error otherwise
                         if (diffA != null && diffB != null) {
                           if (!elA.contains(diffA)) { // diff is applied on a state different from A
-                            conflictKeys.add(key);
+                            if (!elA.contains(
```
ravishetye, line 73: when the end modification was exactly the same we would still call a conflict , this fixes that case

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
conradoverta: I'm not sure I follow what this does and why it's an issue only for notebook paths.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -397,23 +400,28 @@ private static AutogenCodeDiff getCodeConflictBlob(
...
+        new AutogenPathDatasetComponentDiff()
+            .setA(a == null || a.getPath() == null ? null : a.getPath().getB())
```
conradoverta, line 415:
```suggestion
            .setA(a == null || a.getPath() == null ? null : a.getPath().getA())
```
Maybe?

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
conradoverta: Could you expand on why they resolve?

conradoverta (commented):
I am concerned about fixing just for the cases that were reported and leaving similar latent issues behind. How could we verify that is not the case? Is there a way for us to build a property test to verify the new properties we're creating?

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: during the merge attempt, we collect the locations of the conflicts. A location could have a python env blob , which has  the same python version but a conflict in requirements . this code would remove the BLOBs which are not in conflict mainly by `a.equals(b)`.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: bug was the code setting a AutogenPathDatasetComponentDiff without validation. so even if a,b,c are null a stray Status.conflict was present in the conflicts. this is not related directly to the bug you reported , but was identified during the testing

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: no these are the diffs of the commits being merged so get B is correct , getA would be the common parent

ravishetye:
> I am concerned about fixing just for the cases that were reported and leaving similar latent issues behind. How could we verify that is not the case? Is there a way for us to build a property test to verify the new properties we're creating?

actually this is not fixing just the reported cases, there are multiple issues which were identified and fixed.

That aside , I agree the merge tests can be made more targetted as well. Will address the issue in another PR and track it in jira

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
conradoverta: Sg

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
conradoverta: Wouldn't this also be an issue for the Git part of the blob?

ravishetye: the git part is handled in getGitDiff

conradoverta (approved):
LGTM

PR #615, opened 2020-05-08
IN operator with owners verta ID & fuzzy owner search for repositories

anandjakhaniya:


ravishetye:
Find is not using RBAC, this needs to be fixed. Lets fix that in a separate PR

ravishetye (approved):

PR #623, opened 2020-05-12
Account for TF 2.X in Keras integration tests

convoliution:
Apparently `batch_size` and `samples` just aren't exposed to callbacks in TF 2.X

### DIFF
--- a/client/verta/tests/test_integrations.py
+++ b/client/verta/tests/test_integrations.py
@@ -7,6 +7,8 @@ import shutil
 import tempfile
 import time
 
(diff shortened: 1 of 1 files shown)


client/verta/tests/test_integrations.py
```diff
@@ -7,6 +7,8 @@
 import tempfile
 import time
 
+from verta._internal_utils._artifact_utils import TF_MAJOR_VERSION
```
convoliution, line 10: This is defined in #621 (and so this PR is blocked until that one is merged)

PR #627, opened 2020-05-13
VR-4144: Skip spaCy models that sneak into requirements.txt

convoliution:
Filter out spaCy models in `requirements.txt` (previously, they were only filtered out when reading from the environment).

Also adds a test to verify that unsupported options are indeed being filtered out.


client/verta/verta/_internal_utils/_pip_requirements_utils.py
```diff
@@ -53,12 +53,6 @@ def get_pip_freeze():
 
     req_specs = clean_reqs_file_lines(req_specs)
 
-    # remove non-PyPI-installable SpaCy models:
-    req_specs = [
-        req_spec for req_spec in req_specs
-        if not SPACY_MODEL_REGEX.match(req_spec)
-    ]
-
```
convoliution, line 61: Logic moved to `clean_reqs_file_lines()` below.

conradoverta (approved):
LGTM

PR #631, opened 2020-05-15
VR-4127: Ignore S3 folders for versioning blob

convoliution:
Interestingly, the tests [should've failed](https://github.com/VertaAI/modeldb/blob/cf8926a/client/verta/tests/test_versioning/test_dataset.py#L16) since folders (with size 0) were being captured by the versioning utility.

~And they **do** fail on my machine, but they **succeed** in our automated tests. I've verified that it's not related to the installed `boto3` version. I have no idea what is causing the difference.~
edit 1: I've been able to reproduce this in run-pytest 281 (which I was running to test PR #630), so I'm not really sure what S3's deal is, here.
edit 2: Now our automated Client tests are failing. I wonder if—while testing this PR—I had changed the state of the metadata on our test bucket so that folders are now returned, when they weren't before?

### DIFF
--- a/client/verta/verta/dataset/_s3.py
+++ b/client/verta/verta/dataset/_s3.py
@@ -97,10 +97,15 @@ class S3(_dataset._Dataset):
         if s3_loc.key is None:
             # TODO: handle `bucket_name` not found
             for obj in s3.list_object_versions(Bucket=s3_loc.bucket)['Versions']:
+                if obj['Key'].endswith('/'):  # folder, not object
+                    continue
                 if obj['IsLatest']:
                     yield cls._get_s3_obj_metadata(obj, s3_loc.bucket, obj['Key'])
         else:
             # TODO: handle `key` not found
+            if s3_loc.key.endswith('/'):
+                s3_path = cls._S3_PATH.format(s3_loc.bucket, s3_loc.key)
+                raise ValueError("{} must be a bucket or an object, not a folder".format(s3_path))
             if s3_loc.version_id is not None:
                 # TODO: handle `version_id` not found
                 obj = s3.head_object(Bucket=s3_loc.bucket, Key=s3_loc.key, VersionId=s3_loc.version_id)
--- a/client/verta/verta/dataset/_s3.py
+++ b/client/verta/verta/dataset/_s3.py
@@ -95,11 +95,14 @@ class S3(_dataset._Dataset):
 
         if (s3_loc.key is None  # bucket
                 or s3_loc.key.endswith('/')):  # folder
-            # TODO: handle `bucket_name` and `prefix` not found
             if s3_loc.key is None:
+                # TODO: handle `bucket_name` not found
                 obj_versions = s3.list_object_versions(Bucket=s3_loc.bucket)
             else:
                 obj_versions = s3.list_object_versions(Bucket=s3_loc.bucket, Prefix=s3_loc.key)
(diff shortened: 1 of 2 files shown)


client/verta/verta/dataset/_s3.py
```diff
@@ -97,10 +97,16 @@ def _get_s3_loc_metadata(cls, s3_loc):
...
+                if obj['Key'].endswith('/'):  # folder, not object
+                    continue
+                if not obj['IsLatest']:
+                    continue
+                yield cls._get_s3_obj_metadata(obj, s3_loc.bucket, obj['Key'])
         else:
             # TODO: handle `key` not found
+            if s3_loc.key.endswith('/'):
+                s3_path = cls._S3_PATH.format(s3_loc.bucket, s3_loc.key)
+                raise ValueError("{} must be a bucket or an object, not a folder".format(s3_path))
```
conradoverta: Why wouldn't we allow a folder and then we scrape it?

convoliution: Because that's a much more complex change that I think is outside the scope of this ticket.

I can create a followup ticket to scrape folders.

convoliution: Actually, totally doable—took 20 minutes.

client/verta/verta/dataset/_s3.py
```diff
@@ -93,12 +93,23 @@ def _get_s3_loc_metadata(cls, s3_loc):
...
         s3 = boto3.client('s3')
 
-        # TODO: handle prefixes
-        if s3_loc.key is None:
-            # TODO: handle `bucket_name` not found
-            for obj in s3.list_object_versions(Bucket=s3_loc.bucket)['Versions']:
-                if obj['IsLatest']:
-                    yield cls._get_s3_obj_metadata(obj, s3_loc.bucket, obj['Key'])
+        if (s3_loc.key is None  # bucket
+                or s3_loc.key.endswith('/')):  # folder
```
conradoverta, line 97: What happens if I send a folder that doesn't end with `/`? For example, `s3://my-bucket/my-folder`. I image the `prefix` argument work work just the same. In fact, the prefix argument might work the same even if that's a file.

convoliution: Handling prefixes more broadly is already a separate TODO, and I didn't allocate bandwidth this sprint to do that. This PR was just meant to fix a bug that was causing issues downstream.

It's a _substantially_ more complicated change. If a user has a folder `foo/` and an object `foobar.csv` at the same level, then sending `"foo"` as a prefix is going to pick up that object, which is wrong. I have some ideas for how to do it, but it's not what I planned to do here.

conradoverta: Makes sense. Seems messier indeed.

conradoverta (approved):

client/verta/verta/dataset/_s3.py
conradoverta: What would happen if I provide `s3://my-bucket/my-folder` today? Will we still get NaN in the values?

convoliution: Good question!
We won't get `NaN` in the values, because both before _and after_ this fix, the Client will raise a  
```
ClientError: An error occurred (404) when calling the HeadObject operation: Not Found
```
from boto3.

Still doesn't seem like desirable behavior to me. 😦 Actually bothers me a bit, now that I've realized it. I'm going to make a ticket to resolve this folder thing next sprint.

PR #632, opened 2020-05-15
VR-3738: Fix client config merging/discovery

convoliution:
Turns out there were two issues with the implementation of Client config:

- It didn't merge config files. Instead, it only loaded the one with the highest precedence.
- It actually didn't check parent directories for config files; it checked child directories instead.
-----
This PR

- moves config file discovery logic to `_config_utils` so it can be shared between the Python Client and the CLI
- merges discovered config files
- looks in parent directories for config files
- adds much more comprehensive testing

### DIFF
--- a/client/verta/tests/conftest.py
+++ b/client/verta/tests/conftest.py
@@ -7,6 +7,7 @@ import os
 import random
 import shutil
 import string
+import tempfile
 
 import requests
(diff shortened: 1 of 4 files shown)


convoliution:
Rebasing and opening a new PR.

PR #633, opened 2020-05-15
findExperimentRun based on the experimentId and ignore project info

anandjakhaniya:
ticket: https://vertaai.atlassian.net/browse/VR-4304


conradoverta (commented):
Do we check that the user has permission to access that experiment?

anandjakhaniya:
Yes we are checking that.

ravishetye:
> Do we check that the user has permission to access that experiment?

we do that right at the beginning at https://github.com/VertaAI/modeldb/blob/master/backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java#L1841-L1847

ravishetye (approved):

PR #635, opened 2020-05-15
VR-3738: Fix client config merging/discovery

convoliution:
Example usage
```python
with _config_utils.read_merged_config() as config:
    proj_name = config.get('project', "")

with _config_utils.write_local_config() as config:
    config['project'] = proj_name
```

---
Turns out there were two issues with the implementation of Client config:

- It didn't merge config files. Instead, it only loaded the one with the highest precedence.
- It actually didn't check parent directories for config files; it checked child directories instead.

---
This PR

- moves config file discovery logic to `_config_utils` so it can be shared between the Python Client and the CLI
- merges discovered config files
- looks in parent directories for config files
- adds much more comprehensive testing


client/verta/verta/_internal_utils/_config_utils.py
```diff
@@ -72,9 +106,129 @@ def create_empty_config_file(dirpath):
...
+    should only be called after the configs have been validated against the protobuf spec.
+
+    Notes
+    -----
+    Adapted from https://stackoverflow.com/a/20666342/8651995.
+
+    """
+    for key, value in other.items():
+        if isinstance(value, dict):
+            node = accum.setdefault(key, {})
```
conradoverta, line 235: I don't know what `setdefault` does here or below.

client/verta/verta/_internal_utils/_config_utils.py
```diff
@@ -35,25 +48,46 @@ def write_config():
...
+    Yields
+    ------
+    config : dict
+        Contents of the nearest config file.
+
     """
-    # TODO: unify with Client's config methods
-    # TODO: recursively search for nearest config file
-    config_filepath = CONFIG_YAML_FILENAME
+    config_filepath = find_closest_config_file()
```
conradoverta, line 61: If the closest one is in the home folder, then you'll end up setting a global config even for local configs. I'd recommend keeping it local for almost all configs (I don't recall which configs we have).

Also, it seems like it will dump the whole merged config, right? This seems wrong, as now you can't change the global config (e.g. in home).

If this is too complicated to handle, then we can live with a single config for now.

convoliution: 1. That's a good point. How about if the nearest config is the home config, then instead create and use a new empty config in the current directory? I think it makes sense for all the items (listed [here](https://github.com/VertaAI/modeldb/blob/master/protos/protos/public/client/Config.proto))
1. This util will _not_ dump the whole merged config into the nearest config! This only gives the contents of the one nearest config file to edit; the full merged config won't be copied into it. The flow with these utils is:
    1. To get the whole merged config for reading only, use the other `read_config` util.
    1. To make config changes, close the `read_config`, stop using the merged `dict` from that util, use _this_ `write_config` util which _only_ exposes the nearest config, and edit/insert its values.

client/verta/verta/_internal_utils/_config_utils.py
convoliution: You might already know this part, but it's a method on Python dicts [[API doc](https://docs.python.org/3/library/stdtypes.html#dict.setdefault)]. It's like `dict.get(key, {})`, but it'll insert the default value into the dict at `key` before returning it, so it'll always return a reference to a value in `dict`.

So what this is doing here is, if we want to merge an dict item from `other` into `accum`, but `accum` doesn't even have that key in it yet, we'll insert an empty dict into `accum` that will be used for the merge. Same for below, but with extending a list value.

client/verta/verta/_internal_utils/_config_utils.py
conradoverta: (1) sgtm. (2) sounds very complicated. This isn't exposed to the user, right? I'd call it `write_local_config` to clarify.

convoliution: 2) It's not exposed to the user; this is only for the Client to use internally. `write_local_config()` is a good idea! I think I'd rename the read util to `read_merged_config()` as well.

client/verta/verta/_internal_utils/_config_utils.py
conradoverta: Ok, cool. I didn't know. Below we have `set_default` and here we have `setdefault`. Is that a typo?

convoliution: Yes! That is indeed a typo. It should be `setdefault`.

client/verta/verta/_internal_utils/_config_utils.py
convoliution: 1. https://github.com/VertaAI/modeldb/pull/635/commits/0a44b1130a8094051fcda3bacb8a7bf54a70c3a1
2. https://github.com/VertaAI/modeldb/pull/635/commits/53f509df0dead0db19678dfb0499a08f7aa544fd

conradoverta (approved):
LGTM

PR #640, opened 2020-05-18
Run a cron task with a frequency of seconds taking the sql query and updating the timestamp

anandjakhaniya:
**Update config:**
```
cron_job:
  update_parent_timestamp:
    frequency: 60 #frequency to update parent timestamp in second, default 60
    record_update_limit: 100
```

### DIFF
--- a/backend/config.default.yaml
+++ b/backend/config.default.yaml
@@ -68,4 +68,5 @@ feature-flag:
 
 cron_job:
   update_parent_timestamp:
-    frequency: 1 #frequency to update parent timestamp in minute, default 1
+    frequency: 60 #frequency to update parent timestamp in second, default 60
+    record_update_limit: 100
--- a/backend/config.default.yaml
+++ b/backend/config.default.yaml
@@ -65,3 +65,7 @@ migration:
 
 feature-flag:
   disabled-authz: true
+
+cron_job:
+  update_parent_timestamp:
+    frequency: 1 #frequency to update parent timestamp in minute, default 1
--- a/backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
+++ b/backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
@@ -198,7 +198,7 @@ public interface ModelDBConstants {
   // Telemetry Constants
   String TELEMETRY = "telemetry";
   String OPT_IN = "opt_in";
-  String FREQUENCY = "frequency"; // frequency to share data in hours
+  String TELEMENTRY_FREQUENCY = "frequency"; // frequency to share data in hours
   String TELEMETRY_CONSUMER = "consumer";
   String TELEMETRY_CONSUMER_URL =
       "https://app.verta.ai/api/v1/uac-proxy/telemetry/collectTelemetry";
@@ -223,6 +223,8 @@ public interface ModelDBConstants {
   String REPOSITORY = "repository";
 
   // Cron job constant
+  String FREQUENCY = "frequency"; // frequency to run cron job in second
+  String RECORD_UPDATE_LIMIT = "record_update_limit";
   String CRON_JOB = "cron_job";
   String UPDATE_PARENT_TIMESTAMP = "update_parent_timestamp";
 }
--- a/backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
+++ b/backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
@@ -198,7 +198,7 @@ public interface ModelDBConstants {
   // Telemetry Constants
   String TELEMETRY = "telemetry";
   String OPT_IN = "opt_in";
-  String TELEMENTRY_FREQUENCY = "frequency"; // frequency to share data in hours
+  String FREQUENCY = "frequency"; // frequency to share data in hours
   String TELEMETRY_CONSUMER = "consumer";
   String TELEMETRY_CONSUMER_URL =
       "https://app.verta.ai/api/v1/uac-proxy/telemetry/collectTelemetry";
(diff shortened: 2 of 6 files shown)


backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
```diff
@@ -198,7 +198,7 @@
   // Telemetry Constants
   String TELEMETRY = "telemetry";
   String OPT_IN = "opt_in";
-  String TELEMENTRY_FREQUENCY = "frequency"; // frequency to share data in hours
+  String FREQUENCY = "frequency"; // frequency to share data in hours
```
conradoverta: Let's keep this variable and use another one. They have very different meanings in the two contexts right now, as the comment itself says.

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
```diff
@@ -0,0 +1,113 @@
...
+
+  public ParentTimestampUpdateCron() {}
+
+  /** The action to be performed by this timer task. */
+  @Override
+  public void run() {
+    LOGGER.info("ParentTimestampUpdateCron wakeup");
+
+    try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
+      session.beginTransaction();
```
conradoverta, line 25: Don't create long transactions. Also, none of these SQL commands even require a transaction.

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
```diff
@@ -0,0 +1,113 @@
...
+      // Update project timestamp
+      updateProjectByExperimentTimestamp(session);
+
+      // Update experiment timestamp
+      updateExperimentByExperimentRunTimestamp(session);
```
conradoverta: Reverse the order. First update the lower elements of the hierarchy, then the top ones.

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
```diff
@@ -0,0 +1,113 @@
...
+
+      // Update experiment timestamp
```
conradoverta: Remove or update comment.

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
```diff
@@ -0,0 +1,113 @@
...
+      updateDatasetByDatasetVersionTimestamp(session);
+
+      // Update experiment timestamp
```
conradoverta: Remove or update comment.

conradoverta (commented):
Queries look ok, but please check that we have indices on all the affected operations (grouping, joining and filtering). I don't think we do.

backend/config.default.yaml
```diff
@@ -65,3 +65,7 @@ migration:
 
 feature-flag:
   disabled-authz: true
+
+cron_job:
+  update_parent_timestamp:
+    frequency: 1 #frequency to update parent timestamp in minute, default 1
```
conradoverta: Can we make the config be seconds by default? It will give us more flexibility.

conradoverta (commented):
Actually, can we add a LIMIT in the update requests? I want to avoid locking the database for too long. I'd prefer to have the option to update a small batch more frequently than a large batch less frequently.

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
anandjakhaniya: These SQL commands are the update operation so for the hibernate we can not execute update operation without the transaction. but ye definitely I can reduce transaction scope and create a new one for each.

conradoverta: I'm not sure that is correct. I believe hibernate works without transactions. In any case, as long as we put the `LIMIT`, doing a transaction per set of elements seems like a fair trade.

backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
anandjakhaniya: constant reverted and created a new one for frequency

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
anandjakhaniya: Added `LIMIT` in the query.

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
anandjakhaniya: Fixed

backend/config.default.yaml
anandjakhaniya: Change config with the second and LIMIT where default value of the same is 60s & 100 respectively.

anandjakhaniya:
Fixed all review comments with adding indexes, LIMIT in the query, and other review comment fixes.

conradoverta (approved):
LGTM.

PR #643, opened 2020-05-18
Enable Multipart Upload from Client

convoliution:
# DO NOT MERGE
Until https://github.com/VertaAI/modeldb/pull/642 is merged and protos are regenerated; otherwise the Client will be broken.

### DIFF
--- a/client/verta/verta/client.py
+++ b/client/verta/verta/client.py
@@ -2074,8 +2074,16 @@ class ExperimentRun(_ModelDBEntity):
                 # get presigned URL
                 url = self._get_url_for_artifact(key, "PUT", part_num=part_num).url
 
+                # wrap file part into bytestream to avoid OverflowError
+                #     Passing a bytestring >2 GB (num bytes > max val of int32) directly to
+                #     ``requests`` will overwhelm CPython's SSL lib when it tries to sign the
+                #     payload. But passing a buffered bytestream instead of the raw bytestring
+                #     indicates to ``requests`` that it should perform a streaming upload via
+                #     HTTP/1.1 chunked transfer encoding and avoid this issue.
+                #     https://github.com/psf/requests/issues/2717
+                part_stream = six.BytesIO(file_part)
+
                 # upload part
-                part_stream = six.BytesIO(file_part)  # enables streaming by `requests`, otherwise overwhelms SSL
                 response = _utils.make_request("PUT", url, self._conn, data=part_stream)
                 response.raise_for_status()
 
--- a/client/verta/verta/client.py
+++ b/client/verta/verta/client.py
@@ -2041,25 +2041,39 @@ class ExperimentRun(_ModelDBEntity):
 
         self._upload_artifact(key, artifact_stream, response_msg.multipart_upload_ok)
 
-    def _upload_artifact(self, key, artifact_stream, use_multipart):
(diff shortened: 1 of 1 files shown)


convoliution:
@conradoverta
The Client currently retries on `502`, `503`, `504` up to five times with exponential backoff for a total of ~30 sec.

Do you think that should be changed/increased for the multipart upload loop?

client/verta/verta/client.py
```diff
@@ -2038,16 +2045,77 @@ def _log_artifact(self, key, artifact, artifact_type, extension=None, method=Non
...
+            print("[DEBUG] uploading {} bytes ({})".format(len(artifact_stream.read()), key))
             artifact_stream.seek(0)
 
-        response = _utils.make_request("PUT", url, self._conn, data=artifact_stream)
-        _utils.raise_for_http_error(response)
-        print("upload complete ({})".format(basename))
+        # check if multipart upload ok
+        try:
+            url_for_artifact = self._get_url_for_artifact(key, "PUT", part_num=1)
+        except requests.HTTPError:  # backend knows it doesn't support multipart upload, e.g. NFS
```
conradoverta: It shouldn't raise an exception. It should handle it gracefully. We added a flag to the protos specifically for this.

convoliution: Makes sense to me. I'll check on that in the backend

client/verta/verta/client.py
```diff
@@ -2038,16 +2045,71 @@ def _log_artifact(self, key, artifact, artifact_type, extension=None, method=Non
...
+
+        if url_for_artifact.multipart_upload_ok:
+            # TODO: parallelize this
+            file_parts = iter(lambda: artifact_stream.read(part_size), b'')
+            for part_num, file_part in enumerate(file_parts, start=1):
+                # get presigned URL
+                url = self._get_url_for_artifact(key, "PUT", part_num=part_num).url
+
+                # upload part
+                part_stream = six.BytesIO(file_part)  # enables streaming by `requests`, otherwise overwhelms SSL
```
conradoverta: Can you expand on your comment? Seems like something that we should be more fully aware, so good to take note in the code.

client/verta/verta/client.py
```diff
@@ -2038,16 +2045,71 @@ def _log_artifact(self, key, artifact, artifact_type, extension=None, method=Non
...
+            )
+            msg = _CommonService.CommitMultipartArtifact(id=self.id, key=key)
+            data = _utils.proto_to_json(msg)
+            response = _utils.make_request("POST", url, self._conn, json=data)
+            _utils.raise_for_http_error(response)
+        else:
+            # upload full artifact
+            response = _utils.make_request("PUT", url_for_artifact.url, self._conn, data=artifact_stream)
+            _utils.raise_for_http_error(response)
+            # TODO: check artifact length, suggest size problem to user
```
conradoverta: Not sure that's something we want to do. If it's large, then this will be slow. Also, the backend should handle it transparently.

conradoverta (approved):
A couple of changes on the comments alone.

client/verta/verta/client.py
convoliution: Based on further digging, this is actually only a problem when uploading >2 GB at once, but yes I'll definitely expand the comment!

PR #651, opened 2020-05-20
Set delete flag to true and implement cron job for delete entities

anandjakhaniya:
**Config Change:** 
```
mdb_service_user:
  email:
  devKey:

delete_entities:
```

**May have possible Thread concurrency in code:**

- There is a thread concurrency occurred between the fetched entity in endpoint DAO transactions, parent entity time update, and delete entities cron jobs. so we should have to find some proper way to overcome it.

### DIFF
--- a/backend/config.default.yaml
+++ b/backend/config.default.yaml
@@ -70,3 +70,5 @@ cron_job:
   update_parent_timestamp:
     frequency: 60 #frequency to update parent timestamp in second, default 60
     record_update_limit: 100
+  delete_entities:
+    frequency: 60 #frequency to delete entities in second, default 60
--- a/backend/config.default.yaml
+++ b/backend/config.default.yaml
@@ -6,6 +6,10 @@ springServer:
   port: 8086
   shutdownTimeout: 30 #time in second
 
+mdb_service_user:
+  email:
+  devKey:
+
 artifactStoreConfig:
   artifactStoreType: NFS #S3, GCP, NFS
   S3:
--- a/backend/src/main/java/ai/verta/modeldb/authservice/AuthServiceChannel.java
+++ b/backend/src/main/java/ai/verta/modeldb/authservice/AuthServiceChannel.java
@@ -32,7 +32,9 @@ public class AuthServiceChannel implements AutoCloseable {
   private UACServiceGrpc.UACServiceBlockingStub uacServiceBlockingStub;
   private TeamServiceGrpc.TeamServiceBlockingStub teamServiceBlockingStub;
   private OrganizationServiceGrpc.OrganizationServiceBlockingStub organizationServiceBlockingStub;
-  public static boolean isMigrationUtilsCall = false;
+  public static boolean isBackgroundUtilsCall = false;
+  private String serviceUserEmail;
+  private String serviceUserDevKey;
 
   public AuthServiceChannel() {
     App app = App.getInstance();
@@ -44,6 +46,9 @@ public class AuthServiceChannel implements AutoCloseable {
           ManagedChannelBuilder.forTarget(host + ModelDBConstants.STRING_COLON + port)
               .usePlaintext()
               .build();
+
+      this.serviceUserEmail = app.getServiceUserEmail();
+      this.serviceUserDevKey = app.getServiceUserDevKey();
     } else {
       Status status =
           Status.newBuilder()
@@ -56,9 +61,15 @@ public class AuthServiceChannel implements AutoCloseable {
 
   private Metadata getMetadataHeaders() {
     Metadata requestHeaders;
-    if (isMigrationUtilsCall && ModelDBAuthInterceptor.METADATA_INFO.get() == null) {
-      Metadata.Key<String> source_key = Metadata.Key.of("source", Metadata.ASCII_STRING_MARSHALLER);
+    if (isBackgroundUtilsCall && ModelDBAuthInterceptor.METADATA_INFO.get() == null) {
       requestHeaders = new Metadata();
+      Metadata.Key<String> email_key = Metadata.Key.of("email", Metadata.ASCII_STRING_MARSHALLER);
+      Metadata.Key<String> dev_key =
+          Metadata.Key.of("developer_key", Metadata.ASCII_STRING_MARSHALLER);
+      Metadata.Key<String> source_key = Metadata.Key.of("source", Metadata.ASCII_STRING_MARSHALLER);
+
+      requestHeaders.put(email_key, this.serviceUserEmail);
+      requestHeaders.put(dev_key, this.serviceUserDevKey);
       requestHeaders.put(source_key, "PythonClient");
     } else {
       requestHeaders = ModelDBAuthInterceptor.METADATA_INFO.get();
--- a/backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
+++ b/backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
@@ -41,10 +41,14 @@ public class CronJobUtils {
             Map<String, Object> deleteEntitiesCronMap = (Map<String, Object>) cronJob.getValue();
             deleteEntitiesFrequency =
                 (int) deleteEntitiesCronMap.getOrDefault(ModelDBConstants.FREQUENCY, 60);
+            int recordUpdateLimit =
+                (int) deleteEntitiesCronMap.getOrDefault(ModelDBConstants.RECORD_UPDATE_LIMIT, 100);
             // creating an instance of task to be scheduled
-            TimerTask task = new DeleteEntitiesCron(authService, roleService);
+            TimerTask task = new DeleteEntitiesCron(authService, roleService, recordUpdateLimit);
             ModelDBUtils.scheduleTask(task, deleteEntitiesFrequency, TimeUnit.SECONDS);
             LOGGER.info("{} cron job scheduled successfully", ModelDBConstants.DELETE_ENTITIES);
+          } else {
+            LOGGER.warn("Unknown config key ({}) found for the cron job", cronJob.getKey());
           }
         }
       }
--- a/backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
+++ b/backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
@@ -1,6 +1,8 @@
 package ai.verta.modeldb.cron_jobs;
 
 import ai.verta.modeldb.ModelDBConstants;
+import ai.verta.modeldb.authservice.AuthService;
+import ai.verta.modeldb.authservice.RoleService;
 import ai.verta.modeldb.utils.ModelDBUtils;
 import java.util.Map;
 import java.util.TimerTask;
@@ -11,8 +13,10 @@ import org.apache.logging.log4j.Logger;
 public class CronJobUtils {
   private static final Logger LOGGER = LogManager.getLogger(CronJobUtils.class);
   public static Integer updateParentTimestampFrequency = 60;
+  public static Integer deleteEntitiesFrequency = 60;
 
-  public static void initializeBasedOnConfig(Map<String, Object> propertiesMap) {
+  public static void initializeBasedOnConfig(
+      Map<String, Object> propertiesMap, AuthService authService, RoleService roleService) {
     LOGGER.info("Enter in CronJobUtils: initializeBasedOnConfig()");
     if (propertiesMap.containsKey(ModelDBConstants.CRON_JOB)) {
       Map<String, Object> cronJobMap =
@@ -33,6 +37,14 @@ public class CronJobUtils {
             ModelDBUtils.scheduleTask(task, updateParentTimestampFrequency, TimeUnit.SECONDS);
             LOGGER.info(
                 "{} cron job scheduled successfully", ModelDBConstants.UPDATE_PARENT_TIMESTAMP);
+          } else if (cronJob.getKey().equals(ModelDBConstants.DELETE_ENTITIES)) {
+            Map<String, Object> deleteEntitiesCronMap = (Map<String, Object>) cronJob.getValue();
+            deleteEntitiesFrequency =
+                (int) deleteEntitiesCronMap.getOrDefault(ModelDBConstants.FREQUENCY, 60);
+            // creating an instance of task to be scheduled
+            TimerTask task = new DeleteEntitiesCron(authService, roleService);
+            ModelDBUtils.scheduleTask(task, deleteEntitiesFrequency, TimeUnit.SECONDS);
+            LOGGER.info("{} cron job scheduled successfully", ModelDBConstants.DELETE_ENTITIES);
           }
         }
       }
--- a/backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
+++ b/backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
@@ -613,6 +613,7 @@ public class DeleteEntitiesCron extends TimerTask {
                 }
               }
             });
+        session.delete(repository);
         transaction.commit();
       }
     }
--- a/backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
+++ b/backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
@@ -1,5 +1,7 @@
 package ai.verta.modeldb.cron_jobs;
 
+import static ai.verta.modeldb.authservice.AuthServiceChannel.isBackgroundUtilsCall;
+
 import ai.verta.modeldb.DatasetVisibilityEnum;
 import ai.verta.modeldb.ModelDBConstants;
 import ai.verta.modeldb.ProjectVisibility;
@@ -35,12 +37,15 @@ public class DeleteEntitiesCron extends TimerTask {
   private static final Logger LOGGER = LogManager.getLogger(DeleteEntitiesCron.class);
   private final AuthService authService;
   private final RoleService roleService;
+  private final Integer recordUpdateLimit;
   private static final String DATASET_GLOBAL_SHARING = "_DATASET_GLOBAL_SHARING";
   private static final String REPOSITORY_GLOBAL_SHARING = "_REPO_GLOBAL_SHARING";
 
-  public DeleteEntitiesCron(AuthService authService, RoleService roleService) {
+  public DeleteEntitiesCron(
+      AuthService authService, RoleService roleService, Integer recordUpdateLimit) {
     this.authService = authService;
     this.roleService = roleService;
+    this.recordUpdateLimit = recordUpdateLimit;
   }
 
   /** The action to be performed by this timer task. */
@@ -48,6 +53,7 @@ public class DeleteEntitiesCron extends TimerTask {
   public void run() {
     LOGGER.info("DeleteEntitiesCron wakeup");
 
+    isBackgroundUtilsCall = true;
     try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
       // Update project timestamp
       deleteProjects(session);
@@ -67,9 +73,10 @@ public class DeleteEntitiesCron extends TimerTask {
       // Update repository timestamp
       deleteRepositories(session);
     } catch (Exception ex) {
+      ex.printStackTrace();
       LOGGER.error("DeleteEntitiesCron Exception: ", ex);
     }
-
+    isBackgroundUtilsCall = false;
     LOGGER.info("DeleteEntitiesCron finish tasks and reschedule");
   }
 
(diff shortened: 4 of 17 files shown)


anandjakhaniya:
There is a thread concurrency occurred between `parent entity time update` and `delete entities` cron jobs. so we should have to find some proper way to overcome it.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
```diff
@@ -33,6 +37,14 @@ public static void initializeBasedOnConfig(Map<String, Object> propertiesMap) {
...
                 "{} cron job scheduled successfully", ModelDBConstants.UPDATE_PARENT_TIMESTAMP);
+          } else if (cronJob.getKey().equals(ModelDBConstants.DELETE_ENTITIES)) {
+            Map<String, Object> deleteEntitiesCronMap = (Map<String, Object>) cronJob.getValue();
+            deleteEntitiesFrequency =
+                (int) deleteEntitiesCronMap.getOrDefault(ModelDBConstants.FREQUENCY, 60);
+            // creating an instance of task to be scheduled
+            TimerTask task = new DeleteEntitiesCron(authService, roleService);
+            ModelDBUtils.scheduleTask(task, deleteEntitiesFrequency, TimeUnit.SECONDS);
+            LOGGER.info("{} cron job scheduled successfully", ModelDBConstants.DELETE_ENTITIES);
           }
```
conradoverta, line 52: Can we add a warn log if the config has an unknown key here? It will make debugging much easier.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,657 @@
...
+
+    int lowerBound = 0;
+    final int pagesize = 500;
+    Query countQuery =
+        session.createQuery("SELECT COUNT(" + alias + ") " + deleteProjectsQueryString);
+    countQuery.setParameter("deleted", true);
+    Long count = (Long) countQuery.uniqueResult();
+    LOGGER.debug("Total projectEntities {} for deletion", count);
+
+    while (lowerBound < count) {
```
conradoverta: This whole process seems pretty error prone. First you query the DB to get X project IDs, then you issue the deletion of each individually, then you issue update of experiment deletion with the X project IDs, then you delete the role bindings. And this sounds like it should be in a transaction if done in this way.

What about the following:
1. run request to get project IDs
2. delete role bindings. We do it at this step so that we can retry later if we get an error from UAC
3. run one request to update the experiments to set `deleted=1` based on the IDs we got before 
4. run one request to delete the projects based on the IDs we got before

In this order, no transactions are necessary since failure in any step before the last will leave the project with `deleted=1` there to be retried later.

The same logic is applied to the other ones below. The order of the steps remain the same. Always ask the question: if the system crashes at any particular step, can I restart safely? Right now, the answer is "no" unless we make a big transaction.

Side note: you don't need the loop with `lowerBound`. Just remove a configurable number of entries, just like we did on the update. Every time we'll get closer to the final state of having the things actually deleted, but we don't necessarily need to reach that in a single step, which might place a higher burden on the database.

conradoverta (commented):
Some comments in the update section.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
anandjakhaniya: Added warn log for the unknown key

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Change code as per the above suggestion

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -389,40 +366,16 @@ private void deleteRoleBindingsOfAccessibleResources(List<RepositoryEntity> allo
...
+      int updatedCount = deletedRepositoriesQuery.executeUpdate();
+      LOGGER.debug(
+          "Mark Repositories as deleted : {}, count : {}", allowedRepositoryIds, updatedCount);
       // Delete all VersionedInputs for repository ID
       experimentRunDAO.deleteLogVersionedInputs(session, repository.getId(), null);
-
-      deleteRoleBindingsOfAccessibleResources(Collections.singletonList(repository));
+      transaction.commit();
       session.delete(repository);
       session.getTransaction().commit();
```
conradoverta: Looks like this is committing the transaction twice, right?

backend/src/main/java/ai/verta/modeldb/authservice/AuthServiceChannel.java
```diff
@@ -32,7 +32,9 @@
   private UACServiceGrpc.UACServiceBlockingStub uacServiceBlockingStub;
   private TeamServiceGrpc.TeamServiceBlockingStub teamServiceBlockingStub;
   private OrganizationServiceGrpc.OrganizationServiceBlockingStub organizationServiceBlockingStub;
-  public static boolean isMigrationUtilsCall = false;
+  public static boolean isBackgroundUtilsCall = false;
```
conradoverta, line 35: What is this? It seems like it's a single global variable shared by everything using the auth service for calls, which is a **big** no-go. We should not use the service account permissions for non-service account requests.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+              ModelResourceEnum.ModelDBServiceResourceTypes.PROJECT,
+              ProjectVisibility.forNumber(project.getProject_visibility())
+                  .equals(ProjectVisibility.ORG_SCOPED_PUBLIC),
+              "_GLOBAL_SHARING");
+      roleBindingNames.addAll(workspaceRoleBindingNames);
+    }
+
+    roleService.deleteAllResources(
+        allowedProjects.stream().map(ProjectEntity::getId).collect(Collectors.toList()),
+        ModelResourceEnum.ModelDBServiceResourceTypes.PROJECT);
```
conradoverta: Move this somewhere else. This is very sneaky by calling a major delete inside a function named `getRoleBindingsOfAccessibleProjects`. It should only *get* role bindings.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+              .append(" = :deleted ")
+              .append(" WHERE expr.")
+              .append(ModelDBConstants.PROJECT_ID)
+              .append(" IN (:projectIds)")
+              .toString();
+      Query deletedExperimentRunQuery =
+          session.createQuery(updateDeletedStatusExperimentRunQueryString);
+      deletedExperimentRunQuery.setParameter("deleted", true);
+      deletedExperimentRunQuery.setParameter("projectIds", projectIds);
+      deletedExperimentRunQuery.executeUpdate();
```
conradoverta: Experiment run deletion should chain from experiment deletion. Let the hierarchy handle itself one step at a time.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+
+      for (ProjectEntity projectEntity : projectEntities) {
+        session.delete(projectEntity);
+      }
```
conradoverta, line 130: Instead of issuing N database commands, can we just issue a single delete request? You already have the IDs for what you want to delete.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+        experimentIds.add(experimentEntity.getId());
+        String ownerRoleBindingName =
+            roleService.buildRoleBindingName(
+                ModelDBConstants.ROLE_EXPERIMENT_OWNER,
+                experimentEntity.getId(),
+                experimentEntity.getOwner(),
+                ModelResourceEnum.ModelDBServiceResourceTypes.EXPERIMENT.name());
+        if (ownerRoleBindingName != null) {
+          roleBindingNames.add(ownerRoleBindingName);
+        }
```
conradoverta: Be consistent. This is a separate function for projects, so make it a separate function for experiments too.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+              .toString();
+      Query deletedExperimentRunQuery =
+          session.createQuery(updateDeletedStatusExperimentRunQueryString);
+      deletedExperimentRunQuery.setParameter("deleted", true);
+      deletedExperimentRunQuery.setParameter("experimentIds", experimentIds);
+      deletedExperimentRunQuery.executeUpdate();
+
+      for (ExperimentEntity experimentEntity : experimentEntities) {
+        session.delete(experimentEntity);
+      }
```
conradoverta, line 241: Same comment as for projects.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+        experimentRunIds.add(experimentRunEntity.getId());
+        String ownerRoleBindingName =
+            roleService.buildRoleBindingName(
+                ModelDBConstants.ROLE_EXPERIMENT_RUN_OWNER,
+                experimentRunEntity.getId(),
+                experimentRunEntity.getOwner(),
+                ModelResourceEnum.ModelDBServiceResourceTypes.EXPERIMENT_RUN.name());
+        if (ownerRoleBindingName != null) {
+          roleBindingNames.add(ownerRoleBindingName);
+        }
```
conradoverta: Same.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+
+      Transaction transaction = session.beginTransaction();
+      // Delete the ExperimentRun comments
+      if (!experimentRunIds.isEmpty()) {
+        removeEntityComments(session, experimentRunIds, ExperimentRunEntity.class.getSimpleName());
+      }
+
+      for (ExperimentRunEntity experimentRunEntity : experimentRunEntities) {
+        session.delete(experimentRunEntity);
+      }
```
conradoverta, line 298: Same.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+        }
+      }
+      if (!roleBindingNames.isEmpty()) {
+        roleService.deleteRoleBindings(roleBindingNames);
+      }
+
+      Transaction transaction = session.beginTransaction();
+      // Delete the ExperimentRun comments
+      if (!experimentRunIds.isEmpty()) {
+        removeEntityComments(session, experimentRunIds, ExperimentRunEntity.class.getSimpleName());
```
conradoverta, line 293: Shouldn't this happen for the other entities too? I believe I saw this deleted in more than one place, but I could be mistaken.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+            .append(" =:entityName")
+            .toString();
+    Query commentDeleteQuery = session.createQuery(commentDeleteHql);
+    commentDeleteQuery.setParameterList("entityIds", entityIds);
+    commentDeleteQuery.setParameter("entityName", entityName);
+    LOGGER.debug("Comments delete query : {}", commentDeleteQuery.getQueryString());
+    List<CommentEntity> commentEntities = commentDeleteQuery.list();
+    for (CommentEntity commentEntity : commentEntities) {
+      session.delete(commentEntity);
+    }
```
conradoverta, line 342: Issue a single DB command.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+              .toString();
+      Query deletedDatasetVersionQuery =
+          session.createQuery(updateDeletedStatusDatasetVersionQueryString);
+      deletedDatasetVersionQuery.setParameter("deleted", true);
+      deletedDatasetVersionQuery.setParameter("datasetIds", datasetIds);
+      deletedDatasetVersionQuery.executeUpdate();
+
+      for (DatasetEntity datasetEntity : datasetEntities) {
+        session.delete(datasetEntity);
+      }
```
conradoverta, line 392: Same.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+                  datasetEntity.getDataset_visibility()));
+      if (!workspaceRoleBindingNames.isEmpty()) {
+        roleBindingNames.addAll(workspaceRoleBindingNames);
+      }
+    }
+
+    // Remove all datasetEntity collaborators
+    roleService.deleteAllResources(
+        allowedDatasets.stream().map(DatasetEntity::getId).collect(Collectors.toList()),
+        ModelResourceEnum.ModelDBServiceResourceTypes.DATASET);
```
conradoverta: Same.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+    }
+    // Remove all role bindings
+    if (!roleBindingNames.isEmpty()) {
+      roleService.deleteRoleBindings(roleBindingNames);
+    }
+
+    Transaction transaction = session.beginTransaction();
+    for (DatasetVersionEntity datasetVersionEntity : datasetVersionEntities) {
+      session.delete(datasetVersionEntity);
+    }
```
conradoverta: Same.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+    for (DatasetVersionEntity datasetVersionEntity : datasetVersionEntities) {
+      String ownerRoleBindingName =
+          roleService.buildRoleBindingName(
+              ModelDBConstants.ROLE_DATASET_VERSION_OWNER,
+              datasetVersionEntity.getId(),
+              datasetVersionEntity.getOwner(),
+              ModelResourceEnum.ModelDBServiceResourceTypes.DATASET_VERSION.name());
+      if (ownerRoleBindingName != null && !ownerRoleBindingName.isEmpty()) {
+        roleBindingNames.add(ownerRoleBindingName);
+      }
```
conradoverta, line 545: Same.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+            commitEntity -> {
+              if (commitEntity.getRepository().contains(repository)) {
+                commitEntity.getRepository().remove(repository);
+                if (commitEntity.getRepository().isEmpty()) {
+                  session.delete(commitEntity);
+                } else {
+                  session.update(commitEntity);
+                }
+              }
+            });
```
conradoverta, line 625: Commits can exist in more than one repo. The cleanup for them needs to be done more carefully. Just delete the things that are definitely associated with the repo.

conradoverta (commented):
1. be consistent on how the code is written and divided
2. don't sneak deletes in functions named get
3. don't perform N database operations when you can perform one operation on the N items at the same time
4. the global flag in the auth class seems like a huge security hole if it goes through

backend/config.default.yaml
```diff
@@ -6,6 +6,10 @@ springServer:
   port: 8086
   shutdownTimeout: 30 #time in second
 
+mdb_service_user:
```
ravishetye, line 9: not sure how this is being used . some updates in PR description would be great.

ravishetye: how does this work for no auth mode?

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -0,0 +1,666 @@
...
+    this.roleService = roleService;
+    this.recordUpdateLimit = recordUpdateLimit;
+  }
+
+  /** The action to be performed by this timer task. */
+  @Override
+  public void run() {
+    LOGGER.info("DeleteEntitiesCron wakeup");
+
+    isBackgroundUtilsCall = true;
```
ravishetye, line 56: so for the execution of this function auth service will act as isBackgroundUtilsCall, for other calls as well?

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: No, we can not do it in a single delete query. if we do like that it will not delete sub entity list (Not support casecade deletion). So i have used `session.delete(projectEntitiy)` for cascade deletion provided by hibernate

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Comment: https://github.com/VertaAI/modeldb/pull/651#discussion_r429177879

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Same Comment: https://github.com/VertaAI/modeldb/pull/651#discussion_r429177879

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: `removeEntityComments` is a delete `comments` entities. which is only associated with the ExperimentRun entity so it should be fine here. and it is only in a single place.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Same Comment: https://github.com/VertaAI/modeldb/pull/651#discussion_r429177879

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Same Comment: https://github.com/VertaAI/modeldb/pull/651#discussion_r429177879

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Same Comment: https://github.com/VertaAI/modeldb/pull/651#discussion_r429177879

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Yes, the commit will exist in multiple repositories that why I have handled that flow here. if you see line 607, I have checked that commits have a deleted repository. if yes then I am removing it from commit mapping because that commit may have second associated repository. if that commit doesn't have any repository then I will delete commit otherwise just update commit with removing deleted repository from its mapping.

backend/config.default.yaml
ad-47: In No Auth mode, MDB ignore this config settings

anandjakhaniya: In No Auth mode, MDB ignore this config setting

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: Yes, Fixed

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Created separate functions called `deleteROleBindingsFor*` and moved all delete role bindings code in that function for each entity.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Removed it.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Created separate functions called `deleteROleBindingsFor*` and moved all delete role bindings code in that function for each entity.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Created separate functions called `deleteROleBindingsFor*` and moved all delete role bindings code in that function for each entity.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Created separate functions called `deleteRoleBindingsFor*` and moved all delete role bindings code in that function for each entity.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Created separate functions called `deleteRoleBindingsFor*` and moved all delete role bindings code in that function for each entity.

ravishetye (approved):

PR #657, opened 2020-05-21
VR-4345 Update database for new Artifact fields

evgeniilezhnin:


convoliution (approved):
Makes sense to me.
Adding @ravishetye as reviewer to double-check.

backend/src/main/resources/liquibase/db-changelog-1.0.xml
```diff
@@ -231,6 +231,47 @@
...
+    <changeSet id="artifact_table_update_upload_completed" author="lezhevg">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="artifact" columnName="upload_completed"/>
+            </not>
+        </preConditions>
+        <addColumn tableName="artifact">
+            <column name="upload_completed" type="boolean" defaultValue="true" />
+        </addColumn>
+    </changeSet>
```
ravishetye, line 254: given that this is not deployed yet, can be edit the original changeset instead of add columns?

evgeniilezhnin: Do you mean create_artifact changeset? I don't think that it will work well in some cases. When I add columns to create_artifact changeset and try to drop and create the database I get artifact table from createTable-1 changeset, but not from create_artifact changeset, so new columns are missing. That may create some problems. So maybe I should use changesets to add new columns to it? Besides, that's the way it's done in tables-4 changeset.

conradoverta: From my understanding, we need to add 2 columns to the pre-existing `artifacts` table and create a new table `artifact_part`. Am I missing something here @ravishetye ?

ravishetye: this makes sense to me, I misread the chageset

ravishetye (approved):

PR #660, opened 2020-05-22
VR-4348 VR-4349 VR-4350 commit artifact part, get commited part, commit parts

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAO.java
```diff
@@ -505,4 +509,13 @@ ExperimentRun deleteArtifacts(String experimentRunId, String atrifactKey)
   Entry<String, String> getExperimentRunArtifactsS3PathAndMultipartUploadID(
       String experimentRunId, String key, long partNumber, S3KeyFunction initializeMultipart)
       throws ModelDBException;
+
+  Response commitArtifactPart(CommitArtifactPart request) throws ModelDBException;
+
+  GetCommittedArtifactParts.Response getCommitedArtifactParts(GetCommittedArtifactParts request)
```
convoliution:
```suggestion
  GetCommittedArtifactParts.Response getCommittedArtifactParts(GetCommittedArtifactParts request)
```

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,54 +2261,151 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
...
+              artifactEntity, artifactPart.getPartNumber(), artifactPart.getEtag());
+      session.beginTransaction();
+      session.saveOrUpdate(artifactPartEntity);
+      session.getTransaction().commit();
+      return Response.newBuilder().build();
+    }
+  }
+
+  @Override
+  public GetCommittedArtifactParts.Response getCommitedArtifactParts(
```
convoliution:
```suggestion
  public GetCommittedArtifactParts.Response getCommittedArtifactParts(
```

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -2360,4 +2363,91 @@ public void getVersionedInputs(
...
+      } else if (request.getKey().isEmpty()) {
+        errorMessage = "Artifact key not found in CommitArtifactPart request";
+      }
+
+      if (errorMessage != null) {
+        throw new ModelDBException(errorMessage, io.grpc.Status.Code.INVALID_ARGUMENT);
+      }
+
+      GetCommittedArtifactParts.Response response =
+          experimentRunDAO.getCommitedArtifactParts(request);
```
convoliution:
```suggestion
          experimentRunDAO.getCommittedArtifactParts(request);
```

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -2360,4 +2363,91 @@ public void getVersionedInputs(
...
+      GetCommittedArtifactParts request,
+      StreamObserver<GetCommittedArtifactParts.Response> responseObserver) {
+    QPSCountResource.inc();
+    try (RequestLatencyResource latencyResource =
+        new RequestLatencyResource(ModelDBAuthInterceptor.METHOD_NAME.get())) {
+      String errorMessage = null;
+      if (request.getId().isEmpty()) {
+        errorMessage = "ExperimentRun ID not found in CommitArtifactPart request";
+      } else if (request.getKey().isEmpty()) {
+        errorMessage = "Artifact key not found in CommitArtifactPart request";
```
convoliution:
```suggestion
        errorMessage = "ExperimentRun ID not found in GetCommittedArtifactParts request";
      } else if (request.getKey().isEmpty()) {
        errorMessage = "Artifact key not found in GetCommittedArtifactParts request";
```

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -2360,4 +2363,91 @@ public void getVersionedInputs(
...
+      CommitMultipartArtifact request,
+      StreamObserver<CommitMultipartArtifact.Response> responseObserver) {
+    QPSCountResource.inc();
+    try (RequestLatencyResource latencyResource =
+        new RequestLatencyResource(ModelDBAuthInterceptor.METHOD_NAME.get())) {
+      String errorMessage = null;
+      if (request.getId().isEmpty()) {
+        errorMessage = "ExperimentRun ID not found in CommitArtifactPart request";
+      } else if (request.getKey().isEmpty()) {
+        errorMessage = "Artifact key not found in CommitArtifactPart request";
```
convoliution:
```suggestion
        errorMessage = "ExperimentRun ID not found in CommitMultipartArtifact request";
      } else if (request.getKey().isEmpty()) {
        errorMessage = "Artifact key not found in CommitMultipartArtifact request";
```

convoliution (changes requested):
@evgeniilezhnin Based on my testing, it looks like you are removing `artifact_part`s from the database after Client calls `/commitMultipartArtifact`. Is that correct?

cc @conradoverta @ravishetye Do you think `artifact_part`s should remain in the database? The removal is unexpected to me, but maybe there isn't a good reason to keep them.

Screen Shot 2020-05-26 at 2 26 07 PM

---
~edit: This is fine. There's no reason to keep the `artifact_part`s right now.~
edit 2: actually, this would help immensely with an integration test I'm working on

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/S3Service.java
```diff
@@ -94,11 +100,32 @@ public String generatePresignedUrl(String s3Key, String method, long partNumber,
         new GeneratePresignedUrlRequest(bucketName, s3Key)
             .withMethod(reqMethod)
             .withExpiration(expiration);
-    if (partNumber != 0) {
-      request.addRequestParameter("partNumber", String.valueOf(partNumber));
-      request.addRequestParameter("uploadId", uploadId);
-    }
+    request.addRequestParameter("partNumber", String.valueOf(partNumber));
+    request.addRequestParameter("uploadId", uploadId);
```
convoliution: @evgeniilezhnin What was the reason for this change?
Based on my testing, this adds `"uploadId=&partNumber=0"` to the presigned URL when Client sends `part_number: 0`, which causes an error from S3 during upload.

I recommend reverting. Or at least make sure `"uploadId=&partNumber=0"` is not added to the presigned URL.

convoliution (changes requested):

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/S3Service.java
evgeniilezhnin: There wasn't this check in feature/VR-4346_log_initialize_multipart , after that I added it, but forgot to add it in feature/VR-4348_commit_artifact_part . So this change is not needed here, I'll remove it

evgeniilezhnin: done

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,54 +2261,151 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
...
+          } else {
+            uploadId = initializeMultipart.apply(artifactEntity.getPath()).orElse(null);
+          }
+        }
+        if (uploadId == null) {
+          message = "Can't initialize multipart upload";
+        }
+        if (message != null) {
+          LOGGER.info(message);
+          throw new ModelDBException(message, io.grpc.Status.Code.FAILED_PRECONDITION);
```
convoliution, line 2341: When Client sends `part_number` but multipart is not supported:
Instead of raising an error, could we change this to return a normal presigned URL (not multipart)?
And make sure `multipart_upload_ok` is false in the response. Thanks!

(based on my testing, it looks like we could just remove this one line, but I'd like confirmation from a backend eng)

convoliution (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: fixed. There's 2 validation checks: one is used during GetUrlForArtifact, another in commitMultipartArtifact.

I have removed the one for GetUrlForArtifact

convoliution: Makes sense to me!

convoliution (approved):

backend/src/main/resources/liquibase/db-changelog-1.0.xml
```diff
@@ -266,7 +266,7 @@
                   referencedColumnNames="id"/>
             </column>
             <column name="part_number" type="int8">
-                <constraints nullable="false"/>
+                <constraints primaryKey="true" nullable="false"/>
```
conradoverta: Move this to the PR doing liquibase changes because it seems like the original one was buggy.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -2360,4 +2363,91 @@ public void getVersionedInputs(
           responseObserver, e, GetVersionedInput.Response.getDefaultInstance());
     }
   }
+
+  @Override
+  public void commitArtifactPart(
```
conradoverta, line 2368: Missing authorization check.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -2360,4 +2363,91 @@ public void getVersionedInputs(
...
+      responseObserver.onNext(response);
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      ModelDBUtils.observeError(
+          responseObserver, e, CommitArtifactPart.Response.getDefaultInstance());
+    }
+  }
+
+  @Override
+  public void getCommittedArtifactParts(
```
conradoverta, line 2396: Missing authorization check.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -2360,4 +2363,91 @@ public void getVersionedInputs(
...
+      responseObserver.onNext(response);
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      ModelDBUtils.observeError(
+          responseObserver, e, GetCommittedArtifactParts.Response.getDefaultInstance());
+    }
+  }
+
+  @Override
+  public void commitMultipartArtifact(
```
conradoverta, line 2424: Missing authorization check.

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/S3Service.java
```diff
@@ -101,4 +107,27 @@ public String generatePresignedUrl(String s3Key, String method, long partNumber,
...
+    }
+    CompleteMultipartUploadRequest completeMultipartUploadRequest =
+        new CompleteMultipartUploadRequest(bucketName, s3Key, uploadId, partETags);
+    try {
+      CompleteMultipartUploadResult result =
+          s3Client.completeMultipartUpload(completeMultipartUploadRequest);
+      LOGGER.info("upload result: {}", result);
+    } catch (AmazonS3Exception e) {
+      if (e.getStatusCode() == HttpStatusCodes.STATUS_CODE_BAD_REQUEST) {
+        LOGGER.info("message: {} additional details: {}", e.getMessage(), e.getAdditionalDetails());
```
conradoverta, line 127: This doesn't sound like an `info` to me. An actual error happened, right?

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,54 +2261,148 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
     }
   }
 
+  private Optional<ArtifactEntity> getExperimentRunArtifact(
+      Session session, String experimentRunId, String key) {
+    ExperimentRunEntity experimentRunObj = session.get(ExperimentRunEntity.class, experimentRunId);
+    if (experimentRunObj == null) {
+      LOGGER.warn(ModelDBMessages.EXP_RUN_NOT_FOUND_ERROR_MSG);
```
conradoverta: Is the `experimentRunId` and `key` coming from the user? If so, then a bad user input is not a warning for us.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
conradoverta: @convoliution to re-check

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,54 +2261,148 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
...
+          throw new ModelDBException(message, io.grpc.Status.Code.FAILED_PRECONDITION);
+        }
+      } catch (ModelDBException e) {
         Status status =
             Status.newBuilder()
-                .setCode(Code.NOT_FOUND_VALUE)
-                .setMessage(ModelDBMessages.EXP_RUN_NOT_FOUND_ERROR_MSG)
+                .setCode(Code.INVALID_ARGUMENT_VALUE)
+                .setMessage(e.getMessage())
+                .addDetails(Any.pack(CreateJob.Response.getDefaultInstance()))
```
conradoverta: This seems wrong. There are many things that could cause an exception, especially since we're calling `initializeMultipart`. Just saying it's the user's fault seems wrong.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,54 +2261,148 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
...
+      partETags =
+          artifactPartEntities.stream()
+              .map(ArtifactPartEntity::toPartETag)
+              .collect(Collectors.toList());
+      session.beginTransaction();
+      artifactEntity.setUploadCompleted(true);
+      artifactPartEntities.forEach(session::delete);
+      artifactPartEntities.clear();
+      session.getTransaction().commit();
+      result = getS3PathAndMultipartUploadId(session, artifactEntity, true, request.getKey(), null);
```
conradoverta: `getS3PathAndMultipartUploadId` has `artifactEntity.setUploadCompleted(false);`. Seems like calling it here is a bug (or at least something that will probably be a bug in the future).

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,54 +2261,148 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
...
     }
+    commitMultipartFunction.apply(result.getKey(), result.getValue(), partETags);
```
conradoverta: Commit and then save the database. Otherwise you just lost data.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
convoliution: I verified that 90e8007c126a1ff3af1387ebfd7faadb3342737f does the expected behavior (returns valid presigned URL and `multipart_upload_ok = False` w/ NFS setup)

backend/src/main/resources/liquibase/db-changelog-1.0.xml
evgeniilezhnin: moved

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/S3Service.java
evgeniilezhnin: No. That is for errors like: "Your proposed upload is smaller than the minimum allowed size".
It looks like a validation condition that can be checked only during multipart commit call.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: I think I implemented some complex code here. I have made some refactoring because really this exception can happen only during multipart commit. So it can happen if 1. get multipart URL wasn't called before commit multipart. 2. get multipart URL was called for NFS.

So 1 looks like the user's fault, right? But should 2 have some other type of response?

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: what I wanted to do with getS3PathAndMultipartUploadId is to just get the information from artifactEntity. But it can be done without getS3PathAndMultipartUploadId. made a refactoring to not use it.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: fixed

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
evgeniilezhnin: added

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
evgeniilezhnin: added

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
evgeniilezhnin: added

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
conradoverta: Yeah, I recommend removing the specialized part you want to use in two places instead of calling a wrapper that does more than what you want, especially when it's risky.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
conradoverta: We should never hit (2), right? We just ignore the part info from the user and send in the flag back saying we don't support.

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/S3Service.java
conradoverta: Ok, got it. Let's make sure we return something clear and actionable to the user. An s3 error message is not that, since it's an implementation detail of how we store data.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,52 +2262,173 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
...
+              .setMessage(ModelDBMessages.EXP_RUN_NOT_FOUND_ERROR_MSG)
+              .build();
+      throw StatusProto.toStatusRuntimeException(status);
+    }
+
+    String projectId = experimentRunObj.getProject_id();
+
+    // Validate if current user has access to the entity or not
+    roleService.validateEntityUserWithUserInfo(
+        ModelDBServiceResourceTypes.PROJECT, projectId, modelDBServiceActions);
```
conradoverta, line 2286: Always check for permission first. You already returned information to the user.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2251,52 +2262,173 @@ public void deleteLogVersionedInputs(Session session, Long repoId, String commit
...
+
+  @Override
+  public CommitMultipartArtifact.Response commitMultipartArtifact(
+      CommitMultipartArtifact request, CommitMultipartFunction commitMultipartFunction)
+      throws ModelDBException, InvalidProtocolBufferException {
+    List<PartETag> partETags;
+    try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
+      ArtifactEntity artifactEntity =
+          getArtifactEntity(
+              session, request.getId(), request.getKey(), ModelDBServiceActions.UPDATE);
```
conradoverta, line 2413: This will involve a grpc to check permission. You should not do that while holding a db session.

conradoverta (commented):
I'm ok with following-up with the two changes after.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: I guess. But the flag is in GetUrlForArtifact. So it's still possible to call CommitMultipartArtifact after that.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: https://github.com/VertaAI/modeldb/pull/691

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: https://github.com/VertaAI/modeldb/pull/691

PR #665, opened 2020-05-25
remove unique constraint from liquibase indexes for name

anandjakhaniya:


backend/src/main/resources/liquibase/db-changelog-1.0.xml
```diff
@@ -127,10 +127,11 @@
             </not>
         </preConditions>
         <createIndex indexName="p_name_wspace_wspace_type"
-                     tableName="project" unique="true">
+                     tableName="project">
             <column name="name" type="varchar(255)"/>
             <column name="workspace" type="varchar(255)"/>
             <column name="workspace_type" type="int"/>
+            <column name="deleted" type="boolean"/>
```
ravishetye, line 155: why is this change required?

ravishetye (changes requested):
we will have to drop and recreate the index, since this will now modify the index already created.
we will have to do it in a new changeset

anandjakhaniya:
Updated indexes as per review comment

ravishetye:
can you pull in master, I do not expect so many changes

ravishetye (approved):

PR #668, opened 2020-05-26
retry current request on "Communications link failure" in MDB

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/batchProcess/OwnerRoleBindingRepositoryUtils.java
```diff
@@ -124,6 +125,12 @@ private static void migrateRepositories() {
 
         transaction.commit();
         lowerBound += pagesize;
+      } catch (Exception ex) {
```
ravishetye, line 128: not related directly to the PR but can you please add comments on the top of the files in batchProcess/* on what they are doing?

anandjakhaniya: Sure. it makes sense.

ravishetye (commented):
we spoke about the wrapper around session operators, did that not work?

anandjakhaniya:
No, it does not work because each and every `session.*` operations need to retry with new sessions after the old session invalidates so for `session.update`, `session.save` etc. need to open transaction after acquiring new connection and it would create multiple transactions in many endpoints DAO.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
```diff
@@ -480,6 +482,28 @@ public static Throwable findRootCause(Throwable throwable) {
     responseObserver.onError(statusRuntimeException);
   }
 
+  public static boolean needToRetry(Exception ex) {
```
ravishetye, line 484: we need to also cover retrying on Unavailable for UAC right?

anandjakhaniya: `Unavailable for UAC` is a separate thing, we will apply the same logic in AuthService & RoleService for the same.

ravishetye:
not all of mdb/uac APIs will be safe to be retried.  This would have open gate for other bugs which we would be required to fix. Is there anything better we could do in that area?

cc @conradoverta

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
```diff
@@ -480,6 +481,32 @@ public static Throwable findRootCause(Throwable throwable) {
...
+    Throwable communicationsException = findCommunicationsFailedCause(ex);
+    if ((communicationsException.getCause() instanceof CommunicationsException)
+        || (communicationsException.getCause() instanceof SocketException)) {
+      LOGGER.warn(communicationsException.getMessage());
+      if (ModelDBHibernateUtil.ping()) {
+        ModelDBHibernateUtil.resetSessionFactory();
+      }
```
anandjakhaniya: @ravishetye added this new check to reset the session factory

ravishetye (approved):
LGTM , but I will test it more,before merging it.

evgeniilezhnin (approved):

ravishetye:
Screen Shot 2020-06-02 at 8 35 48 PM

ravishetye:
@ad-47 once you fix the NPE , I can merge it.
Good Job.

PR #689, opened 2020-06-01
driver loading bug fix

ravishetye:


backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -181,10 +181,10 @@ public static SessionFactory createOrGetSessionFactory() throws ModelDBException
             (Map<String, Object>) databasePropMap.get("RdbConfiguration");
 
         databaseName = (String) rDBPropMap.get("RdbDatabaseName");
-        if (app.getTraceEnabled()) {
+        if (!app.getTraceEnabled()) {
           rDBDriver = (String) rDBPropMap.get("RdbDriver");
         } else {
-          rDBDriver = (String) rDBPropMap.get("io.opentracing.contrib.jdbc.TracingDriver");
+          rDBDriver = "io.opentracing.contrib.jdbc.TracingDriver";
```
anandjakhaniya, line 187: We can not replace `rDBDriver` because it is used in many places to check DB type. Ex: https://github.com/VertaAI/modeldb/blob/master/backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java#L709 so we should keep it as it is and create a variable for `"io.opentracing.contrib.jdbc.TracingDriver"`

PR #692, opened 2020-06-01
Fix find query with NOT_CONTAIN Operator

anandjakhaniya:


ravishetye (approved):
merge after adding test cases where dataset and dataset version exist together and has find is on not contains

PR #695, opened 2020-06-01
Scala client

nhatsmrt:


client/scala/src/main/scala/ai/verta/client/package.scala
```diff
@@ -0,0 +1,32 @@
+package ai.verta
+
+import scala.concurrent.ExecutionContext
+import java.net.URLEncoder
+
+package object client {
```
conradoverta, line 6: Avoid package objects. It becomes a dumpster for all kinds of random things. Put things in meaningful locations with meaningful names.

client/scala/src/main/scala/ai/verta/client/Client.scala
```diff
@@ -54,4 +56,51 @@ class Client(conn: ClientConnection) {
         }
       }))
   }
+
+  /** Get the repository by name (and workspace). If not exist, create new repository
+   * @param name Name of the Repository
+   * @param workspace Workspace under which the Repository with name name exists. If not provided, the current user’s personal workspace will be used.
+   */
+  def getOrCreateRepository(name: String, workspace: String = null)(implicit ec: ExecutionContext) = {
```
conradoverta, line 64: Don't use `null` for optional arguments. Use `Option`. `null` references are a curse

client/scala/src/main/scala/ai/verta/client/Client.scala
```diff
@@ -54,4 +56,51 @@ class Client(conn: ClientConnection) {
...
+
+  /** Get repository based on id
+   *  @param id id of the repository
+   *  @return the repository
+   */
+  def getRepository(id: String)(implicit ec: ExecutionContext): Try[Repository] = {
+    clientSet.versioningService.GetRepository2(
+      id_repo_id = BigInt(id)
+    )
+    .map(r => if (r.repository.isEmpty) null else new Repository(clientSet, r.repository.get))
```
conradoverta, line 94: Don't return `null`. Return an option of the object or a `Failure` with an exception.

client/scala/src/main/scala/ai/verta/client/Client.scala
```diff
@@ -54,4 +56,51 @@ class Client(conn: ClientConnection) {
...
+  }
+
+
+  /** Get repository based on id
+   *  @param id id of the repository
+   */
+  def deleteRepository(id: String)(implicit ec: ExecutionContext): Try[VersioningDeleteRepositoryRequestResponse] = {
```
conradoverta, line 101: Can you return something more meaningful to the user that doesn't require them to understand our protos? For example, what does the response contain that could be useful?

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -0,0 +1,389 @@
...
+        case Some(v) => List(v)
+        case None => commit.parent_shas.get
+      }
+
+      ids.map(id => load_blobs_from_id(id))
+      loaded_from_remote = true
+    }
+  }
+
+  /** Retrieve blobs associated to commit with given id and update blobs
```
conradoverta, line 37: `and update blobs` seems like an implementation detail, not a documentation to the user

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -0,0 +1,389 @@
...
+   *  @param id id of the commit
+   */
+  private def load_blobs_from_id(id: String)(implicit ec: ExecutionContext): Try[List[Option[VersioningBlob]]] = {
```
conradoverta, line 40: Why is there an `Option` here? I was expecting just `Try[List[VersioningBlob]]`.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -0,0 +1,389 @@
...
+    clientSet.versioningService.ListCommitBlobs2(
+      commit_sha = id,
+      repository_id_repo_id = repo.id.get
+    ) // Try[VersioningListCommitBlobsRequestResponse]
+    .map(_.blobs) // Try[Option[List[VersioningBlobExpanded]]]
+    .map(ls => if (ls.isEmpty) null else ls.get.map(
```
conradoverta, line 46: Don't return `null`

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -0,0 +1,389 @@
...
+  var loaded_from_remote = false // whether blob has been retrieved from remote
+  var blobs = new HashMap[String, VersioningBlob]() // mutable map for storing blobs
+
+  /** Retrieve commit's blobs from remote
+   */
+  private def load_blobs()(implicit ec: ExecutionContext): Unit = {
+    if (!loaded_from_remote) {
+      val ids: List[String] = commit.commit_sha match {
+        case Some(v) => List(v)
+        case None => commit.parent_shas.get
```
conradoverta, line 29: This seems wrong to me. @convoliution do we load all parents?

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -0,0 +1,389 @@
...
+      }
+
+      ids.map(id => load_blobs_from_id(id))
```
conradoverta, line 32: You are ignoring any return from `load_blobs_from_id`. What happens if it has an exception? Always return `Try` for functions that can have exceptions.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -0,0 +1,389 @@
...
+  /** Return ancestors, starting from this Commit until the root of the Repository
+   *  @return a list of ancestors
+   */
+  def log()(implicit ec: ExecutionContext): Try[List[Commit]] = {
+    clientSet.versioningService.ListCommitsLog4(
+      repository_id_repo_id = repo.id.get,
+      commit_sha = commit.commit_sha.get
+    ) // Try[VersioningListCommitsLogRequestResponse]
+    .map(_.commits) // Try[Option[List[VersioningCommit]]]
+    .map(ls => if (ls.isEmpty) null else ls.get.map(c => new Commit(clientSet, repo, c))) // Try[List[Commit]]
```
conradoverta, line 103: Don't return `null`.

conradoverta (commented):
General feedback:
- avoid `null`
- make things private when they shouldn't be accessed by external source
- be consistent with names. Avoid snake case mixed with camel case
- every request that can raise an exception should return a `Try`

client/scala/src/main/scala/ai/verta/repository/Commit.scala
convoliution: Yes; if the `Commit` doesn't have its own ID yet, the Python Client [iterates through the parents and loads in all of their blobs](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_repository/commit.py#L62).

But note that this is (or should be) only the immediate parents, not the entire branch history.

client/scala/src/main/scala/ai/verta/client/Client.scala
nhatsmrt: Hmm for that one the idea is that user should only care about whether the delete operation succeeds or fails (the "VersioningDeleteRepositoryRequestResponse" object is useless, and it's empty anyway).

conradoverta:
@nhatsmrt can we close this? We're breaking this PR into multiple ones.

PR #697, opened 2020-06-02
initial delay to start the cron jobs

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/App.java
```diff
@@ -636,7 +636,7 @@ public static void initializeTelemetryBasedOnConfig(Map<String, Object> properti
     if (optIn) {
       // creating an instance of task to be scheduled
       TimerTask task = new TelemetryCron(consumer);
-      ModelDBUtils.scheduleTask(task, frequency, TimeUnit.HOURS);
+      ModelDBUtils.scheduleTask(task, frequency, frequency, TimeUnit.HOURS);
```
ravishetye, line 639: we need to send frequency twice?

ravishetye (changes requested):
there is a minor issue, please verify
other than that  I assume you have verified that the parameter works as expected.

backend/src/main/java/ai/verta/modeldb/App.java
anandjakhaniya: before these changes, intialDelay and frequency were the same https://github.com/VertaAI/modeldb/pull/697/files#diff-678c1a74bf0f4a56c36bdfba5112c87eL413 So I have changed it for other cron jobs as per our 5 min default constant talk excluding TelemetryCron. for TelemetryCron I have taken as it is like an old one.

ravishetye (approved):
LGTM

PR #698, opened 2020-06-02
getUrlForArtifact should check if current user has access to the modeldb entity

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experiment/ExperimentServiceImpl.java
```diff
@@ -1128,6 +1128,21 @@ public void getUrlForArtifact(
         throw StatusProto.toStatusRuntimeException(status);
       }
 
+      Map<String, String> projectIdFromExperimentMap =
+          experimentDAO.getProjectIdsByExperimentIds(Collections.singletonList(request.getId()));
+      if (projectIdFromExperimentMap.size() == 0) {
+        errorMessage = "Access is denied. Experiment not found for given id : " + request.getId();
+        ModelDBUtils.logAndThrowError(
```
ravishetye, line 1135: this should be a not found? you are not checking for privileges here

ravishetye (changes requested):
IMO needs a change in error code

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentServiceImpl.java
anandjakhaniya: Fixed

ravishetye (approved):

ravishetye:
@ad-47 : any particular reason this is not merged?

PR #700, opened 2020-06-02
added `internal_versioned_path` field in path_dataset_component_blob

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
```diff
@@ -1293,4 +1301,323 @@ private Boolean blobTypeExistsInList(List<BlobType> blobTypeList, Blob.ContentCa
...
+      RepositoryEntity repositoryEntity = repositoryFunction.apply(session);
+      CommitEntity commitEntity = commitFunction.apply(session, session1 -> repositoryEntity);
+
+      Map<String, Map.Entry<BlobExpanded, String>> locationBlobWithHashMap =
+          getCommitBlobMapWithHash(
+              session,
+              commitEntity.getRootSha(),
+              request.getLocationList(),
+              Collections.singletonList(BlobType.DATASET_BLOB));
+      String presignedUrl =
```
ravishetye: could we just pass the value to the function instead of entire HashMap? This way we can also pull in the validation sooner and keep the function more generic

ravishetye (commented):
minor changes , but the PR seems ready for @convoliution 's testing

backend/src/main/java/ai/verta/modeldb/versioning/BlobDAORdbImpl.java
anandjakhaniya: Fixed as per comment

convoliution (changes requested):
When I complete a multipart upload of `"s3://ml-us-east1-test/tiny.txt"`, and then try to upload the same file (maybe different commit, different repo, doesn't matter): backend tries to re-use the old `UploadId` from the existing `UploadStatusEntity`.

Because this `UploadId` was already completed, S3 returns an error with the presigned URL, saying the upload no longer exists.

Backend shouldn't re-use the old `UploadId` after `/commitVersionedBlobArtifactPart` has been called for it.

protos/protos/public/modeldb/versioning/VersioningService.proto
```diff
@@ -479,7 +479,7 @@ message GetUrlForBlobVersioned {
 
     message Response {
         string url = 1;
-        bool multipart_upload_ok = 3;  // for backcompat: new Client, old Backend
+        bool multipart_upload_ok = 2;  // for backcompat: new Client, old Backend
```
convoliution: I actually picked `3` intentionally to match the proto [in `CommonService`](https://github.com/VertaAI/modeldb/blob/c98b1fe/protos/protos/public/modeldb/CommonService.proto#L95)!
And I have already regenerated the protos for this, so I think it may be too late to change it to `2` anyway 😦 
```suggestion
        bool multipart_upload_ok = 3;  // for backcompat: new Client, old Backend
```

convoliution (changes requested):

protos/protos/public/modeldb/versioning/VersioningService.proto
anandjakhaniya: Ok, No issue. I can change it will works because the backend does not depend on it.

backend/src/main/resources/liquibase/db-changelog-1.0.xml
```diff
@@ -445,7 +445,69 @@
...
+            <column name="etag" type="varchar(255)"/>
+            <column name="artifact_type" type="int4" defaultValue="0"><!-- defaultValue="0" = EXP_RUN_ARTIFACT-->
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet id="copy_temp_to_new_artifact_part_table" author="anandJ">
+        <sql>
+            INSERT INTO artifact_part SELECT CAST(artifact_id AS CHAR) artifact_id, part_number, etag,"0"  FROM artifact_part_temp;
```
ravishetye: @ad-47 : this needs to be `'` quotes

convoliution:
```suggestion
            INSERT INTO artifact_part SELECT CAST(artifact_id AS CHAR) artifact_id, part_number, etag, '0' FROM artifact_part_temp;
```

convoliution (changes requested):
I am okay with this PR being merged after this one change.

backend/src/main/resources/liquibase/db-changelog-1.0.xml
anandjakhaniya: Fixed

convoliution (approved):

ravishetye (approved):

PR #701, opened 2020-06-02
get presigned url for versioned blob

anandjakhaniya:


backend/src/main/resources/liquibase/db-changelog-1.0.xml
```diff
@@ -534,6 +535,24 @@
...
+                <tableExists tableName="upload_status"/>
+            </not>
+        </preConditions>
+        <createTable tableName="upload_status">
+            <column name="id" type="int8" autoIncrement="true" incrementBy="1">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="path_dataset_component_blob_id" type="varchar(64)"/>
+            <column name="s3_dataset_component_blob_id" type="varchar(64)"/>
+            <column name="component_blob_type" type="int4"/>
```
anandjakhaniya, line 550: identify column `dataset_component_blob_id` based on the `component_blob_type` instead of two column for ids.

PR #703, opened 2020-06-02
Update the API so that the query map accept list of string instead.

nhatsmrt:


client/scala/src/main/scala/ai/verta/swagger/client/HttpClient.scala
```diff
@@ -21,14 +21,14 @@ class HttpClient(val host: String, val headers: Map[String, String]) {
     URLEncoder.encode(s, "UTF-8")
   }
 
-  private def urlEncodeUTF8(q: Map[String, String]): String = {
+  private def urlEncodeUTF8(q: Map[String, List[String]]): String = {
     if (q.isEmpty) "" else
       q
-        .map(entry => urlEncodeUTF8(entry._1) + "=" + urlEncodeUTF8(entry._2))
+        .map(entry => urlEncodeUTF8(entry._1) + "=" + entry._2.map(urlEncodeUTF8 _).mkString("&" + urlEncodeUTF8(entry._1) + "="))
```
conradoverta: This looks pretty confusing to me due to the duplication. Let me know if this functionality would be equivalent:
```
.flatmap(entry => entry._2.map(x => urlEncodeUTF8(entry._1) + "=" + urlEncodeUTF8(x)))
```
and let the line below handle the `&`.

nhatsmrt: yep it should. I've also tested this in my original scala-client branch and it works for walk (which is the purpose of this change). Resolve.

conradoverta (approved):

PR #705, opened 2020-06-02
Remove unnecessary print message

convoliution:
Method already prints `"upload complete"` on line 2150.
Still need `print()` to go to the next line, because the print in the loop used carriage return.


conradoverta (approved):
Is this py2 compatible?

convoliution:
@conradoverta Yep

All these prints are Python2 compatible, with the
```python
from __future__ import print_function
```
at the top of the file

PR #706, opened 2020-06-03
VR-4523: Read artifact in chunks

convoliution:


conradoverta (approved):
This PR is digging a deeper and deeper hole. I really don't understand the hacks you're using, but I'm assuming you have at least validated it works.

PR #708, opened 2020-06-03
VR-4565 add repository access modifier database

evgeniilezhnin:


ravishetye (approved):
this is fine,since the refactor is minor. however if you think there is any change which should go in master sooner , please have it on a different PR so that it goes to production sooner

PR #709, opened 2020-06-03
VR-4566 check protected

evgeniilezhnin:


ravishetye (commented):
My bad if the ticket /doc was not clear but the point was no APIs from VersioningServiceImpl should be able to process Protected Repos , not just the GetByID

evgeniilezhnin:
> My bad if the ticket /doc was not clear but the point was no APIs from VersioningServiceImpl should be able to process Protected Repos , not just the GetByID

all endpoints from VersioningServiceImpl (commit, tags) use GetByID. Let me recheck

evgeniilezhnin:
> > My bad if the ticket /doc was not clear but the point was no APIs from VersioningServiceImpl should be able to process Protected Repos , not just the GetByID
> 
> all endpoints from VersioningServiceImpl (commit, tags) use GetByID. Let me recheck

Metadata doesn't access repositories. But all commits, branches, repositories endpoints will get needed repository information by calling GetByID

ravishetye (approved):

PR #712, opened 2020-06-03
Implement Commit's tag and branch operations for Scala client

nhatsmrt:
"saved" is currently not used for these operations, but is crucial for operations involving saving/loading blobs. I'm keeping it here for now (along with the check inside tag and newBranch).


client/scala/src/main/scala/ai/verta/utils/URLUtils.scala
```diff
@@ -0,0 +1,10 @@
+package ai.verta.utils
+
+import java.net.URLEncoder
+
+object URLUtils {
+  /** Utility method for HTML form encoding
+   *  @param input input
+   */
+  def urlEncode(input: String): String = URLEncoder.encode(input, "UTF-8").replaceAll("\\+", "%20")
```
conradoverta: Let's put this in the client as I think everything has access to it. Having generic `utils` folders tends to make them into kitchen sinks with everything.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -21,4 +23,45 @@ class Commit(
...
+        tag = URLUtils.urlEncode(tag)
+    ).map(_ => ())
+  }
+
+  /** Set the commit of named branch to current commit
+   *  @param branch branch
+   */
+  private def setBranch(branch: String)(implicit ec: ExecutionContext) = {
+    clientSet.versioningService.SetBranch2(
+      body = "\"" + commit.commit_sha.get + "\"",
```
conradoverta: Use Scala formatting (either inline or not) instead to make this more readable.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -21,4 +23,45 @@ class Commit(
...
+  }
+
+  /** Assigns a tag to this Commit
+   *  @param tag tag
+   */
+  def tag(tag: String)(implicit ec: ExecutionContext) = {
+    if (!saved)
+      Failure(new IllegalStateException("Commit must be saved before it can be tagged"))
+    else clientSet.versioningService.SetTag2(
+        body = "\"" + commit.commit_sha.get + "\"",
```
conradoverta: Same

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -21,4 +23,45 @@ class Commit(
...
+
+  /** Creates a branch at this Commit and returns the checked-out branch
+   *  If the branch already exists, it will be moved to this commit.
+   *  @param branch branch name
+   *  @return if not saved, a failure; otherwise, this commit as the head of `branch`
+   */
+  def newBranch(branch: String)(implicit ec: ExecutionContext) = {
+    if (!saved)
+      Failure(new IllegalStateException("Commit must be saved before it can be attached to a branch"))
+    else setBranch(branch)
```
conradoverta: @convoliution this seems different from the client, right? We decided to return a new commit or something like that since overwriting was confusing.

client/scala/src/main/scala/ai/verta/utils/URLUtils.scala
nhatsmrt: That makes sense. I'll fix that.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: You mean the string formatting/interpolation? I think previously I was having some issue because mixing that will the double quotes is kinda buggy. I'll give it another try.

nhatsmrt: Is this what you're referring to:
https://github.com/scala/bug/issues/6476
It's a really long conversation, but the issue is still open at the end.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: I think you're right actually. Will fix this, and reflect in the tests as well.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
conradoverta: Hm. Let me put in another way. If you send a string to `body`, it should serialize as a string in json, right? You shouldn't have to add the extra quotes. But why doesn't it work?

nhatsmrt: Hmm originally I didn't put the quotations in either. It gave me a really weird JSON-parsing error. Let me try it again and send the error message here.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Resolved. I make a request to set branch, then use repository to get that branch (analogously to the Python client).

client/scala/src/main/scala/ai/verta/utils/URLUtils.scala
nhatsmrt: Resolved. I put the URLUtils in client package now.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: The error object return is this:
ai.verta.swagger.client.HttpException: [400] {"error":"json: cannot unmarshal number into Go value of type string","code":3,"message":"json: cannot unmarshal number into Go value of type string"}

nhatsmrt: Ok I managed to use put together an utility function based on net.lift.json's method. A bit less hacky. Resolved.

conradoverta: It's isn't much less hacky. Take a moment to think about your solution. Do you think it's a good idea to have the person adding a new functionality to have to think whether their particular case will fall into the class of issues you faced here and whether they need to call this method? Or is there a way that you can hide this complexity under the hood and automatically account for all such scenarios?

nhatsmrt: Oh I see what you mean. To avoid having to deal with this for new functionality, I'll have to put it where the query parameters are processed. I'm not quite sure if this hack is required at any other place though (I don't recall having to added double quotes to commit sha when I implemented other methods, only in this particular place), so a blanket fix (apply the method to all incoming parameters) is dangerous. I'll think more about this.

nhatsmrt: Ok so this is what I did: I move the jsonFormat inside the request method of HttpClient instead. This way, when I implement a new functionality involving the passing of a body, it will be automatically handled, instead of me having to manually apply the function again. Does it sound any better?

nhatsmrt: BTW I've also moved the jsonFormat function inside HttpClient, we're not gonna use it outside of this context.

client/scala/src/main/scala/ai/verta/client/FormatUtils.scala
```diff
@@ -0,0 +1,11 @@
+package ai.verta.client
+
+import java.net.URLEncoder
+
+/** Containing the helper functions to format the input string */
+object FormatUtils {
```
conradoverta: Which then leaves this `utils`, which is bound to become a kitchen sink. Can we do something with `urlEncode` similar to what you did with `jsonFormat`?

conradoverta (approved):

client/scala/src/main/scala/ai/verta/client/FormatUtils.scala
nhatsmrt: Hmm I think I can. I just replace your url encode function inside HttpClient (adding the replacement of "+" with "%20"), then split the path string and apply that function to each parts, then join it again. It seems to pass all of my tests (not breaking anything). I've removed FormatUtils entirely. See if that looks right. Resolved.

PR #713, opened 2020-06-03
update version in pom.xml

ravishetye:


backend/pom.xml
```diff
@@ -105,12 +104,6 @@
 			<artifactId>junit-quickcheck-generators</artifactId>
 			<version>0.9.1</version>
 		</dependency>
-		<!-- https://mvnrepository.com/artifact/net.jqwik/jqwik -->
```
ravishetye, line 108: this was duplicate entry

backend/pom.xml
```diff
@@ -179,13 +172,6 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
-		<!-- https://mvnrepository.com/artifact/org.projectlombok/lombok
```
ravishetye, line 182: this was commented but not removed

PR #716, opened 2020-06-04
rename message to not have Blob as suffix

ravishetye:


protos/protos/public/modeldb/versioning/VersioningService.proto
```diff
@@ -468,7 +468,7 @@ message FindRepositoriesBlobs {
     }
 }
 
-message GetUrlForVersionedBlob {
+message GetUrlForBlobVersioned {
```
ravishetye, line 471: only manual changes are in this file

evgeniilezhnin (approved):

PR #719, opened 2020-06-04
 VR-4528 Create a Dataset Blob from local path for Scala client

nhatsmrt:
I've decided to broken the Dataset Blob implementation into 2 PRs, since one is already too large. The shared code in Dataset trait is also used for S3 (another type of Dataset).

### DIFF
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
@@ -3,9 +3,10 @@ package ai.verta.blobs.dataset
 import ai.verta.swagger._public.modeldb.versioning.model._
 
 import scala.collection.mutable.HashMap
+import scala.util.{Failure, Success, Try}
 
 trait Dataset extends Blob {
-  protected var contents = new HashMap[String, FileMetadata]() // for deduplication and comparing
+  protected var contents = new HashMap[String, Try[FileMetadata]]() // for deduplication and comparing
 
   /** Helper to convert VersioningPathDatasetComponentBlob to FileMetadata
    */
@@ -29,10 +30,16 @@ trait Dataset extends Blob {
    *  @param path path to the file
    *  @return None if path is not in dataset blob, or some file metadata.
    */
-  def getMetadata(path: String): Option[FileMetadata] = contents.get(path)
+  def getMetadata(path: String): Try[FileMetadata] = contents.get(path) match {
+    case Some(v) => v
+    case None => Failure(new IllegalArgumentException("Path is not stored, or is a directory."))
+  }
 
-  /** Get all the files' metadata managed by the Dataset blob */
-  protected def components = contents.values.map(toComponent _).toList
+  /** Get all the Dataset blob's corresponding list of components */
+  protected def components = getAllMetadata.map(toComponent _).toList
+
+  /** Get the set of all the files' metadata managed by the Dataset blob  */
+  def getAllMetadata = contents.values.filter(_.isSuccess).map(_.get)
 }
 
 /** Represent a file's metadata stored in Dataset Blob
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
@@ -1,6 +1,7 @@
 package ai.verta.blobs.dataset
 
 import ai.verta.swagger._public.modeldb.versioning.model._
+import ai.verta.blobs._
 
 import scala.collection.mutable.HashMap
 import scala.util.{Failure, Success, Try}
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
@@ -36,7 +36,7 @@ trait Dataset extends Blob {
   }
 
   /** Get all the Dataset blob's corresponding list of components */
-  protected def components = getAllMetadata.map(toComponent _).toList
+  protected def components = getAllMetadata.map(toComponent).toList
 
   /** Get the set of all the files' metadata managed by the Dataset blob  */
   def getAllMetadata = contents.values.filter(_.isSuccess).map(_.get)
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
@@ -6,7 +6,7 @@ import scala.collection.mutable.HashMap
 import scala.util.{Failure, Success, Try}
 
 trait Dataset extends Blob {
-  protected var contents = new HashMap[String, Try[FileMetadata]]() // for deduplication and comparing
+  protected var contents: HashMap[String, FileMetadata] // for deduplication and comparing
 
   /** Helper to convert VersioningPathDatasetComponentBlob to FileMetadata
    */
@@ -30,28 +30,11 @@ trait Dataset extends Blob {
    *  @param path path to the file
    *  @return None if path is not in dataset blob, or some file metadata.
    */
-  def getMetadata(path: String): Try[FileMetadata] = contents.get(path) match {
-    case Some(v) => v
-    case None => Failure(new IllegalArgumentException("Path is not stored, or is a directory."))
-  }
+  def getMetadata(path: String) = contents.get(path)
 
   /** Get all the Dataset blob's corresponding list of components */
   protected def components = getAllMetadata.map(toComponent).toList
 
   /** Get the set of all the files' metadata managed by the Dataset blob  */
-  def getAllMetadata = contents.values.filter(_.isSuccess).map(_.get)
-}
-
-/** Represent a file's metadata stored in Dataset Blob
- *  @param lastModified last time file was modified
- *  @param md5 MD5 hash of the file
- *  @param path path of the file
- *  @param size size of the file
- */
-class FileMetadata(val lastModified: BigInt, val md5: String, val path: String, val size: BigInt) {
-  override def equals(other: Any) = other match {
-    case other: FileMetadata => lastModified == other.lastModified &&
-    md5 == other.md5 && path == other.path && size == other.size
-    case _ => false
-  }
+  def getAllMetadata = contents.values
 }
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
@@ -1,0 +1,50 @@
+package ai.verta.blobs.dataset
+
+import ai.verta.swagger._public.modeldb.versioning.model._
+
+import scala.collection.mutable.HashMap
+
+trait Dataset extends Blob {
+  protected var contents = new HashMap[String, FileMetadata]() // for deduplication and comparing
+
+  /** Helper to convert VersioningPathDatasetComponentBlob to FileMetadata
+   */
+  protected def toMetadata(component: VersioningPathDatasetComponentBlob) = new FileMetadata(
+    component.last_modified_at_source.get,
+    component.md5.get,
+    component.path.get,
+    component.size.get
+  )
+
+  /** Helper to convert VersioningPathDatasetComponentBlob to FileMetadata
+   */
+  protected def toComponent(metadata: FileMetadata) = VersioningPathDatasetComponentBlob(
+    last_modified_at_source = Some(metadata.lastModified),
+    md5 = Some(metadata.md5),
+    path = Some(metadata.path),
+    size = Some(metadata.size)
+  )
+
+  /** Get the metadata of a certain file stored in the dataset blob
+   *  @param path path to the file
+   *  @return None if path is not in dataset blob, or some file metadata.
+   */
+  def getMetadata(path: String): Option[FileMetadata] = contents.get(path)
+
+  /** Get all the files' metadata managed by the Dataset blob */
+  protected def components = contents.values.map(toComponent _).toList
+}
+
+/** Represent a file's metadata stored in Dataset Blob
+ *  @param lastModified last time file was modified
+ *  @param md5 MD5 hash of the file
+ *  @param path path of the file
+ *  @param size size of the file
+ */
+class FileMetadata(val lastModified: BigInt, val md5: String, val path: String, val size: BigInt) {
+  override def equals(other: Any) = other match {
+    case other: FileMetadata => lastModified == other.lastModified &&
+    md5 == other.md5 && path == other.path && size == other.size
+    case _ => false
+  }
+}
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
@@ -1,0 +1,17 @@
+package ai.verta.blobs.dataset
+
+/** Represent a file's metadata stored in Dataset Blob
+ *  @param lastModified last time file was modified
+ *  @param md5 MD5 hash of the file
+ *  @param path path of the file
+ *  @param size size of the file
+ */
+class FileMetadata(val lastModified: BigInt, val md5: String, val path: String, val size: BigInt) {
+  override def equals(other: Any) = other match {
+    case other: FileMetadata => lastModified == other.lastModified &&
+    md5 == other.md5 && path == other.path && size == other.size
+    case _ => false
+  }
+
+  override def toString = lastModified + " " + md5 + " " + path + " " + size
+}
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
@@ -12,6 +12,4 @@ class FileMetadata(val lastModified: BigInt, val md5: String, val path: String, 
     md5 == other.md5 && path == other.path && size == other.size
     case _ => false
   }
-
-  override def toString = lastModified + " " + md5 + " " + path + " " + size
 }
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
@@ -13,10 +13,15 @@ import scala.util.{Failure, Success, Try}
  */
  case class PathBlob(private val paths: List[String]) extends Dataset {
   private val BufferSize = 8192
-  paths.map(expanduser).map((path: String) => processPath(new File(path)))
+
+  private val metadataList = paths.map(expanduser)
+  .flatMap((path: String) => processPath(new File(path)))
+  .map(metadata => metadata.path -> metadata)
+
+  protected var contents = HashMap(metadataList: _*)
 
   override def equals(other: Any) = other match {
-    case other: PathBlob => getAllMetadata.equals(other.getAllMetadata)
+    case other: PathBlob => contents.equals(other.contents)
     case _ => false
   }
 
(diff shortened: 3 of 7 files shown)


client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -0,0 +1,50 @@
...
+  )
+
+  /** Get the metadata of a certain file stored in the dataset blob
+   *  @param path path to the file
+   *  @return None if path is not in dataset blob, or some file metadata.
+   */
+  def getMetadata(path: String): Option[FileMetadata] = contents.get(path)
+
+  /** Get all the files' metadata managed by the Dataset blob */
+  protected def components = contents.values.map(toComponent _).toList
```
conradoverta: I don't think you need `_` here.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/** Represent a file's metadata stored in Dataset Blob
+ *  @param lastModified last time file was modified
+ *  @param md5 MD5 hash of the file
+ *  @param path path of the file
+ *  @param size size of the file
+ */
+class FileMetadata(val lastModified: BigInt, val md5: String, val path: String, val size: BigInt) {
```
conradoverta: Separate file for separate class.

Also, shouldn't this be a case class?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,122 @@
...
+
+import scala.collection.mutable.HashMap
+import scala.util.{Failure, Success, Try}
+
+/** Captures metadata about files
+ *  @param paths list of filepaths or directory paths
+ */
+ case class PathBlob(private val paths: List[String]) extends Dataset {
+  private val BufferSize = 8192
+  paths.map(expanduser _).map((path: String) => processPath(new File(path)))
```
conradoverta: I don't think you need `_` here.

Also, can you try to re-organize this so that it's not working by side-effects?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,122 @@
...
+  override def equals(other: Any) = other match {
+    case other: PathBlob => contents.equals(other.contents)
+    case _ => false
+  }
+
+  /** Hash the file's content
+   *  From https://stackoverflow.com/questions/41642595/scala-file-hashing
+   *  @param path filepath
+   */
+  private def hash(file: File): String = {
```
conradoverta: Try?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,122 @@
...
+
+     // Convert to hexadecimal
+     messageDigest.digest.map("%02x".format(_)).mkString
+   }
+
+   /** Get the metadata of path.
+    *  @param file a file object, representing the path
+    *  @return a list of components of file under the path
+    */
+   private def processPath(file: File) = {
```
conradoverta: Can you think of a more functional way of doing this?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,122 @@
...
+      Try {
+        new FileMetadata(
+          BigInt(file.lastModified()),
+          hash(file),
+          file.getPath(),
+          BigInt(file.length)
+        )
+      } match {
+        case Success(metadata) => {
+          contents.put(file.getPath(), metadata)
```
conradoverta: Can you think of a more functional way of doing this?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,122 @@
...
+      }
+    }
+   }
+
+  /** Analogous to Python's os.path.expanduser
+   *  From https://stackoverflow.com/questions/6803913/java-analogous-to-python-os-path-expanduser-os-path-expandvars
+   *  @param path path
+   *  @return path, but with (first occurence of) ~ replace with user's home directory
+   */
+  private def expanduser(path: String) = path.replaceFirst("~", System.getProperty("user.home"))
```
conradoverta: Seems duplicate, right? I recall seeing this somewhere else.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,122 @@
...
+object PathBlob {
+  /** Factory method to convert a versioning path dataset blob instance
+   *  @param pathVersioningBlob the versioning blob to convert
+   *  @return equivalent PathBlob instance
+   */
+  def apply(pathVersioningBlob: VersioningPathDatasetBlob) = {
+    var pathBlob = new PathBlob(List())
+
+    pathVersioningBlob.components.get.map(
+      comp => pathBlob.contents.put(comp.path.get, pathBlob.toMetadata(comp))
```
conradoverta: Can you think of a way to make this more functional?

client/scala/src/test/scala/ai/verta/blobs/dataset/TestPathBlob.scala
```diff
@@ -0,0 +1,71 @@
...
+  test("PathBlob should not store directory") {
+    val f = fixture
+    val pathBlob = PathBlob(List(f.testDir, f.testSubdir))
+
+    assert(pathBlob.getMetadata(f.testDir).isEmpty)
+    assert(pathBlob.getMetadata(f.testSubdir).isEmpty)
+    assert(pathBlob.getMetadata(f.testfile).isDefined)
+  }
+
+  test("PathBlob should not store invalid paths, but should not stop execution") {
```
conradoverta: I think it should stop. That's an error, no?

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: I don't think so. This is not actually a subclass of Blob or anything. I just use it as a user friendly interface to work with the metadata of a file (It's sort of like the VersioningPathDatasetComponent, but with much shorter name). I don't imagine it having any extra functions or anything, but I'll move it to another file if it will make it look better.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestPathBlob.scala
nhatsmrt: That's what the Python client did too. I was trying to salvage the valid ones with all the Try magics, but if it's not worth the troubles then I can just let it die.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Maybe in the old scala-client PR? I don't recall this being used anywhere else (originally I put it in a package object, but since that has a tendency to become a kitchen sink I just copy pasted it here and delete the object altogether).

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Huh I swear that when you msged me about it I fixed it immediately. Or maybe you posted here first? Anyway resolved.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestPathBlob.scala
nhatsmrt: So this is what I'll do to stop the program: if any of the path is invalid, an exception is thrown. If the user doesn't catch this exception, then the program terminates. This is a runtime problem though, since there is no way to check if a file exists at compile time.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Ok I can make it more functional by mapping from a list of item to list of (key -> value) before converting it to map, but IMO it still looks a bit ugly because I'm using HashMap. I think it's a bit more efficient than immutable map right? But if it's not worth it then I can just call toMap on the list of pair and call it a day. Resolved for now.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Oh I was wrapping the Try down in the processFile or something, since there could be multiple method calls there that cause an exception (not just hash). But I'll add a Try here since in the future someone might re-use this, and should probably know it might cause an exception. Resolved.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Hmm originally I wrote a pretty elegant recursive method. I change it to this DFS with stack implementation because I feared that it might cause a stack overflow error if the directory structure is too deep. I wonder if there's a functional non-recursive DFS traversal?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: I'll just return the FileMetadata here, then create the map once at the end instead of creating the map then adding stuff to it. Resolved.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Resolved for now, unless I remember where I used it, in which case I'll fix that and update here.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Same as what I did above. Also same comment about HashMap vs Immutable Map.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestPathBlob.scala
nhatsmrt: I'll resolve this for now, but if there's a better way than to throw an exception I'm open to change it. I'll keep thinking about it too.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Resolved.

client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
```diff
@@ -0,0 +1,17 @@
...
+ *  @param size size of the file
+ */
+class FileMetadata(val lastModified: BigInt, val md5: String, val path: String, val size: BigInt) {
+  override def equals(other: Any) = other match {
+    case other: FileMetadata => lastModified == other.lastModified &&
+    md5 == other.md5 && path == other.path && size == other.size
+    case _ => false
+  }
+
+  override def toString = lastModified + " " + md5 + " " + path + " " + size
```
conradoverta: Check how other things for `toString`. It needs to be more informative.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,124 @@
...
+
+/** Captures metadata about files
+ *  @param paths list of filepaths or directory paths
+ */
+ case class PathBlob(private val paths: List[String]) extends Dataset {
+  private val BufferSize = 8192
+
+  private val metadataList = paths.map(expanduser)
+  .flatMap((path: String) => processPath(new File(path)))
+  .map(metadata => metadata.path -> metadata)
```
conradoverta: Can we have a more scala-like indentation? This is pretty hard to read.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,124 @@
...
+       else {
+         ret = processFile(fileToProcess) :: ret
+       }
+     }
+
+     ret
+   }
+
+  /** Extract the metadata of the file
+   *  If the file has an invalid path, exception is thrown immediately, and program stops (if not caught outside)
```
conradoverta: Don't throw. Return a try. Always.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,124 @@
...
+   private def processPath(file: File): List[FileMetadata] = {
+     var files = List(file) // stack
+     var ret: List[FileMetadata] = List()
+
+     // non-recursive DFS to prevent stack overflow
+     while (files.length > 0) {
+       var fileToProcess = files.head
+       files = files.drop(1)
+
+       if (fileToProcess.isDirectory()) {
```
conradoverta: So it's not a file, right?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,124 @@
...
+
+     // Convert to hexadecimal
+     messageDigest.digest.map("%02x".format(_)).mkString
+   }
+
+   /** Get the metadata of path.
+    *  @param file a file object, representing the path
+    *  @return a list of components of file under the path
+    */
+   private def processPath(file: File): List[FileMetadata] = {
```
conradoverta: This can fail, so it should return a try. You implicitly can throw things, which is not documented in the function signature. A try will enforce this knowledge at compile time.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: OK implement a tail-recursive dfs traversal. Pretty functional, and should not cause a stack overflow error if I do it right. Please check if it looks good.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  /** Tail-recursive DFS traversal to prevent stack overflow error
+   *  @param stack a stack containing the files/dirs to explore
+   *  @param acc accumulator list of file metadata to return
+   */
+  @tailrec private def dfs(stack: List[File], acc: List[FileMetadata]): List[FileMetadata] = {
+    if (stack.isEmpty) acc
+    else if (stack.head.isDirectory) {
+      val dir = stack.head
+      dfs(dir.listFiles().toList ::: stack.drop(1), acc)
```
conradoverta:
```suggestion
      dfs(dir.listFiles().toList ::: stack.tail, acc)
```

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -0,0 +1,127 @@
...
+    }
+    else {
+      val file = stack.head
+      dfs(stack.drop(1), processFile(file) :: acc)
```
conradoverta:
```suggestion
      dfs(stack.tail, processFile(file) :: acc)
```

client/scala/src/test/scala/ai/verta/blobs/dataset/TestPathBlob.scala
conradoverta: "That's what the Python client did too." @convoliution this seems like a bug to me.

client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
nhatsmrt: OK this is stupid but I never intended to add this function. I only used it for debugging. Must have accidentally added it. I'll remove it (unless you want me to create a professional toString).

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: This is outdated I think. I just worked on the functional version of this as you have requested, and has pushed it I believe?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: oh that was dumb. should be fixed already.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: fixed

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Yeah atom's autoindent betrayed me there. I've aligned the dots. Honestly there are so many people doing different things with method chaining indentation.

client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
nhatsmrt: Removed.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: OK I'll propagate the failure up all the way to the apply function then.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestPathBlob.scala
```diff
@@ -0,0 +1,75 @@
...
+
+    assert(dirAttempt.isEmpty)
+    assert(subdirAttempt.isEmpty)
+    assert(pathBlob.getMetadata(f.testfile).isDefined)
+  }
+
+  test("PathBlob should throw an exception when an invalid path is passed") {
+    val f = fixture
+    val invalid = f.testDir + "/invalid-file"
+    assertThrows[FileNotFoundException] {
```
conradoverta: This needs to be a try test.

nhatsmrt: Yep just fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Resolved.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestPathBlob.scala
nhatsmrt: Oh by "that's what the python did too" what i meant was what you said was what it did. (i.e raise an error). I was trying to do something else.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Wait I found multiple method chaining styles via googling. I think this will look best:
https://alvinalexander.com/scala/how-to-fluent-style-programming-methods-in-scala/

conradoverta (approved):
LGTM

PR #721, opened 2020-06-05
Implement repository description

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
```diff
@@ -144,8 +144,10 @@ public void updateRepository(
     try {
       try (RequestLatencyResource latencyResource =
           new RequestLatencyResource(modelDBAuthInterceptor.getMethodName())) {
-        if (request.getRepository().getName().isEmpty()) {
-          throw new ModelDBException("Repository name is empty", Code.INVALID_ARGUMENT);
+        if (request.getRepository().getName().isEmpty()
+            && request.getRepository().getDescription().isEmpty()) {
```
evgeniilezhnin, line 148: does it make an update repository request with an empty name and not empty description valid? In this case, name will be set to empty string?

backend/src/main/resources/liquibase/db-changelog-1.0.xml
```diff
@@ -518,6 +518,17 @@
...
     </changeSet>
 
+    <changeSet id="add-repository-description-column" author="anandJ">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="repository" columnName="description"/>
+            </not>
+        </preConditions>
+        <addColumn tableName="repository">
+            <column name="description" type="TEXT"/>
```
evgeniilezhnin: will it have default value null which will cause NPE in toProto for existing repos? Please check

anandjakhaniya: no, it will not give an NPE because when we are creating repository it will set `""` into the field at the constructor. so it will never null.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
anandjakhaniya: For both the questions, the answer is yes.

backend/src/main/resources/liquibase/db-changelog-1.0.xml
evgeniilezhnin: I mean repositories that was already created with the old constructor (that didn't have description field). What value for description will it have?

anandjakhaniya: Ok got it your point. updating liquebase

anandjakhaniya: updated

evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
ravishetye: this is not right , a repository can have empty description but should never have a empty name

ravishetye (changes requested):
minor change required, feel free to merge after the comment is resolved

ravishetye (approved):

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
anandjakhaniya: reverted condition for the statement `but should never have an empty name`

PR #723, opened 2020-06-05
VR-4592 dataset blobs description proto

evgeniilezhnin:


ravishetye (changes requested):
description should be added to all blobs, implementation wise we can just implement it for dataset first if that simplifies things. 
Let us add it at 

```
message Blob {
    oneof content {
        DatasetBlob dataset = 2;// Blob of the type dataset
        EnvironmentBlob environment = 3;
        CodeBlob code = 4;
        ConfigBlob config = 5;
    },
    string description = 6;
}
```

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/DiffComputer.java
```diff
@@ -117,6 +117,8 @@ public static AutogenBlobDiff computeBlobDiff(AutogenBlob a, AutogenBlob b) {
             .setEnvironment(
                 computeDiff(
                     a, b, AutogenBlob::getEnvironment, DiffComputer::computeEnvironmentDiff)));
+    // .setDescription(computeDiff(a, b, AutogenBlob::getDescription,
+    // DiffComputer::computeDescriptionDiff)));
```
ravishetye, line 121: this was commented after the file was auto generated?
Could this be resolved in the templating logic?
cc @conradoverta  if he has any thoughts.

backend/src/test/java/ai/verta/modeldb/blobs/BlobEquality.java
```diff
@@ -61,6 +61,11 @@ public void equalityAutogenDatasetDiff(AutogenDatasetDiff b) {
     assertTrue(b.equals(b));
   }
 
+  @Property
```
ravishetye, line 64: @conradoverta : should the metadata associated with the Blob affect the diff?

PR #724, opened 2020-06-05
VR-4592 dataset blobs description implementation

evgeniilezhnin:


ravishetye (changes requested):
description needs to be at Blob level , dataset version corresponds to DatasetBlob and the [dataset part info](https://github.com/VertaAI/modeldb/blob/master/protos/protos/public/modeldb/DatasetVersionService.proto#L71) correspond to datasetComponentBlob

evgeniilezhnin:
> description needs to be at Blob level , dataset version corresponds to DatasetBlob and the [dataset part info](https://github.com/VertaAI/modeldb/blob/master/protos/protos/public/modeldb/DatasetVersionService.proto#L71) correspond to datasetComponentBlob

done

PR #726, opened 2020-06-05
Added tag support on (blob,commit,repo) combo

anandjakhaniya:


ravishetye (approved):

ravishetye:
this is putting the responsibility of creating a concatenated string onto the client/ user  of labelling service , this is fine for dataset integration since backend is the user, however if we have to GA labeling of repositories , is this expectation valied ? @conradoverta

PR #728, opened 2020-06-05
VR-4530 Create a Dataset Blob from S3 location

nhatsmrt:


client/scala/src/main/scala/ai/verta/blobs/dataset/S3Location.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import scala.util.{Try, Success, Failure}
+
+/** A location in S3
+ *  To initialize, pass in an S3 path and (optionally) a version ID:
+ *  {{{
+ *  val S3Location: Try[S3Location] = S3Location("some-path-1")
+ *  }}}
+ */
+class S3Location(val bucketName: String, val key: Option[String], val versionID: Option[String] = None) {}
```
conradoverta:
```suggestion
class S3Location(val bucketName: String, val key: Option[String] = None, val versionID: Option[String] = None) {}
```

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  override def equals(other: Any) = other match {
+    case other: S3 => contents.equals(other.contents) && versionMap.equals(other.versionMap)
+    case _ => false
+  }
+}
+
+/** Companion object to handle interaction with versioning blob */
+object S3 {
+  private val s3: AmazonS3 = AmazonS3ClientBuilder.standard().withRegion(Regions.DEFAULT_REGION).build()
```
conradoverta: What is `Regions.DEFAULT_REGION`? Is that loaded from the environment?

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,173 @@
...
+  private def getVersionMetadata(version: S3VersionSummary) = Try {
+    val metadata = new FileMetadata(
+      BigInt(version.getLastModified().getTime()),
+      version.getETag(),
+      getPath(version.getBucketName(), version.getKey()),
+      BigInt(version.getSize())
+    )
+
+    if (version.getVersionId() != null)
+      new Tuple2(metadata, Some(version.getVersionId()))
```
conradoverta: I recommend using a quick `case class` instead of tuples, since Tuples are usually associated with key-pair in maps. It makes reading the code much easier, especially when something is a few levels deeper.

nhatsmrt: Will do.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: It should be. However, according to:
https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html
User probably should set up AWS_REGION, which will be taken by default without the call to withRegion. That looks cleaner, and also aligns better with the Python client I think. I've also added this to the doc. Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3Location.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed.

nhatsmrt: Update: Make versionId an optional field of FileMetadata instead. This makes more semantic sense than a separate class + is cleaner to work with.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3Location.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import scala.util.{Try, Success, Failure}
+
+/** A location in S3
+ *  To initialize, pass in an S3 path and (optionally) a version ID:
+ *  {{{
+ *  val S3Location: Try[S3Location] = S3Location("some-path-1")
+ *  }}}
+ */
+class S3Location(
```
conradoverta, line 13:
```suggestion
case class S3Location(
```
?

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,168 @@
...
+  /** Helper function to extract metadata from a version summary
+   */
+  private def getVersionMetadata(version: S3VersionSummary) = Try {
+    new FileMetadata(
+      BigInt(version.getLastModified().getTime()),
+      version.getETag(),
+      getPath(version.getBucketName(), version.getKey()),
+      BigInt(version.getSize()),
+      if (version.getVersionId() != null) Some(version.getVersionId())
+      else None
```
conradoverta:
```suggestion
      Option(version.getVersionId())
```

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,168 @@
...
+  /** Helper function to extract metadata from the return object
+   */
+  private def getObjectMetadata(obj: ObjectMetadata, bucketName: String, key: String) = Try {
+    new FileMetadata(
+      BigInt(obj.getLastModified().getTime()), // convert time to UNIX timestamp (ms)
+      obj.getETag(),
+      getPath(bucketName, key),
+      BigInt(obj.getContentLength()),
+      if (obj.getVersionId() != null) Some(obj.getVersionId())
+      else None
```
conradoverta:
```suggestion
      Option(obj.getVersionId())
```

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,168 @@
...
+    else {
+      val request =
+        if (loc.versionID.isDefined) new GetObjectMetadataRequest(loc.bucketName, loc.key.get, loc.versionID.get)
+        else new GetObjectMetadataRequest(loc.bucketName, loc.key.get)
+
+      getObjectMetadata(s3.getObjectMetadata(request), loc.bucketName, loc.key.get).map(List(_))
+    }
+  }
+
+  /** Helper function to deal with VersionListing
```
conradoverta: This comment is not useful. Consider expanding on what the function actually does.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,168 @@
...
+                    .filter(_.isLatest())
+                    .map(getVersionMetadata) // List[Try]
+                    .map(_.get)
+      )
+
+    batchAttempt match {
+      case Failure(e) => Failure(e)
+      case Success(batch) =>
+        if (versionListing.isTruncated())
+          handleVersionListing(s3.listNextBatchOfVersions(versionListing), batch ::: acc)
```
conradoverta, line 143: Maintain order. It might matter.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /** Helper function to query metadata of S3 Location
+   */
+  private def getS3LocMetadata(loc: S3Location) = {
+    if (loc.key.isEmpty || loc.key.get.endsWith("/")) {
+      // no key (bucket), or is a folder
+      val versionListing =
+        if (loc.key.isEmpty) s3.listVersions(loc.bucketName, null)
+        else s3.listVersions(loc.bucketName, loc.key.get)
```
conradoverta:
```suggestion
        s3.listVersions(loc.bucketName, loc.key.orNull)
```

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,168 @@
...
+
+      handleVersionListing(versionListing, List())
+    }
+    else {
+      val request =
+        if (loc.versionID.isDefined) new GetObjectMetadataRequest(loc.bucketName, loc.key.get, loc.versionID.get)
+        else new GetObjectMetadataRequest(loc.bucketName, loc.key.get)
```
conradoverta, line 117: Can we collapse these?

client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
```diff
@@ -5,11 +5,19 @@ package ai.verta.blobs.dataset
...
+  val md5: String,
+  val path: String,
+  val size: BigInt,
+  val versionId: Option[String] = None
+) {
   override def equals(other: Any) = other match {
     case other: FileMetadata => lastModified == other.lastModified &&
-    md5 == other.md5 && path == other.path && size == other.size
+    md5 == other.md5 && path == other.path && size == other.size &&
+    ((versionId.isEmpty && other.versionId.isEmpty) || (versionId.get == other.versionId.get))
```
conradoverta: This will raise an exception if one of them is empty and the other isn't.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+/** Companion object to handle interaction with versioning blob */
+object S3 {
+  private val s3: AmazonS3 = AmazonS3ClientBuilder.standard().build()
+
+  /** Constructor that user should use:
+   */
+  def apply(paths: List[S3Location]): Try[S3] = {
```
conradoverta, line 51: 1. Provide a constructor that receives a simple location too
2. Provide a reducer that combines two S3 locations into a new one (also do that for Path)

client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
nhatsmrt: Oops I was dumb. Forgot to add the condition that both are defined.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Hmm these are two different constructors (with and without versionId). I don't think there's a shorter way to branch here?

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Hmm Ok. I'll reverse batch here, which makes the final list in reverse order:
(3, 4).reverse :::  (1, 2).reverse = (4, 3, 2, 1)
Finally, I'll reverse the result list once more. This will give the correct order.
Note that the order of concatenation is for efficiency reason. I believe that it's better to prepending the new stuff rather than appending them?

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Should be fixed, but do tell me if there's a better way.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3Location.scala
nhatsmrt: Making S3Location a case class is gonna create conflict when calling:
```
S3Location("some-string")
```
because it's not gonna know whether to pick which of the following:
```
apply(bucketName, key = None, versionId = None): S3Location // auto-provided
apply(path, versionID = None): Try[S3Location] // mine
```
We need my apply method to check the validity of path.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Add a bit of explanation. See if it's better. It doesn't do much though.

client/scala/src/main/scala/ai/verta/blobs/dataset/FileMetadata.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Ok will do. Shouldn't be too hard.

nhatsmrt: For 2 do you mean reducing two S3 blob instance into one right? Combining S3Locations themselves doesn't make sense to me. I'll work on that instead.

nhatsmrt: Reducers implemented (via apply factory method) and tested.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -38,4 +42,13 @@ trait Dataset extends Blob {
 
   /** Get the set of all the files' metadata managed by the Dataset blob  */
   def getAllMetadata = contents.values
+
+  /** Check if the other dataset is combinable (i.e no conflicting entries)
+   *  @param other other dataset to combine
+   *  @return whether there is a conflict in the contents of two dataset
+   */
+  protected def canCombine(other: Dataset) = {
```
conradoverta: `conflicts` would be a better name here from the description, right?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -46,6 +53,20 @@ object PathBlob {
     }
   }
 
+  /** Factory method to combine two PathBlob instances
+   *  @param firstBlob: first pathblob
+   *  @param secondBlob: second pathblob
+   *  @return failure if the two blobs have conflicting entries; the combined blob otherwise.
+   */
+  def apply(firstBlob: PathBlob, secondBlob: PathBlob): Try[PathBlob] = {
```
conradoverta: Call this a more meaningful name, like `reduce`. This gets very confusing since it's mixing a reducer and a constructor.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -0,0 +1,174 @@
...
+
+    s3Blob
+  }
+
+  /** Factory method to combine two S3 instances
+   *  @param firstBlob: first S3 blob
+   *  @param secondBlob: second S3 blob
+   *  @return failure if the two blobs have conflicting entries; the combined blob otherwise.
+   */
+  def apply(firstBlob: S3, secondBlob: S3): Try[S3] = {
```
conradoverta: Same.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
conradoverta: Why do you have to reverse? Why doesn't `acc ::: batch` work?

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
conradoverta: Interesting. Ok.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: I swear I name it conflicts initially, then change it to canCombine =) Ok I'll fix.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Resolved in chat. Added clarifying comments.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -38,4 +42,13 @@ trait Dataset extends Blob {
 
   /** Get the set of all the files' metadata managed by the Dataset blob  */
   def getAllMetadata = contents.values
+
+  /** Check if the other dataset is combinable (i.e no conflicting entries)
+   *  @param other other dataset to combine
+   *  @return whether there is a conflict in the contents of two dataset
+   */
+  protected def conflicts(other: Dataset) = {
```
conradoverta: Actually, this is checking for `notConflicts`, right? I read wrong the first time, I think.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -46,6 +53,20 @@ object PathBlob {
     }
   }
 
+  /** Factory method to combine two PathBlob instances
+   *  @param firstBlob: first pathblob
+   *  @param secondBlob: second pathblob
+   *  @return failure if the two blobs have conflicting entries; the combined blob otherwise.
+   */
+  def reduce(firstBlob: PathBlob, secondBlob: PathBlob): Try[PathBlob] = {
+    if (firstBlob.conflicts(secondBlob)) {
```
conradoverta: This should have tipped you off that the function name is wrong, since you want to fail on conflict.

conradoverta (commented):
Just one naming issue left.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Yeah you're right. I should have looked at this more closely.

conradoverta (approved):

PR #729, opened 2020-06-06
VR-4525: Retry part uploads on ConnectionErrors

convoliution:
I wanted to specifically only catch broken pipes, but unwrapping a `requests` exception is tedious and undocumented! This is how deep I had to go in my investigation:
Screen Shot 2020-06-08 at 3 57 28 PM
Rather than have a series of `if`s that I can't even justify with docs because there are none, I'm comfortable with retrying on all `ConnectionError`s for just this presigned URL upload.

### DIFF
--- a/client/verta/verta/client.py
+++ b/client/verta/verta/client.py
@@ -2116,7 +2116,14 @@ class ExperimentRun(_ModelDBEntity):
                 part_stream = six.BytesIO(file_part)
 
                 # upload part
-                response = _utils.make_request("PUT", url, self._conn, data=part_stream)
+                #     Retry connection errors, to make large multipart uploads more robust.
+                for _ in range(3):
+                    try:
+                        response = _utils.make_request("PUT", url, self._conn, data=part_stream)
+                    except requests.ConnectionError:
+                        continue  # try again
+                    else:
+                        break
                 response.raise_for_status()
 
                 # commit part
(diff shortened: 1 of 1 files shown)


client/verta/verta/client.py
```diff
@@ -2116,7 +2116,15 @@ def _upload_artifact(self, key, artifact_stream, part_size=64*(10**6)):
                 part_stream = six.BytesIO(file_part)
 
                 # upload part
-                response = _utils.make_request("PUT", url, self._conn, data=part_stream)
+                #     Retry connection errors, to make large multipart uploads more robust.
+                for _ in range(3):
+                    try:
+                        response = _utils.make_request("PUT", url, self._conn, data=part_stream)
+                    except requests.ConnectionError:
+                        time.sleep(10)
```
conradoverta: This seems way too long. Is there any specific reason for this choice?

convoliution: No particular reason at all. Should I remove the sleep?

Wanted a bit of backoff delay for connection problems, and 10s seemed small to me in comparison to how long a 64 MB part upload can take.

conradoverta: Maybe 1s? Consider that the user would have to sit 30s before knowing something is wrong.

conradoverta (approved):

PR #733, opened 2020-06-08
VR-4710 Scala client commit update and get operations

nhatsmrt:


client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -23,6 +28,99 @@ class Commit(
...
+    )
+  }
+
+  /** Adds blob to this commit at path
+   *  If path is already in this Commit, it will be updated to the new blob
+   *  @param path Location to add blob to
+   *  @param blob Instance of Blob subclass.
+   *  @return whether the update attempt succeeds.
+   */
+  def update[T <: Blob](path: String, blob: T)(implicit ec: ExecutionContext): Try[Unit] = {
```
conradoverta: How hard would it be to return the modified commit? I know the python client modifies in place, but different languages have different coding patterns.

conradoverta (commented):
Try to come up with a way to have this class be immutable.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Hmm OK. How about return a whole new commit for every update/remove operation? I'll change the blobs map to immutable Map, so that each such operation return a new version of it, but keep it a private var so that it's easier for initializing. How does that sound?

get doesn't modify or anything so we can disregard it.

conradoverta: Sounds good

nhatsmrt:
I just make the minimal change to make commit (effectively) immutable. Updating and removing should not modify the commit. The only side effect is loading blobs, which is basically a lazy operation. I do not make it a "lazy val" because then I lose the ability to initialize it however I want. 

However, this side effect is not observable by user. All the APIs exposed to the user do not allow them to probe into these mutable fields (e.g get always return the same result). To their perspective, update operation does not modify the old commit instance, but return a new commit instance instead.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -14,15 +17,113 @@ class Commit(
...
+        case Some(v) => List(v)
+        case None => commit.parent_shas.get
+      }
+
+      Try(ids.map(id => loadBlobsFromId(id)).map(_.get).reduce(_ ++ _)) match {
+        case Failure(e) => Failure(e)
+        case Success(map) => Success {
+          loadedFromRemote = true
+          blobs = map
+        }
```
conradoverta: This information is being lost since you're converting to Unit. Is this information not used anywhere? I find it suspicious that the result of `Success(map)` is not used anywhere and the tests pass. Either something is mutable or the tests are wrong. Even the places that call `loadBlobs` ignore the return values.

nhatsmrt: Hmm I think I mentioned it somewhere, but this is to simulate the lazy loading behavior of blobs. loadedFromRemote indicate whether the blobs has been loaded from DB. So here, if the "map" is successfully loaded, it will be set to true, and the map itself assign to blobs.

nhatsmrt: I'm aware that scala has "lazy val", which is basically this. But using that means I cannot reassign the variable blobs to another value, which makes initializing a new Commit instance difficult.

nhatsmrt: This is also very similar to the initializing of S3/PathBlob btw. The extra complexity comes from the fact that we're trying to have the blobs map lazy. 

Otherwise I can use a very similar pattern with an apply method in the companion object: retrieve the blobs from DB, turn it into a map, then assign it to the blobs var of a the new Commit instance.

nhatsmrt: To demonstrate my idea:

```
  def apply(
    clientSet: ClientSet,
    repo: Repository,
    versioningCommit: VersioningCommit,
    commitBranch: Option[String] = None,
    saved: Boolean = true,
    blobs: Option[Map[String, VersioningBlob]] = None
  ) = Try {
    // wrap in Try because loading blobs from DB might fail
    var commit = new Commit(clientSet, repo, versioningCommit, commitBranch, saved)
    commit.blobs = blobs.getOrElse(commit.loadBlobs().get)
    commit
  }
```
Something like this? I haven't tried actually using this, but the general idea would be this (assuming we sacrifice the laziness of commit.blobs)

nhatsmrt:
@conradoverta See if the new design makes sense. I basically just sacrifice the laziness of loading blobs, and move all the initialization to companion object. If it looks good then I'll add docs and stuff to it.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
conradoverta: It's fine for now. But if I had to ask and you have to use 3 responses to clarify, then it deserves at least a few comments in the code explaining what's going on. We should also not return a struct that is discarded as that's very confusing.

conradoverta:
> @conradoverta See if the new design makes sense. I basically just sacrifice the laziness of loading blobs, and move all the initialization to companion object. If it looks good then I'll add docs and stuff to it.

The original idea was fine, it was just **very** not clear to me what was going on from the code. Lazy loading is the correct way to go to avoid loading the content of every commit during `log` operations and similar.

nhatsmrt:
@conradoverta Oh I see. Oh well good attempt I guess. I'll undo the recent changes, and add some clarifying comments instead.

conradoverta (approved):

PR #734, opened 2020-06-08
VR-4712 Support saving modifications to Commit

nhatsmrt:


client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -66,6 +66,79 @@ class Commit(
...
+        body = VersioningCreateCommitRequest(
+          commit = Some(addMessage(message)),
+          blobs = blobs,
+          commit_base = commitBase,
+          diffs = diffs
+        ),
+        repository_id_repo_id = repo.id
+      )
+      .flatMap(r => {
+        val newCom = new Commit(clientSet, repo, r.commit.get, commitBranch)
```
conradoverta: Avoid cutting words with non-standard terms. `cfg` is a common contraction for `config`, but `com` is not for `commit`.

nhatsmrt: You're right. I think I used it in my other branch as well. Will have to fix there too.

conradoverta (approved):

PR #739, opened 2020-06-09
VR-4592 dataset blobs description metadata proto

evgeniilezhnin:


ravishetye (changes requested):
this will change based on the implementation PR, we could use the VersioningCompositeIdentifier to attach label or description to it. So we need to add 3 more APIs for add,get,delete description

evgeniilezhnin:
> this will change based on the implementation PR, we could use the VersioningCompositeIdentifier to attach label or description to it. So we need to add 3 more APIs for add,get,delete description

created separate endpoints for metadata parameters instead of labels (this will be used only for description parameter for now)

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -51,6 +58,35 @@ message DeleteLabelsRequest {
     }
 }
 
+message GetParameterRequest {
```
ravishetye: replace Parameter with Metadata and then this looks good to me.

ravishetye: or `Property` to avoid the repetition of metadata/metadata

ravishetye: feel free to merge this if you end up making the change to Property

PR #743, opened 2020-06-09
VR-4711 Support removing blobs from Commit

nhatsmrt:


client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -181,15 +188,23 @@ class Commit(
...
-    val newCommit = VersioningCommit(
+    val newVersioningCommit = VersioningCommit(
       parent_shas = if (saved) commit.commit_sha.map(List(_)) else commit.parent_shas
     )
-    new Commit(clientSet, repo, newCommit, commitBranch)
+
+    val child = new Commit(clientSet, repo, newVersioningCommit, commitBranch)
+    child.blobs = childBlobs
+    child.saved = false
+    child.loadedFromRemote = true
```
conradoverta: You're making a dangerous assumption here. You're assuming that `getVersioningBlob` has been called before, which might not be true.

nhatsmrt: Oops, I should have put loadBlobs before that. I'll fix then.

conradoverta (approved):

PR #744, opened 2020-06-09
VR-4505: Add test to verify multipart upload

convoliution:


client/verta/tests/test_artifacts.py
```diff
@@ -88,6 +91,42 @@ def test_upload_dir(self, experiment_run, strs, dir_and_files):
         with zipfile.ZipFile(experiment_run.get_artifact(key), 'r') as zipf:
             assert filepaths == set(zipf.namelist())
 
+    @pytest.mark.not_oss
```
conradoverta, line 96: Why not?

client/verta/tests/test_artifacts.py
```diff
@@ -88,6 +91,42 @@ def test_upload_dir(self, experiment_run, strs, dir_and_files):
...
+            assert response.status_code == 404
+            assert "Can't find specified artifact" in response.text
+
+            # log artifact
+            experiment_run.log_artifact(key, tempf)
+
+            # Client entered multipart loop
+            assert "uploading part " in capsys.readouterr().out
+
+            # artifact parts ModelDB entry exists
```
conradoverta: Can we make a stronger assertion?

client/verta/tests/test_artifacts.py
convoliution: Because multipart upload doesn't occur in our OSS setup, so that is what I'm using to determine whether or not to run this test.

client/verta/tests/test_artifacts.py
convoliution: I couldn't come up with a stronger assertion to make, since the committed artifact parts are deleted from the database once the multipart upload is completed (see [my comment on Evgenii's backend PR](https://github.com/VertaAI/modeldb/pull/660#pullrequestreview-418686850) for context).

Do you suppose this would be a good reason to keep them in the database?

conradoverta: Maybe. But if that's the case, I don't know what the call below is doing.

client/verta/tests/test_artifacts.py
conradoverta: Why it doesn't occur?

convoliution: MDB doesn't use the multipart upload procedure in our OSS setup with NFS as its artifact store. It just gives a single URL to upload the whole artifact.

client/verta/tests/test_artifacts.py
convoliution: It just checks that `getCommittedArtifactParts` now returns a non-error response (it returned a `404` before `log_artifact()`), suggesting (weakly) that MDB registered committed artifact points at some point

It's indirect, but it's the most I could do with the information I seem to have.

conradoverta: That's a very weak test that doesn't verify it was actually uploaded with multiple parts. If you're missing information, then we should consider exposing that information.

It's also weird to me that it returning not-404 while also deleting the data.

client/verta/tests/test_artifacts.py
conradoverta: Oh yeah, I forgot about that.

conradoverta (approved):

PR #750, opened 2020-06-10
fix null pointer exception when DB communication link fail

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -302,6 +302,7 @@ private static SessionFactory loopBack(SessionFactory sessionFactory) {
   }
 
   public static SessionFactory resetSessionFactory() {
+    isReady = false;
```
ravishetye, line 305: why was this required? the original code was not referring to isReady anyway

anandjakhaniya: null pointer came from https://github.com/VertaAI/modeldb/pull/750/files#diff-aeb0311276570eecf18287327eea57fbR492 so while  `resetSessionFactory` it is not in the ready Health status that's why I have marked it false as per our last discussion.

ravishetye (approved):

PR #760, opened 2020-06-10
VR-4713 Support merging commits

nhatsmrt:


client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -243,4 +234,52 @@ class Commit(
...
+   *  This method creates and returns a new Commit in ModelDB, and assigns a new ID to this object
+   *  @param other Commit to be merged
+   *  @param message Description of the merge. If not provided, a default message will be used
+   *  @return Failure if this commit or other has not yet been saved, or if they do not belong to the same Repository; the merged commit otherwise.
+   */
+  def merge(other: Commit, message: Option[String] = None)(implicit ec: ExecutionContext) = {
+    if (!saved)
+      Failure(new IllegalStateException("This commit must be saved"))
+    else if (!other.saved)
+      Failure(new IllegalStateException("Other commit must be saved"))
```
conradoverta: Wrong type of exception?

client/scala/src/test/scala/ai/verta/repository/TestCommit.scala
```diff
@@ -177,4 +179,116 @@ class TestCommit extends FunSuite {
...
+
+    try {
+        val branch1 = f.repo.getCommitByBranch()
+                       .flatMap(_.newBranch("a"))
+                       .flatMap(_.update("abc/cde", f.pathBlob))
+                       .flatMap(_.save("Some message 1")).get
+
+        // touch the file
+        var file = new File(f"${System.getProperty("user.dir")}/src/test/scala/ai/verta/blobs/testdir/testfile")
+        file.setLastModified(System.currentTimeMillis())
```
conradoverta, line 236: This test will fail if it runs fast enough, right?

conradoverta (commented):
Can you check that the branch is updated on merge?

client/scala/src/test/scala/ai/verta/repository/TestCommit.scala
nhatsmrt: Yep =) I'll add some waiting. Ideally I want to make a test out of S3 vs Path or sth, but since that is still problematic I'll have to stick to this for now.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Yeah this one is my pet peeves as well. The other commit exception probably fits IllegalArgumentException, but I'm not too sure about the the one for this commit. It's technically a "state" (hence the IllegalState). Maybe I can write some custom Exception like CommitNotSavedException?

conradoverta: Sounds like a good idea.

client/scala/src/test/scala/ai/verta/repository/TestCommit.scala
nhatsmrt: Updated.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Ok done. I call it IllegalCommitSavedStateException. Quite mouthful, but this applies to save operation as well (which requires the commit to be **not** saved).

nhatsmrt:
I have added a check in the following format:
- branch a merge branch b
- check that head of branch a is equals to the returned merge commit
- check that head of branch b is not the returned merge commit, and is still the old branch b commit.

client/scala/src/test/scala/ai/verta/repository/TestCommit.scala
```diff
@@ -177,4 +179,123 @@ class TestCommit extends FunSuite {
...
+      val branch2 = f.repo.getCommitByBranch()
+                     .flatMap(_.newBranch("b"))
+                     .flatMap(_.update("def/ghi", f.s3Blob))
+                     .flatMap(_.save("Some message 2")).get
+
+      val mergeAttempt = branch1.merge(branch2, message = Some("Merge test"))
+      assert(mergeAttempt.isFailure)
+      assert(mergeAttempt match {case Failure(e) => e.getMessage contains "Two commits must belong to the same repository"})
+
+      f.client.deleteRepository(repo2.id)
```
conradoverta: This should be in the `finally` block.

conradoverta (commented):
One minor cleanup change.

client/scala/src/test/scala/ai/verta/repository/TestCommit.scala
nhatsmrt: Done.

conradoverta (approved):

PR #761, opened 2020-06-10
VR-4786: Typecheck tags everywhere

convoliution:
If someone calls
```python
client.find_datasets(tags="banana")  # instead of tags=["banana"]
```
The Client will iterate over the string and produce
```
tags EQ 'b'
&& tags EQ 'a'
&& tags EQ 'n'
&& tags EQ 'a'
&& tags EQ 'n'
&& tags EQ 'a'
```

This typecheck prevents that.


client/verta/verta/client.py
```diff
@@ -668,12 +668,18 @@ def find_datasets(self,
         """
         predicates = []
         if tags is not None:
+            if isinstance(tags, six.string_types):
+                raise TypeError("`tags` must be iterable of str, not {}".format(type(tags)))
```
conradoverta: Can we just put them on a list for them? Seems pretty strong to raise an error if it's a common enough mistake that we can absorb and hide.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -941,3 +941,34 @@ def is_org(workspace_name, conn):
...
+
+    Raises
+    ------
+    TypeError
+        If `tags` is neither str nor list of str.
+
+    """
+    if isinstance(tags, six.string_types):
+        tags = [tags]
+    else:
```
conradoverta, line 969: It might not be a list, right? For example, it could be a number or None, since some places are missing type checks. I'd recommend checking it here and raising another exception if it's none of the expected types.

convoliution: I intended for this to take any iterable of `str`, not just `list`s.
If it's, say, a number or `None`, then when the `for` loop begins, Python will raise
```
TypeError: 'int' object is not iterable
```

Two questions:

1. Is this adequate, or should I give the user a more specific error message, like "Expected `tags` to be a list of str, but got {}"?
1. Would be better to specifically only allow `list`s?

conradoverta: 1. As a user, I'd appreciate a more specific error message. If I get that one, I'll have to look into the client code to figure out what happened.
2. Lists and tuples, probably. Not sure, but seems like the base cases. Maybe generator or arbitrary iterables if easy enough.

convoliution: I've decided to cast the arg to a `list` and catch the potential `TypeError`.

That way I can know what type is actually being returned from this fn, avoid any potential nasty interactions between arbitrary iterables and protobuf messages I pass them into, and avoid having to reset a generator after exhausting it in the `for` loop. Should be fine as long as users limit their tags to below a couple few hundred.

convoliution: Ugh no, that would accept `dict`s by only capturing its keys.

convoliution: `list`s, `tuple`s, `set`s. Final answer.

conradoverta (approved):

PR #763, opened 2020-06-11
protos for dataset integration

ravishetye:


protos/protos/public/modeldb/DatasetVersionService.proto
```diff
@@ -301,12 +283,6 @@ service DatasetVersionService {
         };
     };
 
-    rpc getDatasetVersionTags(GetTags) returns (GetTags.Response) {
-        option (google.api.http) = {
-            get: "/v1/dataset-version/getDatasetVersionTags"
```
conradoverta, line 306: I thought we were supporting this, no?

protos/protos/public/modeldb/DatasetVersionService.proto
```diff
@@ -273,12 +261,6 @@ service DatasetVersionService {
         };
     };
 
-    rpc getDatasetVersionById (GetDatasetVersionById) returns (GetDatasetVersionById.Response) {
-        option (google.api.http) ={
-            get: "/v1/dataset-version/getDatasetVersionById"
```
conradoverta, line 278: Why remove?

protos/protos/public/modeldb/DatasetVersionService.proto
```diff
@@ -211,14 +207,6 @@ message UpdateDatasetVersionAttributes {
     }
 }
 
-message GetDatasetVersionAttributes {
```
conradoverta, line 214: I thought we were supporting this, no?

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -12,14 +12,22 @@ message IDTypeEnum {
         UNKNOWN = 0; // Type 0 should always cause an error
         VERSIONING_REPOSITORY = 1;
         VERSIONING_COMMIT = 2;
+        VERSIONING_REPO_COMMIT_BLOB = 3;
     }
 }
 
+message VersioningCompositeIdentifier {
```
conradoverta, line 19: Can we just use a string? The idea of this was exactly to avoid having to add one entry (that I believe will map to database changes and a ton of other overhead) every time we want to add something.

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -50,6 +58,35 @@ message DeleteLabelsRequest {
     }
 }
 
+message GetMetadataRequest {
```
conradoverta, line 61: Can we give this a more meaningful name? "Metadata" is way too broad.

protos/protos/public/modeldb/versioning/VersioningService.proto
```diff
@@ -106,8 +115,10 @@ message Repository {
     common.WorkspaceTypeEnum.WorkspaceType workspace_type = 6;
 
     RepositoryVisibilityEnum.RepositoryVisibility repository_visibility = 7;
-
+    RepositoryAccessModifierEnum.RepositoryAccessModifier repositoryAccessModifier = 8;
```
conradoverta, line 118: Why do we need this?

protos/protos/public/modeldb/versioning/VersioningService.proto
```diff
@@ -233,6 +244,20 @@ message ListCommitBlobsRequest {
     }
 }
 
+message FindBlobsRequest {
```
conradoverta, line 247: Not sure why we need this for this change.

protos/protos/public/modeldb/DatasetVersionService.proto
ravishetye: we are not supporting the get..Attribute, since no one was using it. 
webapp was using get Hydrated calls but with the decision, of moving to graphql may be we can not remove it

protos/protos/public/modeldb/DatasetVersionService.proto
ravishetye: no one was usinf it  and then having an equivalent get blob by id did not make sense, the find call has a dataset_id parameter which helps us go through the repo-> commit-> blob route

protos/protos/public/modeldb/DatasetVersionService.proto
ravishetye: we are not supporting the get..Tag, since no one was using it. 
webapp was using get Hydrated calls but with the decision, of moving to graphql may be we can not remove it

protos/protos/public/modeldb/metadata/MetadataService.proto
ravishetye: db entry is still a string ,  this is just the representation over wire.

protos/protos/public/modeldb/metadata/MetadataService.proto
ravishetye: already changes to Properties in the actual branch supporting this

protos/protos/public/modeldb/versioning/VersioningService.proto
ravishetye: agreed not required

ravishetye:
obsolete

PR #768, opened 2020-06-11
VR-4781 Update right after saving did not properly load the blobs

nhatsmrt:
Found a bug in update: load blobs in the wrong place. getChild does not need to load blob, because a blob is passed to it. Update does (and I didn't load the blobs for it). Fix, and add a test to demonstrate the case where it breaks.


client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -186,19 +186,18 @@ class Commit(
...
+  private def getChild(childBlobs: Map[String, VersioningBlob])(implicit ec: ExecutionContext) = {
       /** TODO: Deal with author, date_created */
       val newVersioningCommit = VersioningCommit(
         parent_shas = if (saved) commit.commit_sha.map(List(_)) else commit.parent_shas
       )
 
       val child = new Commit(clientSet, repo, newVersioningCommit, commitBranch)
       child.blobs = childBlobs
-      child.loadedFromRemote = true
+      child.loadedFromRemote = true  // child's blobs is already provided
```
```suggestion
      child.loadedFromRemote = true  // child's blobs are already provided
```

conradoverta (approved):

PR #769, opened 2020-06-12
Enable managed data versioning through Client

convoliution:
```python
# download files from S3 to local disk
s3_blob = S3("s3://verta-starter/", enable_mdb_versioning=True)

commit.update('s3-file', s3_blob)

# upload files from local disk to MDB
commit.save("Adding tiny file")

s3_blob = commit.get('s3-file')

# download file from MDB to local disk
s3_blob.download(
    "s3://verta-starter/census-train.csv",
    "/Users/me/Downloads/census-train.csv",
)
```


client/verta/verta/_repository/commit.py
```diff
@@ -123,6 +127,154 @@ def _raise_lookup_error(path):
         e = LookupError("Commit does not contain path \"{}\"".format(path))
         six.raise_from(e, None)
 
+    # TODO: consolidate this with similar method in `_ModelDBEntity`
```
conradoverta, line 130: Could you at least point to the differences? It's hard to review code that was already reviewed and spot the differences.

client/verta/verta/dataset/_s3.py
```diff
@@ -72,6 +79,40 @@ def __init__(self, paths):
 
         s3_metadata = six.viewvalues(obj_paths_to_metadata)
         self._msg.s3.components.extend(s3_metadata)  # pylint: disable=no-member
+        self._components_to_upload = dict()
+
+        # TODO: maybe make this happen in the above loop to help avoid race condition
+        # TODO: or it should happen in commit.save() so that's it's closer to upload time
```
conradoverta: I think putting it closer to save makes sense. Otherwise you're downloading things that you might not use.

PR #771, opened 2020-06-12
cleaned up logs

ravishetye:
change warn to info where it is not a backend isssue
removed Job related classes


backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
```diff
@@ -63,7 +63,7 @@ public static void initializeBasedOnConfig(
                 task, initialDelay, deleteEntitiesFrequency, TimeUnit.SECONDS);
             LOGGER.info("{} cron job scheduled successfully", ModelDBConstants.DELETE_ENTITIES);
           } else {
-            LOGGER.warn("Unknown config key ({}) found for the cron job", cronJob.getKey());
```
conradoverta, line 66: This seems like a proper error, right?

conradoverta (approved):
LGTM. I think that invalid config should cause an immediate crash to avoid a hard time debugging.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
ravishetye: this would mean there is a cron job in config without actual implementation support. This is similar to having random config entries but backend ignoring them, so I am fine logging and proceeding with this.

PR #776, opened 2020-06-12
VR-4522: Add methods for uploading artifacts in commit

convoliution:
Copied from [_ModelDBEntity._get_url_for_artifact()](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py#L768) and [ExperimentRun._upload_artifact()](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py#L2085).

Logic is identical.
Differences are really only the `Message`s, endpoints, and params.

PR #777, opened 2020-06-12
Use _path_component_blobs attribute for dataset blobs

convoliution:
On top of #776

Use a standard interface for accessing the inner component blobs.

This avoids requiring
```python
if S3:
    blob._msg.s3.components
elif Path:
    blob._msg.path.components
else:
    raise TypeError
```
whenever a caller wants to access the components.


client/verta/tests/test_versioning/test_dataset.py
```diff
@@ -9,9 +9,9 @@ def test_s3_bucket(self):
         pytest.importorskip("boto3")
 
         dataset = verta.dataset.S3("s3://verta-starter")
-        assert len(dataset._msg.s3.components) > 1
```
convoliution, line 12: Skip this file when reviewing; it's just a find-and-replace.

conradoverta (approved):
Nice!

client/verta/tests/test_versioning/test_dataset.py
convoliution: Unfortunately, it's a bit more complex than a find-and-replace for the S3 tests now, but it's fairly trivial and they all still pass.

convoliution:
@conradoverta Re-requesting review because I realized `S3DatasetBlob`, compared to `PathDatasetBlob`, has its `PathDatasetComponentBlob` nested one level deeper.

conradoverta (approved):

PR #778, opened 2020-06-12
Download objects from S3 for managed versioning

convoliution:
On top of #777

Equivalent for `Path` is still in progress.


conradoverta (commented):
I'd recommend adding a note to `save`'s documentation that it will trigger a (potentially big) download. Otherwise LGTM.

conradoverta (approved):
LGTM

PR #779, opened 2020-06-12
VR-4441: Upload data to ModelDB

convoliution:
On top of #778


client/verta/verta/_repository/commit.py
```diff
@@ -499,7 +499,16 @@ def save(self, message):
...
+        # upload ModelDB-versioned blobs
+        for blob_path, blob in mdb_versioned_blobs.items():
+            for component_blob in blob._component_blobs:
+                if component_blob.path.internal_versioned_path:
+                    component_path = component_blob.path.path
+                    downloaded_filepath = blob._components_to_upload[component_path]
+                    with open(downloaded_filepath, 'rb') as f:
+                        self._upload_artifact(blob_path, component_path, f)
+
+                    os.remove(downloaded_filepath)
```
convoliution: As a reminder: the `mdb_versioned_blobs` dict was created [just above](https://github.com/VertaAI/modeldb/blob/8eae5c527fc674cc4971cafcd8cc0d338ab808fa/client/verta/verta/_repository/commit.py#L490), before `_save()`.

conradoverta: I'd recommend leaving the cleaning to the class that knows about it (S3 and path). I imagine it would make the cleaning process simpler since you don't have to clean for path.

convoliution: That's a really good point! Will make the change to this PR.

conradoverta (approved):
LGTM

PR #780, opened 2020-06-12
VR-{4439, 4443, 4444}: Download a single versioned file from ModelDB

convoliution:
On top of #779


client/verta/verta/dataset/_dataset.py
```diff
@@ -53,3 +59,62 @@ def _path_component_to_repr_lines(path_component_msg):
...
+    def download(self, component_path, filepath):
+        """
+        Downloads `component_path` from this dataset if ModelDB-managed versioning was enabled.
+
+        Parameters
+        ----------
+        component_path : str
+            Original path of the file in this dataset to download.
+        filepath : str
+            Filepath to download `component_path` to.
```
conradoverta: Default to local di.

client/verta/verta/dataset/_dataset.py
```diff
@@ -53,3 +59,62 @@ def _path_component_to_repr_lines(path_component_msg):
...
+
+        # backend will return error if `component_path` not found/versioned
+        url = self._commit._get_url_for_artifact(self._blob_path, component_path, "GET").url
+
+        # stream download to avoid overwhelming memory
+        response = _utils.make_request("GET", url, self._commit._conn, stream=True)
+        try:
+            _utils.raise_for_http_error(response)
+
+            # decode responses that have Content-Encoding
```
conradoverta: Check if this is compatible with the backend. For example, does it send gz automatically?

convoliution: Note that this request is being made to the artifact store, not to the ModelDB backend.

Regardless, I've decided to use `response.iter_content()` instead because I feel more comfortable with `requests`'s public API than poking into `urllib3`. S3 sends `application/octet-stream` anyway.

client/verta/verta/dataset/_dataset.py
```diff
@@ -53,3 +58,62 @@ def _path_component_to_repr_lines(path_component_msg):
...
+        try:
+            _utils.raise_for_http_error(response)
+
+            # create parent dirs
+            pathlib2.Path(filepath).parent.mkdir(parents=True, exist_ok=True)  # pylint: disable=no-member
+
+            # read response stream into file
+            with open(filepath, 'wb') as f:
+                for chunk in response.iter_content(chunk_size=chunk_size):
+                    f.write(chunk)
```
conradoverta: I'd recommend writing to a location that has a different name and then renaming. The use case is when people check a given location to see if the file is there and skip the download (that we might even implement ourselves). So people only get the final file if everything goes fine.

client/verta/verta/dataset/_dataset.py
```diff
@@ -62,3 +68,77 @@ def _path_component_to_repr_lines(path_component_msg):
...
+        component_path : str
+            Original path of the file in this dataset to download.
+        filepath : str, optional
+            Filepath to download `component_path` to. If not provided, the file will be downloaded
+            to the current directory.
+        chunk_size : int, default 32 kB
+            Number of bytes to download at a time.
+
+        """
+        if self._commit is None and self._blob_path is None:
```
conradoverta:
```suggestion
        if self._commit is None or self._blob_path is None:
```
Maybe?

client/verta/verta/dataset/_dataset.py
```diff
@@ -62,3 +68,77 @@ def _path_component_to_repr_lines(path_component_msg):
...
+                "this dataset cannot be used for downloads;"
+                " consider using `commit.get()` to obtain a download-capable dataset"
+                " if ModelDB-managed versioning was enabled"
+            )
+        if filepath is None:
+            filepath = os.path.basename(component_path)
+
+        # TODO: remove this check when the os.rename() step is fixed to sidestep collisions
+        if os.path.exists(filepath):
+            raise OSError("{} already exists".format(filepath))
```
conradoverta: This is the contrary of what we had agreed, right?

client/verta/verta/dataset/_dataset.py
```diff
@@ -62,3 +68,77 @@ def _path_component_to_repr_lines(path_component_msg):
...
+
+            print("downloading {} from ModelDB".format(component_path))
+            try:
+                # read response stream into temp file
+                with tempfile.NamedTemporaryFile('wb', delete=False) as tempf:
+                    for chunk in response.iter_content(chunk_size=chunk_size):
+                        tempf.write(chunk)
+            except Exception as e:
+                # delete partially-downloaded file
+                os.remove(tempf.name)
```
conradoverta: Is the variable `tempf` actually defined here? Looks wrong to me, but maybe it's a quirk of python.

client/verta/verta/dataset/_dataset.py
```diff
@@ -62,3 +68,77 @@ def _path_component_to_repr_lines(path_component_msg):
...
+                raise e
+            else:
+                # move written contents to `filepath`
+                # TODO: prevent collision if `filepath` already exists
+                os.rename(tempf.name, filepath)
```
conradoverta: Same.

client/verta/verta/dataset/_dataset.py
convoliution: It should be. The `try` block doesn't create its own scope; `tempf` will be accessible in the `except` block.

Although I now realize that if `tempfile.NamedTemporaryFile('wb', delete=False)` fails, then we'll end up here before `tempf` has been assigned a value. I wouldn't expect it, but that's probably worth handling.

convoliution: See 7d0717385a6562d376e57513bef84b1485eb25e5

client/verta/verta/dataset/_dataset.py
convoliution: Done

client/verta/verta/dataset/_dataset.py
convoliution: You're correct. Was in the middle of updating.

convoliution:
@conradoverta Re-requesting review. This PR is for downloading a single file only.

I intend to implement "download folders" and "download all" in subsequent PRs.
Let me know if you'd prefer them all together

client/verta/verta/dataset/_dataset.py
```diff
@@ -62,3 +69,88 @@ def _path_component_to_repr_lines(path_component_msg):
...
+        try:
+            _utils.raise_for_http_error(response)
+
+            print("downloading {} from ModelDB".format(component_path))
+            tempf = None  # declare first in case NamedTemporaryFile init fails
+            try:
+                # read response stream into temp file
+                with tempfile.NamedTemporaryFile('wb', delete=False) as tempf:
+                    for chunk in response.iter_content(chunk_size=chunk_size):
+                        tempf.write(chunk)
```
conradoverta, line 143: We should download after all the next commands. If they fail, we want to fail early. I'd suggest moving this to right before the rename.

conradoverta:
One small recommendation

client/verta/verta/dataset/_dataset.py
convoliution: @conradoverta I've just remembered: I put this download step _before_ the "avoid collisions with existing files" step because a file could have been placed at the desired path during download, so I wanted to generate the destination filepath after download to lessen the chances of an accidental overwrite.

What do you think?

convoliution: Moved collision avoidance + parent dir creation to before the download
https://github.com/VertaAI/modeldb/pull/780/commits/d08b434a926ca3fc4b54b155852813bc1165a82c

conradoverta (approved):

PR #789, opened 2020-06-15
Add a table which holds the list of Repos which are Dataset

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/entities/versioning/DatasetRepositoryMappingEntity.java
```diff
@@ -0,0 +1,49 @@
...
+    this.isDatasetRepository = isDatasetRepository;
+  }
+
+  @Id
+  @OneToOne(cascade = CascadeType.ALL)
+  @JoinColumn(name = "repository_id")
+  private RepositoryEntity repositoryEntity;
+
+  @Column(name = "is_dataset_repository")
+  private Boolean isDatasetRepository = false;
```
ravishetye: is the second column required?
we can only store the repos which are dataset , why store rows which have this as false?

backend/src/main/java/ai/verta/modeldb/entities/versioning/RepositoryEntity.java
```diff
@@ -231,7 +239,6 @@ private void updateAttribute(List<KeyValue> attributes) throws InvalidProtocolBu
   }
 
   public boolean isDataset() {
-    // TODO: keep track if it is a dataset https://vertaai.atlassian.net/browse/VR-4571
-    return true;
+    return datasetRepositoryMappingEntity.getIsDatasetRepository();
```
ravishetye: return datasetRepositoryMappingEntity!= null 

would this work?

ravishetye (changes requested):
can merge if comment are resolved as suggested

backend/src/main/java/ai/verta/modeldb/entities/versioning/DatasetRepositoryMappingEntity.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/entities/versioning/RepositoryEntity.java
anandjakhaniya: Yes, it works. I have changed it as per your suggestion

PR #790, opened 2020-06-15
VR-4715 Support reverting commit

nhatsmrt:
Revert a merge commit is still non-deterministic. Not sure what the cause is.


client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -304,6 +304,32 @@ class Commit(
...
+   *  @param message Description of the revert. If not provided, a default message will be used
+   *  @return The new commit, with the changes in other reverted, if suceeds. Failure if this commit or other has not yet been saved, or if they do not belong to the same Repository.
+   */
+  def revert(other: Commit = this, message: Option[String] = None)(implicit ec: ExecutionContext) = {
+    if (!saved)
+      Failure(new IllegalCommitSavedStateException("This commit must be saved"))
+    else if (!other.saved)
+      Failure(new IllegalCommitSavedStateException("Other commit must be saved"))
+    else if (other.repo.id != repo.id)
+      Failure(new IllegalArgumentException("Two commits must belong to the same repository"))
```
conradoverta: Seems like we repeat this check in a few places. Can we extract this into its own lib?

conradoverta (commented):
Can we check that the branch is correctly updated too? I don't think it is, from the code.

nhatsmrt:
I'm not sure if this is what you're thinking of, but I just add a small test in the following format:
- Some updates
- newBranch, then revert on that branch
- check that the revert commit is head of the new branch, but the master branch's head is still the update commit.

Do tell me if you have a particular failure case that you have in mind. The branching behavior is controlled by versioningCommitToCommit helper function, which check if current instance of commit has a branch associated with it, in which case that branch's head is updated.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Hmm what do you have in mind? I can write a helper function that (optionally) take the other commit, check if either is unsaved (in which case a Failure is return). It can either just use a default error message (i.e "this commit is not saved", "other commit is not saved"), or require error messages to be passed to it. I'm not sure if it's worth the overhead though?

nhatsmrt: I just make the changes in the latest commit. See if it looks good.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -244,13 +244,7 @@ class Commit(
...
    */
   def merge(other: Commit, message: Option[String] = None)(implicit ec: ExecutionContext) = {
-    if (!saved)
-      Failure(new IllegalCommitSavedStateException("This commit must be saved"))
-    else if (!other.saved)
-      Failure(new IllegalCommitSavedStateException("Other commit must be saved"))
-    else if (other.repo.id != repo.id)
-      Failure(new IllegalArgumentException("Two commits must belong to the same repository"))
-    else
+    checkCommit("This commit must be saved", Some(other), Some("Other commit must be saved")).flatMap(_ => {
```
```suggestion
    checkCommit("This commit must be saved").flatMap(_ => other.checkCommit("Other commit must be saved")).flatMap(_ => {
```
seems more idiomatic, right? No need to worry about arbitrary number of arguments.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -333,9 +346,24 @@ class Commit(
...
+      createCommit(message = message, diffs = diff.blobDiffs, commitBase = id)
+    )
+  }
+
+  /** Check that the commit is saved. If other commit is passed, check that it is saved, and both are in the same repository
+   *  @param message error message if this commit is not saved
+   *  @param other (optional) the other commit
+   *  @param otherMessage (optional) the error message if the other commit is not saved
+   */
+  private def checkCommit(message: String, other: Option[Commit] = None, otherMessage: Option[String] = None): Try[Unit] = {
```
conradoverta: This method is pretty complex and trying to cover a particular case. I'd recommend building a base functionality and seeing if you can re-use that component instead of trying to cover everything possible in a single method. Composition is key.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Yeah it does, let me fix that.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Hmm how about split this into two methods: one for checking saved state, and one for checking if two commits belogn to the same repository?

nhatsmrt: I just separate it into checkSaved and checkSameRepository. See if it looks better.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Done.

conradoverta (approved):

PR #791, opened 2020-06-15
VR-4550 Support walking operation for Commit

nhatsmrt:
This should be merged after https://github.com/VertaAI/modeldb/pull/790, as there might be conflict in TestCommit.scala, but the core code should be ready for review.


client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -338,4 +338,59 @@ class Commit(
...
+      Stream(Failure(new IllegalCommitSavedStateException("Commit must be saved before it can be walked")))
+    else
+      continueWalk(List(List()))
+  }
+
+  /** Continue the walk from the locations passed
+   *  @param locations remaining locations to explore. Each location is in list of string format.
+   *  @return Stream of Trys of WalkOutputs. If the returned WalkOutput fails, abort the remaining locations.
+   */
+  def continueWalk(locations: List[List[String]])(implicit ec: ExecutionContext): Stream[Try[WalkOutput]] = {
```
conradoverta: Use stronger types. I think `List[String]` means something specific here, not a generic list. It also simplifies understanding.

client/scala/src/main/scala/ai/verta/repository/WalkOutput.scala
```diff
@@ -0,0 +1,14 @@
+package ai.verta.repository
+
+/** A class to represent the output of Commit's walk method
+ *  @param folderPath path to current folder
+ *  @param folderNames names of subfolders in current folder
+ *  @param blobNames names of blobs in current folder
+ *  @param remainingLocations list of remaining locations
+ */
+class WalkOutput(
```
conradoverta: What do you expect the user to do with this class? I'd recommend adding some methods to make it simpler for the user to know what to do.

nhatsmrt: Hmm I'm not quite sure what the user can do other than accessing the fields of WalkOutput. This is just a fancy version of Tuple4, making it more semantically meaningful.

client/scala/src/main/scala/ai/verta/repository/Commit.scala

client/scala/src/main/scala/ai/verta/repository/WalkOutput.scala
nhatsmrt: Update: I've added two helper functions to construct full path to blobs and subfolders. The first one is particularly useful because user can use it to retrieve blobs from commit.

conradoverta: Essentially, imagine the user is typing in their IDE and presses `.` in this object. This should list values that they can access and common actions they can make. For example, `Try.get` can be implemented via a `match`, but that's very verbose. So the action of getting the value is provided. Similarly for `toOption` and many others.

When you define a class, I recommend thinking about what the user will do with the class and give them a method for the most common cases, so that they don't have to see the objects available, interpret what they mean and their types, then thing which actions are available on those types.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -338,4 +338,78 @@ class Commit(
...
+    clientSet.versioningService.GetCommitComponent2(
+      commit_sha = id.get,
+      repository_id_repo_id = repo.id,
+      location = if (location.components.length > 0) Some(location.components) else None
+    ).flatMap(r => {
+        val folderPath = location.path
+        val responseFolder = r.folder
+
+        val folderNames = responseFolder
+        .flatMap(_.sub_folders) // Option[List[VersioningFolderElement]]
```
conradoverta: Fix indentation all cross this PR. Not sure if this was automatic, but doesn't look like the standard indentation.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -338,4 +338,78 @@ class Commit(
...
+        .map(_.map(folder => folder.element_name.get))
```
conradoverta: I believe this get might fail if you have no subfolders. I think the backend doesn't send the key instead of sending an empty list. Could you double check? Right now, it wouldn't be caught in a try.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -338,4 +338,78 @@ class Commit(
...
+        // Option[List[String]]
+        .map(_.sortWith((name1: String, name2: String) => name1 < name2))
```
conradoverta: There should be a standard sorter, no? No big deal, I was just surprised.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -338,4 +338,78 @@ class Commit(
...
+      case Failure(e) => Stream(Failure(e))
+      case Success(sf) => {
+        blobs match {
+          case Failure(e) => Stream(Failure(e))
+          case Success(bl) => {
+            val blobResults: Stream[Try[T]] =
+              filteredFolder.blobs.zip(bl).toStream.map(pair => Success(replacedWalker.visitBlob(pair._1, pair._2)))
+            val subfolderResults: Stream[Try[T]] =
+              filteredFolder.subfolders.zip(sf).toStream.map(pair => Success(replacedWalker.visitFolder(pair._1, pair._2)))
+
```
conradoverta, line 451: I recommend to generally avoiding `_` in compound functions like this. It's not clear if `_` will be the argument from the first or second function (e.g. if continue walk took a function with no arguments and returned a function with one argument, this line would look the same or very similar).

client/scala/src/main/scala/ai/verta/repository/PathList.scala
```diff
@@ -0,0 +1,15 @@
+package ai.verta.repository
+
+/** A class to represent a path in the list of strings format
+ *  @param components a list of components of the path
+ */
+final case class PathList(val components: List[String]) {
```
conradoverta: Final? Did you mean sealed? What is the limitation you had in mind?

conradoverta (commented):
Very minor things. Otherwise lgtm

client/scala/src/main/scala/ai/verta/repository/PathList.scala
nhatsmrt: I read somewhere that if I don't make case class final, user can extend it which leads to weird equality behavior. Probably overkill though. I'll remove it.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: You're right actually, as String has an implicit ordering we can just use List.sorted. Fixed.

client/scala/src/main/scala/ai/verta/repository/PathList.scala
nhatsmrt: Removed.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Huh weird. I'll manually align the method chaining. See if it looks better.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: I believe you meant to comment on the line below (the one with continueWalk function call). Anyway should be fixed.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Hmm responseFolder itself is an option, so I was trying to flatMap it to flatten the nested option structure and avoid exception at the same time. If my understanding is correct, then line 368 logic is:
- If responseFolder is None, output is None
- If responseFolder is not None, but subfolders is, then output is still None
- If both are defined, then it will be Some(List(...))

Let me write a test on empty commit to see if it cause exception.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
conradoverta: What formatter do you use?

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Wrote a test on empty commit. See if it covers the case we're thinking of.

conradoverta: Ok, maybe `element_name` can't actually be `None`. I checked this pretty late. Let me double check

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: None actually. I was just using the default scala plugin of Atom, along with an auto-indent plugin, which always seems to fail me, so I have to manually fix the indentation a lot of times. Is there one that you would recommend for Atom?

client/scala/src/main/scala/ai/verta/repository/PathList.scala
conradoverta: Ah, ok. `final` is ok here indeed. You can keep or remove.

conradoverta (approved):
LGTM

PR #792, opened 2020-06-15
Fix import

convoliution:
Followup to https://github.com/VertaAI/modeldb/pull/749

This wasn't caught by testing because it's for logging artifacts as observations, which the Client doesn't actually even support.


convoliution:
fyi @ravishetye

PR #796, opened 2020-06-16
VR-4578 remove support get attributes

evgeniilezhnin:


ravishetye (approved):
remove it from protos as well, I have verified that it is not being used

PR #798, opened 2020-06-16
VR-4744: Don't follow 302s

convoliution:
Neither `requests` nor `urllib3` has a filter for specific redirect codes, so the redirect chain has to be manually inspected.

### DIFF
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -266,22 +266,27 @@ def make_request(method, url, conn, **kwargs):
             # manually inspect redirects to stop on 302s
             for response in itertools.chain([response], s.resolve_redirects(response, request)):
                 if response.status_code == 302:
-                    # TODO: break without error if conn.ignore_conn_err
-                    raise RuntimeError(
-                        "received status 302 from {},"
-                        " which is not supported by the Client".format(response.url)
-                    )
-            # insert initial response back into history, b/c `resolve_redirects()` removes
+                    if not conn.ignore_conn_err:
+                        raise RuntimeError(
+                            "received status 302 from {},"
+                            " which is not supported by the Client".format(response.url)
+                        )
+                    else:
+                        # TODO: make it so this actually falls through
+                        break  # fall through to fabricate 200 response
             if response is not initial_response:
+                # insert initial response back into history, b/c `resolve_redirects()` removed it
                 response.history.insert(0, initial_response)
         except (requests.exceptions.BaseHTTPError,
                 requests.exceptions.RequestException) as e:
             if not conn.ignore_conn_err:
                 raise e
+            # else fall through to fabricate 200 response
         else:
             if response.ok or not conn.ignore_conn_err:
                 return response
-        # fabricate successful response
+            # else fall through to fabricate 200 response
+        # fabricate 200 response
         response = requests.Response()
         response.status_code = 200  # success
         response._content = six.ensure_binary("{}")  # empty contents
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -263,7 +263,7 @@ def make_request(method, url, conn, **kwargs):
             request = requests.Request(method, url, **kwargs).prepare()
             response = s.send(request, allow_redirects=False)
 
-            # manually inspect redirects to stop on 302s
+            # manually inspect initial response and subsequent redirects to stop on 302s
             history = []  # track history because `requests` doesn't since we're redirecting manually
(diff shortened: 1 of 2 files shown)


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -259,19 +260,55 @@ def make_request(method, url, conn, **kwargs):
...
         try:
-            response = s.request(method, url, **kwargs)
+            request = requests.Request(method, url, **kwargs).prepare()
+            response = s.send(request, allow_redirects=False)
+
+            # store initial response to add back into final history
+            initial_response = response
+
+            # manually inspect redirects to stop on 302s
+            responses = itertools.chain([initial_response], s.resolve_redirects(initial_response, request))
```
conradoverta: I don't understand what this is doing.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -259,19 +260,55 @@ def make_request(method, url, conn, **kwargs):
...
+                        raise RuntimeError(
+                            "received status 302 from {},"
+                            " which is not supported by the Client".format(response.url)
+                        )
+                    else:
+                        return fabricate_200()
+
+            if response is not initial_response:
+                # insert initial response back into history, b/c `resolve_redirects()` removed it
+                response.history.insert(0, initial_response)
```
conradoverta: I don't understand why this is necessary or what it is doing exactly

convoliution: `requests.Response` has a `history` sequence attribute that keeps track of redirects that occurred.
Because of how `requests` implemented `resolve_redirects()`, the initial response isn't included in the beginning of that history.

This is only necessary if something downstream needs an accurate response history, which my tests do.

client/verta/verta/_internal_utils/_utils.py
convoliution: `s.resolve_redirects(initial_response, request)` will yield the sequence of redirects that occur _after_ the initial response for the `for` loop below.
But I need to check the status code of that initial response too.

So I'm using `itertools.chain()` to yield the initial response for the first iteration of the loop, then proceed to iterate through `s.resolve_redirects()`.

client/verta/verta/_internal_utils/_utils.py
conradoverta: Oh, interesting. I'd assume history is irrelevant because it's not following redirects. Can you add a comment clarifying this?

client/verta/verta/_internal_utils/_utils.py
conradoverta: Got it

conradoverta (approved):

client/verta/verta/_internal_utils/_utils.py
convoliution: Oh ugh you're right, thanks. I failed to catch this but there actually isn't _any_ history recorded in the final response.

I'd still prefer to have that full history intact though, both for testing and because I think the caller shouldn't lose access to it just because redirects are being done manually.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -259,19 +260,53 @@ def make_request(method, url, conn, **kwargs):
...
+            responses = itertools.chain([response], s.resolve_redirects(response, request))
+            for response in responses:
+                if response.status_code == 302:
+                    if not conn.ignore_conn_err:
+                        raise RuntimeError(
+                            "received status 302 from {},"
+                            " which is not supported by the Client".format(response.url)
+                        )
+                    else:
+                        return fabricate_200()
```
conradoverta, line 277: Why would we fabricate a 200 on a 302? If the user wants to ignore errors, shouldn't we just leave it be and return the original response?

convoliution: For context, the way that the Client's `ignore_conn_err` flag is intended to function is that this util returns a non-error HTTP response with an empty JSON object in its body. Then when the body is loaded into protobuf objects by the caller in the Client, the message's fields will be populated with their default values. So the Client can continue to execute as much as possible even if it's receiving `4XX`s, `5XX`s, or connection errors.

So for me, the key part here as that empty JSON body, because it enables that default protobuf object behavior in the caller. The original `302` might not have that, but `fabricate_200()` does.

~Alternatively, I could reassign the content of the original `302`.~
Actually, altering its content is probably a bad idea since it might interfere with how `requests` handles connections. I think a new object has to be created.

conradoverta: Ok

conradoverta (approved):

PR #801, opened 2020-06-17
add support for empty commit

ravishetye:


conradoverta (commented):
We had fixed this before, no? I recall having a discussion with you on this case of empty commit while we were still in the office.

ravishetye:
i remember the discussion as well, my best guess is we did not follow through on it.

PR #805, opened 2020-06-17
VR-4871 applyDiff attach the new commit to the wrong parent.

nhatsmrt:
Found a bug in applyDiff: I attached the new commit to the wrong parent. With my old implementation, the new commit's parent will be the same as the old commit's parent. This behavior is wrong; the new commit's parent should be the old commit. I have also added 2 tests that demonstrate the case where my old implementation would break, using log and revert to demonstrate.


conradoverta (approved):
Nice catch

PR #806, opened 2020-06-17
Handle filepath in data_blob.download() to avoid collisions

convoliution:
On top of #780


convoliution:
Actually, I'll just put these commits directly into #780

PR #807, opened 2020-06-17
VR-4538, VR-4539 Support associating a commit to an experiment run and getting it

nhatsmrt:


nhatsmrt:
Right now, ExperimentRunCommit's toCommit method takes in a Client to convert itself into a Commit instance. This is because ExperimentRun doesn't have a field for Client. 

A more natural way would be to include Client as a field for ExperimentRun, then convert it right away, but this a pretty big change. Let me know if you have a preference, or there's a design principle that favors one way over the other.

conradoverta (approved):
LGTM

conradoverta (commented):
Actually, I have some feedback

client/scala/src/main/scala/ai/verta/client/entities/ExperiementRunCommit.scala
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Represent a commit associated with an experiment run.
+ *  User should not initialize this. Instances of this class are returned from ExperimentRun's getCommit
+ *  @param commitSHA commit SHA of the commit
+ *  @param repoId the id of the commit's repository
+ *  @param keyPaths A mapping between descriptive keys and paths of particular interest within commit.
+ */
+case class ExperimentRunCommit(
+  val commitSHA: String,
+  val repoId: BigInt,
```
conradoverta: Why don't you just save this as a commit? It resolves the issue of passing the client to `toCommit`. It also would remove the need for that method at all.

nhatsmrt: It's addressed in the comment I made above. ExperimentRun doesn't have client, so I cannot get repository, then get commit. Well, I can use clientSet to make API calls, but that's repetitive code. Other than that, there are a few options:
1. Pass client as a parameter to a method of ExperimentRunCommit (what I'm doing here)
2. Include client as a field of ExperimentRun, then use inside getCommit() of ExperimentRun. I'll have to make some changes to existing code/design though.
3. Pass client as a parameter to ExperimentRun's getCommit(). This keeps the change local, but looks kinda... unnatural to me, and also break consistency with Python client.

I'm not sure which one is the best design-wise, so I just went with the one with least change in existing code and waited for your input. What do you think?

nhatsmrt: In short, I need to decide where to pass "client" to call this:
```
client.getRepository(repository_id).flatMap(_.getCommitById(commitSHA))
```

nhatsmrt: OK I'll just use clientSet to make an API call. Doesn't look too bad. Fixed.

conradoverta (approved):

PR #810, opened 2020-06-17
VR-4874: Enable downloading folders

convoliution:
On top of #780 

I had forgotten how wild path manipulations can be.


client/verta/verta/dataset/_dataset.py
```diff
@@ -113,44 +169,48 @@ def download(self, component_path, download_to_path=None, chunk_size=32*(10**3))
...
+                        os.remove(tempf.name)
+                    raise e
+                else:
+                    print("download complete; file written to {}".format(components_to_download[path]))
+            finally:
+                response.close()
+
+        # TODO: return dir, not path, when user passed dir containing one file
+        # TODO: os.path.commonprefix is not the right tool here
+        return os.path.abspath(os.path.commonprefix(list(components_to_download.values())))
```
convoliution: This doesn't work for folder downloads

convoliution:
I'll bet there are some corner cases, but the core functionality is sufficient for upload, download, and rollback.

client/verta/verta/dataset/_dataset.py
```diff
@@ -93,14 +93,106 @@ def _set_commit_and_blob_path(self, commit, blob_path):
...
+                    local_path = download_to_path
+
+                return ({component_blob.path: local_path}, local_path)
+            elif component_blob.path.startswith(component_path_as_dir):
+                if implicit_download_to_path:
+                    # default to path relative to parent of `component_path`
+                    #     For example:
+                    #     component_blob.path = "coworker/downloads/data/info.csv"
+                    #     component_path      = "coworker/downloads"
+                    #     local_path          =          "downloads/data/info.csv"
```
conradoverta: ~So helpful!~ I crossed this after I noticed you set `local_path` twice and they have different values, so this was kind of helpful only.

client/verta/verta/dataset/_dataset.py
```diff
@@ -93,14 +93,106 @@ def _set_commit_and_blob_path(self, commit, blob_path):
...
+                    local_path = os.path.relpath(
```
conradoverta: Can we call this something else and add to your example above? I think it would be `data/info.csv`, right?

client/verta/verta/dataset/_dataset.py
```diff
@@ -93,14 +93,106 @@ def _set_commit_and_blob_path(self, commit, blob_path):
...
+                    )
+
+                components_to_download[component_blob.path] = local_path
+
+        if not components_to_download:
+            raise KeyError("no components found for path {}".format(component_path))
+
+        # determine the root path everything is being downloaded to
+        local_paths = map(os.path.relpath, components_to_download.values())  # relative to cwd
+        root_dirnames = set(local_path.split('/')[0] for local_path in local_paths)
```
conradoverta: I don't understand what this is doing.

conradoverta (commented):
I had a hard time with the end of the `_get_components_to_download`. Since you spent time to create an example in the comments, it would be useful to re-use it to give an example.

client/verta/tests/test_versioning/test_dataset.py
```diff
@@ -165,6 +170,72 @@ def test_repr(self):
...
+        assert os.path.isfile(filepath2)
+        assert filepath2 != filepath
+
+        # download to explicit path with overwrite
+        last_updated = os.path.getmtime(filepath)
+        filepath3 = dataset.download(s3_key, filepath)
+        assert filepath3 == filepath
+        assert os.path.getmtime(filepath) > last_updated
+
+    def test_mngd_ver_folder(self, commit, in_tempdir):
```
conradoverta, line 218: Check the content matches? It could have just created the path. Same all across.

client/verta/verta/dataset/_dataset.py
```diff
@@ -93,14 +93,106 @@ def _set_commit_and_blob_path(self, commit, blob_path):
...
+            downloaded_to_path = root_dirnames.pop()
+        else:
+            # everything is relative to cwd, so cwd is our root
+            # NOTE: we shouldn't end up here if `download_to_path` was provided
+            # TODO: handle if a `local_path` was actually at system root or a parent dir
+            downloaded_to_path = os.curdir
+
+        return (components_to_download, downloaded_to_path)
+
     def download(self, component_path, download_to_path=None, chunk_size=32*(10**3)):
```
conradoverta: Should we allow a component path to be empty and mean "everything"? I'd say yes.

client/verta/tests/test_versioning/test_dataset.py
```diff
@@ -165,6 +170,72 @@ def test_repr(self):
...
+        dirname = "tiny-files/"
+        s3_folder = "s3://verta-versioned-bucket/{}".format(dirname)
+        blob_path = "data"
+
+        dataset = verta.dataset.S3(s3_folder, enable_mdb_versioning=True)
+        commit.update(blob_path, dataset)
+        commit.save("Version data.")
+        dataset = commit.get(blob_path)
+
+        dirpath = dataset.download("s3://")
```
conradoverta, line 281: What happens when it's empty?

client/verta/verta/dataset/_dataset.py
convoliution: Ah will do.

I wrote this comment before I had the second assignment to `local_path` and forgot to update it

client/verta/verta/dataset/_dataset.py
```diff
@@ -93,14 +93,106 @@ def _set_commit_and_blob_path(self, commit, blob_path):
...
+                        download_to_path,
+                        os.path.relpath(component_blob.path, component_path),
+                    )
+
+                components_to_download[component_blob.path] = local_path
+
+        if not components_to_download:
+            raise KeyError("no components found for path {}".format(component_path))
+
+        # determine the root path everything is being downloaded to
```
conradoverta: I'm not sure if the segment of code below is properly covered in the tests, but probably because I don't understand the line below. I'll re-check later.

convoliution: It is not covered in tests! Will do

client/verta/tests/test_versioning/test_dataset.py
convoliution: If the string is empty? The user gets a
```
KeyError: 'no components found for path '
```

See my comment above about download-all behavior

client/verta/verta/dataset/_dataset.py
convoliution: I have VR-4875 for `dataset.download()` with no args to mean "download all"

client/verta/tests/test_versioning/test_dataset.py
convoliution: Will do

client/verta/verta/dataset/_dataset.py
convoliution: It is actually already `downloads/data/info.csv` here. The reassignment later is in case `downloads/` already exists, in which case `local_path` becomes `downloads 1/data/info.csv`. I'll try to clarify that in the comments.

client/verta/tests/test_versioning/test_dataset.py
convoliution: The "comment above" disappeared.
>I have VR-4875 for dataset.download() with no args to mean "download all"

client/verta/verta/dataset/_dataset.py
```diff
@@ -93,24 +93,107 @@ def _set_commit_and_blob_path(self, commit, blob_path):
...
+                else:
+                    # exactly where the user requests
+                    local_path = download_to_path
+
+                return ({component_blob.path: local_path}, os.path.abspath(local_path))
+
+        # figure out where files are going to be downloaded to
+        if implicit_download_to_path:
+            # automatically determine directory
+            # TODO: this is going have issues with `component_path` in {".", "..", "/"}
```
convoliution, line 137: The problem here is if the blob contains
```python
"../a.txt"
"../b.txt"
"../c.txt"
```
then this logic as-is would write to
```python
".. 1/a.txt"
".. 1/b.txt"
".. 1/c.txt"
```

In these cases, I'm going to need to go through the blobs and check their outermost path components against path collisions, and be able to write to
```python
"a.txt"
"b 1.txt"  # b.txt already exists
"c.txt"
```

conradoverta (approved):

convoliution:
After reviewing together during a sync with Conrado, this is to be merged-in as-is.

PR #811, opened 2020-06-18
modify add, update, get, delete datasetVersion attributes api

anandjakhaniya:


protos/protos/public/modeldb/DatasetVersionService.proto
```diff
@@ -209,6 +211,9 @@ message UpdateDatasetVersionAttributes {
 
 message GetDatasetVersionAttributes {
     string id = 1;
+    repeated string attribute_keys = 2;
+    bool get_all = 3;
```
ravishetye, line 215: why was this required, lets not do extra stuff without discussing on zoom / slack

ravishetye (approved):

PR #815, opened 2020-06-18
Make sure stream is correctly passed

convoliution:
#798 (don't follow `302`s) had split the logic in `make_request` into "build request" –> "send request", so the `stream` param needs to be specifically passed to the correct step.


convoliution:
Note that this bug only impacts managed versioning downloads, which hasn't been announced yet as a new feature.

PR #816, opened 2020-06-18
support for merge conflicts when different blobs are added/modified at same locatiom

ravishetye:


backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -69,43 +70,147 @@ public static BlobDiff setConflictBlobsInDiff(
...
+          diff2.setDataset(
+              getDatasetConflictBlob(a.getDataset(), b.getDataset(), c.getDataset()).get(0));
+          break;
+        case ENVIRONMENT:
+          diff2.setEnvironment(
+              getEnvironmentonflictBlob(a.getEnvironment(), b.getEnvironment(), c.getEnvironment())
+                  .get(0));
+          break;
+        case CONTENT_NOT_SET:
+          break;
```
anandjakhaniya: Should we extract this in the separate function to reduce complexity?

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -69,43 +70,147 @@ public static BlobDiff setConflictBlobsInDiff(
...
+          diff1.setDataset(
+              getDatasetConflictBlob(a.getDataset(), b.getDataset(), c.getDataset()).get(0));
+          break;
+        case ENVIRONMENT:
+          diff1.setEnvironment(
+              getEnvironmentonflictBlob(a.getEnvironment(), b.getEnvironment(), c.getEnvironment())
+                  .get(0));
+          break;
+        case CONTENT_NOT_SET:
+          break;
```
anandjakhaniya: Should we extract this and below comment in the separate function to reduce complexity? I think these two cases look the same.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -69,43 +70,147 @@ public static BlobDiff setConflictBlobsInDiff(
...
+    AutogenDockerEnvironmentDiff dockerDiff =
+        getDockerDiff(
+            a == null ? null : a.getDocker(),
+            b == null ? null : b.getDocker(),
+            c == null ? null : c.getDocker());
+    AutogenCommandLineEnvironmentDiff cliDiff =
+        getCommandLineDiff(
+            a == null ? null : a.getCommandLine(),
+            b == null ? null : b.getCommandLine(),
+            c == null ? Collections.emptyList() : c.getCommandLine());
```
anandjakhaniya, line 188: Instead of just `c == null ? Collections.emptyList() : c.getCommandLine());` can we add the same line for the `a` & `b`? i am not sure how much it will helpful.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -385,20 +497,29 @@ private static AutogenConfigDiff getConfigConflictBlob(
...
+            c == null ? null : c.getNotebook());
+    if (gitConflict != null && notebookConflict != null) {
+
+      return Arrays.asList(
+          Utils.removeEmpty(new AutogenCodeDiff().setGit(gitConflict)),
+          Utils.removeEmpty(new AutogenCodeDiff().setNotebook(notebookConflict)));
+    } else {
+      return Collections.singletonList(
+          Utils.removeEmpty(
+              new AutogenCodeDiff().setGit(gitConflict).setNotebook(notebookConflict)));
```
anandjakhaniya, line 523: If `gitConflict` & `notebookConflict` i think this line should throw an error because proto does not accept the `null` value. just need to cross-check.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
```diff
@@ -432,20 +553,28 @@ private static AutogenGitCodeDiff getGitDiff(
...
+            b == null ? null : b.getS3(),
+            c == null ? null : c.getS3());
+    if (pathConflict != null && s3Conflict != null) {
+
+      return Arrays.asList(
+          Utils.removeEmpty(new AutogenDatasetDiff().setPath(pathConflict)),
+          Utils.removeEmpty(new AutogenDatasetDiff().setS3(s3Conflict)));
+    } else {
+      return Collections.singletonList(
+          Utils.removeEmpty(new AutogenDatasetDiff().setPath(pathConflict).setS3(s3Conflict)));
```
anandjakhaniya, line 578: If `pathConflict` & `s3Conflict` i think this line should throw an error because proto does not accept the null value. just need to cross-check.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: that was the advantage of the Autogen classes.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: that was the advantage of the Autogen classes.

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: I am not sure if that will help a lot, also I would avoid touching it in this PR since it is just moved code with no new logic

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: done

backend/src/main/java/ai/verta/modeldb/versioning/blob/diff/ConflictGenerator.java
ravishetye: done

PR #821, opened 2020-06-19
VR-4906 Replace Any in KVHandler with a Union-like type

nhatsmrt:
I'll undraft this once I have added more documentation, tests, and a few helper stuff for conversion. But the gist of it is to use implicit conversion method as described in this:
https://stackoverflow.com/questions/48869110/collection-of-union-types-scala
I have also consulted this (the link that you sent):
https://stackoverflow.com/questions/3508077/how-to-define-type-disjunction-union-types
But I'm not sure how to make it work for Map[String, ValueType]... 
Anyway do take a look. This is completely new for me, so I might be missing some obvious things.


client/scala/src/test/scala/ai/verta/client/entities/TestExperimentRun.scala
```diff
@@ -35,11 +36,15 @@ class TestExperimentRun extends FunSuite {
 
     try {
        f.expRun.logMetric("some-metric", 0.5)
+       f.expRun.logMetric("int-metric", 4)
        f.expRun.logMetrics(Map("other-metric" -> 0.3, "other-metric-2" -> 0.1))
 
-       assert(f.expRun.getMetric("some-metric").get.get equals 0.5)
-       assert(f.expRun.getMetric("other-metric").get.get equals 0.3)
-       assert(f.expRun.getMetric("other-metric-2").get.get equals 0.1)
+       assertTypeError("f.expRun.logMetric(\"should-not-compile\", true)")
```
conradoverta, line 42: Nice

conradoverta (approved):
Lgtm

PR #823, opened 2020-06-19
Enable clientside artifact store

convoliution:


client/verta/verta/client.py
```diff
@@ -3324,13 +3338,21 @@ def get_artifact(self, key):
         """
         artifact, path_only = self._get_artifact(key)
         if path_only:
+            # might be clientside storage
+            # TODO: can cause problem if accidentally picks up unrelated file w/ same name
```
conradoverta: Which is somewhat likely since it's logging as path. But I'd say that, in this case, we can probably assume that either the user is using the location-based one or not, so mixup is less likely.

conradoverta (commented):
LGTM so far

conradoverta (approved):

PR #828, opened 2020-06-20
Implement run.download_artifact()

convoliution:


client/verta/verta/client.py
```diff
@@ -3360,7 +3360,81 @@ def get_artifact(self, key):
...
+        # get a stream of the file bytes, without loading into memory, and write to file
+        # TODO: consolidate this with _get_artifact() and get_artifact()
+        print("downloading {} from ModelDB".format(key))
+        if artifact.path_only:
+            if os.path.exists(artifact.path):
+                # copy from clientside storage
+                # TODO: use a tempfile first, and also delete if failed
+                with open(artifact.path, 'rb') as src_f:
+                    with open(download_to_path, 'wb') as dest_f:
+                        shutil.copyfileobj(src_f, dest_f)
```
conradoverta: Why not `copyfile`?

client/verta/verta/client.py
```diff
@@ -3360,7 +3360,81 @@ def get_artifact(self, key):
             except:
                 return six.BytesIO(artifact)
 
-    # TODO: download_artifact(self, key, download_to_path)
+    def download_artifact(self, key, download_to_path, chunk_size=32*(10**3)):
```
conradoverta:
```suggestion
    def download_artifact(self, key, download_to_path, chunk_size=32*(10**6)):
```
KB will cause way too much IO

client/verta/verta/client.py
convoliution: I think I just really keep gravitating towards using streams.

Also, my understanding was that `copyfile` just uses `copyfileobj` under the hood anyway, rather than an OS-level `cp`.

Actually, the Py3 docs say `copyfile` uses "the most efficient way possible", so I'll use that.

client/verta/verta/client.py
```diff
@@ -3360,7 +3360,78 @@ def get_artifact(self, key):
...
+        """
+        Downloads the artifact with name `key` to path `download_to_path`.
+
+        Parameters
+        ----------
+        key : str
+            Name of the artifact.
+        download_to_path : str
+            Path to download to.
+        chunk_size : int, default 32 kB
```
```suggestion
        chunk_size : int, default 32 MB
```

client/verta/verta/client.py
conradoverta: If it doesn't use OS-level `cp`, I'd be surprised to say the least.

PR #829, opened 2020-06-21
Check status code first to help debug test failure

convoliution:
These tests are failing because of [an issue with httpbin](https://github.com/postmanlabs/httpbin/issues/617).

If the status check happened first, it would've given a better indication as to what happened (a `404`).


convoliution:
Oh boy now that issue on httpbin has a link to this PR. That's awkward.

PR #830, opened 2020-06-21
delete cron job for oss

ravishetye:


backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
```diff
@@ -46,8 +46,8 @@ public static void initializeBasedOnConfig(
             LOGGER.info(
                 "{} cron job scheduled successfully", ModelDBConstants.UPDATE_PARENT_TIMESTAMP);
           } else if (cronJob.getKey().equals(ModelDBConstants.DELETE_ENTITIES)
-              && app.getServiceUserEmail() != null
-              && app.getServiceUserDevKey() != null) {
+              && ((app.getServiceUserEmail() != null && app.getServiceUserDevKey() != null)
+                  || !roleService.IsImplemented())) {
```
anandjakhaniya, line 50: If you are adding condition here then it will be Logging like https://github.com/VertaAI/modeldb/pull/830/files#diff-1386aca3d9ddfa5e45d08f7902d87dedR66 So if you can add `|| !roleService.IsImplemented()` in these `else if` statement then it will be helpful.

anandjakhaniya (approved):
Approved with minor change if it is make sense.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CronJobUtils.java
ravishetye: which is fine i think, since the corn job is not configured completely

PR #832, opened 2020-06-21
Unpickle clientside artifacts on get()

convoliution:
Followup to #823 

I neglected to include the unpickling logic for retrieving clientside artifacts.


conradoverta (approved):
LGTM

PR #835, opened 2020-06-22
Split charts

conradoverta:


chart/modeldb/charts/backend/templates/hpa.yaml
```diff
@@ -0,0 +1,13 @@
+{{ if and .Values.hpa.enabled (not .Values.artifactStore.enabled) }}
```
ravishetye, line 1: backend will come up in HA by default?

chart/modeldb/charts/backend/templates/service.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  ports:
+    - name: backend
+      port: 8085
+    - name: artifactstore
+      port: 8086
+    - port: 3000
+      name: http
+    - port: 3001
+      name: grpc
```
ravishetye: this was mainly because backend and proxy were on the same port , this is needed for OSS?

chart/modeldb/charts/backend/templates/statefulset.yaml
```diff
@@ -0,0 +1,116 @@
+{{- if .Values.artifactStore.enabled -}}
```
ravishetye, line 1: for the users it would be clear when to use stateful set versus deployment?

chart/modeldb/charts/backend/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+        RdbDialect: "org.hibernate.dialect.PostgreSQLDialect"
+        RdbUrl: "jdbc:postgresql://modeldb-postgresql:5432"
+        RdbUsername: postgres
+        RdbPassword: postgres
+
+  #ArtifactStore Properties
+  artifactStore_grpcServer:
+    host: localhost
+    port: 8086
+
```
ravishetye, line 89:
```suggestion
    port: 8086

  cron_job:
    update_parent_timestamp:
      frequency: 10 #frequency to update parent timestamp in second, default 60
      record_update_limit: 100
    delete_entities:
      frequency: 60 #frequency to update parent timestamp in second, default 60
      record_update_limit: 100
```

chart/modeldb/charts/postgres/templates/NOTES.txt
```diff
@@ -0,0 +1,43 @@
+** Please be patient while the chart is being deployed **
```
ravishetye, line 1: earlier when we did not have chart for postgres , were the charts being picked from some central repository automatically? was that better ? since now we will have to keep updating the charts?

chart/modeldb/charts/backend/templates/hpa.yaml
ravishetye: ok never mind , reread the comment and this will get enabled from values

chart/modeldb/charts/webapp/templates/deployment.yaml
```diff
@@ -0,0 +1,81 @@
...
+{{- end}}
+      annotations:
+        verta.ai/disable_anti_affinity: "true"
+    spec:
+      containers:
+        - image: {{ .Values.image }}:{{ .Values.imageTag }}
+          imagePullPolicy: Always
+          name: {{ .Release.Name }}
+          ports:
+            - containerPort: 3000
```
ravishetye, line 33: we had these settings for AMI : https://github.com/VertaAI/modeldb/blob/master/chart/modeldb/templates/modeldb-frontend-service.yaml#L9-L12 do you think they will no longer be needed ?

ravishetye (approved):
Mainly comments / questions due to lack of familiarity. LGTM if this works

chart/modeldb/charts/backend/templates/service.yaml
conradoverta: This were the ports for the proxy, it wasn't about collision. But the 3001 is indeed not exposed. Removing.

chart/modeldb/charts/backend/templates/statefulset.yaml
conradoverta: No, but that's why the flag is not called `statefulset` =) It's about a specific thing in MDB that, when enabled, forces one or the other.

chart/modeldb/charts/backend/values.yaml
conradoverta: @ravishetye please update the config for docker compose too. I just based off of that.

chart/modeldb/charts/postgres/templates/NOTES.txt
conradoverta: I'm not sure what "earlier" you are referring to. Charts should not be picked automatically. You need to pass the repository or location you want to use. Using an external one is doable, but I'm not sure it would work with subcharts. So it would be on the user to install the two completely separate systems and connect them, which is not nice.

chart/modeldb/charts/webapp/templates/deployment.yaml
conradoverta: AMI is different, since the system is self contained, so opening a node port is less of an issue than doing that in your cluster (e.g. with that config by default, you couldn't have two separate sets of the webapp running).

PR #839, opened 2020-06-22
Webapp update

AndreyKuznetsovQS:
1) removed owner info
2) Fixed displaying string for an observation.
3) Improved styles for the experiment run table pagination
4) added epoch number to observations
5) changed observations chart label


webapp/client/src/shared/models/Error.ts
```diff
@@ -12,6 +12,7 @@ export function isAppError(error: any): error is AppError<undefined> {
 export const createCodeErrorFromError = (error: Error): CodeError => {
   const codeError = new CodeError({ message: error.message });
   codeError.stack = error.stack;
+  console.error(codeError);
```
CaptEmulation, line 15: nope

CaptEmulation: though maybe it is needed because otherwise we wouldn't see any error anywhere....

PR #843, opened 2020-06-22
VR-4885: date created & updated on ExperimentRun

convoliution:
Note that the `__repr__` shows the time in the user's system's local time, like the WebApp does.
But the methods return the Unix timestamps in UTC for potentially programmatic applications.

Screen Shot 2020-06-22 at 2 36 03 PM


conradoverta (approved):
LGTM

PR #851, opened 2020-06-23
New data versioning interface for Dataset trait + prepareForUpload for S3

nhatsmrt:


client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -18,7 +17,12 @@ import scala.annotation.tailrec
  *  }}}
  *  If an invalid path is passed to the constructor, it will return a failure.
  */
-case class PathBlob(protected val contents: HashMap[String, FileMetadata]) extends Dataset {
+case class PathBlob(
+  protected val contents: HashMap[String, FileMetadata],
+  protected val enableMDBVersioning: Boolean = false
+) extends Dataset {
+  override def prepareForUpload() = ???
```
conradoverta, line 24: This will return an error. Is that the intention?

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -27,28 +31,42 @@ case class PathBlob(protected val contents: HashMap[String, FileMetadata]) exten
...
+   *  @return if the path is invalid, a failure along with exception message. Otherwise, the pathblob (wrapped in success)
+   */
+  def apply(path: String, enableMDBVersioning: Boolean): Try[PathBlob] = apply(List(path), enableMDBVersioning)
+
+  /** Constructor taking only one path. Does not version the data in the blob with ModelDB
    *  @param path a single path
    *  @return if the path is invalid, a failure along with exception message. Otherwise, the pathblob (wrapped in success)
    */
-  def apply(path: String): Try[PathBlob] = apply(List(path))
+  def apply(path: String): Try[PathBlob] = apply(List(path), false)
```
conradoverta:
```suggestion
  def apply(path: String): Try[PathBlob] = apply(List(path))
```
I recommend doing one step at a time

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
```diff
@@ -60,7 +78,10 @@ object PathBlob {
    */
   def reduce(firstBlob: PathBlob, secondBlob: PathBlob): Try[PathBlob] = {
     if (firstBlob.notConflicts(secondBlob))
-      Success(new PathBlob(firstBlob.contents ++ secondBlob.contents))
+      Success(new PathBlob(
+        firstBlob.contents ++ secondBlob.contents,
+        firstBlob.enableMDBVersioning && secondBlob.enableMDBVersioning
```
conradoverta: Maybe cause an error if they're different? Silencing such an important and explicit request from the user seems bad.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -20,13 +22,61 @@ import scala.annotation.tailrec
...
+    val request =
+      if(location.versionID.isDefined)
+        new GetObjectRequest(location.bucketName, location.key.get, location.versionID.get)
+      else
+        new GetObjectRequest(location.bucketName, location.key.get)
+
+    val obj: S3Object = s3.getObject(request)
+    val s3InputStream: S3ObjectInputStream = obj.getObjectContent()
+    val fileOutputStream: FileOutputStream = new FileOutputStream(file)
+    val buffer = new Array[Byte](1024)
```
conradoverta: At least 1MB. If you use 1KB, you'll drown on IO. The current max disk IO in a single request (which gives max throughput) is 32MB.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -20,13 +22,61 @@ import scala.annotation.tailrec
...
+    var readLen = s3InputStream.read(buffer)
+
+    while (readLen > 0) {
+        fileOutputStream.write(buffer, 0, readLen)
+        readLen = s3InputStream.read(buffer)
+    }
```
conradoverta: Isn't there a function to do this for you automatically? I bet there's some optimized thing to copy form input to output stream.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -20,13 +22,61 @@ import scala.annotation.tailrec
...
+
+    s3InputStream.close()
+    fileOutputStream.close()
```
conradoverta: Can any of the functions above raise an exception? I'd say probably yes. If yes, we should make this close safe.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -37,23 +87,41 @@ case class S3(protected val contents: HashMap[String, FileMetadata]) extends Dat
 object S3 {
   private val s3: AmazonS3 = AmazonS3ClientBuilder.standard().build()
 
-  /** Constructor taking only one S3Location
+  /** Constructor taking only one S3Location. Does not version with ModelDB.
    *  @param location a single S3Location
    *  @return if location is invalid, a failure along with exception message. Otherwise, the blob (wrapped in success)
    */
-  def apply(location: S3Location): Try[S3] = apply(List(location))
+  def apply(location: S3Location): Try[S3] = apply(List(location), false)
```
conradoverta:
```suggestion
  def apply(location: S3Location): Try[S3] = apply(List(location))
```

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
```diff
@@ -76,7 +144,10 @@ object S3 {
    */
   def reduce(firstBlob: S3, secondBlob: S3): Try[S3] = {
     if (firstBlob.notConflicts(secondBlob))
-      Success(new S3(firstBlob.contents ++ secondBlob.contents))
+      Success(new S3(
+        firstBlob.contents ++ secondBlob.contents,
+        firstBlob.enableMDBVersioning && secondBlob.enableMDBVersioning
```
conradoverta: Same

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Yes, but the whole section is wrapped in a Try. If there is an exception it should be caught by it, right?

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: No idea lol. I literally copy pasted this section from Amazon S3 documentation =) I'll figure sth out.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Yes. I plan to implement one thing at a time. This PR is already way too big.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/PathBlob.scala
nhatsmrt: Fixed + added a test that demonstrates the new expected behavior.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed + added a test that demonstrates the new expected behavior.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: There is, but it only exists for Java 9 I think:
https://stackoverflow.com/questions/43157/easy-way-to-write-contents-of-a-java-inputstream-to-an-outputstream
Other answers mostly involve some third party libraries.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
conradoverta: Got it.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
conradoverta: Yes, but the streams won't be properly closed. You need a `try {} finally {}` block inside the `Try` constructor.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -38,9 +50,14 @@ trait Dataset extends Blob {
     val shared = contents.keySet.intersect(other.contents.keySet)
     contents.filterKeys(shared).equals(other.contents.filterKeys(shared))
   }
+
+  /** Clean up the uploaded components */
+  protected def cleanUpUploadedComponents(uploadMap: Map[String, UploadComponent]): Try[Unit] = Success(())
 }
 
 object Dataset {
+  private val BufferSize = 8192
```
conradoverta: We should also increase this buffer for the same reason. I missed it before. I also generally recommend placing the constant close to where they're used.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -53,4 +70,34 @@ object Dataset {
...
+     val dis = new DigestInputStream(
+       new FileInputStream(file),
+       messageDigest
+     )
+
+     try {
+       while (dis.read(buffer) != -1) {}
+     } finally {
+       dis.close()
+     }
```
conradoverta, line 97: Not that here you did the right thing to close the stream.

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: Oh I see what you're saying. This is annoying.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+  test("The downloaded s3 file should not be corrupted") {
+    val f = fixture
+    val s3Blob = S3(f.bucketLoc).get
+    val s3PathToLocalMap = s3Blob.prepareForUpload().get.mapValues(comp => comp.filePath)
+
+    val localPath = s3PathToLocalMap.get(f.testfilePath).get
+    val pathBlob = PathBlob(localPath).get
+    assert(pathBlob.getMetadata(localPath).get.md5 equals s3Blob.getMetadata(f.testfilePath).get.md5)
```
conradoverta: This is checking the two md5s are equal, right? What happens if the md5 computation wrapper that you have is wrong and returns empty strings always?

client/scala/src/main/scala/ai/verta/blobs/dataset/S3.scala
nhatsmrt: OK fixed. Variables declared in try block cannot be referred to in finally block, so I have to use a getaround via Option[]. Very interesting pattern, but not sure if it's sound design-wise. See if it looks good.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
```diff
@@ -0,0 +1,80 @@
...
+      val testDir = workingDir + "/src/test/scala/ai/verta/blobs/testdir"
+      val testfile = testDir + "/testfile"
+      val testSubdir = testDir + "/testsubdir"
+      val testfile2 = testSubdir + "/testfile2"
+    }
+
+  test("The downloaded s3 file should not be corrupted") {
+    val f = fixture
+    val s3Blob = S3(f.bucketLoc).get
+    val s3PathToLocalMap = s3Blob.prepareForUpload().get.mapValues(comp => comp.filePath)
```
conradoverta: I'd recommend testing this at the commit level since that's what customers will use. Right now, there's some code between `commit.save` and `blob. prepareForUpload` that is not tested.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
nhatsmrt: The s3Blob's MD5 computation is actually handled by AWS S3 (I populate the MD5 field of the Metadata by the one returned by S3 getETag() method), so if the computation is wrong for PathBlob, or if downloading the file from S3 corrupts the file, this would fail the test (which is the correct expected behavior). 

(Also for what it's worth I actually checked the output with Michael's Python client some time ago. That was how I found a bug in my hash method =) ).

We're removing this test anyway, but I'll keep this in mind.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
nhatsmrt: Hmm OK. I'll remove this test and test this more carefully at Commit level when we actually implemented download. Right now I agree that this is not very helpful.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Acknowledged.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
nhatsmrt: Removed.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Fixed.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
conradoverta: ETags are not exactly the same as the MD5 in the full file

nhatsmrt: Huh really. I was under the impression that they were the same. In fact I performed a very similar check before: 
- Upload a file to S3
- Construct PathBlob and S3 blob of this same file
- Compare the MD5 of the two blobs

Maybe the file is too small? BTW I think that Python client is also using this:
https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/dataset/_s3.py#L146

nhatsmrt: Maybe this is relevant:
https://teppen.io/2018/06/23/aws_s3_etags/
So it's not comparable in the case of multi upload. My small files are singly uploaded, so the MD5 will = MD5 of the single file.

But yeah other than this check there's no other way to check the MD5 computation that I can think of. It's quite dangerous.

nhatsmrt: Update: It seems there is a way to get MD5 from S3 Metadata Object:
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/ObjectMetadata.html#getContentMD5--
The encoding is different though (not sure if it's convertable between the two?). And not sure if I should replace the MD5 field from ETag to this (considering it would break consistency with Python Client).

Update2: This is not provided by VersionSummary (the class we use when user want to retrieve an S3 object with version ID). This is sad. 
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/S3VersionSummary.html

conradoverta: It is fine to use small files and etags. I was just pointing out it's not a universal truth. But my original question remained.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
conradoverta: @nhatsmrt please file a follow-up

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
nhatsmrt: Hmm your question was if my MD5 computation is wrong (e.g always return empty string) what would happen right?
There are the following cases:
- Small file: Test would fail. ETag return the correct MD5. My computation gives the wrong MD5.
- Big file: Test would always fail (unless somehow ETag magically returns the same wrong MD5 as well).

Not sure how to verify for big file yet. This check is rather defective in that sense. But at least it won't hide the problem (i.e false positive - passing tests even though there is a bug).

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
nhatsmrt: @conradoverta wait what do you mean by follow-up? Do you want me to write a test now or wait until the stuff inside Commit is implemented? I was under the impression that you just wanted me to remove this test and wait until we have more stuff to do integration tests.

Or a Jira ticket to keep track of this test?

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
conradoverta: Well, there's also a third case: the code is buggy. Replace the hash computation with returning an empty string and you'll see.

client/scala/src/test/scala/ai/verta/blobs/dataset/TestVersioning.scala
conradoverta: Jira ticket. We can merge this in as-is, but we should have a reminder that this needs to be properly tested once commit is in.

conradoverta (approved):

PR #853, opened 2020-06-23
Fix webapp BACKEND_API_DOMAIN for logging

conradoverta:
This doesn't change functionality, it just fixes the logs.

### DIFF
--- a/docker-compose-all.yaml
+++ b/docker-compose-all.yaml
@@ -61,7 +61,7 @@ services:
         environment:
             DEPLOYED: "yes"
             BACKEND_API_PROTOCOL: "http"
-            BACKEND_API_DOMAIN: "modeldb-proxy:8080"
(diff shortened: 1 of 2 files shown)


docker-compose-all.yaml
```diff
@@ -61,7 +61,7 @@ services:
         environment:
             DEPLOYED: "yes"
             BACKEND_API_PROTOCOL: "http"
-            BACKEND_API_DOMAIN: "modeldb-proxy:8080"
+            BACKEND_API_DOMAIN: "modeldb-webapp:8080"
```
ravishetye: shouldn't the port change as well?

conradoverta:
```suggestion
            BACKEND_API_DOMAIN: "modeldb-webapp:3000"
```

conradoverta: Good idea.

ravishetye (approved):

PR #861, opened 2020-06-24
Fix review comments of integration dataset version

anandjakhaniya:
- Make dataset_id optional and look up the repo and error out if the commit belongs to multiple dataset repository
- Removed RepositoryAccessModifierEnum.RepositoryAccessModifier


ravishetye:
we need to move these checks to Dataset{Version} Service instead of generic {Blob,Repo}DAO and error out specific to Dataset

anandjakhaniya:
Follow up PR for the above [comment](https://github.com/VertaAI/modeldb/pull/861#issuecomment-649017192) PR: https://github.com/VertaAI/modeldb/pull/885

backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
```diff
@@ -507,25 +544,77 @@ public boolean deleteCommits(
...
+                  + "' associated with multiple datasets",
+              Code.INTERNAL);
+        }
+        Long newRepoId = new ArrayList<>(commitEntity.getRepository()).get(0).getId();
+        repositoryIdentification.setRepoId(newRepoId);
+      } else {
+        repositoryIdentification.setRepoId(Long.parseLong(datasetId));
+      }
+      repositoryEntity = repositoryDAO.getRepositoryById(repositoryIdentification.build(), true);
+      addDeleteCommitTags(
```
ravishetye: should this be label?

ravishetye: ok verfied that the function is actually adding label , it is just named incorrectly

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -279,10 +272,6 @@ public RepositoryEntity getRepositoryById(
                           "Couldn't find repository by id " + id.getRepoId(), Code.NOT_FOUND));
     }
     try {
-      if (checkProtected && repository.isProtected()) {
-        throw new ModelDBException(
-            "Can't access repository because it's protected", Code.PERMISSION_DENIED);
-      }
```
ravishetye, line 285: I do not think this is right, we had this because we did not want the protected repos to be accessed via RepoService

backend/src/main/java/ai/verta/modeldb/entities/versioning/RepositoryEntity.java
```diff
@@ -214,10 +204,6 @@ public Integer getRepository_visibility() {
     return repository_visibility;
   }
 
-  public boolean isProtected() {
-    return Objects.equals(repositoryAccessModifier, RepositoryAccessModifier.PROTECTED.getNumber());
-  }
```
ravishetye, line 219: we should have the flag just that we should not have the enum in protos

backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
anandjakhaniya: Ok, Change makes sense to me. I was set `tag` because for dataset version we called it to tags.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: Ok, Your point is right. I misunderstood that point.

backend/src/main/java/ai/verta/modeldb/entities/versioning/RepositoryEntity.java
anandjakhaniya: Ok, Got it. then column related fixes we should keep.

anandjakhaniya:
All changes and your points are valid. I am approving PR.

PR #866, opened 2020-06-24
Use walk_file() and assert_dirs_match() utils in tests

convoliution:
Cut down on that code duplication!

### DIFF
--- a/client/verta/tests/test_versioning/test_dataset.py
+++ b/client/verta/tests/test_versioning/test_dataset.py
@@ -15,6 +15,18 @@ def walk_files(dirpath):
             yield os.path.join(root, filename)
 
 
+def assert_dirs_match(dirpath1, dirpath2):
+    for filepath1 in walk_files(dirpath1):
+        # get corresponding path in dirpath2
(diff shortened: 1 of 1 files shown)


client/verta/tests/test_versioning/test_dataset.py
```diff
@@ -9,6 +9,24 @@
 import verta.dataset
 
 
+def walk_files(dirpath):
+    for root, _, filenames in os.walk(dirpath):
+        for filename in filenames:
+            yield os.path.join(root, filename)
+
+
+def assert_dirs_match(dirpath1, dirpath2):
```
conradoverta, line 25: If I'd be a little bit annoying, I could say that `dirpath2` can have more files than `dirpath1`. Not sure if it's worth covering that.

conradoverta (approved):

client/verta/tests/test_versioning/test_dataset.py
convoliution: Ooooooh how did I miss that??
That's definitely something I'd want covered by this.

PR #876, opened 2020-06-26
No more log_dataset() as artifact

convoliution:
We're deprecating dataset-as-artifact, and it's causing problems, so I'm removing it from examples.


convoliution:
Shoot I did not mean to merge that

PR #879, opened 2020-06-26
VR-4763: Enable dropping `base_path` for versioning

convoliution:


client/verta/verta/dataset/_path.py
```diff
@@ -45,17 +48,39 @@ class Path(_dataset._Dataset):
         ])
 
     """
-    def __init__(self, paths, enable_mdb_versioning=False):
+    def __init__(self, paths, base_path=None, enable_mdb_versioning=False):
```
conradoverta, line 50: Be careful about breaking API (e.g. order of arguments). No big deal in this case

client/verta/verta/dataset/_path.py
```diff
@@ -117,8 +142,14 @@ def _prepare_components_to_upload(self):
         for component_blob in self._path_component_blobs:
             component_path = component_blob.path
 
-            # TODO: when stripping base path is implemented, reconstruct original path here
-            filepath = os.path.abspath(component_path)
+            # reconstruct original filepaths with removed `base_path`s
+            for base_path, component_paths in self._base_path_to_component_paths.items():
+                if component_path in component_paths:
+                    filepath = os.path.join(base_path, component_path)
```
conradoverta: This seems like a recipe for disaster. If I understand correctly, this code is trying to guess which components had a base path or not. Besides being confusing, it's also storing the same information in two separate locations. What about this alternative: we don't store everything in a protobuf directly and try to compensate for the lack of flexibility later. Using the protobuf as storage mechanism doesn't really gain us anything besides typing a few letters. If you keep track of the components yourself, then you can have each component be a NamedTuple like `ComponentPath(file_location, base_path, ...)`. You can then just add more information for each component isolated from the others, but with everything you need to operate on that specific component. Distributing information about the same thing on multiple locations with some business logic to glue them together is more error prone than not doing it.

Let me know if it makes sense or if you'd like some clarification of the suggestion.

client/verta/verta/dataset/_path.py
convoliution: ack

client/verta/verta/dataset/_path.py
convoliution: Certainly, I'm onboard with that.

The reason I didn't shift away from the protobuf for this ticket is because _all_ the blob types use protobufs as storage, and the [implementation of](https://github.com/VertaAI/modeldb/blob/405a01a/client/verta/verta/_repository/commit.py#L872) `commit.get()` relies on that uniformity to return blobs. It's going to take a wider refactor to accommodate this, but I can try to get that done today

conradoverta: Maybe something for future Michael to consider is to avoid using wire format for internal storage ;)

convoliution: Yep yep 2000% agreed—we've talked about this before, too, but it was just after the blobs were originally implemented 😭

convoliution: PRs #891, #892, #893, and #896 consolidate `dataset._msg`, `dataset._components_to_upload`, and this PR's `dataset._base_path_to_component_paths` into a more flexible (and picklable!) `Component` struct.

conradoverta (approved):
Much cleaner, thanks =)

PR #886, opened 2020-06-26
Downloading Dataset blobs from ModelDB, Part 1

nhatsmrt:
A huge chunk of this PR is just testing code. Note that I haven't implemented in this PR:
- Implicit local path to download to (in https://github.com/VertaAI/modeldb/pull/914)
- Remove partially downloaded files if the download attempt fails in (in https://github.com/VertaAI/modeldb/pull/914)


client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -5,14 +5,110 @@ import ai.verta.blobs._
...
+   *  @return Whether the download attempts succeed.
+   */
+  def download(
+    componentPath: Option[String] = None,
+    downloadToPath: String
+  )(implicit ec: ExecutionContext): Try[Unit] = {
+    /** TODO: Make downloadToPath optional */
+
+    if (!enableMDBVersioning)
+      Failure(new IllegalStateException("This blob did not allow for versioning"))
```
conradoverta, line 38: Is there a nice way the user can ask if a blob had versioning enabled?

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -5,14 +5,110 @@ import ai.verta.blobs._
...
+    else if (downloadFunction.isEmpty || blobPath.isEmpty)
+      Failure(new IllegalStateException(
+        "This dataset cannot be used for downloads. Consider using `commit.get()` to obtain a download-capable dataset"
```
conradoverta, line 41: Under which scenarios would this happen?

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -5,14 +5,110 @@ import ai.verta.blobs._
...
+   *  @return Map of component paths to local destination paths
+   */
+  private def determineComponentAndLocalPaths(
+    componentPath: Option[String] = None,
+    downloadToPath: String
+  ): Map[String, String] = {
+    if (componentPath.isEmpty) {
+      // download entire blob
+      val downloadToPaths =
+        listPaths.map(comp => joinPaths(downloadToPath, removePrefixDir(comp, "s3:")))
```
conradoverta, line 87: Does this match what the client does?

conradoverta (commented):
Looks reasonable. A couple of comments

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: I can expose this field to the user (but keep it immutable, so that user can't mess with it)

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: I believe so. This is the corresponding section (according to my understanding of Michael's implementation):
https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/dataset/_dataset.py#L138

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
conradoverta: I'd recommend that. Otherwise the user can't escape the failure here.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: So currently the way Python client work (which I'm trying to emulate) is that:
```scala
val originalBlob = PathBlob("some-path", true).get
val newCommit = commit.update("path-blob", originalBlob).flatMap(_.save("some-msg")).get

val downloadingBlob = newCommit.get("path-blob").get
downloadingBlob.download()

originalBlob.download() // this would fail
```
So only a blob obtained via commit.get can download stuff

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Fixed.

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: Added a small test to demonstrate the expected behavior (i.e failures).

nhatsmrt:
I'm temporarily drafting this PR, as there are some super weird bugs, which I'm not 100% sure the cause.

conradoverta (approved):

PR #867, opened 2020-06-24
VR-4857 Support uploading data to ModelDB (all at once)

nhatsmrt:
On top of https://github.com/VertaAI/modeldb/pull/854
Note that no real test is provided for this PR (again). I did check the artifact_store in my local filesystem to make sure that the uploaded files are there, but without download I cannot write proper test. I'll write tests once I've implemented download (which should be the next PR)

PR #888, opened 2020-06-26
VR-4935 Clean up after uploading data

nhatsmrt:
On top of https://github.com/VertaAI/modeldb/pull/867.
Not sure how to write test for this one yet, other than visual inspection, or poking on private field.


conradoverta:
> On top of #867.
> Not sure how to write test for this one yet, other than visual inspection, or poking on private field.

You could allow the user to setup where the temp files would be located and check that it was empty after.

conradoverta (approved):

conradoverta (changes requested):
Actually, can you do a try per element first? Let's delete as much as we can first. Right now, it will stop at the first error, I believe, right?

nhatsmrt:
I believe this would do the trick:
```
      Try(contents.values.map(_.localPath.get).map(path => Try((new File(path)).delete())).map(_.get))
```
Very subtle difference, but the delete call is now wrapped in a Try, so if one fails the rest will still be deleted. Looks good?

conradoverta (approved):

PR #889, opened 2020-06-27
Implement changes to support a custom Minio Endpoint

Atharex:
The changes include some pom.xml adaptations from a minio S3 java client example

The modifications support using a minio endpoint, when it is specified in the configuration

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -25,25 +25,11 @@
 		<liquibase.version>3.8.0</liquibase.version>
 		<prometheus.version>0.8.1</prometheus.version>
 		<protobuf.version>3.8.0</protobuf.version>
-		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<aws.sdk.version>1.11.106</aws.sdk.version>
 
 		<sonar.core.codeCoveragePlugin>jacoco</sonar.core.codeCoveragePlugin>
 		<sonar.jacoco.reportPaths>${project.build.directory}/coverage-reports/jacoco-ut.exec</sonar.jacoco.reportPaths>
 		<sonar.language>java</sonar.language>
 	</properties>
-
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>com.amazonaws</groupId>
-                <artifactId>aws-java-sdk-bom</artifactId>
-                <version>1.11.794</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-        </dependencies>
-    </dependencyManagement>
 
 	<dependencies>
 
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -25,8 +25,8 @@
 		<liquibase.version>3.8.0</liquibase.version>
 		<prometheus.version>0.8.1</prometheus.version>
 		<protobuf.version>3.8.0</protobuf.version>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <aws.sdk.version>1.11.106</aws.sdk.version>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<aws.sdk.version>1.11.106</aws.sdk.version>
 
 		<sonar.core.codeCoveragePlugin>jacoco</sonar.core.codeCoveragePlugin>
 		<sonar.jacoco.reportPaths>${project.build.directory}/coverage-reports/jacoco-ut.exec</sonar.jacoco.reportPaths>
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -263,14 +263,14 @@
 					</execution>
 				</executions>
 			</plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
-                </configuration>
-            </plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<source>1.8</source>
+					<target>1.8</target>
+				</configuration>
+			</plugin>
 			<plugin>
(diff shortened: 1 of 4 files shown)


backend/pom.xml
```diff
@@ -25,12 +25,26 @@
 		<liquibase.version>3.8.0</liquibase.version>
 		<prometheus.version>0.8.1</prometheus.version>
 		<protobuf.version>3.8.0</protobuf.version>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ravishetye: indentation is off

backend/pom.xml
```diff
@@ -218,7 +232,7 @@
 	</dependencies>
 
 	<build>
-		<finalName>modeldb-${version}-client-build</finalName>
+		<finalName>modeldb-${project.version}-client-build</finalName>
```
ravishetye, line 221: was this necessary?

backend/pom.xml
```diff
@@ -25,12 +25,26 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>com.amazonaws</groupId>
+                <artifactId>aws-java-sdk-bom</artifactId>
+                <version>1.11.794</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
ravishetye: around line 139 we do have dependency on
```
<artifactId>aws-java-sdk-s3</artifactId>
<version>1.11.794</version>
```

A quick read on sdk-bom tells me it is useful when downloading multiple artifacts from com.amazonaws, that does not seem to be the case in modeldb. Was this change necessary?

backend/pom.xml
```diff
@@ -249,6 +263,14 @@
 					</execution>
 				</executions>
 			</plugin>
+            <plugin>
```
ravishetye: indentation

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/S3Service.java
```diff
@@ -30,18 +33,37 @@
...
+  private String keyName;
+  private String UploadFileName;
 
   public S3Service(String cloudBucketName) {
     App app = App.getInstance();
     String cloudAccessKey = app.getCloudAccessKey();
     String cloudSecretKey = app.getCloudSecretKey();
-    if (cloudAccessKey != null && cloudSecretKey != null) {
+    String minioEndpoint = app.getMinioEndpoint();
+    if (cloudAccessKey != null && cloudSecretKey != null && minioEndpoint == null) {
```
ravishetye: I think this will improve the readability

```
if (cloudAccessKey != null && cloudSecretKey != null){
 if(minioEndpoint ==null){
 ...
 } else {
  //your new code 
 }
} else { 
  //read from env
}
```

ravishetye (changes requested):
Thanks for your contribution @Atharex .
I had few questions and suggestions.

backend/pom.xml
```diff
@@ -25,12 +25,26 @@
 		<liquibase.version>3.8.0</liquibase.version>
 		<prometheus.version>0.8.1</prometheus.version>
 		<protobuf.version>3.8.0</protobuf.version>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <aws.sdk.version>1.11.106</aws.sdk.version>
```
ravishetye: this is unused?

backend/pom.xml
Atharex: True, I used spaces there... Replaced them with tabs. Should be ok now

backend/pom.xml
Atharex: when running `mvn package -Dmaven.test.skip=true` I got this warning in the output:

```
[WARNING]
[WARNING] Some problems were encountered while building the effective model for ai.verta.modeldb:modeldb:jar:1.0-SNAPSHOT
[WARNING] The expression ${version} is deprecated. Please use ${project.version} instead.
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
```

This is the reason why I've changed it. But can also revert it, if you believe this shouldn't be included!

backend/pom.xml
Atharex: Also used spaces there... replaced them now with tabs

backend/pom.xml
Atharex: No, it was not necessary. I've tested it also without sdk-bom and it worked, so I'll remove this bit

backend/pom.xml
Atharex: seems so... I was using a prebuilt minio example from another project and inadvertently transferred over more than was needed it seems. I've also removed this now

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/S3Service.java
Atharex: Agreed. I've adapted the file to this structure and I've also removed a couple of variables which were not used 
```
-  private String keyName;
-  private String UploadFileName;
```

Atharex:
Even though this adaption enables artifact storage to Minio, the artifacts can only be downloaded from Minio itself, but not from the ModelDB UI.

I guess the problem lies with the path to the file stored in the DB. Even when using a custom Minio endpoint, the path to the artifact stored in Postgres still points to Amazon s3 as the base URL
`s3://modeldb-bucket/0b4c3469bc2751f33d7f32c17ba28e92efc4eb17449f25d4283f032e38d71549/dataset.pkl`

backend/pom.xml
ravishetye: ok makes sense , thanks for the change and the explanation

ravishetye:
Thanks for the analysis. Would you like to fix that in a separate PR? someone from the team can guide you with the relevant code section.

Out team could fix this as well, but if will be helpful how the URL for minio should look like.

ravishetye (approved):
Thanks @Atharex for the fix . This LGTM to be merged in.

PR #890, opened 2020-06-28
VR-5007 Getting a PathBlob with empty component fails

nhatsmrt:
Small bug: When the size of a component is none, my current implementation fails when I attempt to get that blob back. I also provided a test that actually failed when I ran against my old code, and passed when I ran against my new implementation.

### DIFF
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
@@ -61,9 +61,9 @@ object Dataset {
      component: VersioningPathDatasetComponentBlob,
      versionId: Option[String] = None
    ) = new FileMetadata(
-     component.last_modified_at_source.get,
-     component.md5.get,
-     component.path.get,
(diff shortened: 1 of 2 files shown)


client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -64,7 +64,7 @@ object Dataset {
      component.last_modified_at_source.get,
      component.md5.get,
```
conradoverta: More of these fields can be empty. I recommend using getOrElse in most places for safety.

conradoverta (approved):

client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
nhatsmrt: I'll use 0 for int values, and empty string for strings. Sounds good?

conradoverta (approved):

PR #891, opened 2020-06-28
Add Blob._from_proto() and Blob._as_proto()

convoliution:
On top of #879

Until now, interacting with a blob's `_msg` attribute was the way to translate between these Python objects and protobuf messages.

This PR begins the shift away from that practice by implementing a `_from_proto()` and `_as_proto()` interface for blobs. This is only a start, as these methods still rely on the `_msg` attribute.

### DIFF
--- a/client/verta/verta/_repository/blob.py
+++ b/client/verta/verta/_repository/blob.py
@@ -2,6 +2,38 @@
 
 from __future__ import print_function
 
+import abc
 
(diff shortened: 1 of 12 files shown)


conradoverta (approved):
We should merge this to master directly.

convoliution:
Rebased; see #901

PR #892, opened 2020-06-28
No more protobuf-as-attr in dataset blobs

convoliution:
On top of #891

`S3` and `Path` dataset blobs no longer have a protobuf `_msg` attribute, and instead have a collection of `Component` objects.

This accomplishes the primary goal of this refactor which was to have `_base_path` (and as a bonus, the `_local_path` used for MDB versioning upload!) grouped together with the components' information.

I chose to make this a class instead of a `namedtuple` for these reasons:
- optional fields (though this turned out to not be that important)
- control over `__repr__`
- ability to define `_from_proto()` and `_as_proto()`

### DIFF
--- a/client/verta/verta/_repository/commit.py
+++ b/client/verta/verta/_repository/commit.py
@@ -488,7 +488,7 @@ class Commit(object):
 
         """
         # prepare ModelDB-versioned blobs, and track for upload after commit save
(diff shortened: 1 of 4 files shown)


convoliution:
Rebased; see #902

PR #894, opened 2020-06-28
Fix Dataset Versioning Tests

convoliution:
On top of #893

`dataset._msg` is gone so I guess I have no choice but to reply on public APIs for tests 😞


convoliution:
Rebased; see #905

PR #896, opened 2020-06-28
Fix Notebook blob

convoliution:
On top of #893

`Notebook` blobs include a `Path` blob, so it needs to be updated as well.


convoliution:
Rebased; see #904

PR #899, opened 2020-06-29
VR-4855 Implement listing path for Dataset blobs.

nhatsmrt:


client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
```diff
@@ -52,6 +52,11 @@ trait Dataset extends Blob {
 
   /** Clean up the uploaded components */
   protected def cleanUpUploadedComponents(): Try[Unit] = Success(())
+
+  /** Returns the paths of all components in this dataset
+   *  @return Paths of all components
+   */
+  def listPaths: List[String] = contents.keySet.toList
```
conradoverta: I think this will be unsorted. If yes, let's sort.

nhatsmrt: Yep it was unsorted. Fixed.

conradoverta (approved):

PR #902, opened 2020-06-29
No more protobuf-as-attr in dataset blobs

convoliution:
On top of #901

`S3` and `Path` dataset blobs no longer have a protobuf `_msg` attribute, and instead have a collection of `Component` objects.

This accomplishes the primary goal of this refactor which was to have `_base_path` (and as a bonus, the `_local_path` used for MDB versioning upload!) grouped together with the components' information.

I chose to make this a class instead of a `namedtuple` for these reasons:
- optional fields (though this turned out to not be that important)
- control over `__repr__`
- ability to define `_from_proto()` and `_as_proto()`

### DIFF
--- a/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
+++ b/client/scala/src/main/scala/ai/verta/blobs/dataset/Dataset.scala
@@ -11,7 +11,7 @@ import scala.util.{Failure, Success, Try}
 
 trait Dataset extends Blob {
   protected val contents: HashMap[String, FileMetadata] // for deduplication and comparing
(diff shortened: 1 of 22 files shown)


conradoverta (approved):
LGTM. Quick question, do we ever use the key from `_components_map`? I only see operations setting or getting the values.

convoliution:
@conradoverta Currently, the key isn't used.

What I have in mind is that when blob addition and `blob.add(path)` are implemented, it'll be used for lookups to check if paths are already present so I wouldn't have to write logic to iterate through the collection whenever I need to.

Although I could just write a function for that...

PR #905, opened 2020-06-29
Fix Dataset Versioning Tests

convoliution:
On top of #903

`dataset._msg` is gone so I guess I have no choice but to reply on public APIs for tests 😞


conradoverta (approved):
Using public APIs is good!

PR #914, opened 2020-06-30
Downloading Dataset blobs from ModelDB, Part 2

nhatsmrt:
On top of https://github.com/VertaAI/modeldb/pull/886. Should not be reviewed before it. This PR implements: 
- downloadToPath inference when it is not provided
- remove (partially) downloaded file(s) if the download attempt fails.


client/scala/src/main/scala/ai/verta/blobs/dataset/ComponentToLocalPath.scala
```diff
@@ -0,0 +1,9 @@
+package ai.verta.blobs.dataset
+
+/** Output of Dataset's determineComponentAndLocalPaths method
+ *  @param componentToLocalPath a map from component path to local path to download to
+ *  @param absoluteLocalPath absolute local path to the downloaded file(s)
```
conradoverta, line 5: I'm not sure why we have these two and what's their relationship. Why don't we just use the absolute local path inside the map?

nhatsmrt: OK this is going to be complicated. Say I want to download a folder in the blob, called blobFolder, containing:
blobFolder/testfile
blobFolder/testfile2

to a folder called testdir.

The componentToLocalPath map will look like: 
blobFolder/testfile -> testdir/testfile
blobFolder/testfile2 -> testdir/testfile

where as absoluteLocalPath will be the (absolute version of) testdir. I think the idea is that absoluteLocalPath will eventually become the output of download, and the user can use this to access the downloaded files. 

@convoliution Is this explanation correct?

nhatsmrt:
Close PR in favor of 2 smaller ones. Continue review at https://github.com/VertaAI/modeldb/pull/942 and https://github.com/VertaAI/modeldb/pull/943

client/scala/src/main/scala/ai/verta/blobs/dataset/ComponentToLocalPath.scala
conradoverta: I believe Michael has simplified this on the python side. I'll review with this in mind, but let's not copy bad designs from the python client just because they're there.

PR #917, opened 2020-06-30
Add util for downloading from HTTP responses

convoliution:


conradoverta (commented):
Is there a reason you're not closing the response anymore?

At this point, I think it makes sense to have a Downloder class instead of scattered methods that people need to figure out how to glue together. It can be done later.

convoliution:
@conradoverta I changed all the
```python
response = _utils.make_request("GET", url, self._conn, stream=True)
...
```
to contexts
```python
with _utils.make_request("GET", url, self._conn, stream=True) as response:
    ...
```
They will close on their own.

conradoverta (approved):
So magical =)

PR #926, opened 2020-07-03
VR-4198 overwrite check test cases fix

evgeniilezhnin:


backend/src/test/java/ai/verta/modeldb/ExperimentRunTest.java
```diff
@@ -8489,29 +8490,36 @@ public void logGetVersioningExperimentRunCreateTest()
...
+    }
+
+    try {
+      experimentRunServiceStub.logVersionedInput(logVersionedInputFail);
+      fail();
+    } catch (StatusRuntimeException e) {
+      Status status = Status.fromThrowable(e);
+      LOGGER.warn("Error Code : " + status.getCode() + " Description : " + status.getDescription());
+      assertEquals(Status.ALREADY_EXISTS.getCode(), status.getCode());
+    }
```
anandjakhaniya: I think this is redundant. this should be a single case.

anandjakhaniya (changes requested):
Looks like two same cases added in the test class. added comment on it

backend/src/test/java/ai/verta/modeldb/ExperimentRunTest.java
evgeniilezhnin: removed it

anandjakhaniya (approved):
LGTM

PR #933, opened 2020-07-07
VR-5095 Upload request needs content-length to be provided.

nhatsmrt:
@conradoverta Do you think that using the size of the file here would work? It seems to pass all of my test.


conradoverta (approved):
It works since we have files. It wouldn't otherwise.

PR #934, opened 2020-07-07
VR-4950 Support multi-part uploading of data

nhatsmrt:
On top of https://github.com/VertaAI/modeldb/pull/943.

Seems to work for the older tests, as well as a bigger test (uploading a 128 MB file, in two parts), which I commented out.


client/scala/src/main/scala/ai/verta/swagger/client/HttpClient.scala
```diff
@@ -71,7 +71,27 @@ class HttpClient(val host: String, val headers: Map[String, String]) {
...
 
+  /** Make a (raw) request, returning the body of the response
+   *  @param method method of the request
+   *  @param url url of the request
+   *  @param query query parameters
+   *  @param localHeaders local headers
+   *  @param body body of the request
+   *  @return body of the response
+   */
   def requestRaw(method: String, url: String, query: Map[String, List[String]], localHeaders: Map[String, String], body: InputStream)(implicit ec: ExecutionContext) = {
```
conradoverta, line 82: Why was this split? I think it makes more sense to return a case class with the response+headers, no? It seems weird to have different functions, one to fetch body, one to fetch the headers.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
```diff
@@ -514,43 +514,122 @@ class Commit(
...
+          3, // number of upload attempts
+          f"Uploading part ${partNum} of component ${datasetComponentPath} of blob at ${blobPath} fails."
+        )
+    })
+  }
+
+  /** Helper function to retry failable function.
+   */
+  private def retry[T](f: => Try[T], attempt: Int, errorMessage: String): Try[T] = {
+    if (attempt <= 0)
```
```suggestion
    if (attempt < 0)
```
0 just means not retry, right?

nhatsmrt: It's just semantic, but in my current set up 1 would mean "no retry". Here, attempt represents the number of attempts left (including the current attempt). The trace looks sth like this:
- retry(3): If succeeds, done. If fail, call retry(2)
- retry(2): If succeeds, done. If fail, call retry(1)
- retry(1): If succeeds, done. If fail, call retry(0), which fails automaticaly.
So we would try uploading for 3 times.

client/scala/src/main/scala/ai/verta/swagger/client/HttpClient.scala
nhatsmrt: Total didn't think about that. Fixed. Have also run the tests again to make sure that all the tests still passed.

client/scala/src/main/scala/ai/verta/repository/Commit.scala
nhatsmrt: Added some documentation to clarify the semantic.

conradoverta: Could we change the logic instead of adding documentation? I generally favor strong semantics. Maybe changing from `attempt` to `attemptsLeft` would also help clarify.

nhatsmrt: Personally I don't have any strong preference. I'll modify this into attemptsLeft. That sounds better right?

conradoverta (approved):

PR #939, opened 2020-07-07
VR-4770: Implement dataset.add()

convoliution:


client/verta/verta/dataset/_dataset.py
```diff
@@ -188,6 +186,23 @@ def _get_components_to_download(self, component_path=None, download_to_path=None
 
         return (components_to_download, os.path.abspath(downloaded_to_path))
 
+    def _add_dataset_blob(self, other):
```
conradoverta: Why not re-use the `+` you added before? This seems to essentially be doing the same thing, no?

convoliution: Because `+` creates a new object whereas `add()` needs to modify `self` in-place.

I could reuse `+` by doing a
```python
new = self + other
self.__dict__ = new.__dict__  # replace self's state with new's
```
But that feels hacky to me. Do you have an opinion on that?

conradoverta: Overload `__iadd__` and have `__add__ == clone().__iadd__`. That's how most systems do this kind of functionalities.

convoliution: Oh I forgot about `__iadd__` thanks!

conradoverta (approved):

PR #942, opened 2020-07-07
Downloading Dataset blobs from ModelDB, Part 2a

nhatsmrt:
As @conradoverta suggested, I have broken part 2 of download into two halves. This half contain downloadToPath inference.
This part implements downloadToPath inference only. For handling the failure to download, see  https://github.com/VertaAI/modeldb/pull/943.


nhatsmrt:
Copying my reply to your earlier question about the relationship between the map and the (absolute) local path:

Say I want to download a folder in the blob, called blobFolder, containing:
blobFolder/testfile
blobFolder/testfile2

to a folder called testdir in the local filesystem.

The componentToLocalPath map will look like:
blobFolder/testfile -> testdir/testfile
blobFolder/testfile2 -> testdir/testfile

where as absoluteLocalPath will be the (absolute version of) testdir. I think the idea is that absoluteLocalPath will eventually become the output of download, and the user can use this to access the downloaded files.

For reference, I was trying to follow the output signature of python client as in here:
https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/dataset/_dataset.py#L170
Hopefully I got the nitty gritty details correct.

conradoverta (approved):
LGTM, but I reviewed it so many times in the Python client that I might be missing something. If we find something off, we can fix later.

PR #943, opened 2020-07-07
Downloading Dataset blobs from ModelDB, Part 2b

nhatsmrt:
On top of https://github.com/VertaAI/modeldb/pull/942. This PR does the second bullet point of the original part 2. Specifically, it makes the following modifications:
- Download to temporary files first
- If download fails, delete the temporary files
- Otherwise, move the temporary files to correct locations


conradoverta (approved):
LGTM

PR #945, opened 2020-07-08
VR-5110 s3 region setting

evgeniilezhnin:


ravishetye:
minor comment lets call the parameter in config and code as awsRegion

evgeniilezhnin:
> minor comment lets call the parameter in config and code as awsRegion

renamed

ravishetye (approved):

PR #947, opened 2020-07-08
VR-5121: Try torch.load() first in run.get_artifact()

convoliution:


client/verta/verta/client.py
```diff
@@ -3366,22 +3372,35 @@ def get_artifact(self, key):
...
+
+        try:
+            obj = pickle.load(artifact_stream)
+        except:  # not something pickle can deserialize
+            artifact_stream.seek(0)
+        else:
+            artifact_stream.close()
+            return obj
+
+        return artifact_stream
```
convoliution, line 3402: Before, this function had an `if local artifact store` block and an `else (downloaded artifact)` block that basically duplicated the unpickling logic.

Now, I've rotated it to unify those two pathways by line 3386.

conradoverta (approved):
LGTM. Let's get this to out. Also, let's create a ticket to add docs on this.

PR #960, opened 2020-07-10
dataset protos

ravishetye:


ravishetye:
merging protos for dataset integration, builds and tests may be flaky for a while

convoliution:
There's a possibility that the removal of `DatasetTypeEnum` will require some changes on the Client, fyi myself

conradoverta:
It actually breaks the client currently.

PR #963, opened 2020-07-11
VR-5185 Commit reuploads versioned blobs everytime it is saved

nhatsmrt:
On top of https://github.com/VertaAI/modeldb/pull/962.

@conradoverta I found a new bug that I didn't think of before: Every time the commit saves, it attempts to reupload the blobs that were previously designated to enableMDBVersioning, even though user does not update that blob. 

This is because unlike the Python client, I used the flag enableMDBVersioning for both upload-able and download-able check. 
This is wrong; a blob might be downloadable, but no need to be uploaded when the commit is saved (as it has been uploaded before, and there is no update since).

The solution is to decouple these two states. I add another (immutable) field called downloadable, which is a user-friendly name for internal_versioned_path.isDefined. 

I've also updated one of the tests, which would now fail under my old implementation, but pass under my new implementation.


conradoverta (approved):
Nice catch!

PR #965, opened 2020-07-12
Provide abstraction for set/get on MDB objects in the client

conradoverta:
Minor updates to the tests were also done to remove cases that didn't make much sense according to
what we do (e.g. we don't show repo associated with a project). We can easily reverse those later if
desired.

This is mainly focused on:
- isolating functionalities
- removing any network access on creation
- reducing some complexity on network calls


client/verta/verta/_tracking/entity.py
```diff
@@ -57,6 +47,70 @@ def __setstate__(self, state):
...
+        if msg:
+            return cls(conn, conf, msg)
+        else:
+            raise ValueError("{} with ID {} not found".format(cls.__name__, id))
+
+    @classmethod
+    def _get_proto_by_id(cls, conn, id):
+        raise NotImplementedError
+
+    # TODO: add prints about status
```
conradoverta, line 66: Print when using something that is already on MDB is missing.

convoliution (approved):

PR #969, opened 2020-07-13
VR-4386 delete projects of a deleted org

evgeniilezhnin:


conradoverta (commented):
1) hydrated API will be deprecated
2) we should probably avoid making these calls to uac

After a second thought, this might be slightly lower priority since deleting orgs is not a normal operation

backend/src/test/java/ai/verta/modeldb/HydratedServiceTest.java
```diff
@@ -126,8 +125,7 @@ public static void setServerAndService() throws Exception {
       channelBuilder.intercept(authClientInterceptor.getClient1AuthInterceptor());
       client2ChannelBuilder.intercept(authClientInterceptor.getClient2AuthInterceptor());
     }
-    deleteEntitiesCron =
-        new DeleteEntitiesCron(authService, roleService, CronJobUtils.deleteEntitiesFrequency);
+    deleteEntitiesCron = new DeleteEntitiesCron(authService, roleService, 100);
```
evgeniilezhnin, line 128: wrong value was used here, because this parameter is not frequency, but updateLimit. Also it looks like this public constant is used only in tests, probably better to use some local value

jkwatson-verta:

PR #972, opened 2020-07-13
Update helm subcharts

siriuslee:
Mostly minor updates. The version value is required in all subcharts in Helm 3. For our use-case it is nice to not have the namespace specified since we are generating files that will then be applied using `kubectl apply`. This is my first time working with helm, and I'm open to any suggested changes!

Changes:
- Add version strings to subcharts
- Remove namespace specification from all specs (useful for creating namespace-agnostic specs using helm template)
- Add artifact volume options and securityContext options for backend

### DIFF
--- a/chart/modeldb/charts/backend/Chart.yaml
+++ b/chart/modeldb/charts/backend/Chart.yaml
@@ -1,1 +1,2 @@
 name: backend
+version: 0.2.0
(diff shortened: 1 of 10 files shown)


ravishetye:
Verified this on helm 2 and 3. This looks good, thanks @siriuslee for the contribution

PR #975, opened 2020-07-13
Update proto message locations

convoliution:


convoliution:
This was based on a find-and-replace while referencing #960; running test suite to see if I missed anything.

conradoverta (approved):

PR #979, opened 2020-07-13
Add listing for projects and experiments

conradoverta:


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -234,32 +234,128 @@ def __len__(self):
...
+        new_list = copy.deepcopy(self)
+
+        new_list._msg.sort_key = key
+        new_list._msg.ascending = not descending
+
+        return new_list
+
+    def _call_back_end(self, msg):
+        """Find the request in the backend and returns (elements, total count)."""
         raise NotImplementedError
```
convoliution, line 359: The [subclasses for `Dataset` and `DatasetVersion`](_call_back_end) probably also need to implement this.
I could do that in a separate PR, if this should be merged in first.

For my reference, [`Dataset`'s](https://github.com/VertaAI/modeldb/blob/c882224/client/verta/verta/client.py#L684) is used here, [`DatasetVersion`'s](https://github.com/VertaAI/modeldb/blob/c882224/client/verta/verta/_dataset.py#L196) is used here.

conradoverta: Yeah, it can be done after. Also, be mindful about not manipulating internal fields of other objects. Use things like `with_*` instead to provide API between objects.

conradoverta: I'd also recommend sticking with the API of getting the full list by default and then filtering. It will make things easier to build docs.

convoliution (approved):

PR #982, opened 2020-07-14
Add initial registry skeleton

conradoverta:


client/verta/verta/_registry/modelversion.py
```diff
@@ -0,0 +1,52 @@
...
+    @classmethod
+    def _generate_default_name(cls):
+        return "ModelVersion {}".format(_utils.generate_default_name())
+
+    @classmethod
+    def _get_proto_by_id(cls, conn, id):
+        raise NotImplementedError
+
+    @classmethod
+    def _get_proto_by_name(cls, conn, name, workspace):
```
convoliution:
```suggestion
    def _get_proto_by_name(cls, conn, name, model_id):
```
Similar to how [the signature for `ExperimentRun`](https://github.com/VertaAI/modeldb/blob/cm/registry-skeleton/client/verta/verta/_tracking/experimentrun.py#L170) uses `name`+`expt_id`

convoliution (changes requested):

client/verta/verta/_registry/modelversion.py
```diff
@@ -0,0 +1,52 @@
...
+        raise NotImplementedError
+
+    @classmethod
+    def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=None, date_created=None, public_within_org=None):
```
convoliution, line 34: @conradoverta My understanding is that `RegisteredModelVersion` being created from `RegisteredModel`s, rather than the `Client`, right?

Based on that, I don't think there's a way for it to receive a `ctx` arg, since that's limited to `Client`.

conradoverta: is the ctx necessary?

conradoverta (commented):
LGTM

convoliution (approved):

PR #987, opened 2020-07-14
Use cache for more fields of experiment run

conradoverta:


client/verta/verta/_tracking/experimentrun.py
```diff
@@ -666,16 +670,8 @@ def get_tags(self):
...
-        msg = Message(id=self.id)
-        data = _utils.proto_to_json(msg)
-        response = _utils.make_request("GET",
-                                       "{}://{}/api/v1/modeldb/experiment-run/getExperimentRunTags".format(self._conn.scheme, self._conn.socket),
-                                       self._conn, params=data)
-        _utils.raise_for_http_error(response)
-
-        response_msg = _utils.json_to_proto(_utils.body_to_json(response), Message.Response)
-        return response_msg.tags
+        self._clear_cache()
```
convoliution, line 673:
```suggestion
        self._refresh_cache()
```

convoliution, line 673:
```suggestion
        self._refresh_cache()
```

convoliution (changes requested):

convoliution:
I noticed that the `get_tags()` test fails (returns fewer items than expected). Need to check if it's a quirk of the test or caching tags.

conradoverta:
> I noticed that the `get_tags()` test fails (returns fewer items than expected). Need to check if it's a quirk of the test or caching tags.

This is weird =)

PR #991, opened 2020-07-14
VR-5192 generate registry files

evgeniilezhnin:


convoliution:
@evgeniilezhnin @ravishetye It looks like this PR doesn't include RegistryService
I ran `build_with_docker.sh` and got these additional files: https://github.com/VertaAI/modeldb/pull/996

PR #992, opened 2020-07-14
VR-5192 registered model get and set

evgeniilezhnin:


client/verta/verta/client.py
```diff
@@ -732,12 +743,91 @@ def set_repository(self, *args, **kwargs):
...
+    def get_or_create_registered_model(self, *args, **kwargs):
+        """
+        Alias for :meth:`Client.set_project()`.
+
+        """
+        return self.set_registered_model(*args, **kwargs)
 
     def get_registered_model(self, name=None, workspace=None, id=None):
-        raise NotImplementedError
+        name = self._set_from_config_if_none(name, "registered_model")
```
convoliution: This is from a different PR I'm working on. Let's add it here, too, if you don't mind.
```suggestion
        if name is not None and id is not None:
            raise ValueError("cannot specify both `name` and `id`")

        name = self._set_from_config_if_none(name, "registered_model")
        if name is None and id is None:
            raise ValueError("must specify either `name` or `id`")
```

client/verta/verta/_registry/model.py
```diff
@@ -37,12 +41,46 @@ def _generate_default_name(cls):
...
     @classmethod
     def _get_proto_by_name(cls, conn, name, workspace):
-        raise NotImplementedError
+        Message = _RegisteredModelService.GetRegisteredModelRequest
+        if name and workspace :
+            response = conn.make_proto_request("GET",
+                                               "/api/v1/registry/workspaces/{}/registered_models/{}".format(workspace, name))
+        else:
+            raise RuntimeError("the Client has encountered an error;"
+                               " please notify the Verta development team: registered model name or workspace not specified")
```
convoliution: I think we do not need this `if-else` check, because `name` and `workspace` should be valid when they are passed into this function.
```suggestion
        response = conn.make_proto_request("GET",
                                           "/api/v1/registry/workspaces/{}/registered_models/{}".format(workspace, name))
```

client/verta/verta/client.py
```diff
@@ -732,12 +743,91 @@ def set_repository(self, *args, **kwargs):
...
+        public_within_org : bool, default False
+            If creating a registered_model in an organization's workspace, whether to make this registered_model
+            accessible to all members of that organization.
+        id : str, optional
+            ID of the registered_model. This parameter cannot be provided alongside `name`, and other
+            parameters will be ignored.
+
+        Returns
+        -------
+        :class:`registered_model`
```
```suggestion
        :class:`~verta._registry.model.RegisteredModel`
```

client/verta/verta/client.py
```diff
@@ -732,12 +743,91 @@ def set_repository(self, *args, **kwargs):
 
         """
         return self.get_or_create_repository(*args, **kwargs)
+    # set aliases for get-or-create functions for API compatibility
 
-    def get_or_create_registered_model(self, name=None, desc=None, tags=None, attrs=None, workspace=None, public_within_org=None, id=None):
-        raise NotImplementedError
+    def get_or_create_registered_model(self, *args, **kwargs):
+        """
+        Alias for :meth:`Client.set_project()`.
```
convoliution: Could we actually implement the logic here in `get_or_create_registered_model()`,
and then have `set_registered_model()` call this?

It will be different from the way `set_project()` etc are written, because we are trying to switch them.

convoliution (changes requested):

client/verta/verta/client.py
evgeniilezhnin: moved implementations

convoliution (approved):

PR #994, opened 2020-07-14
Added endpoint for fetch label ids for registry find calls

anandjakhaniya:


ravishetye (changes requested):
let's add support for negative operators

ravishetye (approved):

PR #997, opened 2020-07-14
feat: added env vars for proxy changeOrigin for web

Vafilor:
This allows the `changeOrigin` property of the web proxies to be set via environment variables. 
It defaults to false, so if no variable is set, it is false. This preserves backward compatibility.


CaptEmulation:
Thanks for your first time contribution :-)

CaptEmulation (commented):
LGTM

CaptEmulation:
I already have some time scheduled later this week to release new docker containers and I will release a new webapp container with this change by EOD July 17

PR #1002, opened 2020-07-14
Fix checks and tests for client.get_*()s

convoliution:
Followup to https://github.com/VertaAI/modeldb/pull/966

### DIFF
--- a/client/verta/tests/test_entities.py
+++ b/client/verta/tests/test_entities.py
@@ -262,8 +262,8 @@ class TestExperiment:
         assert expt.id == client.get_experiment(expt.name).id
         assert expt.id == client.get_experiment(id=expt.id).id
 
-        # clear client state to test setting parent
-        client.proj = client.expt = None
+        # test parents are restored
+        client.set_project()
         client.get_experiment(id=expt.id)
         assert client.proj.id == proj.id
         assert client.expt.id == expt.id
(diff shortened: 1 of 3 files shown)


client/verta/tests/test_entities.py
```diff
@@ -323,8 +323,9 @@ def test_get(self, client):
         assert run.id == client.get_experiment_run(run.name).id
         assert run.id == client.get_experiment_run(id=run.id).id
 
-        # clear client state to test setting parent
-        client.proj = client.expt = None
+        # test parents are restored
+        client.set_project()
```
conradoverta, line 327: Shouldn't these be restored automatically?

convoliution: Yes. This test is setting a new Project and Experiment, and then testing that `get_experiment_run()` restores the _old_ ones.

client/verta/tests/test_entities.py
```diff
@@ -323,8 +323,9 @@ def test_get(self, client):
         assert run.id == client.get_experiment_run(run.name).id
         assert run.id == client.get_experiment_run(id=run.id).id
 
-        # clear client state to test setting parent
-        client.proj = client.expt = None
+        # test parents are restored
```
convoliution:
```suggestion
        # test parents are restored by first setting new, unrelated ones
```

client/verta/tests/test_entities.py
conradoverta: Fair enough.

conradoverta (approved):

PR #1003, opened 2020-07-15
Properly import bigquery

convoliution:


conradoverta (approved):
I think this is `Fix properly imported bigquery`, right? ;)

PR #1004, opened 2020-07-15
Convert Datasets and DatasetVersions to new LazyList interface

convoliution:


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -138,7 +138,8 @@ class LazyList(object):
     # number of items to fetch per back end call in __iter__()
     _ITER_PAGE_LIMIT = 100
 
-    _OP_MAP = {'==': _CommonCommonService.OperatorEnum.EQ,
+    _OP_MAP = {'~=': _CommonCommonService.OperatorEnum.CONTAIN,
```
conradoverta: "in" is a more common term, if I remember correctly what CONTAINS does

client/verta/verta/client.py
```diff
@@ -660,27 +662,29 @@ def find_datasets(self,
         list of :class:`Dataset`
 
         """
+        datasets = Datasets(self._conn, self._conf)
+        if dataset_ids:
+            datasets._msg.dataset_ids.extend(_utils.as_list_of_str(dataset_ids))
+        if sort_key:
+            datasets.sort(sort_key, not ascending)
+        if workspace:
+            datasets.with_workspace(workspace)
```
conradoverta: Are these don't in-place or do they return a new object? How does that compare with the similar ones in tracking?

convoliution: Oops! They should (and do) return new objects. This needs to be fixed here.

client/verta/verta/_internal_utils/_utils.py
convoliution: I've experienced `CONTAIN` being used more often for fuzzy matching strings (like names).
I could _also_ add `in` as a key for `CONTAIN`.

conradoverta: So in python, it's equivalent to `"abc" in project.name`, right? `~=` is very prometheus-based and I'm not sure most people would get it (since the right hand side is different from what prometheus enables, which is full regex)

convoliution: Hmm okay, I see how `~=` would not be the best choice.

I don't see it as _equivalent_ to `"abc" in project.name` since substring checking is different from fuzzy matching (unless that's what the backend does, but I thought it does character substitutions too).

But I'll agree that it's conceptually similar enough to use `in` instead.

conradoverta: I believe that's precisely what the backend does.

conradoverta (approved):

client/verta/verta/client.py
```diff
@@ -660,27 +662,29 @@ def find_datasets(self,
         list of :class:`Dataset`
 
         """
+        datasets = Datasets(self._conn, self._conf)
+        if dataset_ids:
+            datasets._msg.dataset_ids.extend(_utils.as_list_of_str(dataset_ids))
```
conradoverta: I'd recommend adding a public interface to avoid cross-object coupling of internals.

PR #1009, opened 2020-07-15
VR-5193 registry model labels

evgeniilezhnin:


client/verta/verta/_registry/model.py
```diff
@@ -84,3 +84,33 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
...
+    def _get_existing(self):
+        registered_model = RegisteredModel(self._conn, self._conf, self._get_proto_by_id(self._conn, self.id))
+        return registered_model
+
+    def add_label(self, label):
+        if label is None:
+            raise ValueError("label is not specified")
+        registered_model = self._get_existing()
+        registered_model._msg.labels.append(label)
+        return registered_model._update()
```
convoliution: Don't return anything for `add_label()`

client/verta/verta/_registry/model.py
```diff
@@ -84,3 +84,33 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
...
+
+    def del_label(self, label):
+        if label is None:
+            raise ValueError("label is not specified")
+        registered_model = self._get_existing()
+        if label in registered_model._msg.labels:
+            registered_model._msg.labels.remove(label)
+            return registered_model._update()
+        return registered_model
```
convoliution: Don't return anything for `del_label()`

client/verta/verta/_registry/model.py
```diff
@@ -84,3 +84,33 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
...
+
+    def get_labels(self):
+        registered_model = self._get_existing()
+        return registered_model._msg.labels
+
+    def _update(self):
+        response = self._conn.make_proto_request("PUT",
+                                           "/api/v1/registry/{}".format(self.id),
+                                           body=self._msg)
+        return RegisteredModel(self._conn, self._conf, self._conn.must_proto_response(response, _RegisteredModelService.SetRegisteredModel.Response).registered_model)
```
convoliution: Don't return anything.

Actually, since we're updating the backend entity, we should also clear the cache.
```suggestion
        self._conn.must_proto_response(response, _RegisteredModelService.SetRegisteredModel.Response)
        self._clear_cache()
```

client/verta/verta/_registry/model.py
```diff
@@ -84,3 +84,33 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
 
         print("created new RegisteredModel: {} in {}".format(registered_model.name, WORKSPACE_PRINT_MSG))
         return registered_model
+
+    def _get_existing(self):
+        registered_model = RegisteredModel(self._conn, self._conf, self._get_proto_by_id(self._conn, self.id))
+        return registered_model
```
convoliution: We should not add this function.
```suggestion
```

Instead of
```python
registered_model = self._get_existing()
registered_model._msg.labels.append(label)
```
let's use
```python
self._refresh_cache()
self._msg.labels.append(label)
```
etc.

convoliution (changes requested):

client/verta/verta/_registry/model.py
evgeniilezhnin: removed return

client/verta/verta/_registry/model.py
evgeniilezhnin: removed return

client/verta/verta/_registry/model.py
evgeniilezhnin: removed return

client/verta/verta/_registry/model.py
evgeniilezhnin: replaced with clear refresh cache

convoliution (approved):

PR #1010, opened 2020-07-15
VR-5194 model find

evgeniilezhnin:


client/verta/verta/_registry/models.py
```diff
@@ -33,3 +45,17 @@ def with_workspace(self, workspace_name=None):
...
+
+    def set_page_number(self, msg, param):
+        msg.pagination.page_number = param
+        return msg
+
+    def page_limit(self, msg):
+        return msg.pagination.page_limit
+
+    def page_number(self, msg):
+        return msg.pagination.page_number
```
convoliution, line 61: Interesting, I didn't realize these APIs used a nested pagination. I will need to follow up on this later to use a cleaner interface.

convoliution (approved):
Implementation looks good. Please add some tests, especially once the labels PR is also added. We'd like to cover cases such as
```python
client.registered_models.find(["labels == \"foo\"])
```

evgeniilezhnin:
> Implementation looks good. Please add some tests, especially once the labels PR is also added. We'd like to cover cases such as
> 
> ```python
> client.registered_models.find(["labels == \"foo\"])
> ```

added

convoliution (approved):

PR #1011, opened 2020-07-15
VR-5201 with model method

evgeniilezhnin:


client/verta/verta/_registry/modelversions.py
```diff
@@ -29,5 +29,7 @@ def _call_back_end(self, msg):
     def _create_element(self, msg):
         return RegisteredModelVersion(self._conn, self._conf, msg)
 
-    def with_model(self, registered_model=None):
-        raise NotImplementedError
+    def with_model(self, registered_model):
```
convoliution: Let's keep the `registered_model=None` in the parameters.
```suggestion
    def with_model(self, registered_model=None):
```
This will be useful to reset the object to not use a model.

client/verta/verta/_registry/modelversions.py
```diff
@@ -29,5 +29,7 @@ def _call_back_end(self, msg):
...
+        new_list = copy.deepcopy(self)
+        new_list._registered_model = registered_model
```
convoliution: This should look more like [`Experiments.with_project()`](https://github.com/VertaAI/modeldb/blob/5cce4c4/client/verta/verta/_tracking/experiments.py#L45-L48)

Something like
```python
if registered_model:
    new_list._msg.id.registered_model_id = registered_model.id
...
```
But check [the protos](https://github.com/VertaAI/protos-all/blob/master/protos/protos/private/registry/RegistryService.proto#L171) (`new_list._msg` is `FindModelVersionRequest`) to verify the right field is assigned the ID.

convoliution (changes requested):

client/verta/verta/_registry/modelversions.py
evgeniilezhnin: done

client/verta/verta/_registry/modelversions.py
evgeniilezhnin: done

convoliution (approved):

PR #1014, opened 2020-07-15
VR-5195, VR-5261 Create and Get ModelVersion

nhatsmrt:


client/verta/verta/_registry/model.py
```diff
@@ -21,11 +21,25 @@ def name(self):
...
+        else:
+            ctx = _Context(self._conn, self._conf)
+            ctx.registered_model = self
+            return RegisteredModelVersion._get_or_create_by_name(self._conn, name,
+                                                       lambda name: RegisteredModelVersion._get_by_name(self._conn, self._conf, name, self.id),
+                                                       lambda name: RegisteredModelVersion._create(self._conn, self._conf, ctx, name, desc=desc, tags=tags, attrs=attrs, date_created=time_created))
 
     def get_version(self, name=None, id=None):
-        raise NotImplementedError
+        if id is not None:
```
convoliution, line 43:
```suggestion
        if name is not None and id is not None:
            raise ValueError("cannot specify both `name` and `id`")
        if name is None and id is None:
            raise ValueError("must specify either `name` or `id`")

        if id is not None:
```

client/verta/verta/client.py
```diff
@@ -739,8 +739,10 @@ def get_or_create_registered_model(self, name=None, desc=None, tags=None, attrs=
     def get_registered_model(self, name=None, workspace=None, id=None):
         raise NotImplementedError
 
-    def get_registered_model_version(self, id=None):
-        raise NotImplementedError
+    def get_registered_model_version(self, name=None, id=None):
+        if self._ctx.registered_model is None:
+            raise RuntimeError("Registered model is not set. Please call `client.set_registered_model()`")
+        return self._ctx.registered_model.get_version(name, id)
```
convoliution, line 745: We should actually be able to get by ID without a RegisteredModel being set, right?
I'm going to create a follow-up ticket for this, since it's not a blocker for tomorrow, and also because I think it'll involve a change to `Context.populate()`.

convoliution (approved):
One small suggestion, then good to merge.
I'm assuming these tests will pass when merged with Evgenii's PRs.

client/verta/verta/client.py
nhatsmrt: Correct. I'm assuming we'll deal with it in a following PR?

client/verta/verta/_registry/model.py
nhatsmrt: Fixed as suggested.

convoliution (approved):
🚢

client/verta/verta/client.py
convoliution: Yep. I've created a new lower-priority ticket that should be done in a different PR.

PR #1016, opened 2020-07-15
VR-5198 Client: ModelVersion: upload artifacts

nhatsmrt:
Need documentation. 
@convoliution 
For better tests, please see https://github.com/VertaAI/modeldb/pull/1019


convoliution (approved):

client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -72,6 +75,10 @@ def test_log_artifact(self, registered_model):
         assert "The key has been set" in str(excinfo.value)
 
     def test_del_artifact(self, registered_model):
+        np = pytest.importorskip("numpy")
+        sklearn = pytest.importorskip("sklearn")
+        from sklearn.linear_model import LogisticRegression
```
convoliution, line 80: This works perfectly fine, but note (in case you didn't know) that Python's `from` statement doesn't use the `sklearn` variable; it still looks up `"sklearn"` in the environment.

For example,
```python
    from np import array
```
wouldn't work.

PR #1019, opened 2020-07-16
VR-5276 Client: ModelVersion: retrieve artifacts

nhatsmrt:
On top of https://github.com/VertaAI/modeldb/pull/1016. Provide better testing.


client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -1,9 +1,11 @@
 import pytest
+import verta
 
 from .. import utils
+import os
 
-import verta.dataset
-import verta.environment
+import numpy as np
+from sklearn.linear_model import LogisticRegression
```
convoliution: Directly importing third-party libraries like NumPy and scikit-learn will actually end up leading to test failures (or an error initializing the entire suite) if they're not installed!
Refer to other test scripts for using `pytest.importorskip()` within test bodies instead.

client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -33,6 +35,87 @@ def test_get_by_clent(self, client):
         if registered_model:
             utils.delete_registered_model(registered_model.id, client._conn)
 
+    def test_log_model(self, model_version):
```
convoliution, line 38: Minor question: did these tests all make sense to you as you migrated them from the artifact tests script? I know they can be a little convoluted in some places.

client/verta/verta/_registry/modelversion.py
```diff
@@ -70,24 +80,110 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, labels=No
...
-        # similar to ExperimentRun.get_model
-        raise NotImplementedError
+        return _artifact_utils.deserialize_model(self._get_artifact("model", _CommonCommonService.ArtifactTypeEnum.MODEL))
 
-    def add_asset(self, key, asset, overwrite=False):
-        # similar to ExperimentRun.log_artifact
-        raise NotImplementedError
+    def log_artifact(self, key, asset, overwrite=False):
+        if key == "model":
+            raise ValueError("The key `model` is reserved for model. Please use `set_model`")
```
convoliution: Minor wording change based on the convention I've been using for (most of) the Client:
```suggestion
            raise ValueError("the key \"model\" is reserved for model; consider using set_model() instead")
```

convoliution (changes requested):
Couple of very minor changes; looks good otherwise, assuming test coverage is solid.

client/verta/tests/test_model_registry/test_model_version.py
nhatsmrt: Will do.

client/verta/tests/test_model_registry/test_model_version.py
nhatsmrt: I actually wrote the log_model and log_asset myself (so if it coincides with your tests, then great minds think alike :P). I did port the artifact file test from one of your tests, but it's pretty straightforward: create some file, log it, then check the contents.

client/verta/verta/_registry/modelversion.py
nhatsmrt: Will do.

client/verta/tests/test_model_registry/test_model_version.py
convoliution: Actually, I've just noticed that this PR seems to include https://github.com/VertaAI/modeldb/pull/1016, so maybe it should be fixed there instead.

client/verta/verta/_registry/modelversion.py
nhatsmrt: Updated.

client/verta/tests/test_model_registry/test_model_version.py
nhatsmrt: Fixed here and there.

client/verta/verta/_registry/modelversion.py
```diff
@@ -313,6 +345,24 @@ def _create_artifact_msg(self, key, artifact_stream, artifact_type, extension=No
...
         return artifact_msg
 
+    def _get_artifact(self, key, artifact_type):
+        # check to see if key exists
+        self._refresh_cache()
+        if key == "model":
+            # get model artifact
+            if not self.has_model:
+                raise KeyError("no model associated with this version")
+        elif len(filter(lambda artifact: artifact.key == key, self._msg.artifacts)) == 0:
```
convoliution: `filter` isn't `len`able in Python 3.
I now realize the specific gotchas that we're going to have moving forward.
```suggestion
        elif len(list(filter(lambda artifact: artifact.key == key, self._msg.artifacts))) == 0:
```

client/verta/verta/_registry/modelversion.py
```diff
@@ -122,9 +127,12 @@ def log_model(self, model, overwrite=False):
             _CommonCommonService.ArtifactTypeEnum.MODEL,
         )
 
+    def get_model(self):
+        return _artifact_utils.deserialize_model(self._get_artifact("model", _CommonCommonService.ArtifactTypeEnum.MODEL))
```
convoliution: Since I'm requesting another minor change anyway...
Could we divide this into two lines? It'd help with readability.

convoliution (changes requested):
Couple (more) small changes, sorrythanks!

client/verta/verta/_registry/modelversion.py
nhatsmrt: Updated.

client/verta/verta/_registry/modelversion.py
nhatsmrt: Updated.

convoliution (approved):

PR #1023, opened 2020-07-16
VR-5279: verta registry get registeredmodel

convoliution:
Demonstration of how to implement CLI functionality + tests

### DIFF
--- a/client/verta/verta/_cli/registry/get.py
+++ b/client/verta/verta/_cli/registry/get.py
@@ -1,8 +1,14 @@
 # -*- coding: utf-8 -*-
 
+import json
+
 import click
 
+from ..._internal_utils import _utils
+from ... import Client
+
 from .registry import registry
+
 
 @registry.group(name="get")
 def get():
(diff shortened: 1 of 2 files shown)


client/verta/verta/_cli/registry/get.py
```diff
@@ -27,7 +33,20 @@ def get():
 def get_model(model_name, output, workspace):
     """Get detailed information about a model.
     """
-    pass
+    client = Client()
```
convoliution, line 36: @conradoverta This prints all the usual `set * from environment` messages. Should these prints be suppressed in the CLI?

conradoverta: Yes, but can be done later. Ideally those messages are not `print` but with a logger so that we can control that better. For later.

client/verta/verta/_cli/registry/get.py
```diff
@@ -27,7 +33,20 @@ def get():
...
+
+    try:
+        model = client.get_registered_model(model_name, workspace=workspace)
+    except ValueError:
+        raise click.BadParameter("model {} not found".format(model_name))
+
+    if output == "json":
+        model_repr = json.dumps(_utils.proto_to_json(model._msg))
+    else:
+        model_repr = model._msg
```
conradoverta, line 46: We'll need a better representation I think =) Can be done later.

conradoverta (approved):
Nice

PR #1028, opened 2020-07-16
Use run ID as index in expt_runs.as_dataframe()

convoliution:
This makes it easier to fetch rows!
Changes
Screen Shot 2020-07-16 at 8 42 24 AM
to
Screen Shot 2020-07-16 at 8 42 29 AM

(and adds test coverage)

### DIFF
--- a/client/verta/verta/_tracking/experimentruns.py
+++ b/client/verta/verta/_tracking/experimentruns.py
@@ -68,6 +68,14 @@ class ExperimentRuns(_utils.LazyList):
         return ExperimentRun(self._conn, self._conf, msg)
 
     def as_dataframe(self):
+        """
+        Returns this collection of Experiment Runs as a table.
+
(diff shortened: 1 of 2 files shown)


client/verta/verta/_tracking/experimentruns.py
```diff
@@ -86,7 +94,7 @@ def as_dataframe(self):
 
             data.append(run_data)
 
-        return pd.DataFrame(data, columns=['id'] + sorted(list(columns)))
+        return pd.DataFrame(data, columns=['id'] + sorted(list(columns))).set_index('id')
```
conradoverta: Can we do it at creation time to avoid re-index?

convoliution: Done.

conradoverta (approved):

PR #1029, opened 2020-07-16
VR-5199 Client: ModelVersion: associate requirements

nhatsmrt:


client/verta/verta/_registry/modelversion.py
```diff
@@ -89,12 +94,15 @@ def get_asset(self, key):
     def del_asset(self, key):
         raise NotImplementedError
 
-    def set_environment(self, env):
-        # Env must be an EnvironmentBlob. Let's re-use the functionality from there
-        raise NotImplementedError
+    def log_environment(self, env):
+        self._refresh_cache()
+        self._msg.environment.CopyFrom(env._msg)
+        self._update()
```
convoliution: Let's add a typecheck on `env`. I think it'll be
```python
if not isinstance(env, python._environment._Environment):
    raise TypeError("`env` must be of type Environment, not {}".format(type(env))

```

convoliution (changes requested):

client/verta/verta/_registry/modelversion.py
nhatsmrt: I believe this is what you meant?
```python
from ..environment import _Environment
...
if not isinstance(env, _Environment):
        raise TypeError("`env` must be of type Environment, not {}".format(type(env)))
```

convoliution: That is indeed cleaner, thanks

convoliution (approved):

PR #1032, opened 2020-07-16
VR-5274 Enable getting ModelVersion by ID without setting RegisteredModel first

nhatsmrt:


client/verta/verta/client.py
```diff
@@ -64,6 +64,12 @@
     RegisteredModelVersions,
 )
 from ._dataset_versioning import Datasets
+from ._registry import (
+    RegisteredModel,
+    RegisteredModels,
+    RegisteredModelVersion,
+    RegisteredModelVersions,
+)
```
convoliution: These imports are already present a couple lines above
```suggestion
```

client/verta/verta/client.py
```diff
@@ -856,9 +862,16 @@ def set_registered_model(self, *args, **kwargs):
...
+        if id is not None:
+            self._ctx.registered_model_version = RegisteredModelVersion._get_by_id(self._conn, self._conf, id)
+            self._ctx.populate()
+        else:
+            if self._ctx.registered_model is None:
+                self.set_registered_model()
+
+            self._ctx.registered_model_version = RegisteredModelVersion._get_by_name(self._conn, self._conf, name, self._ctx.registered_model.id)
+
+        return self._ctx.registered_model_version
```
convoliution: 1. I don't think we should assign `registered_model_version` to the context, since `RegisteredModelVersion`s can be created from `RegisteredModel`s, independent of the Client context.
1. With that in mind, and since `Context.populate()` [doesn't handle `RegisteredModel`](https://github.com/VertaAI/modeldb/blob/ecc7211/client/verta/verta/_tracking/context.py#L20), and it can't really without a `registered_model_version`, let's just directly get the `RegisteredModel` and assign to `self_ctx.registered_model` on line 867 with a TODO to support it in the context.

convoliution (changes requested):

client/verta/verta/client.py
nhatsmrt: Gotcha. Let me fix this.

client/verta/verta/client.py
nhatsmrt: Fixed.

client/verta/verta/client.py
nhatsmrt: Updated. See if it makes more sense.

client/verta/verta/_registry/modelversion.py
```diff
@@ -23,6 +23,10 @@ def name(self):
         self._refresh_cache()
         return self._msg.version
 
+    @property
+    def registered_model_id(self):
+        return self._msg.registered_model_id
```
convoliution, line 29:
```suggestion
        self._refresh_cache()
        return self._msg.registered_model_id
```

client/verta/verta/client.py
```diff
@@ -854,9 +854,17 @@ def set_registered_model(self, *args, **kwargs):
         return self.get_or_create_registered_model(*args, **kwargs)
 
     def get_registered_model_version(self, name=None, id=None):
-        if self._ctx.registered_model is None:
-            raise RuntimeError("Registered model is not set. Please call `client.set_registered_model()`")
-        return self._ctx.registered_model.get_version(name, id)
+        if id is not None:
+            # TODO: Support registered_model in populate
+            model_version = RegisteredModelVersion._get_by_id(self._conn, self._conf, id)
+            self.set_registered_model(id=model_version.registered_model_id)
```
```suggestion
            self.get_registered_model(id=model_version.registered_model_id)
```

convoliution (changes requested):
Could we also have one `return` statement at the end of the method?

client/verta/verta/_registry/modelversion.py
nhatsmrt: Added.

client/verta/verta/client.py
nhatsmrt: Fixed.

nhatsmrt:
Return is moved to the end of the method.

convoliution (approved):

PR #1042, opened 2020-07-17
delete experimentrun metadata [protos]

ravishetye:


protos/protos/public/modeldb/ExperimentRunService.proto
```diff
@@ -381,6 +402,17 @@ message GetHyperparameters {
     }
 }
 
+message DeleteHyperparameters {
+    string id = 1;
+    // delete HyperParameters corresponding to the keys
+    repeated string hyperparameter_keys = 2;
+    // if true delete all hyper patrameters corresponding to the run, ignoring hyperparameter_keys
```
ravishetye:
```suggestion
    // if true delete all hyper parameters corresponding to the run, ignoring hyperparameter_keys
```

conradoverta (approved):
Attributes? Let's make sure we have APIs for all.

anandjakhaniya (approved):

ravishetye:
> Attributes? Let's make sure we have APIs for all.

already supported

PR #1044, opened 2020-07-17
VR-5277 create model in CLI

evgeniilezhnin:


client/verta/tests/test_cli/test_registry.py
```diff
@@ -4,14 +4,29 @@
 from click.testing import CliRunner
 
 from verta._cli import cli
+from verta._registry.model import RegisteredModel
 
 
-pytest.skip("registry in Client not yet implemented", allow_module_level=True)
```
convoliution: Actually, we still need to skip these on `master`, because registry is not yet available on dev. Please use the `skip()` in `master` when you resolve the conflict (has an updated message)

client/verta/verta/_cli/registry/create.py
```diff
@@ -28,6 +30,14 @@ def create_model(model_name, label, visibility, workspace):
...
     """
     public_within_org = visibility == "org"
+    client = Client()
+
+    try:
+        model = client.get_or_create_registered_model(model_name, workspace=workspace, public_within_org=public_within_org)
+        for l in label:
+            model.add_label(l)
+    except ValueError:
+        raise click.BadParameter("model {} not found".format(model_name))
```
convoliution:
```suggestion
    model = client.get_or_create_registered_model(model_name, workspace=workspace, public_within_org=public_within_org)
    for l in label:
        model.add_label(l)
```
No need for the `try-except` here, since they are creating a _new_ model.

convoliution (changes requested):

client/verta/tests/test_cli/test_registry.py
evgeniilezhnin: reverted

client/verta/verta/_cli/registry/get.py
```diff
@@ -40,6 +40,9 @@ def get_model(model_name, output, workspace):
     except ValueError:
         raise click.BadParameter("model {} not found".format(model_name))
 
+    if model is None:
+        raise click.BadParameter("model {} not found".format(model_name))
+
```
convoliution: I think this can be removed now, after https://github.com/VertaAI/modeldb/pull/1077 is merged

convoliution (approved):

client/verta/verta/_cli/registry/get.py
evgeniilezhnin: removed

PR #1045, opened 2020-07-17
VR-5281 list models cli

evgeniilezhnin:


client/verta/tests/test_cli/test_registry.py
```diff
@@ -3,11 +3,10 @@
 import pytest
 from click.testing import CliRunner
 
+from verta import Client
 from verta._cli import cli
 
 
-pytest.skip("registry in Client not yet implemented", allow_module_level=True)
```
convoliution: Actually, we still need to skip these on master, because registry is not yet available on dev. Please use the skip() in master when you resolve the conflict (has an updated message)

client/verta/verta/_registry/models.py
```diff
@@ -42,9 +42,11 @@ def _create_element(self, msg):
...
 
     def with_workspace(self, workspace_name=None):
-        new_list = copy.deepcopy(self)
-        new_list._msg.workspace_name = workspace_name
-        return new_list
+        if workspace_name is not None:
+            new_list = copy.deepcopy(self)
+            new_list._msg.workspace_name = workspace_name
+            return new_list
+        return self
```
convoliution: Was this causing a bug? If not, I prefer we always `return new_list`, because other `LazyList.with_*()` methods always return a new object.

Suggesting revert:
```suggestion
                new_list = copy.deepcopy(self)
        new_list._msg.workspace_name = workspace_name
        return new_list
```

convoliution (changes requested):

client/verta/verta/_registry/models.py
evgeniilezhnin: None can't be assigned to _msg.workspace_name because _msg is immutable object, but I can copy an empty string to it. Used new_list._msg.workspace_name = '' here.

client/verta/tests/test_cli/test_registry.py
evgeniilezhnin: reverted

client/verta/verta/_cli/registry/list.py
```diff
@@ -25,11 +30,28 @@ def lst():
...
     click.echo("output: {}".format(output))
 
+    click.echo("result count: {}".format(len(models)))
+    click.echo("Listing all models")
+    if output == "json":
+        array = reduce(lambda a, b: "{}, {}".format(a, b),
+               map(lambda model: json.dumps(_utils.proto_to_json(model._msg)), models))
+        result = "\"models\": [{}]".format(array)
+        result = '{' + result + '}'
+        click.echo()
```
convoliution:
```suggestion
```

client/verta/verta/_cli/registry/list.py
```diff
@@ -25,11 +30,28 @@ def lst():
...
+        click.echo(result)
+    else:
+        for model in models:
+            model_repr = model._msg.name
+            click.echo()
```
convoliution:
```suggestion
```

client/verta/verta/_cli/registry/list.py
```diff
@@ -25,11 +30,28 @@ def lst():
...
-    click.echo("Listing all models")
+    client = Client()
+
+    models = client.registered_models.with_workspace(workspace).find(filter)
 
     click.echo("filter: {}".format(filter))
     click.echo("output: {}".format(output))
 
+    click.echo("result count: {}".format(len(models)))
+    click.echo("Listing all models")
```
convoliution: I realized that these echos are actually unnecessary, and not good as CLI output.
```suggestion
    click.echo()
```

client/verta/tests/test_cli/test_registry.py
```diff
@@ -52,8 +52,37 @@ def test_get_model_wrong_name_error(self, strs):
...
+
+        result = runner.invoke(
+            cli,
+            ['registry', 'list', 'registeredmodel', '--filter', "labels == \"{}\"".format(label), "--output=json"],
+        )
+
+        assert not result.exception
+        assert 'result count: 2' in result.output
+        assert str(model1._msg.name) in result.output
+        assert str(model._msg.name) in result.output
```
convoliution, line 231: Could you also add one more test case that does _not_ use `--filter`?

convoliution (changes requested):
A few small requests

client/verta/tests/test_cli/test_registry.py
evgeniilezhnin: yes, I added it now. But it lists all the models that I have created before (works ~ 1 minute). Are you sure that it is needed? When you run these tests using a test account can you delete these unnecessary entities?

convoliution: Hmm you are right that it takes more time to run, but I would like to make sure the command works properly even when the user does not pass `--filter`.

To reduce execution time, we could maybe remove `--filter` from the `--output=json` case.

I will make another PR to make sure test entities are deleted!

convoliution (approved):

client/verta/tests/test_cli/test_registry.py
evgeniilezhnin: moved this case to --filter

PR #1066, opened 2020-07-20
VR-5102 registered models collaborator endpoints public

evgeniilezhnin:


ravishetye:
@conradoverta : i do not think we need collaborator for registry in public right?

ravishetye (changes requested):
is there a discussion that triggered this change?

evgeniilezhnin:
> @conradoverta : i do not think we need collaborator for registry in public right?

yes, you are right

PR #1071, opened 2020-07-20
Add skeleton to register a MDB Run to Registry

convoliution:


client/verta/verta/_registry/modelversion.py
```diff
@@ -97,6 +98,13 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
         print("Created new ModelVersion: {}".format(model_version.version))
         return model_version
 
+    @classmethod
+    def from_run(cls, run_id):
```
conradoverta: After some thought, maybe it should be `create_version_from_run` and be a method of `RegisteredModel`? I think we'll need the registered model info to attach, right?

client/verta/verta/_registry/modelversion.py
```diff
@@ -97,6 +98,13 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
...
+        if isinstance(run_id, ExperimentRun):
+            run_id = run_id.id
```
conradoverta: As always, avoid multi-typed arguments. Pick one.

client/verta/verta/_registry/modelversion.py
convoliution: After some independent thought, I arrived at the same conclusion.

client/verta/verta/_cli/registry/create.py
```diff
@@ -36,7 +36,8 @@ def create_model(model_name, label, visibility, workspace):
 @click.option("--model", help="Path to the model.")
 @click.option("--artifact", type=(str, str), multiple=True, help="Path to the artifact required for the model. The format is --artifact artifact_key path_to_artifact.")
 @click.option("--workspace", "-w", help="Workspace to use.")
-def create_model_version(model_name, version_name, label, model, artifact, workspace):
+@click.option("--from-run", help="ID of the Experiment Run to enter into the model registry")
+def create_model_version(model_name, version_name, label, model, artifact, workspace, from_run):
```
convoliution:
```bash
verta registry create registeredmodelversion MODEL_NAME VERSION_NAME --run-id="xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
```

conradoverta: We should somehow highlight this option will ignore many others

conradoverta (approved):

client/verta/verta/_cli/registry/create.py
ravishetye: nit: registeredmodelversion is this common usage , no snake_case not camelCase? 
personal opinion aside , fine to do if this is common practice

ravishetye: aws cli seems to use hyphenated nouns [edit: sporadically. Knowing how was functions it probably is whateach team thought was right]

convoliution: As an opposing example, kubectl doesn't separate words (`kubectl get persistentvolumeclaims`), though they do offer short names.

I have a loose preference for hyphenated nouns because I think they're more readable, though harder to type.

I'll start a thread about this

PR #1075, opened 2020-07-21
VR-5284  CLI: verta registry update registeredmodelversion

nhatsmrt:
Separate https://github.com/VertaAI/modeldb/pull/1069 from implementation


client/verta/verta/_cli/registry/update.py
```diff
@@ -30,4 +32,34 @@ def update_model(model_name, label, workspace):
...
+    if model is not None:
+        model_version.log_model(model, True)
+
+    if artifact is not None:
+        for (key, _) in artifact:
+            if key == "model":
+                raise click.BadParameter("the key \"model\" is reserved for model")
+
+        for (key, path) in artifact:
+            model_version.log_artifact(key, path, True)
```
convoliution: Could we check for collisions involving `model` and `artifact`'s keys  at the beginning of function, before any changes are made? That was if the user supplies a bad argument, they'll know before any difficult-to-reverse changes are made to their `ModelVersion`.

It's going to involve iterating through `model_version._msg.artifacts` (or you could add something like a `get_artifact_keys()`)

convoliution (changes requested):

client/verta/verta/_cli/registry/update.py
nhatsmrt: Done. Also check if the user passes duplicate artifact keys. Add a test to reflect these.

convoliution (approved):

PR #1080, opened 2020-07-21
Propose global FindModelVersion endpoint

convoliution:


ravishetye (commented):
this makes sense though the protos for private component should originate in protos-all

convoliution:
Closed in favor of the protos-all PR

PR #1081, opened 2020-07-21
VR-5381 clear cache for write operations

nhatsmrt:
Still pass tests.

### DIFF
--- a/client/verta/tests/test_model_registry/test_model_version.py
+++ b/client/verta/tests/test_model_registry/test_model_version.py
@@ -175,9 +175,9 @@ class TestModelVersion:
         np = pytest.importorskip("numpy")
         artifact = np.random.random((36, 12))
 
-        for i in range(5):
+        for i in range(2):
             model_version.log_artifact("artifact_{}".format(2 * i), artifact)
             model_version_2.log_artifact("artifact_{}".format(2 * i + 1), artifact)
 
         model_version = registered_model.get_version(id=model_version.id) # re-retrieve the version
-        assert len(model_version._msg.artifacts) == 10
+        assert len(model_version._msg.artifacts) == 4
--- a/client/verta/tests/test_model_registry/test_model_version.py
+++ b/client/verta/tests/test_model_registry/test_model_version.py
@@ -166,3 +166,18 @@ class TestModelVersion:
             model_version.archive()
 
         assert "the version has already been archived" in str(excinfo.value)
+
+    def test_clear_cache(self, registered_model):
+        # Multiple log_artifacts calls, which would potentially fail without clear_cache
+        model_version = registered_model.get_or_create_version(name="my version")
+        model_version_2 = registered_model.get_or_create_version(name="my version")
+
+        np = pytest.importorskip("numpy")
+        artifact = np.random.random((36, 12))
+
+        for i in range(5):
+            model_version.log_artifact("artifact_{}".format(2 * i), artifact)
+            model_version_2.log_artifact("artifact_{}".format(2 * i + 1), artifact)
+
+        model_version = registered_model.get_or_create_version(name="my version")
+        assert len(model_version._msg.artifacts) == 10
--- a/client/verta/tests/test_model_registry/test_model_version.py
+++ b/client/verta/tests/test_model_registry/test_model_version.py
@@ -170,7 +170,7 @@ class TestModelVersion:
     def test_clear_cache(self, registered_model):
         # Multiple log_artifacts calls, which would potentially fail without clear_cache
         model_version = registered_model.get_or_create_version(name="my version")
-        model_version_2 = registered_model.get_or_create_version(name="my version")
+        model_version_2 = registered_model.get_version(id=model_version.id) # same version object
 
         np = pytest.importorskip("numpy")
         artifact = np.random.random((36, 12))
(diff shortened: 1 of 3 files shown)


conradoverta (commented):
If it passed tests before, then the tests are missing something.

nhatsmrt:
> If it passed tests before, then the tests are missing something.

@conradoverta I have talked with @convoliution about this and have an idea about a test that would potentially fail, but it is time-sensitive, so it might pass if the commands go through slow enough (which they often do in my underpowered laptop). I will add the test anyway for demonstration purpose.

conradoverta:
> > If it passed tests before, then the tests are missing something.
> 
> @conradoverta I have talked with @convoliution about this and have an idea about a test that would potentially fail, but it is time-sensitive, so it might pass if the commands go through slow enough (which they often do in my underpowered laptop). I will add the test anyway for demonstration purpose.

I'm not sure they need to be time sensitive. For example, save the object id instantiated twice in variables A and B. If you do log in A then log in B, you bypass the cache.

We can also have `_fetch_with_no_cache` that populates `_msg` while also invalidating the cache. The problem here seems to be we're re-using a functionality that is useful on the read path but has negative effects on the write path. So we can change the functionality =)

nhatsmrt:
@conradoverta I believe that refresh_cache, if waited long enough, will automatically re-retrieve the msg, no?
Anyway that's the gist of my test too! I created 2 model version objects (but still the same version), then logging artifacts with both of them in an interleaving manner, then check at the end to see how many artifacts are logged. It did fail under my old implementation, and pass under my new implementation.

nhatsmrt:
Test added.

convoliution:
>we're re-using a functionality that is useful on the read path but has negative effects on the write path
That's becoming more and more apparent to me 😬 

Adding something like `Entity._fetch_with_no_cache()` makes sense to me.
@conradoverta Would that be preferable over, say, adding a `force` flag to `_refresh_cache()`?
```diff
--- a/client/verta/verta/_tracking/entity.py
+++ b/client/verta/verta/_tracking/entity.py
@@ -40,7 +40,10 @@ class _ModelDBEntity(object):
         self._msg = None
         self._cache_time = None
 
-    def _refresh_cache(self):
+    def _refresh_cache(self, force=False):
+        if force:
+            self._cache_time = None
+
         now = time.time()
         if self._cache_time is None:
             self._msg = None
```

nhatsmrt:
If we're making changes to the Entity superclass, do let me know, because there is another functionality also affected (del_model) that has not been merged yet.

conradoverta:
> > we're re-using a functionality that is useful on the read path but has negative effects on the write path
> > That's becoming more and more apparent to me 😬
> 
> Adding something like `Entity._fetch_with_no_cache()` makes sense to me.
> @conradoverta Would that be preferable over, say, adding a `force` flag to `_refresh_cache()`?
> 
> ```diff
> --- a/client/verta/verta/_tracking/entity.py
> +++ b/client/verta/verta/_tracking/entity.py
> @@ -40,7 +40,10 @@ class _ModelDBEntity(object):
>          self._msg = None
>          self._cache_time = None
>  
> -    def _refresh_cache(self):
> +    def _refresh_cache(self, force=False):
> +        if force:
> +            self._cache_time = None
> +
>          now = time.time()
>          if self._cache_time is None:
>              self._msg = None
> ```

I'd say yes, it would. It's a small implementation and it makes things much clearer than adding a parameter. It's also a different functionality (i.e. it doesn't leave the cache as valid and it forces the next `_refresh_cache` to fetch).

nhatsmrt:
@conradoverta @convoliution I just implemented fetch with no cache and incorporate it into ModelVersion methods. See if they make sense.

client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -166,3 +166,18 @@ def test_archive(self, model_version):
...
+
+        np = pytest.importorskip("numpy")
+        artifact = np.random.random((36, 12))
+
+        for i in range(5):
+            model_version.log_artifact("artifact_{}".format(2 * i), artifact)
+            model_version_2.log_artifact("artifact_{}".format(2 * i + 1), artifact)
+
+        model_version = registered_model.get_version(id=model_version.id) # re-retrieve the version
+        assert len(model_version._msg.artifacts) == 10
```
convoliution: I think 10 artifacts is a bit overkill, and gives the impression that this test isn't really confident in its ability to reliably catch what it's testing, if that makes sense.

Based on what you've seen, do you think we could accomplish this with, say, just 2 or 3 calls to `log_*()`? If not, this is fine for now. Let me know what you think.

convoliution (changes requested):

client/verta/tests/test_model_registry/test_model_version.py
nhatsmrt: 2 is a bit more hit or miss, but 4 does fail quite consistently.

conradoverta: No flaky tests. What is the source of flakiness? Are you relying on a race condition? All the code should be deterministic (up to where we can).

nhatsmrt: My guess is that due to these lines:
https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_tracking/entity.py#L49-L55
If the time between the declaration of model_version_2 and its first log_artifact call is more than 5 seconds, then it will automatically re-retrieve the message (L53), so the test will still pass even with refresh_cache instead of fetch_msg_no_cache.

nhatsmrt: fetch_msg_no_cache will always reliably pass. refresh_cache, well if it's fast enough will fail. But if the calls are slow then it will pass.

convoliution (approved):

PR #1092, opened 2020-07-22
VR-5401 endpoint create json

evgeniilezhnin:


client/verta/verta/_deployment/endpoint.py
```diff
@@ -29,10 +31,15 @@ def _create(cls, conn, conf, workspace, path, description=None):
...
+        json = {}
+        if description:
+            json["description"] = description
         if not path.startswith('/'):
             path = '/' + path
-
-        raise NotImplementedError
+        json["path"] = path
+        response = _utils.make_request("POST", "{}://{}{}".format(conn.scheme, conn.socket, "/api/v1/deployment/workspace/{}/endpoints".format(workspace)), conn, json=json)
+        assert response.status_code == 200
```
convoliution: I failed to mention this in the ticket! We have a util for checking successful HTTP responses.
```suggestion
        _utils.raise_for_http_error(response)
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -29,10 +31,15 @@ def _create(cls, conn, conf, workspace, path, description=None):
...
     def _create_json(cls, conn, workspace, path, description=None):
+        json = {}
+        if description:
+            json["description"] = description
         if not path.startswith('/'):
             path = '/' + path
-
-        raise NotImplementedError
+        json["path"] = path
+        response = _utils.make_request("POST", "{}://{}{}".format(conn.scheme, conn.socket, "/api/v1/deployment/workspace/{}/endpoints".format(workspace)), conn, json=json)
```
convoliution: Moving the URL to a separate line will make it easier to read
```suggestion
        url = "{}://{}/api/v1/deployment/workspace/{}/endpoints".format(conn.scheme, conn.socket, workspace)
        response = _utils.make_request("POST", url, conn, json=json)
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -29,10 +31,15 @@ def _create(cls, conn, conf, workspace, path, description=None):
 
     @classmethod
     def _create_json(cls, conn, workspace, path, description=None):
+        json = {}
```
convoliution: `json` is the name for a built-in Python module; I like using `data` instead to prevent name collision. Same change for the `json`s below, please!
```suggestion
        data = {}
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import print_function
 
+from verta._internal_utils import _utils
```
convoliution: Relative imports are a bit safer than starting from `verta.`
```suggestion
from .._internal_utils import _utils
```

convoliution (changes requested):
Few minor suggestions, thanks Evgenii!

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: renamed

client/verta/verta/client.py
```diff
@@ -890,6 +890,7 @@ def get_or_create_endpoint(self, path=None, description=None, workspace=None, id
         workspace = self._set_from_config_if_none(workspace, "workspace")
         if workspace is None:
             workspace = self._get_personal_workspace()
+        Endpoint._get_or_create_by_name(self._conn, path, lambda name: None, lambda name: Endpoint._create( self._conn, self._conf, workspace, path, description))
```
convoliution, line 893: I am assuming the `lambda name: None` is replaced in https://github.com/VertaAI/modeldb/pull/1113!

client/verta/verta/_deployment/endpoint.py
```diff
@@ -32,10 +34,16 @@ def _create(cls, conn, conf, workspace, path, description=None):
 
     @classmethod
     def _create_json(cls, conn, workspace, path, description=None):
+        input = {}
```
convoliution: I would prefer `data` instead of `input` because the Client uses `data` everywhere else, but this is fine for now.

convoliution (approved):

PR #1101, opened 2020-07-22
Clear cache at the end of run._log_artifact()

convoliution:
Tests involving `run.fetch_artifacts()` were failing because
1. `run.fetch_artifacts()` directly accesses `run._msg` to check what artifact keys have been logged
1. Unlike several other `log_*()` calls, `log_artifact()` doesn't clear the cache
1. So `run.fetch_artifacts()` could (and does, in tests) check an outdated `self._msg`

Now, logging artifacts clears the cache, so `run.fetch_artifacts()` will get an up-to-date record of artifact keys.


convoliution:
Actually, let's clear the cache at the end of every `run.log_*()`. I realized that was also leading to _other_ test failures.

PR #1105, opened 2020-07-22
Enable registry tests

convoliution:


client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -8,9 +8,7 @@
 from verta.environment import Python
 
 
-pytest.skip("registry not yet available in backend", allow_module_level=True)
-
-
+@pytest.mark.skip("not implemented yet")
```
convoliution, line 11: This one test needs to be implemented alongside VR-5396

PR #1109, opened 2020-07-23
Add model overwriting check for CLI ModelVersion update/create

nhatsmrt:
Current implementation automatically overwrites the model. This is inconsistent with artifacts. A small fix, plus a test to demonstrate the new behavior.

### DIFF
--- a/client/verta/tests/test_cli/test_registry.py
+++ b/client/verta/tests/test_cli/test_registry.py
@@ -279,3 +279,28 @@ class TestUpdate:
 
         os.remove(filename)
         os.remove(classifier_name)
+
+    def test_model_overwrite(self, registered_model):
+        model_name = registered_model.name
+        version_name = "my version"
+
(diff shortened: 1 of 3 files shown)


client/verta/tests/test_cli/test_registry.py
```diff
@@ -279,3 +279,28 @@ def test_update_version_invalid_key(self, registered_model):
 
         os.remove(filename)
         os.remove(classifier_name)
+
+    def test_model_overwrite(self, registered_model):
```
convoliution:
```suggestion
    def test_model_already_logged_error(self, registered_model):
```

convoliution (changes requested):

client/verta/tests/test_cli/test_registry.py
nhatsmrt: Updated.

convoliution (approved):

PR #1112, opened 2020-07-23
VR-5402 get by id

evgeniilezhnin:


client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+import verta
+
+
+@pytest.mark.skip("endpoint not yet available in backend")
```
convoliution: I thought Endpoint is available in `dev`. Could you do me a favor and confirm this? If it's not available in dev, then this is good 👍

convoliution (approved):

client/verta/tests/test_endpoint/test_endpoint.py
evgeniilezhnin: it works on dev

PR #1113, opened 2020-07-23
VR-5403 VR-5405 VR-5406 get by name

evgeniilezhnin:


client/verta/verta/_deployment/endpoint.py
```diff
@@ -9,26 +9,24 @@
...
+        self._path = path
 
     def __repr__(self):
         # TODO: print full info
         return "<Endpoint \"{}\">".format(self.path)
 
     @property
     def path(self):
-        raise NotImplementedError
+        return self._path
```
convoliution: This is essentially what our other objects (`Project`, `RegisteredModel`) do with their `_refresh_cache()`. I think it's fine to make the extra backend call to get `self.path`, to keep a consistent design pattern.
```suggestion
        self._get_json_by_id(…)['creator_…']['path']
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -9,26 +9,24 @@
...
 
 class Endpoint(object):
-    def __init__(self, conn, conf, workspace, id):
-        self._conn = conn
-        self._conf = conf
-
+    def __init__(self, conn, conf, workspace, id, path):
         self.workspace = workspace
         self.id = id
+        self._path = path
```
convoliution: I would strongly prefer to not change this (suggesting revert), to keep a consistent internal interface compared to our other objects (`Project`, `RegisteredModel`)
```suggestion
    def __init__(self, conn, conf, workspace, id):
        self._conn = conn
        self._conf = conf

        self.workspace = workspace
        self.id = id
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -73,18 +71,27 @@ def _get_or_create_by_name(cls, conn, name, getter, creator):
...
         endpoint_json = cls._get_json_by_path(conn, workspace, path)
         if endpoint_json:
-            return cls(conn, conf, workspace, endpoint_json['id'])
+            return Endpoint._new(conf, conn, workspace, endpoint_json)
         else:
             return None
 
+    @classmethod
+    def _new(cls, conf, conn, workspace, endpoint_json):
+        return cls(conn, conf, workspace, endpoint_json['id'], endpoint_json['creator_request']['path'])
```
convoliution: I do not think it is a good design pattern to use a `obj._new()` instead of `obj.__init__()`. Also see my comment on `__init__()`
```suggestion
```

convoliution (changes requested):
I recommend against adding `_new()`. I think it does not fit with Python's design patterns, and does not fit with the design patterns of our other objects (for example `Project` and `RegisteredModel`).

But if you have a strong opinion for `_new()`, please let me know!

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: replaced with get json by id

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: removed path

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: changed to constructor

client/verta/verta/_deployment/endpoint.py
```diff
@@ -9,26 +9,25 @@
 
 
 class Endpoint(object):
-    def __init__(self, conn, conf, workspace, id):
+    def __init__(self, conn, conf, workspace, endpoint_json):
```
convoliution: Small request: simply the `id` would be preferred for now, thanks!
```suggestion
    def __init__(self, conn, conf, workspace, id):
```

convoliution (approved):

PR #1123, opened 2020-07-24
VR-5457 list endpoints

evgeniilezhnin:


client/verta/requirements.txt
```diff
@@ -1,3 +1,5 @@
+terminaltables
+
```
convoliution: Unfortunately, the output of `terminaltables` looks a bit outdated to me. I would also prefer to not add a dependency to the Client for a relatively small piece of functionality.

See VR-5448 for an example of what we'd like the output of `list`s to look like!
```suggestion
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -21,7 +21,13 @@ def __repr__(self):
...
     @property
     def path(self):
-        return Endpoint._get_json_by_id(self._conn, self.workspace, self.id)['creator_request']['path']
+        return self.path(Endpoint._get_json_by_id(self._conn, self.workspace, self.id))
+
+    def _path(self, data):
+        return data['creator_request']['path']
+
+    def _date_updated(self, data):
+        return data['date_updated']
```
convoliution, line 29: I would prefer to _not_ add new functions simply for dictionary access, because I believe it makes the interface more complicated for developers.

But I may be wrong; if having these functions is easier, feel free to keep them.

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: I think these things like where path is written in a response sometimes change in an API so its better to keep that in one place.

convoliution: Hmm makes sense, okay!

convoliution (approved):
Looks great, thanks!

PR #1125, opened 2020-07-24
Expreriment runs bugs fixes

AndreyKuznetsovQS:
# Summary

- Fixed bulk deletion menu
- Hid empty model from compare observations chart
- Fixed diff colors amount in compare observations chart


CaptEmulation:
Fixed in #1129

PR #1130, opened 2020-07-24
Use shutil.move() instead of os.rename()

convoliution:
If a direct `os.rename()` fails, `shutil.move()` [falls back](https://github.com/python/cpython/blob/2.7/Lib/shutil.py#L316-L326) to copying the file then deleting the original.
This is [supposed to](https://docs.python.org/2/library/shutil.html#shutil.move) have better support for moving files across filesystems.

### DIFF
--- a/client/verta/tests/test_config.py
+++ b/client/verta/tests/test_config.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import tempfile
 
 import yaml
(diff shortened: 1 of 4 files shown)


convoliution:
`shutil.move()` indeed succeeds where `os.rename()` does not:
Screen Shot 2020-07-24 at 4 04 16 PM

convoliution:
**Before:**
Screen Shot 2020-07-24 at 4 09 41 PM
**After:**
Screen Shot 2020-07-24 at 4 10 07 PM

conradoverta (approved):

PR #1134, opened 2020-07-27
VR-5486 get access token

evgeniilezhnin:


client/verta/verta/_deployment/endpoint.py
```diff
@@ -172,3 +172,19 @@ def get_status(self):
...
+            self.id,
+            self._get_or_create_stage(),
+        )
+        response = _utils.make_request("GET", url, self._conn)
+        _utils.raise_for_http_error(response)
+        data = response.json()
+        tokens = data["tokens"]
+        if len(tokens) == 0:
+            return None
+        return tokens[0]
```
convoliution: Needs to return the token itself as str
```suggestion
        return tokens[0]['creator_request']['value']
```

convoliution (changes requested):

convoliution (approved):
Thanks!

PR #1136, opened 2020-07-27
Protos for keyvalue properties support

anandjakhaniya:


protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -90,6 +90,37 @@ message DeletePropertyRequest {
     }
 }
 
+message AddKeyValuePropertiesRequest {
+    IdentificationType id = 1;
+    string property_name = 2;
+    repeated common.KeyValue key_value = 3; // Only string value expected
```
ravishetye: split this up into key and value both of which are string, to avoid this confiusion

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -90,6 +90,37 @@ message DeletePropertyRequest {
...
+
+    message Response {
+    }
+}
+
+message GetKeyValuePropertiesRequest {
+    // If only id_type is provided, gather from all of such sources
```
ravishetye: remove comment?

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -90,6 +90,37 @@ message DeletePropertyRequest {
...
+    IdentificationType id = 1;
+    string property_name = 2;
+    repeated string key = 3;
+    bool getAll = 4;
```
ravishetye, line 112: add a comment if this will 
`get all key value corresponding to id and property`
OR 
`get all properties with their key and value for a given id`

I think it is the first one, but please add a comment, in which case property name is compulsory

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -90,6 +90,37 @@ message DeletePropertyRequest {
...
+    message Response {
+        repeated common.KeyValue key_value = 1;
+    }
+}
+
+message DeleteKeyValuePropertiesRequest {
+    IdentificationType id = 1;
+    string property_name = 2;
+    repeated string key = 3;
+    bool deleteAll = 4;
```
ravishetye, line 124: same as Get

ravishetye (changes requested):
need changes

protos/protos/public/modeldb/metadata/MetadataService.proto
anandjakhaniya: But if I am separating key-value then how I will return it in GetAll API response within a repeated form?

anandjakhaniya: We need one message to identify key-value pairs for a repeated key-value response. if you think, instead of KeyValue I have to create a new message for key and just string value then that would be fine. what's your thought?

ravishetye: message StringKeyValue {
  string key =1;
  string value =2;
} 

that would solve the problem?

anandjakhaniya: Yes.

anandjakhaniya: Changed

protos/protos/public/modeldb/metadata/MetadataService.proto
anandjakhaniya: Added comment on message

protos/protos/public/modeldb/metadata/MetadataService.proto
anandjakhaniya: Added comment on the message

protos/protos/public/modeldb/metadata/MetadataService.proto
anandjakhaniya: Removed

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -90,6 +90,43 @@ message DeletePropertyRequest {
...
+message AddKeyValuePropertiesRequest {
+    IdentificationType id = 1;
+    string property_name = 2;
+    repeated KeyValueStringProperty key_value_property = 3;
+
+    message Response {
+    }
+}
+
+// Ignore keys and get all key value properties corresponding to id and property_name if getAll flag is true
```
ravishetye:
```suggestion
//  if getAll flag is true ignores keys and gets all key value properties corresponding to id and property_name
```

protos/protos/public/modeldb/metadata/MetadataService.proto
```diff
@@ -90,6 +90,43 @@ message DeletePropertyRequest {
...
+    string property_name = 2;
+    repeated string keys = 3;
+    bool getAll = 4;
+
+    message Response {
+        repeated KeyValueStringProperty key_value_property = 1;
+    }
+}
+
+// Ignore keys and delete all key value properties corresponding to id and property_name if deleteAll flag is true
```
ravishetye:
```suggestion
// if deleteAll flag is true ignores keys and deletes all key value properties corresponding to id and property_name
```

ravishetye (approved):
approve post suggestions

protos/protos/public/modeldb/metadata/MetadataService.proto
anandjakhaniya: Changed

protos/protos/public/modeldb/metadata/MetadataService.proto
anandjakhaniya: Changed

PR #1138, opened 2020-07-27
VR-5479 Expand __repr__()s for RegisteredModel and RegisteredModelVersion

nhatsmrt:


client/verta/verta/_registry/modelversion.py
```diff
@@ -32,7 +32,24 @@ def __init__(self, conn, conf, msg):
         super(RegisteredModelVersion, self).__init__(conn, conf, _ModelVersionService, "registered_model_version", msg)
 
     def __repr__(self):
-        return "<ModelVersion \"{}\">".format(self.name)
+        self._refresh_cache()
+        msg = self._msg
+        artifact_keys = list(map(lambda artifact: artifact.key, self._msg.artifacts))
```
convoliution: I think we can re-use our existing method for this.
```suggestion
        artifact_keys = self.get_artifact_keys()
```

convoliution (changes requested):
Minor suggestion

client/verta/verta/_registry/modelversion.py
nhatsmrt: I was thinking about this too, but get_artifact_keys return a set, so the order is not deterministic. Do you think it will be a problem? I can change it to list (at a cost of really minor performance hit).

convoliution (approved):

PR #1139, opened 2020-07-27
VR-5513 Add wait param to endpoint.update()

nhatsmrt:


client/verta/verta/_deployment/endpoint.py
```diff
@@ -124,6 +126,17 @@ def update(self, run, strategy):
         response = _utils.make_request("PUT", url, self._conn, json=strategy._as_build_update_req_body(build_id))
         _utils.raise_for_http_error(response)
 
+        if wait:
+            print("waiting for updating...", end='')
```
```suggestion
            print("waiting for update...", end='')
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -124,6 +126,17 @@ def update(self, run, strategy):
...
+            sys.stdout.flush()
+            while self.get_status()['status'] not in ("active", "error"):
+                print(".", end='')
+                sys.stdout.flush()
+                time.sleep(5)
+            print()
+            if self.get_status()['status'] == "error":
+                raise RuntimeError("endpoint updating failed")
```
```suggestion
                raise RuntimeError("endpoint update failed")
```

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -82,6 +82,17 @@ def test_direct_update(self, client, experiment_run, model_for_deployment):
         new_build_ids = get_build_ids(endpoint.get_status())
         assert len(new_build_ids) - len(new_build_ids.intersection(original_build_ids)) > 0
 
+    @pytest.mark.skip(reason="this test takes too long to complete")
```
convoliution: How long is "too long"? I'd expect this to take maybe 2 minutes, which would be acceptable for our nightly tests.

convoliution (changes requested):

client/verta/tests/test_endpoint/test_endpoint.py
nhatsmrt: 100 seconds. I'll un-skip.

nhatsmrt: Unskipped.

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Fixed.

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Fixed.

client/verta/tests/test_endpoint/test_endpoint.py
convoliution: Thanks! Good thinking though—definitely keep an eye out for those slow-running tests

client/verta/verta/_deployment/endpoint.py
```diff
@@ -124,6 +126,17 @@ def update(self, run, strategy):
...
+            print("waiting for update...", end='')
+            sys.stdout.flush()
+            while self.get_status()['status'] not in ("active", "error"):
+                print(".", end='')
+                sys.stdout.flush()
+                time.sleep(5)
+            print()
+            if self.get_status()['status'] == "error":
+                raise RuntimeError("endpoint update failed")
+
```
convoliution, line 139: Sorry—I know this is a bit outside the scope of this ticket, but I just realized it would make sense to return the endpoint status, similar-ish to what `run.deploy()` does. Can we add this in? Sorrythanks!
```suggestion

        return self.get_status()
```

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Added. I also updated the tests to use the returned status (instead of calling get_status() again).

convoliution: Awesome, thanks a bunch

convoliution (approved):

PR #1141, opened 2020-07-27
Check if key is "model" in del_artifact

nhatsmrt:
Forgot to check this... User might accidentally delete model via del_artifact.

### DIFF
--- a/client/verta/verta/_registry/modelversion.py
+++ b/client/verta/verta/_registry/modelversion.py
@@ -209,7 +209,7 @@ class RegisteredModelVersion(_ModelDBEntity):
 
     def del_artifact(self, key):
         if key == "model":
-            raise ValueError("the key \"model\" is reserved for model; consider using del_model() instead")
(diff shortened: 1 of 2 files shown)


client/verta/verta/_registry/modelversion.py
```diff
@@ -208,6 +208,9 @@ def get_artifact(self, key):
         return artifact_stream
 
     def del_artifact(self, key):
+        if key == "model":
+            raise ValueError("the key \"model\" is reserved for model; consider using del_model() instead")
```
convoliution: I think this sounds a little more straightforward than "key is reserved" for deletion
```suggestion
            raise ValueError("model can't be deleted through del_artifact(); consider using del_model() instead")
```

convoliution (changes requested):
Small wording suggestion

client/verta/verta/_registry/modelversion.py

convoliution (approved):
Good catch!

PR #1148, opened 2020-07-28
Convert client.endpoints into property

convoliution:
For API consistency with the other collection properties (`client.projects`, `client.registered_models`, etc.)

### DIFF
--- a/client/verta/tests/test_endpoint/test_endpoint.py
+++ b/client/verta/tests/test_endpoint/test_endpoint.py
@@ -48,7 +48,7 @@ class TestEndpoint:
         name = _utils.generate_default_name()
         endpoint = client.set_endpoint(name)
 
-        endpoints = client.endpoints()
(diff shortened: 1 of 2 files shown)


client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -48,7 +48,7 @@ def test_list(self, client):
         name = _utils.generate_default_name()
         endpoint = client.set_endpoint(name)
 
-        endpoints = client.endpoints()
+        endpoints = client.endpoints
```
evgeniilezhnin, line 51: Sorry, missed this PR notification. We can't do that because:

1. There is a required argument workspace in Endpoints constructor.
2. Endpoints.workspace shouldn't be None in constructor, because we need to get ids, but the method for default value for workspace (user workspace) is in Client instance

PR #1157, opened 2020-07-28
VR-5477 Move logic for get-workspace-name-by-id to _ModelDBEntity

nhatsmrt:


convoliution (approved):
No need for an actual test, but were you able to verify that this doesn't break anything?

nhatsmrt:
> No need for an actual test, but were you able to verify that this doesn't break anything?

I did try running some tests. Also the URL in https://github.com/VertaAI/modeldb/pull/1160 shows the correct workspace name (Nhat_Pham).

PR #1160, opened 2020-07-28
VR-5478 Add URLs to_ModelDBEntity.__repr__()s

nhatsmrt:
Need testing. get_workspace_name_by_id is implemented in https://github.com/VertaAI/modeldb/pull/1157


convoliution (approved):
Looks good to me. Addind tests might be tricky (not even sure if we can `GET` these URLs with the Client's auth headers)!

nhatsmrt:
Should be merge-able.

PR #1165, opened 2020-07-28
VR-5455  verta deployment update endpoint

nhatsmrt:


client/verta/verta/deployment/update/rules.py
```diff
@@ -24,6 +25,26 @@ def _as_json(self):
...
+        rule_value = float(rule_dict['rule_parameters'][0]['value'])
+
+        if rule_name == "latency_avg":
+            rule = AverageLatencyThresholdRule(rule_value)
+        elif rule_name == "latency_p90":
+            rule = P90LatencyThresholdRule(rule_value)
+        elif rule_name == "error_rate":
+            rule = ErrorRateRule(rule_value)
+        else:
+            raise ValueError("no rule with name {} exists".format(rule_name))
```
convoliution: Could we re-write this to use a mapping that checks the classes' `_NAME` directly?
Something like
```python
for Subclass in (...):
    if rule_name == Subclass._NAME:
        rule = Subclass(rule_value)
        break
else:
    raise ValueError("no rule with name {} exists".format(rule_name))
```
That way, if we ever change the name of a rule, we'd only have to change it in one place (the class attr).

client/verta/verta/deployment/update/rules.py
```diff
@@ -24,6 +25,26 @@ def _as_json(self):
...
+
+        if rule._RULE_ID != rule_id:
+            raise ValueError("wrong rule id.")
```
convoliution: We should say something like
> expected rule ID {id} for rule {name}, not {given id}

client/verta/verta/_cli/deployment/update.py
```diff
@@ -15,9 +17,40 @@ def update():
 @click.option("--run-id", "-r", help="Experiment Run to deploy.")
 @click.option("--strategy", "-s", type=click.Choice(['direct', 'canary'], case_sensitive=False), help="Strategy to use to roll out new deployment.")
 @click.option("--canary-rule", "-c", multiple=True, help="Rule to use for canary deployment. Can only be used alongside --strategy=canary")
+@click.option("--interval", "-i", type=click.IntRange(min=0), help="Rollout interval, in seconds. Can only be used alongside --strategy=canary")
+@click.option("--step", type=click.FloatRange(min=0.0, max=1.0), help="Rollout interval, in seconds. Can only be used alongside --strategy=canary")
```
convoliution:
```suggestion
@click.option("--canary-interval", "-i", type=click.IntRange(min=0), help="Rollout interval, in seconds. Can only be used alongside --strategy=canary")
@click.option("--canary-step", type=click.FloatRange(min=0.0, max=1.0), help="Rollout interval, in seconds. Can only be used alongside --strategy=canary")
```

client/verta/verta/_cli/deployment/update.py
```diff
@@ -15,9 +17,40 @@ def update():
...
 @click.option("--workspace", "-w", help="Workspace to use.")
 # TODO: more options
-def update_endpoint(path, run_id, strategy, canary_rule, workspace):
+def update_endpoint(path, run_id, strategy, canary_rule, interval, step, workspace):
     """List all endpoints available.
     """
-    raise NotImplementedError
+    if strategy != "canary" and (canary_rule or interval is not None or step is not None):
```
convoliution: `interval` and `step` can't be `0`, so I think we can do this to be less verbose. Maybe do the zero-check before.
```suggestion
    if strategy != "canary" and (canary_rule or interval or step):
```

convoliution (changes requested):

client/verta/verta/deployment/update/rules.py
nhatsmrt: Updated! See if the new _from_json makes sense.
I separate the core logic into _from_dict because I want to reuse it for update_from_config.

client/verta/verta/deployment/update/rules.py
nhatsmrt: Fixed.

client/verta/verta/_cli/deployment/update.py
nhatsmrt: Fixed.

client/verta/verta/_cli/deployment/update.py
nhatsmrt: fixed.

client/verta/verta/_cli/deployment/update.py
```diff
@@ -17,21 +17,24 @@ def update():
...
 @click.option("--workspace", "-w", help="Workspace to use.")
 # TODO: more options
-def update_endpoint(path, run_id, strategy, canary_rule, interval, step, workspace):
+def update_endpoint(path, run_id, strategy, canary_rule, canary_interval, canary_step, workspace):
     """List all endpoints available.
     """
-    if strategy != "canary" and (canary_rule or interval is not None or step is not None):
-        raise click.BadParameter("--canary-rule, --interval, and --step can only be used alongside --strategy=canary")
+    if canary_step == 0.0:
+        raise click.BadParameter("--canary-step must be non-zero.")
```
convoliution:
```suggestion
        raise click.BadParameter("--canary-step must be positive.")
```

client/verta/verta/_cli/deployment/update.py
```diff
@@ -17,21 +17,24 @@ def update():
...
 
-    if strategy == "canary" and (not canary_rule or interval is None or step is None):
-        raise click.BadParameter("--canary-rule, --interval, and --step must be provided alongside --strategy=canary")
+    if canary_interval == 0:
+        raise click.BadParameter("--canary-interval must be non-zero.")
```
convoliution:
```suggestion
        raise click.BadParameter("--canary-interval must be positive.")
```

client/verta/verta/deployment/update/rules.py
```diff
@@ -27,22 +27,31 @@ def _as_json(self):
...
-            raise ValueError("no rule with name {} exists".format(rule_name))
+        RULE_SUBCLASSES = [AverageLatencyThresholdRule, P90LatencyThresholdRule, ErrorRateRule]
+
+        for i, Subclass in enumerate(RULE_SUBCLASSES):
+            if rule_name == Subclass._NAME:
+                rule = Subclass(rule_value)
+                break
+            elif i == len(RULE_SUBCLASSES) - 1:
+                # does not match any rule
+                raise ValueError("no rule with name {} exists".format(rule_name))
```
convoliution: The `enumerate` looks a little unnecessarily complicated, and makes me a little nervous about off-by-one errors (and the fact that it needs a comment here may mean it's not be as intuitive as we'd like!).

Try using the `for-else` loop I had suggested previously. It's a Python construct that's perfectly handy for situations like these.

client/verta/verta/deployment/update/rules.py
```diff
@@ -27,22 +27,31 @@ def _as_json(self):
 
     @staticmethod
     def _from_json(json_str):
-        rule_dict = json.loads(json_str)
+        return _UpdateRule._from_dict(json.loads(json_str))
+
+    @staticmethod
+    def _from_dict(rule_dict):
```
convoliution, line 29: That's a pretty good idea! Though now I realize it's awkward having `_as_json()` also return a dict, and `_from_json()` being basically useless.

I've filed VR-5602 to follow up on that.

convoliution (changes requested):

client/verta/verta/deployment/update/rules.py
nhatsmrt: Yikes, more tech debts!
But good to know that _something_ works.

client/verta/verta/_cli/deployment/update.py
nhatsmrt: Fixed.

client/verta/verta/_cli/deployment/update.py
nhatsmrt: Fixed.

client/verta/verta/deployment/update/rules.py
nhatsmrt: Holy crap it's an actual thing? I thought you were just sketching out ideas. Learned a new thing lol!
Fixed.

client/verta/verta/deployment/update/rules.py
nhatsmrt: Actually, if you don't mind, I'll remove from_json here too. I'll fix as_json later on.

client/verta/verta/deployment/update/rules.py
convoliution: The meme about Python being basically pseudocode gets too real sometimes haha

convoliution (approved):

PR #1169, opened 2020-07-29
VR-5398 download_docker_context

evgeniilezhnin:


convoliution (approved):
@evgeniilezhnin
Would you prefer we merge this in now, and then revisit when it's implemented in the backend?
Or we can wait until the backend is merged so we can actually test this, then merge?

PR #1171, opened 2020-07-29
VR-5548 endpoint update resources

evgeniilezhnin:


client/verta/verta/deployment/resources.py
```diff
@@ -1,18 +1,39 @@
...
 
 
 class CpuMilli(_Resource):
-    raise NotImplementedError
+    def __init__(self, parameter):
+        try:
+            if parameter <= 0:
+                raise ValueError('Wrong parameter provided')
+        except TypeError:
+            raise ValueError('Wrong parameter type provided')
```
convoliution: Let's have a slightly more specific error message to help the user, for example:
https://github.com/VertaAI/modeldb/blob/ef91ff1/client/verta/verta/deployment/update/_strategies.py#L46-L50

client/verta/verta/_deployment/endpoint.py
```diff
@@ -213,3 +213,9 @@ def get_access_token(self):
         if len(tokens) == 0:
             return None
         return tokens[0]['creator_request']['value']
+
+    def form_update_body(self, resources, strategy, build_id):
```
convoliution: This seems like something that should be considered private.
```suggestion
    def _form_update_body(self, resources, strategy, build_id):
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -213,3 +213,9 @@ def get_access_token(self):
...
+        update_body = strategy._as_build_update_req_body(build_id)
+        update_body["resources"] = reduce(lambda resource_a, resource_b: {**resource_a, **resource_b}, map(lambda resource: resource.to_dict(), resources))
```
```python
{**resource_a, **resource_b}
```
This syntax is not supported in Python 2. I think we will need a different way to merge dicts.

convoliution (changes requested):

client/verta/verta/deployment/resources.py
evgeniilezhnin: replaced

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: made private

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: replaced to python2 syntax

convoliution (approved):
Thanks, Evgenii!

PR #1172, opened 2020-07-29
VR-5453 Endpoint.update_from_config()

nhatsmrt:
_UpdateRule._from_dict is from https://github.com/VertaAI/modeldb/pull/1165.

### DIFF
--- a/client/verta/verta/_deployment/endpoint.py
+++ b/client/verta/verta/_deployment/endpoint.py
@@ -205,7 +205,6 @@ class Endpoint(object):
     def _update_from_dict(self, update_dict):
         if update_dict["strategy"] == "direct":
             strategy = DirectUpdateStrategy()
-            update_body = strategy._as_build_update_req_body(update_dict["build_id"])
         elif update_dict["strategy"] == "canary":
             strategy = CanaryUpdateStrategy(
                 interval=int(update_dict["canary_strategy"]["progress_interval_seconds"]),
@@ -214,11 +213,10 @@ class Endpoint(object):
 
             for rule in update_dict["canary_strategy"]["rules"]:
                 strategy.add_rule(_UpdateRule._from_dict(rule))
-
-            update_body = strategy._as_build_update_req_body(update_dict["build_id"])
         else:
             raise ValueError("update strategy must be \"direct\" or \"canary\"")
 
+        update_body = strategy._as_build_update_req_body(update_dict["build_id"])
         self._update(update_body)
 
     def _update(self, update_body):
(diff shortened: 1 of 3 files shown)


client/verta/verta/_deployment/endpoint.py
```diff
@@ -185,7 +188,44 @@ def _create_stage(self, name="production"):
...
+            is_json = False
+
+        if not is_json:
+            try:
+                update_dict = yaml.safe_load(config)
+            except:
+                is_yaml = False
+
+        if not (is_json or is_yaml):
+            raise ValueError("input must be a json or yaml")
```
convoliution: I would recommend making this a bit cleaner and removing levels of indentation by starting with
```python
update_dict = None
```
instead of
```python
is_json = True
is_yaml = True
```
Then checking if it's still `None` each step of the way.
This would also help prevent a potential undefined variable error

convoliution: Also, I recommend catching specific errors with the `except`s (even though it's seemingly obvious what the error would be, it's still a good practice!)

For `json`, use `ValueError` (`JSONDecodeError` doesn't exist in Py2)
For `yaml`, use `yaml.YAMLError`

convoliution (changes requested):
Minor suggestion for #CodingBestPractices

client/verta/verta/_deployment/endpoint.py
nhatsmrt: @convoliution Hmm I actualy don't quite understand what you meant. Is this what you're looking for:
```python
        update_dict = None

        with open(filepath, 'r') as f:
            config = f.read()

        try:
            update_dict = json.loads(config)
        except ValueError:
            try:
                update_dict = yaml.safe_load(config)
            except yaml.YAMLError:
                raise ValueError("input file must be a json or yaml")

        return self._update_from_dict(update_dict)
```

convoliution: @nhatsmrt I just realized my suggestion doesn't actually remove a level of nesting, so disregard that statement 😆 

Something like that. Specifically, I was imagining
```python
        update_dict = None

        with open(filepath, 'r') as f:
            config = f.read()

        try:
            update_dict = json.loads(config)
        except ValueError:
            pass

        if update_dict is None:
            try:
                update_dict = yaml.safe_load(config)
            except yaml.YAMLError:
                pass

        if update_dict is None:
            raise ValueError("input file must be a json or yaml")

        return self._update_from_dict(update_dict)
```

nhatsmrt: Done.

convoliution (approved):

PR #1174, opened 2020-07-29
VR-5554 Support autoscaling parameter in Endpoint.update()

nhatsmrt:
Requires Autoscaling class implementation from https://github.com/VertaAI/modeldb/pull/1173 and metric subclasses from https://github.com/VertaAI/modeldb/pull/1177

PR #1177, opened 2020-07-30
VR-5552 Implement subclasses for _AutoscalingMetric

nhatsmrt:


nhatsmrt:
@convoliution I have implemented all the metrics. I'm thinking of doing tests in https://github.com/VertaAI/modeldb/pull/1174.
For reference, the metrics are given by:
```
{
    "metrics": [
        {
            "parameters": [
                {
                    "type": "percentage", 
                    "description": "CPU utilization target", 
                    "name": "cpu_target"
                }
            ], 
            "description": "Trigger autoscaling based on CPU utilization", 
            "name": "cpu_utilization", 
            "metric_id": 1001
        }, 
        {
            "parameters": [
                {
                    "type": "count", 
                    "description": "Requests per worker target", 
                    "name": "requests_per_worker_target"
                }
            ], 
            "description": "Trigger autoscaling based on number of requests per worker", 
            "name": "requests_per_worker", 
            "metric_id": 1002
        }, 
        {
            "parameters": [
                {
                    "type": "percentage", 
                    "description": "Memory utilization target", 
                    "name": "memory_target"
                }
            ], 
            "description": "Trigger autoscaling based on memory utilization", 
            "name": "memory_utilization", 
            "metric_id": 1003
        }
    ]
}
```

client/verta/verta/deployment/autoscaling/metrics.py
```diff
@@ -7,5 +7,63 @@
...
+    def _parameters_dict(self):
         raise NotImplementedError
+
+
+class CpuUtilization(_AutoscalingMetric):
+    _METRIC_ID = 1001
+    _NAME = "cpu_utilization"
+
+    def __init__(self, cpu_target):
+        self._cpu_target = cpu_target
```
convoliution: For all of these, could we also cast them to `str` for the user? I think it's reasonable for a user to expect to pass in a numerical value.

Plus, the fact that the backend wants `str`s is more of an implementation detail that the user shouldn't need to worry about (esp. if the backend's error message is that vague thing about Go type marshalling!)
```suggestion
        self._cpu_target = str(cpu_target)
```

convoliution (changes requested):

client/verta/verta/deployment/autoscaling/metrics.py
nhatsmrt: That makes perfect sense. Fixed.

convoliution (changes requested):
Based on our discussion on Canary rules, can we make `_NAME` the name of the parameter rather than the parent metric?

I think this should be able to resemble [rules.py](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/deployment/update/rules.py) pretty closely, which puts all the logic in the parent class.

nhatsmrt:
> Based on our discussion on Canary rules, can we make `_NAME` the name of the parameter rather than the parent metric?
> 
> I think this should be able to resemble [rules.py](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/deployment/update/rules.py) pretty closely, which puts all the logic in the parent class.

Done.

client/verta/verta/deployment/autoscaling/metrics.py
```diff
@@ -7,5 +7,36 @@
 
 @six.add_metaclass(abc.ABCMeta)
 class _AutoscalingMetric(object):
-    def _as_json(self):
-        raise NotImplementedError
+    _METRIC_ID = 0
+    _NAME = ""
```
convoliution:
```suggestion
    _METRIC_ID = 0
    _PARENT_NAME = ""
    _NAME = ""
```

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -318,9 +318,9 @@ def test_update_autoscaling(self, client, created_endpoints, experiment_run, mod
         created_endpoints.append(endpoint)
 
         autoscaling = Autoscaling(min_replicas=0, max_replicas=2, min_scale=0.5, max_scale=2.0)
-        autoscaling.add_metric(CpuUtilization("0.5"))
-        autoscaling.add_metric(MemoryUtilization("0.7"))
-        autoscaling.add_metric(RequestsPerWorker("100"))
+        autoscaling.add_metric(CpuTarget("0.5"))
+        autoscaling.add_metric(MemoryTarget("0.7"))
+        autoscaling.add_metric(RequestsPerWorkerTarget("100"))
```
convoliution: Could we support letting the user pass in numbers (at least for these metrics), and we convert it to a `str` as a courtesy? The fact that the backend API wants `str`s is an implementation detail; accepting numbers in numerical form seems nicer for the user. And the backend will let us know if the value doesn't make sense.

...Also I'm feeling some déjà vu here, but I genuinely can't remember. Was there somewhere else I brought up this idea?
```suggestion
        autoscaling.add_metric(CpuTarget(0.5))
        autoscaling.add_metric(MemoryTarget(0.7))
        autoscaling.add_metric(RequestsPerWorkerTarget(100))
```

client/verta/verta/deployment/autoscaling/metrics.py
```diff
@@ -7,5 +7,36 @@
...
+
+    def __init__(self, value):
+        self._value = value
```
convoliution:
```suggestion
        self._value = str(value)
```

convoliution (changes requested):
Let me know if these make sense

client/verta/verta/deployment/autoscaling/metrics.py
nhatsmrt: Oh yeah the conversion was in the deleted code.

nhatsmrt: Fixed.

client/verta/tests/test_endpoint/test_endpoint.py
nhatsmrt: Fixed.

client/verta/verta/deployment/autoscaling/metrics.py
nhatsmrt: Fixed.

convoliution (approved):

PR #1181, opened 2020-07-30
Remove param restrictions on _ModelDBEntity._create_proto()

convoliution:
**tl;dr** subclasses can now specify (almost) whatever they want for `_create_proto_internal()`

-----
`_ModelDBEntity._create_proto()` specifies its expected parameters, which made it inflexible if a subclass doesn't have those exact same ones.

This PR removes those hard parameters and leaves it entirely to the subclass to define its interface for proto creation with `_create_proto_internal()`

### DIFF
--- a/client/verta/verta/_tracking/entity.py
+++ b/client/verta/verta/_tracking/entity.py
@@ -137,17 +137,22 @@ class _ModelDBEntity(object):
             return None
 
     @classmethod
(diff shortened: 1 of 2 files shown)


convoliution:
Relevant tests pass

convoliution:
Relevant tests pass

PR #1182, opened 2020-07-30
VR-5577 Remove support for client.get_reg_mod_ver() by name

nhatsmrt:


client/verta/tests/conftest.py
```diff
@@ -355,15 +351,13 @@ def created_registered_models(client):
...
-            client._ctx.registered_model = None
-
         utils.delete_registered_model(registered_model.id, client._conn)
 
 
 @pytest.fixture
-def model_version(registered_model):
+def model_version(registered_model, created_registered_models):
     yield registered_model.get_or_create_version()
+    created_registered_models.append(registered_model)
```
convoliution: It might be less verbose if we call `utils.delete_registered_model()` at the end of the `registered_model` fixture, and only use `created_registered_models` in tests where we create _additional_ models.

Actually, if we did that then we might not need to change any of the test code, since `registered_model` would clean itself up basically like it used to. What do you think?

convoliution (changes requested):

convoliution:
fyi @evgeniilezhnin: removing Client's get model version by name (from now on, ID only 🔥).
This means we're also removing RegisteredModel from Context.

client/verta/tests/conftest.py
nhatsmrt: That is actually a great idea. Done.

evgeniilezhnin (approved):

convoliution (approved):

PR #1185, opened 2020-07-31
VR-5589 Implement Dataset._get_*() and _create_*()s

nhatsmrt:
Testing is done in https://github.com/VertaAI/modeldb/pull/1186 where client methods are implemented.

### DIFF
--- a/client/verta/verta/_dataset_versioning/dataset.py
+++ b/client/verta/verta/_dataset_versioning/dataset.py
@@ -59,19 +59,18 @@ class Dataset(entity._ModelDBEntity):
     def _get_proto_by_id(cls, conn, id):
         Message = _DatasetService.GetDatasetById
         msg = Message(id=id)
-        response = conn.make_proto_request("GET",
-                                           "/api/v1/modeldb/dataset/getDatasetById",
-                                           params=msg)
+        endpoint = "/api/v1/modeldb/dataset/getDatasetById"
+        response = conn.make_proto_request("GET", endpoint, params=msg)
         return conn.maybe_proto_response(response, Message.Response).dataset
 
     @classmethod
     def _get_proto_by_name(cls, conn, name, workspace):
         Message = _DatasetService.GetDatasetByName
         msg = Message(name=name, workspace_name=workspace)
-        response = conn.make_proto_request("GET",
-                                           "/api/v1/modeldb/dataset/getDatasetByName",
-                                           params=msg)
-        response = conn.maybe_proto_response(response, Message.Response)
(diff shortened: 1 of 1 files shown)


client/verta/verta/_dataset_versioning/dataset.py
```diff
@@ -57,15 +57,54 @@ def _generate_default_name(cls):
...
+            return response.shared_datasets[0]
+        else:
+            return None
 
     @classmethod
     def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=None, date_created=None, public_within_org=None):
-        raise NotImplementedError
+        Message = _DatasetService.CreateDataset
+        msg = Message(name=name, description=desc, tags=tags, attributes=attrs, workspace_name=ctx.workspace_name)
+        endpoint = "/api/v1/modeldb/dataset/createDataset"
```
convoliution: Move this line a bit closer to the `make_proto_request()` call, please!

convoliution (changes requested):

client/verta/verta/_dataset_versioning/dataset.py
nhatsmrt: You mean the endpoint line? The msg line must precede the "if public_within_org" block, because under the "if" block a field of the message is set.

convoliution: Yep, just the `endpoint` line. Thanks!

client/verta/verta/_dataset_versioning/dataset.py
```diff
@@ -57,15 +57,54 @@ def _generate_default_name(cls):
...
+
+        if workspace is None or response.HasField("dataset_by_user"):
+            return response.dataset_by_user
+        elif hasattr(response, "shared_datasets") and response.shared_datasets:
+            return response.shared_datasets[0]
+        else:
+            return None
 
     @classmethod
     def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=None, date_created=None, public_within_org=None):
```
nhatsmrt: @convoliution I noticed that Dataset has a field "type":
https://github.com/VertaAI/modeldb/blob/master/protos/protos/public/modeldb/DatasetService.proto#L21
It was also set-able in the old Dataset:
https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py#L586
Should I add a field for it here, or just ignore it for now?

convoliution: We are 100% ignoring it for now and forever because it's being deprecated.

convoliution (approved):

PR #1188, opened 2020-07-31
VR-5550 config variables

evgeniilezhnin:


client/verta/verta/_deployment/endpoint.py
```diff
@@ -126,13 +126,22 @@ def _get_json_by_path(cls, conn, workspace, path):
                 return endpoint
         return None
 
+    env_vars_err_msg = "env_vars must be (string, string) dictionary"
```
convoliution: **Also**, could we move this to _inside_ `update()`? I do not think this is useful as a class attribute if it is only used in one function.
```suggestion
    env_vars_err_msg = "`env_vars` must be dictionary of str keys and str values"
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -126,13 +126,22 @@ def _get_json_by_path(cls, conn, workspace, path):
...
             raise TypeError("run must be an ExperimentRun")
 
         if not isinstance(strategy, _UpdateStrategy):
             raise TypeError("strategy must be an object from verta.deployment.strategies")
 
+        if env_vars:
+            if not isinstance(env_vars, dict):
+                raise TypeError(self.env_vars_err_msg)
+            for key, value in env_vars.items():
+                if not isinstance(key, str) or not isinstance(value, str):
```
convoliution: Because Python 2 and Python 3 have multiple string types 😞 
Will also require `from ..external import six` near line 8.
```suggestion
                if not isinstance(key, six.string_types) or not isinstance(value, six.string_types):
```

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -198,8 +199,11 @@ def test_form_update_body(self):
             Memory("500Mi"),
         ]
 
-        parameter_json = endpoint._form_update_body(resources, DirectUpdateStrategy(), 0)
-        assert parameter_json == {'build_id': 0, 'resources': {'cpu_millis': 500, 'memory': '500Mi'}, 'strategy': 'rollout'}
+        env_vars = {'CUDA_VISIBLE_DEVICES': "1,2", "VERTA_HOST": "dev.verta.ai"}
```
convoliution: Please write "app", not "dev" for **all** tests.
```suggestion
        env_vars = {'CUDA_VISIBLE_DEVICES': "1,2", "VERTA_HOST": "app.verta.ai"}
```

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: moved it inside update. It was a class attribute to not initialize it multiple times

client/verta/verta/_deployment/endpoint.py
evgeniilezhnin: added import and your change

client/verta/verta/_deployment/endpoint.py
convoliution: That makes sense, but I think if it is just a `str` then it is a relatively small cost, and it is more helpful to have the variable isolated to the function where it's needed.

If it was a more complex object, then it may be better to have it as a class attr

client/verta/verta/_deployment/endpoint.py
```diff
@@ -7,6 +7,8 @@
 import yaml
 from functools import reduce
 
+from verta.external import six
```
convoliution: Relative import is _strongly_ preferred for this project.
```suggestion
from ..external import six
```

convoliution (approved):
One minor request, otherwise looks good. Thanks, Evgenii!

evgeniilezhnin:
found my mistake in request body json format, fixed it

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -281,9 +283,14 @@ def test_form_update_body(self):
             Memory("500Mi"),
         ]
 
-        parameter_json = endpoint._form_update_body(resources, DirectUpdateStrategy(), 0)
-        assert parameter_json == {'build_id': 0, 'resources': {'cpu_millis': 500, 'memory': '500Mi'}, 'strategy': 'rollout'}
-        
+        env_vars = {'CUDA_VISIBLE_DEVICES': "1.2", "VERTA_HOST": "app.verta.ai"}
```
convoliution: Not important, but this environment variable intentionally uses a comma!
```suggestion
        env_vars = {'CUDA_VISIBLE_DEVICES': "1,2", "VERTA_HOST": "app.verta.ai"}
```

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -245,6 +245,7 @@ def test_update_from_yaml_config(self, client, in_tempdir, created_endpoints, ex
         new_build_ids = get_build_ids(updated_status)
         assert len(new_build_ids) - len(new_build_ids.intersection(original_build_ids)) > 0
 
+    @pytest.mark.skip("bug in dev with env parameters")
```
convoliution: I believe Marc's fix to enable multiple environment variables has been merged. If it is fixed, could you verify, and un-skip this test? Thanks Evgenii.

convoliution (approved):
Small comments

client/verta/tests/test_endpoint/test_endpoint.py
evgeniilezhnin: done

PR #1192, opened 2020-07-31
Fix error for mapping does not exist in entities

anandjakhaniya:


ravishetye (approved):
For the future please add Jira id to PR description.

PR #1186, opened 2020-07-31
VR-5590 Implement Client.*_dataset2()s

nhatsmrt:
Provide testing for https://github.com/VertaAI/modeldb/pull/1185.
Small bug in backend about naming duplicate.

PR #1197, opened 2020-07-31
VR-5586 Enable Dataset description RU

nhatsmrt:
Test requires client methods in https://github.com/VertaAI/modeldb/pull/1186.

### DIFF
--- a/client/verta/verta/_dataset_versioning/dataset.py
+++ b/client/verta/verta/_dataset_versioning/dataset.py
@@ -77,11 +77,12 @@ class Dataset(entity._ModelDBEntity):
             Description to add.
 
         """
+        self._fetch_with_no_cache()
         Message = _DatasetService.UpdateDatasetDescription
         msg = Message(id=self.id, description=desc)
         endpoint = "/api/v1/modeldb/dataset/updateDatasetDescription"
         response = self._conn.make_proto_request("POST", endpoint, body=msg)
-        self._conn.must_proto_response(response, Message.Response)
+        self._msg = self._conn.must_proto_response(response, Message.Response).dataset
 
     def get_description(self):
         """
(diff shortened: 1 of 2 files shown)


client/verta/verta/_dataset_versioning/dataset.py
```diff
@@ -68,10 +68,32 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
...
 
     def set_description(self, desc):
-        raise NotImplementedError
+        """
+        Sets the description of this Dataset.
+
+        Parameters
+        ----------
+        desc : str
+            Description to add.
```
convoliution:
```suggestion
            Description to set.
```

client/verta/verta/_dataset_versioning/dataset.py
```diff
@@ -121,3 +143,9 @@ def create_path_version(self):  # TODO: same params as Path.__init__()
 
     def get_latest_version(self):
         raise NotImplementedError
+
+    def _update(self, msg, response_proto, endpoint):
+        self._fetch_with_no_cache()
```
convoliution: I don't think we need this, since`Dataset` doesn't use the get-entire-msg-and-send-it-back mechanism that Registry entities do. So this would be an unnecessary backend call per write.
```suggestion
```

convoliution (changes requested):
Small suggestions

client/verta/verta/_dataset_versioning/dataset.py
nhatsmrt: Fixed.

client/verta/verta/_dataset_versioning/dataset.py
nhatsmrt: Yeah you're right. Fixed.

nhatsmrt:
Small fixes and conflicts resolved.

convoliution (approved):

PR #1204, opened 2020-08-03
VR-5633 create organization

evgeniilezhnin:


convoliution (approved):
Evgenii, I've made a mistake and didn't hit Enter when I typed the description for this ticket 😰 

This PR is more complex than I intended. It is good as a first version, but we may need to revisit later.

PR #1209, opened 2020-08-03
VR-5658 Enable deploying ModelVersion NOT created from ExperimentRun

nhatsmrt:


client/verta/verta/_deployment/endpoint.py
```diff
@@ -136,15 +137,21 @@ def update(self, run, strategy, wait=False, resources=None, autoscaling=None, en
...
-            self._conn.socket,
-            self.workspace,
-        )
-        response = _utils.make_request("POST", url, self._conn, json={"run_id": run.id})
-        _utils.raise_for_http_error(response)
-        build_id = response.json()["id"]
+        build_id = self._create_build(run_id=run.id)
+        return self._update_from_build(build_id, strategy, wait, resources, autoscaling, env_vars)
+
+    def update_from_model_version(self, model_version, strategy, wait=False, resources=None, autoscaling=None, env_vars=None):
```
convoliution: Let's re-use the same `update()` method to support both run and model_version. I'd be okay with changing the name of the first parameter to `run_or_model_version` for now. I'll ask around for further opinions on the name.

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Yep I really wanted to use the same method too =)) We could either rename the first parameter, or have both of them, and check them inside the update method.

convoliution: Rename the first parameter. Did you run into an issue trying to do that?

nhatsmrt: Not really. Just feel a bit iffy about a parameter having potentially two types... Anyway, fixed!

client/verta/verta/_deployment/endpoint.py
```diff
@@ -128,23 +129,21 @@ def _get_json_by_path(cls, conn, workspace, path):
                 return endpoint
         return None
 
-    def update(self, run, strategy, wait=False, resources=None, autoscaling=None, env_vars=None):
-        if not isinstance(run, experimentrun.ExperimentRun):
-            raise TypeError("run must be an ExperimentRun")
+    def update(self, run_or_model_version, strategy, wait=False, resources=None, autoscaling=None, env_vars=None):
+        if not (isinstance(run_or_model_version, experimentrun.ExperimentRun) or isinstance(run_or_model_version, RegisteredModelVersion)):
```
convoliution: `isinstance()` supports tuples for its second argument to check against multiple types!
```suggestion
        if not isinstance(model_reference, (experimentrun.ExperimentRun, RegisteredModelVersion)):
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -128,23 +129,21 @@ def _get_json_by_path(cls, conn, workspace, path):
                 return endpoint
         return None
 
-    def update(self, run, strategy, wait=False, resources=None, autoscaling=None, env_vars=None):
-        if not isinstance(run, experimentrun.ExperimentRun):
-            raise TypeError("run must be an ExperimentRun")
+    def update(self, model_reference, strategy, wait=False, resources=None, autoscaling=None, env_vars=None):
+        if not (isinstance(model_reference, experimentrun.ExperimentRun) or isinstance(model_reference, RegisteredModelVersion)):
```
convoliution, line 133: `isinstance()` supports a tuple as its second argument to check against multiple types
```suggestion
        if not isinstance(model_reference, (experimentrun.ExperimentRun, RegisteredModelVersion)):
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -171,6 +170,24 @@ def update(self, run, strategy, wait=False, resources=None, autoscaling=None, en
 
         return self.get_status()
 
+    def _create_build(self, model_version_id=None, run_id=None):
```
convoliution: I think this should _also_ just take `model_reference` as a single parameter. Otherwise, we're basically duplicating the same general check on lines 140, 174, and 183. Instead, we can just have one check at line 183.

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: oh wow nice!

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Yeah but how can we tell if we should pass it to "model_version_id" or "run_id" here:
```python
        if model_version_id:
            response = _utils.make_request("POST", url, self._conn, json={"model_version_id": model_version_id})
        else:
            response = _utils.make_request("POST", url, self._conn, json={"run_id": run_id})
```

convoliution: If it takes `model_reference`, check its type like you're doing right now on line 140.

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Fixed.

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Oh I see. Will fix.

convoliution: Basically, my suggestion is to consolidate these three checks that are being done right now into one by passing the `model_reference` object into this helper function, and doing all the checking logic inside.

nhatsmrt: Fixed. Not quite sure how to remove the check in _update_from_dict though.

nhatsmrt:
All test passes (including ModelVersion tests). Undraft.

client/verta/verta/_deployment/endpoint.py
convoliution: The way I see it is that both `ExperimentRun` and `RegisteredModelVersion` are actually subtypes of a "buildable/deployable" type. Conceptually, they both have the same artifacts, and they both represent something that can be deployed in the same way. If Python had traits/typeclasses (or a more robust subtyping system...) and we were rewriting the Client, they'd probably both inherit an interface with `log_model()`, `log_artifact()`, etc.

I think it's within reason for the interface (both theoretically and practically for the user) to take a Run or a ModelVersion for the first arg 😄

client/verta/verta/_deployment/endpoint.py
```diff
@@ -227,6 +237,10 @@ def update_from_config(self, filepath):
         return self._update_from_dict(update_dict)
 
     def _update_from_dict(self, update_dict):
+        if (not "model_version_id" in update_dict and not "run_id" in update_dict) or \
+                ("model_version_id" in update_dict and "run_id" in update_dict):
```
convoliution: Line breaks `\` are discouraged in Python; using parentheses are preferred
```suggestion
        if ((not "model_version_id" in update_dict and not "run_id" in update_dict)
                or ("model_version_id" in update_dict and "run_id" in update_dict)):
```

convoliution: Actually, here's an alternative that I personally find easier to read
```suggestion
        id_keys = ('model_version_id', 'run_id')
        if (all(key in update_dict for key in id_keys)
                or not any(key in update_dict for key in id_keys)):
```

client/verta/verta/_deployment/endpoint.py
```diff
@@ -240,8 +254,12 @@ def _update_from_dict(self, update_dict):
         else:
             raise ValueError("update strategy must be \"direct\" or \"canary\"")
 
-        run = experimentrun.ExperimentRun._get_by_id(self._conn, self._conf, id=update_dict["run_id"])
-        return self.update(run, strategy)
+        if "run_id" in update_dict:
+            model_reference = experimentrun.ExperimentRun._get_by_id(self._conn, self._conf, id=update_dict["run_id"])
+        else:
+            model_reference = RegisteredModelVersion._get_by_id(self._conn, self._conf, id=update_dict["model_version_id"])
```
convoliution: Actually actually, we can combine the arg-handling with this block so we have it all in one place. I don't know if it's a best practice, but I like it to keep code chunked. What do you think?
```suggestion
        if "run_id" in update_dict and "model_version_id" in update_dict:
            raise ValueError("cannot provide both run_id and model_version_id")
        elif "run_id" in update_dict:
            model_reference = experimentrun.ExperimentRun._get_by_id(self._conn, self._conf, id=update_dict["run_id"])
        elif "model_version_id" in update_dict:
            model_reference = RegisteredModelVersion._get_by_id(self._conn, self._conf, id=update_dict["model_version_id"])
        else:
            raise RuntimeError("must provide either model_version_id or run_id")
```

convoliution (changes requested):
Code formatting suggestion

client/verta/verta/_deployment/endpoint.py
nhatsmrt: The last suggestion looks good to me! Let me give it a try.

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Resolved below.

nhatsmrt: Resolved below.

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Fixed.

client/verta/verta/_deployment/endpoint.py
```diff
@@ -171,6 +164,23 @@ def update(self, run, strategy, wait=False, resources=None, autoscaling=None, en
...
+            self._conn.socket,
+            self.workspace,
+        )
+
+        if isinstance(model_reference, RegisteredModelVersion):
+            response = _utils.make_request("POST", url, self._conn, json={"model_version_id": model_reference.id})
+        elif isinstance(model_reference, experimentrun.ExperimentRun):
+            response = _utils.make_request("POST", url, self._conn, json={"run_id": model_reference.id})
+        else:
+            raise TypeError("model_reference must be an ExperimentRun or RegisteredModelVersion")
```
```suggestion
            raise TypeError("`model_reference` must be an ExperimentRun or RegisteredModelVersion")
```

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Fixed.

client/verta/verta/_deployment/endpoint.py
```diff
@@ -131,11 +132,7 @@ def _get_json_by_path(cls, conn, workspace, path):
                 return endpoint
         return None
 
-
-    def update(self, run, strategy, wait=False, resources=None, autoscaling=None, env_vars=None):
-        if not isinstance(run, experimentrun.ExperimentRun):
-            raise TypeError("`run` must be an ExperimentRun")
```
convoliution, line 137: I think we should keep the check here at the beginning of the function.

I suggested consolidating the _other_ three checks because they were basically doing the same thing at the same place (right before sending a build request), but keeping _this_ check is important because we need to do an initial validation on params as soon as possible!

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: @convoliution Check added back in.

convoliution (approved):
Thanks a bunch, Jim!

PR #1219, opened 2020-08-04
VR-5617: Make model container paths configurable through environment variables

convoliution:
Best I can tell, these are the only hard-coded model service paths (searched with `['"]/(?!api)`)

Custom modules will need a larger change to actually be supported; filed VR-5617 to follow up.
`TFSavedModel` works; I've verified this manually, and made an aesthetic change that helps clarify this.
`setup_script` doesn't involve a hard-coded path.

I've filed VR-5693 to actually document the probably-almost-a-dozen environment variables checked by the Client.


conradoverta (commented):
I believe I saw the custom modules dir in 2 places

convoliution:
@conradoverta One was removed in https://github.com/VertaAI/modeldb/pull/1216

conradoverta (approved):

PR #1221, opened 2020-08-05
VR-5709 env vars cli update

evgeniilezhnin:


client/verta/verta/_cli/deployment/update.py
```diff
@@ -51,10 +51,19 @@ def update_endpoint(path, run_id, strategy, canary_rule, canary_interval, canary
...
-        endpoint.update(run, strategy_obj)
+
+    if env_vars:
+        env_vars_dict = json.loads(env_vars)
+    else:
+        env_vars_dict = None
+    try:
+        endpoint.update(run, strategy_obj, env_vars=env_vars_dict)
+    except TypeError as ex:
+        raise click.BadParameter(str(ex))
```
convoliution: At this point, it's okay to just have Python raise the full stack trace instead of wrapping in `BadParameter`. I recommend only using `BadParameter` for simple cases where the user's input to the CLI is wrong.
```suggestion

    endpoint.update(run, strategy_obj, env_vars=env_vars_dict)
```

convoliution (approved):
Minor comment

PR #1229, opened 2020-08-05
Fix test by dumping update strategies in text mode

convoliution:


convoliution:
Had Jim look over this 👍

PR #1231, opened 2020-08-05
Rename CpuMilli to CpuMillis

convoliution:
To be more consistent with the backend API schema, since the field is actually `cpu_millis`
This is my bad for not realizing this when designing the original Client API 😦


convoliution:
Merging before approval because Jim will need this change for some work.

PR #1233, opened 2020-08-05
VR-5722 Change endpoint.update_from_config to take canary rule names rather than rule_id

nhatsmrt:
@convoliution I believe this is what you are expecting, correct?


convoliution:
@nhatsmrt I would ask the user for `"rule"` instead of `"parent_name"` in their provided JSON. "Parent name" is just something I'm having the Client use internally for parameters to refer to their parent rules.

nhatsmrt:
> @nhatsmrt I would ask the user for `"rule"` instead of `"parent_name"` in their provided JSON. "Parent name" is just something I'm having the Client use internally for parameters to refer to their parent rules.

Fixed.

nhatsmrt:
Fix another test.

convoliution (approved):
Thanks Jim!

PR #1234, opened 2020-08-06
remove base_path from convert response to PathDatasetVersionInfo

anandjakhaniya:


ravishetye (changes requested):
in addition to this when creating blob from partinfo if base path is not null let us add that to the path .
let us ensure this is also done while migration
this will only be an issue for old clients

ravishetye (approved):

PR #1236, opened 2020-08-06
Generate default name in Entity._create()

convoliution:
@evgeniilezhnin added various `.create_*()` methods for our entities, where `name` is optional as always.
So I am adding a call to `_generate_default_name()` in `Entity._create()`.

I have commented on the notable diffs in this PR.

### DIFF
--- a/client/verta/verta/_registry/model.py
+++ b/client/verta/verta/_registry/model.py
@@ -102,7 +102,7 @@ class RegisteredModel(_ModelDBEntity):
             ctx.registered_model = self
             return RegisteredModelVersion._get_or_create_by_name(self._conn, name,
                                                        lambda name: RegisteredModelVersion._get_by_name(self._conn, self._conf, name, self.id),
-                                                       lambda name: RegisteredModelVersion._create(self._conn, self._conf, ctx, name, desc=desc, tags=labels, date_created=time_created))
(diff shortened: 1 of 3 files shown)


client/verta/verta/_registry/model.py
```diff
@@ -124,17 +124,17 @@ def create_version(self, name=None, desc=None, labels=None, time_created=None):
         """
         ctx = _Context(self._conn, self._conf)
         ctx.registered_model = self
-        return RegisteredModelVersion._create(self._conn, self._conf, ctx, name, desc=desc, tags=labels, date_created=time_created)
+        return RegisteredModelVersion._create(self._conn, self._conf, ctx, name=name, desc=desc, tags=labels, date_created=time_created)
 
 
-    def create_version_from_run(self, run_id, name):
+    def create_version_from_run(self, run_id, name=None):
```
convoliution, line 130: `name` here should be optional to match other create-entity methods.

client/verta/verta/_tracking/entity.py
```diff
@@ -132,6 +132,9 @@ def _get_proto_by_name(cls, conn, name, parent):
 
     @classmethod
     def _create(cls, conn, conf, *args, **kwargs):
+        if 'name' in kwargs and kwargs['name'] is None:
+            kwargs['name'] = cls._generate_default_name()
+
```
convoliution, line 137: Adding the call to `_generate_default_name()` if `name` as a keyword-arg is `None`.
The trivial changes in this PR are to ensure that `name` is set as a keyword.

PR #1239, opened 2020-08-06
VR-5858 endpoint.update(wait=True) to wait until only one build is present

nhatsmrt:
When wait=True, the client waits until stage status becomes "active". However, there are still 2 builds:
```
{u'components': [{u'build_id': 150549, u'ratio': 1, u'status': u'creating'},
  {u'build_id': 150557, u'status': u'running'}],
 u'creator_request': {u'name': u'production'},
 u'date_created': u'2020-08-06T14:22:56.000Z',
 u'date_updated': u'2020-08-06T15:58:24.000Z',
 'stage_id': 211751,
 u'status': u'active'}
```
If we keep waiting, until there is only one build, the result will be correct.

### DIFF
--- a/client/verta/verta/_deployment/endpoint.py
+++ b/client/verta/verta/_deployment/endpoint.py
@@ -176,9 +176,9 @@ class Endpoint(object):
             while status_dict['status'] not in ("active", "error") \
                     or (status_dict['status'] == "active" and len(self.get_status()['components']) > 1):
                 print(".", end='')
-                status_dict = self.get_status()
                 sys.stdout.flush()
                 time.sleep(5)
+                status_dict = self.get_status()
 
             print()
             if self.get_status()['status'] == "error":
--- a/client/verta/verta/_deployment/endpoint.py
+++ b/client/verta/verta/_deployment/endpoint.py
@@ -175,9 +175,16 @@ class Endpoint(object):
                 print(".", end='')
                 sys.stdout.flush()
                 time.sleep(5)
-            print()
+
             if self.get_status()['status'] == "error":
+                print()
                 raise RuntimeError("endpoint update failed")
+
+            while len(self.get_status()['components']) > 1:
+                print(".", end='')
+                sys.stdout.flush()
+                time.sleep(5)
+            print()
 
         return self.get_status()
 
--- a/client/verta/verta/_deployment/endpoint.py
+++ b/client/verta/verta/_deployment/endpoint.py
@@ -171,20 +171,18 @@ class Endpoint(object):
         if wait:
             print("waiting for update...", end='')
             sys.stdout.flush()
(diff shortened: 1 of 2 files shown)


client/verta/verta/_deployment/endpoint.py
```diff
@@ -175,10 +175,17 @@ def _update_from_build(self, build_id, strategy, wait=False, resources=None, aut
...
+
             if self.get_status()['status'] == "error":
+                print()
                 raise RuntimeError("endpoint update failed")
 
+            while len(self.get_status()['components']) > 1:
+                print(".", end='')
+                sys.stdout.flush()
+                time.sleep(5)
+            print()
```
convoliution: I have a vague feeling that a second `while` loop might make us susceptible to a race condition or potentially inconsistent state in the backend.

Do you think it would be possible to combine this condition into the previous while loop (line 174)? Not sure if that would also cause any issues.
Might have to be something like
```python
status_dict = self.get_status()  # if you think of a better var name, please do use it haha
while ((status_dict['status'] == "active" and len(status_dict['components']) > 1)
       or status_dict['status'] == "error"):
```

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: I think I know what you meant. Sth like this:
```
            while status_dict['status'] not in ("active", "error") \
                    or (status_dict['status'] == "active" and len(self.get_status()['components']) > 1):
                print(".", end='')
                status_dict = self.get_status()
                sys.stdout.flush()
                time.sleep(5)
```
Fixed. And test still passed. Please double check to see if the logic makes sense.

client/verta/verta/_deployment/endpoint.py
```diff
@@ -171,21 +171,19 @@ def _update_from_build(self, build_id, strategy, wait=False, resources=None, aut
         if wait:
             print("waiting for update...", end='')
             sys.stdout.flush()
-            while self.get_status()['status'] not in ("active", "error"):
+
+            status_dict = self.get_status()  # if you think of a better var name, please do use it haha
+            while status_dict['status'] not in ("active", "error") \
+                    or (status_dict['status'] == "active" and len(self.get_status()['components']) > 1):
                 print(".", end='')
+                status_dict = self.get_status()
```
convoliution: Let's move this to after the `sleep` (so that the `get_status()` and the loop's condition aren't 5 seconds apart).

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Fixed. And god damn the test is slow :(

convoliution (approved):

PR #1241, opened 2020-08-06
Fix various issues in Registry & Endpoint & CLI

convoliution:
Software bugs are like onions. They induce tears.
Changes make more sense if you go commit-by-commit


convoliution:
fyi @evgeniilezhnin @nhatsmrt

PR #1248, opened 2020-08-07
RDS support

ravishetye:
verified on local mysql clean
verified on local postgres clean
verified on local postgres with liquibase on master already run
verified on rds
verified on tidb with liquibase already run

the issue was because of duplicate foreign key names
modified them to be not collide
has to pin the checksum of edited tags so that existing deployments do not break.


conradoverta (commented):
Nice!

anandjakhaniya (approved):
Tested on the local system with existing DB. it works fine for me also.

PR #1253, opened 2020-08-07
VR-5735 Add documentation for Endpoint

nhatsmrt:


nhatsmrt:
I think this is all, but let's review the documentation carefully.

client/verta/verta/_deployment/endpoint.py
```diff
@@ -133,6 +147,29 @@ def _get_json_by_path(cls, conn, workspace, path):
         return None
 
     def update(self, model_reference, strategy, wait=False, resources=None, autoscaling=None, env_vars=None):
+        """
+        Updates the Endpoint with a model logged in an Experiment Run or a Model Version.
+
+        Parameters
+        ----------
+        model_reference : `verta._tracking.experimentrun.ExperimentRun` or `verta._registry.modelversion.RegisteredModelVersion`
```
convoliution: When mentioning a class in docs, also add these (can be used just about anywhere):

- putting `:class:` right before the first backtick tells the docs generator to insert a link to that class's location in the docs, if available
- putting `~` right after the first backtick makes it so the rendered text is just `ExperimentRun` instead of `verta._tracking.experimentrun.ExperimentRun`, which can help with readability

```suggestion
        model_reference : :class:`~verta._tracking.experimentrun.ExperimentRun` or `verta._registry.modelversion.RegisteredModelVersion`
```

client/verta/verta/client.py
```diff
@@ -932,6 +932,23 @@ def get_or_create_endpoint(self, path=None, description=None, workspace=None, id
 
 
     def get_endpoint(self, path=None, workspace=None, id=None):
+        """
+        Retrieves an already created Endpoint. Only one of path or id can be provided.
```
convoliution: I like to use backticks in the docstring's summary whenever referring to parameters.
```suggestion
        Retrieves an already created Endpoint. Only one of `path` or `id` can be provided.
```

convoliution (changes requested):

client/verta/verta/client.py
nhatsmrt: Fixed.

convoliution (approved):

PR #1254, opened 2020-08-10
Fix update cron query for the MySql

anandjakhaniya:


ravishetye:
changes make sense to me , can you comment on platform you have tested it on?
- [x] MySQL
- [x] TiDB
- [x] Postgres

anandjakhaniya:
We have a separate query for the Postgres so it works. but changes are generally in a non-postgres query so I have tested it with MySQL & TiDB production version.

ravishetye (approved):

ravishetye:
Please make similar changes in registry

anandjakhaniya:
Sure.

PR #1258, opened 2020-08-10
VR-5847 log file type for ModelVersion artifacts

nhatsmrt:
The mechanism is already baked in, I just forgot to expose it + use it for model_api.json.
Model already has "pkl" as filename_extension. Is there any other cases that we could be missing?


convoliution:
👍 I think that's it for now.

convoliution (approved):

PR #1260, opened 2020-08-10
VR-5854 Allow DeployedModel to make predictions while an endpoint is "updating"

nhatsmrt:


client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -303,13 +303,22 @@ def test_get_deployed_model(self, client, experiment_run, model_for_deployment,
...
         created_endpoints.append(endpoint)
-        endpoint.update(experiment_run, DirectUpdateStrategy())
+        endpoint.update(experiment_run, DirectUpdateStrategy(), wait=True)
 
-        while not endpoint.get_status()['status'] == "active":
-            time.sleep(3)
         x = model_for_deployment['train_features'].iloc[1].values
         assert endpoint.get_deployed_model().predict([x]) == [2]
 
+        np = pytest.importorskip("numpy")
```
convoliution: `importorskip()`s should be as close to the beginning of the test as possible, so we know what's required and so pytest can skip tests without wasting time running the first part.

If `numpy` isn't installed, iirc it would invalidate the results of the test thus far.
_And_ this whole test happens to rely on NumPy anyway, so it'll work fine.

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -292,7 +292,7 @@ def test_form_update_body(self):
                                   'resources': {'cpu_millis': 500, 'memory': '500Mi'}, 'strategy': 'rollout'}
 
 
-    def test_get_deployed_model(self, client, experiment_run, model_for_deployment, created_endpoints):
+    def test_get_deployed_model(self, client, experiment_run, model_version, model_for_deployment, created_endpoints):
```
convoliution, line 295: This test is kinda checking two things now: We should mention that for future devs!
```suggestion
    def test_get_deployed_model(self, client, experiment_run, model_version, model_for_deployment, created_endpoints):
    """
    Verifies prediction for a finished deployment, as well as for an endpoint
    in the middle of being updated.

    """
```

convoliution (changes requested):

client/verta/tests/test_endpoint/test_endpoint.py
nhatsmrt: Added.

client/verta/tests/test_endpoint/test_endpoint.py
nhatsmrt: Fixed.

convoliution (approved):

PR #1261, opened 2020-08-10
Revert "Don't update query keys for datasets"

convoliution:
Reverts VertaAI/modeldb#1244

Because backend has been fixed


convoliution:
fyi @ravishetye

PR #1264, opened 2020-08-11
VR-5749 use workspace service

evgeniilezhnin:


client/verta/verta/_registry/entity_with_workspace_service.py
```diff
@@ -0,0 +1,36 @@
+from verta._internal_utils import _utils
+from verta._tracking.entity import _ModelDBEntity
+
+
+class _ModelDBEntityWithWorkspaceService(_ModelDBEntity):
```
convoliution: Could we change this class and file name to `_ModelDBRegistryEntity` so it's more extensible?

I know `_ModelDBEntityWithWorkspaceService` is more specific, but I don't think future developers will know how to use it.

client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -39,7 +39,9 @@ class TestModelVersion:
 
     def test_create(self, registered_model):
         name = verta._internal_utils._utils.generate_default_name()
-        assert registered_model.create_version(name)
+        version = registered_model.create_version(name)
+        assert version
+        print(version)
```
convoliution: The `assert` will help `pytest` generate its test failure message
```suggestion
        assert repr(version)
```

convoliution (changes requested):

client/verta/verta/_registry/entity_with_workspace_service.py
evgeniilezhnin: Yes, I like your idea better. Renamed it

client/verta/tests/test_model_registry/test_model_version.py
evgeniilezhnin: Actually this change is not needed because test_repr checks the same thing. Removed it

client/verta/verta/_registry/entity_with_workspace_service.py
evgeniilezhnin: Thank you

convoliution (approved):
Thanks, Evgenii!

PR #1265, opened 2020-08-11
VR-5686 use patch in registry update

evgeniilezhnin:


convoliution (approved):
Makes sense to me so far, if it still passes tests.

ravishetye (approved):

evgeniilezhnin:
added a flag for removing default fields in json (needed for patch) and also sometimes use PUT

PR #1273, opened 2020-08-11
VR-5874 Add tests to create entities with the same name in different workspaces

nhatsmrt:


nhatsmrt:
Test passes now.

client/verta/tests/test_organization.py
```diff
@@ -24,9 +26,75 @@ def test_create_msg():
...
+        # Switching workspace name:
+        client_config = {
+            "workspace": organization.name
+        }
+
+        filepath = "verta_config.json"
+        with open(filepath, "w") as f:
+            json.dump(client_config, f)
+
+        client = Client()
```
convoliution: Could we use the `workspace` parameter on the client methods instead? I think I'd prefer making this test independent of the config file mechanisms.

client/verta/tests/test_organization.py
```diff
@@ -24,9 +26,75 @@ def test_create_msg():
...
+        assert model.id != new_model.id
+        assert version.id != new_version.id
+        assert endpoint.id != new_endpoint.id
+        assert project.id != new_project.id
+        assert exp.id != new_exp.id
+        assert run.id != new_run.id
+        assert dataset.id != new_dataset.id
+        assert repository.id != new_repository.id
+
+        project.delete()  # have to delete manually because creating dataset makes project untracked by client context.
```
convoliution, line 87: > have to delete manually because creating dataset makes project untracked by client context.
Does it really?? That seems odd. Could you do me a favor and point out where in the client code that happens?

convoliution (changes requested):

client/verta/tests/test_organization.py
nhatsmrt: https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py#L1380
This line in particular. In general, project and dataset, whichever is created after the other will re-initialize the context field of client.

client/verta/tests/test_organization.py
nhatsmrt: Fixed.

evgeniilezhnin (approved):

conradoverta (approved):
LGTM

PR #1275, opened 2020-08-11
Add Registry and Endpoint API pages

convoliution:
fyi @nhatsmrt 

[registry.rst](https://github.com/VertaAI/modeldb/blob/master/client/verta/docs/api/api/registry.rst) already exists


conradoverta (approved):
We should consider removing "deployment" or changing the name, as it's confusing when you have "endpoint". I think the old deployment api will go away, right?

PR #1280, opened 2020-08-12
VR-5851 attribute CRUD

evgeniilezhnin:


convoliution (approved):
This seems right to me; we'll see if there is a backend bug or not.

evgeniilezhnin:
There's a deletion bug: https://vertaai.slack.com/archives/C016QL3TEUT/p1597322605034600

convoliution (approved):
Attribute creation is blocking VR-5848, so I am approving this for now. We can follow up on the bugs later.

Because there seems to be a bug with deletion, **please comment out ModelVersion.del_attribute()** since it does not work, before merging.

evgeniilezhnin:
> Attribute creation is blocking VR-5848, so I am approving this for now. We can follow up on the bugs later.
> 
> Because there seems to be a bug with deletion, **please comment out ModelVersion.del_attribute()** since it does not work, before merging.

seem to work on aj.dev after fix

PR #1284, opened 2020-08-12
VR-5908 CLI: verta deployment predict endpoint

nhatsmrt:
@convoliution do you have any preference for output format? Right now I'm just click.echo the returned output list.

### DIFF
--- a/client/verta/verta/_cli/deployment/predict.py
+++ b/client/verta/verta/_cli/deployment/predict.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+
+import click
+import json
+
+from .deployment import deployment
+from ... import Client
+
+
+@deployment.group()
+def predict():
+    pass
+
+
+@predict.command(name="endpoint")
+@click.argument("path", nargs=1, required=True)
+@click.option("--data", "-d", help="Input for prediction. Must be a valid JSON string.")
+@click.option("--workspace", "-w", help="Workspace to use.")
+def predict_endpoint(path, data, workspace):
+    if not data:
+        raise click.BadParameter("--data must be provided.")  # maybe make this an argument instead?
+
+    client = Client()
+
+    try:
+        endpoint = client.get_endpoint(path=path, workspace=workspace)
+    except ValueError:
+        raise click.BadParameter("endpoint with path {} not found".format(path))
+
+    deployed_model = endpoint.get_deployed_model()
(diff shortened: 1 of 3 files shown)


convoliution:
@nhatsmrt Does the output have single quotes `'` or double quotes `"`?
If it has single quotes, let's wrap the prediction output in a `json.dumps()` before `click.echo()`ing it to convert the single quotes to double quotes.

(This is important because the CLI's displayed output should be JSON-compliant, which requires double quotes)

nhatsmrt:
@convoliution deployed_model.predict returns a list, not a string 🤔 So when click.echo-ing, it doesn't have quotes or anything. Printing output.result in test gives me:
```
set host from environment
set email from environment
set developer key from environment
[1.0, 1.0, 0.0, 0.0]
```

 Should I forcibly introduce quotes, or call json.dumps on output, or sth else?

convoliution:
Oh right, our test models use numerical outputs.
Use `json.dumps()` anyway!

nhatsmrt:
Fixed. And update test to reflect new expected output.

client/verta/verta/_cli/deployment/predict.py
```diff
@@ -0,0 +1,32 @@
...
+    pass
+
+
+@predict.command(name="endpoint")
+@click.argument("path", nargs=1, required=True)
+@click.option("--data", "-d", help="Input for prediction. Must be a valid JSON string.")
+@click.option("--workspace", "-w", help="Workspace to use.")
+def predict_endpoint(path, data, workspace):
+    if not data:
+        raise click.BadParameter("--data must be provided.")  # maybe make this an argument instead?
```
convoliution: I decided to make it a non-optional option (:laughing:) because `-d` matches the syntax of the corresponding cURL, plus all our other `endpoint` CLI commands take the endpoint path as their one-and-only argument, so this felt more consistent.

`click.option` takes a `required` parameter
```suggestion
@click.option("--data", "-d", required=True, help="Input for prediction. Must be a valid JSON string.")
@click.option("--workspace", "-w", help="Workspace to use.")
def predict_endpoint(path, data, workspace):
```

convoliution (changes requested):

client/verta/verta/_cli/deployment/predict.py
nhatsmrt: Nice. Fixed.

convoliution (approved):

PR #1291, opened 2020-08-13
versionadded for Registry and Endpoint Client

convoliution:
Add note for what Client version will have live Registry and Endpoint


conradoverta (approved):
FYI, you can say you'll deprecate in a version and not do it without changing references.

PR #1293, opened 2020-08-13
VR-5934 Investigate extracting CLI documentation

nhatsmrt:


convoliution:
@nhatsmrt Is there a particular reason you put the CLI docs on a separate page instead of the same page? I think it would be fine for them to be on the same page, but I don't feel too strongly about it.

nhatsmrt:
> @nhatsmrt Is there a particular reason you put the CLI docs on a separate page instead of the same page? I think it would be fine for them to be on the same page, but I don't feel too strongly about it.

No particular reason. I just separate them for easier inspection. Let me try to merge the two pages.

nhatsmrt:
Added documentation for predict.

convoliution (approved):

PR #1296, opened 2020-08-13
Copy get_artifact_keys() from ModelVersion to ExperimentRun

convoliution:


convoliution:
@nhatsmrt Since you wrote this for ModelVersion: how's this look to you?

nhatsmrt:
LGTM. Also that smells like things we can refactor in the future.

PR #1309, opened 2020-08-14
Only show actual status in not-deployed message

convoliution:
@nhatsmrt to "review" this.

I think this should just show the actual status string (instead of the entire status dict) like ExperimentRun does, unless we had a particular reason for showing the whole dict.


convoliution:
@nhatsmrt Off the top of your head, do you know if this would break any tests?

convoliution:
Actually, it is useful for now to show the full status dict, since it would include the error message from a failed deployment.

PR #1310, opened 2020-08-14
Rename --autoscaling-metrics to singular --autoscaling-metric

convoliution:
Since it's a `multiple=True` option, it makes more grammatical sense for it to be singular.


convoliution:
@nhatsmrt to "review" (I think it was supposed to be singular to begin with!)

PR #1311, opened 2020-08-14
Compare model params instead of pickle dump

convoliution:
@nhatsmrt to "review"

Some tests this morning failed because the `CLASSIFIER_CONTENTS` comparison failed, even though it should by all means be the same model; it might not be a good idea to rely on `pickle.dump()` being deterministic, so this change compares the actual model parameters.


nhatsmrt:
Interesting, didn't know that this could happen. If the tests don't fail otherwise, LGTM.

PR #1314, opened 2020-08-14
Add RegisteredModelVersion.log_training_data()

convoliution:
@conradoverta Is there a stronger test I can use for this from the Client side?


conradoverta (commented):
You can get the same endpoint

conradoverta (approved):
LGTM

PR #1315, opened 2020-08-15
VR-5936 Add brief examples to Strategy, Resource, Metric, and Rule docstrings

nhatsmrt:
@convoliution See if the example makes sense.
Also fix some small mix-up in docstrings for autoscaling metrics, and docstring for strategy.

### DIFF
--- a/client/verta/verta/deployment/resources.py
+++ b/client/verta/verta/deployment/resources.py
@@ -36,6 +36,13 @@ class CpuMillis(_Resource):
     """
     Number of CPU milli allowed for Endpoint. Must be an integer greater than 0.
 
+    Examples
+    --------
+    .. code-block:: python
+
+        from verta.deployment.resources import CpuMillis
+        cpu = CpuMillis(128)
+
     """
     milli_err_msg = "`cpu_millis` must be int greater than 0"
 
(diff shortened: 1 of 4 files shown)


client/verta/verta/deployment/resources.py
```diff
@@ -36,6 +36,13 @@ class CpuMillis(_Resource):
     """
     Number of CPU milli allowed for Endpoint. Must be an integer greater than 0.
 
+    Examples
+    --------
+    .. code-block:: python
+
+        from verta.deployment.resources import CpuMillis
+        cpu = CpuMillis(128)
```
convoliution: 1000 millis = 1 CPU core, so a nice fraction of 1000 like `250` makes more sense than `128`.
```suggestion
        cpu = CpuMillis(250)
```

client/verta/verta/deployment/update/_strategies.py
```diff
@@ -30,7 +42,19 @@ def _as_build_update_req_body(self, build_id):
             'strategy': self._STRATEGY,
         }
 
+
 class CanaryUpdateStrategy(_UpdateStrategy):
+    """
+    Represents canary update strategy for Endpoint.
+
```
convoliution, line 49: Could you also do me a favor and move the **Parameters** from the init docstring into this class docstring (and then remove the init docstring)? See [`Client`](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py) for what that would look like.

convoliution (changes requested):

client/verta/verta/deployment/update/_strategies.py
nhatsmrt: Done.

client/verta/verta/deployment/resources.py
nhatsmrt: Fixed.

convoliution (approved):

PR #1321, opened 2020-08-17
VR-5919 Implement endpoint.create_access_token()

nhatsmrt:


client/verta/verta/_deployment/endpoint.py
```diff
@@ -407,6 +407,26 @@ def get_access_token(self):
             return None
         return tokens[0]['creator_request']['value']
 
+    def create_access_token(self, token):
```
convoliution, line 399: Let's also add a typecheck for token: is `six.string_types`

convoliution (changes requested):

client/verta/verta/_deployment/endpoint.py
nhatsmrt: Fixed.

client/verta/verta/_deployment/endpoint.py
```diff
@@ -396,6 +396,29 @@ def get_access_token(self):
...
+        Creates an access token for the Endpoint.
+
+        Parameters
+        ----------
+        token : str
+            Token to create.
+
+        """
+        if not isinstance(token, six.string_types):
+            raise TypeError("\"token\" must be a string.")
```
convoliution: Convention around this parts is to use backticks for parameter names
```suggestion
            raise TypeError("`token` must be a string.")
```

convoliution (approved):

PR #1327, opened 2020-08-17
Import RegistryService_pb2 as _RegistryService

convoliution:
Also @nhatsmrt to review.

`RegistryService_pb2` was being imported as either `_RegisteredModelService` or `_ModelVersionService`, so I'm renaming it to `_RegistryService` for consistency (also to be consistent with the rest of the Client, which tries to keep the original proto file name).


nhatsmrt:
LGTM.

evgeniilezhnin (approved):

PR #1329, opened 2020-08-17
VR-6008 Add tutorial for using a configuration file for endpoint update

nhatsmrt:


client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,57 @@
+Updating Endpoint with Config File
+==================================
+
+.. TODO: link to Endpoint.update tutorial
+
+Sometimes, it is more convenient to specify update configuration in a separate, human-readable file, than to cram everything in the Python script or the CLI command.
```
conradoverta:
```suggestion
Sometimes, it is more convenient to specify update configuration in a separate, human-readable file, than to have everything in a Python script or a split CLI command.
```

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,57 @@
+Updating Endpoint with Config File
+==================================
```
conradoverta: Please follow the template

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,57 @@
...
+                          "value": "0.1"}
+                    ]},
+                    {"rule": "error_4xx_rate",
+                     "rule_parameters": [
+                        {"name": "threshold",
+                         "value": "1"}
+                    ]}
+                ]
+            }
+        }
```
conradoverta: This is a mixed formatted JSON, it seems. I'd recommend using a standard formatting instead, as this is pretty unusual and harder to consume.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,57 @@
...
+
+This allows for easier inspection and modification later on.
+
+Verta's Client and CLI support updating Endpoint with JSON or YAML config file. This tutorial will show how it can be done, using the Client and CLI.
+
+Updating via Client
+-------------------
+
+With the client, instead of using ``endpoint.update()``, we will use the ``endpoint.update_from_config()`` method, which takes only the path to the config file:
```
conradoverta: API links?

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. code-block:: python
+
+    endpoint.update_from_config("config_file.json")
+
+
+Updating via the CLI
+--------------------
+
+Updating the Endpoint with a config file via the CLI can be done using the `filename` option, or `f` for short:
```
conradoverta: Do single backticks work as expected?

nhatsmrt: I thought so; it "greyed out" the variables when I tried to build the html file. It doesn't show in GitHub preview though. I'll just double backtick just to be safe.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Done.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Hmm see if it looks better. Also the "experiment run id" part breaks the code block. I have to put it inside a string (i.e double quotes) so that highlighting is done properly.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Fixed.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Not really sure what you meant... I only noticed that my "with" should probably be capitalized, based on https://github.com/VertaAI/modeldb/blob/master/client/verta/docs/tutorials/packaging/custom_model_with_dependencies.rst.

Update: "Config" -> "Configuration".

conradoverta: Check the organization of the template. It should start with a simple statement of the problem being solved without getting into specifics. This example should probably be in a section like "Configuration JSON" with links on how to find out the fields available.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
conradoverta: Yeah, makes sense to require string

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
conradoverta: Hm. I see a mix of the two. My main goal is to be consistent all accross.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Hmm the statement "Sometimes, it is more convenient to specify update configuration in a separate, human-readable file, than to have everything in a Python script or a split CLI command." was meant to be the motivation of this tutorial. 

I think the problem being solved here is described in "Verta's Client and CLI support updating Endpoint with JSON or YAML config file. This tutorial will show how it can be done, using the Client and CLI." I'll move that to the top, and separate the rest into a new section.

nhatsmrt: For the field description, I believe Michael is working on that list. I'll add a TODO comment for updating it later.

conradoverta: > I think the problem being solved here is described in "Verta's Client and CLI support updating Endpoint with JSON or YAML config file. This tutorial will show how it can be done, using the Client and CLI." I'll move that to the top, and separate the rest into a new section.

That seems more correct.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,72 @@
...
+                ]
+            }
+        }
+
+.. TODO: Link to configuration file fields.
+
+This allows for easier inspection and modification later on. Users can find out more about the fields of the configuration file here.
+
+In the next sections, we will go through how to update the Endpoint with this configuration file, using the Client and the CLI.
+
```
conradoverta:
```suggestion
```

The good thing about tutorials is that they are short enough that these sort of self-references can be skipped =)

conradoverta (commented):
LGTM

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Done. Yeah that's kinda redundant.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,70 @@
...
+                        "rule_parameters": [
+                            {
+                                "name": "threshold",
+                                "value": "1"
+                            }
+                        ]
+                    }
+                ]
+            }
+        }
```
convoliution, line 35: Would you be able to also squeeze in autoscaling, resources, and environment variables in this example?

convoliution (changes requested):

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Yes I can. I have them prepared, but it was getting a bit long, so I cut them out. Will re-add back in.

nhatsmrt: Added.

convoliution: I don't think we can cut them out because I'd like this doc to show the full config file JSON 😆 But you're right that it runs a little long.

@conradoverta What do you think we should do? Cut things from this JSON, or show all of it (with consistent formatting)?

conradoverta: It's fine to have everything, but keep in mind that's a maintenance burden. I'd recommend figuring out how we can fetch the format of the full file automatically based on the real API. Otherwise this will become stale as a complete reference in a week or two.

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,79 @@
...
+                    {
+                        "rule": "error_4xx_rate",
+                        "rule_parameters": [
+                            {
+                                "name": "threshold",
+                                "value": "1"
+                            }
+                        ]
+                    }
+                ]
```
convoliution: I _think_ we should make this format more similar to the autoscaling metrics below. The fields are analogous, so we could make this more compact
```suggestion
                "rules": [
                    {"rule": "latency_avg_max", "rule_parameters": [{"name": "threshold", "value": "0.1"}]},
                    {"rule": "error_4xx_rate", "rule_parameters": [{"name": "threshold", "value": "1"}]}
                ]
```

convoliution (changes requested):

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
nhatsmrt: Done. Also users don't have to put the value in string.

convoliution (approved):

client/verta/docs/tutorials/deployment/endpoint_update_config.rst
```diff
@@ -0,0 +1,63 @@
+Updating endpoint with a configuration file
+===========================================
+
+.. TODO: link to Endpoint.update tutorial
```
convoliution: @nhatsmrt You should be able to link to \`update <endpoint_update.html>`__ now

nhatsmrt: Done.

convoliution (approved):

PR #1331, opened 2020-08-17
Fix endpoint update test

convoliution:
@nhatsmrt to review.

Model predictions may vary run to run, so we should check the deployed model against the original model, rather than hardcoding `[2]`


nhatsmrt:
Interesting. If the test passes, then everything looks good to me!

Side Note: Apparently "assert a == b" works if a, b are list/array of 1 element? 🤔

evgeniilezhnin (approved):

convoliution:
@nhatsmrt 
>Apparently "assert a == b" works if a, b are list/array of 1 element?

`assert a == b` would work for any `list` of any number of primitive elements (unless the elements are `floats`, which makes it risky), and it happens to work for `np.array`s of 1 element.

But that brings up a good point, which is that this should really be using `np.allclose()`

PR #1332, opened 2020-08-17
VR-5968 [workspace tests] Add tests that link things across workspaces

nhatsmrt:
Test:
- Endpoint v.s Model Version/Experiment Run
- Model Version v.s Experiment Run
- Repository/Commit v.s Experiment Run
- Dataset Version v.s Experiment Run (this fails)


convoliution:
@nhatsmrt Do you recall why this PR is in draft?

nhatsmrt:
> @nhatsmrt Do you recall why this PR is in draft?

@convoliution The final test failed. I remembered that a ticket was created for it. It's skipped anyway, so I'll undraft it.

nhatsmrt:
Fixed conflicts and re-ran tests.

convoliution:
@nhatsmrt I am so sorry
Screen Shot 2020-09-01 at 4 57 38 PM

convoliution:
>It's skipped anyway, so I'll undraft it.

I don't see a skipped test here 🤔 but I might be overlooking it.

~I'm going to see if I can find the related backend ticket, and check its status~
it's https://github.com/VertaAI/modeldb/pull/1403, which is still open today.
I think we can afford to wait for it before resolving this PRs conflicts.

convoliution (approved):

client/verta/tests/conftest.py
```diff
@@ -356,14 +356,17 @@ def commit(repository):
...
     created_datasets = []
 
     yield created_datasets
 
     if created_datasets:
-        utils.delete_datasets(list(set(dataset.id for dataset in created_datasets)), client._conn)
+        try:
+            utils.delete_datasets(list(set(dataset.id for dataset in created_datasets)), client._conn)
+        except requests.HTTPError:
+            utils.delete_datasets(list(set(dataset.id for dataset in created_datasets)), client_2._conn)
```
convoliution:
```suggestion
        utils.delete_datasets(list(set(dataset.id for dataset in created_datasets)), dataset._conn)
```

client/verta/tests/conftest.py
```diff
@@ -356,14 +356,17 @@ def commit(repository):
 
 
 @pytest.fixture
-def created_datasets(client):
+def created_datasets(client, client_2):
```
convoliution: We can't add `client_2` as a requisite for this fixture because `client_2` causes the test to skip if its auth creds aren't present, so this would skip tests that it shouldn't.

Fortunately, I do believe `dataset`s have their own `_conn` that should be usable for the deletion (see below).
```suggestion
def created_datasets():
```

convoliution (changes requested):
Good lordy, I never hit enter on this one

client/verta/tests/conftest.py
nhatsmrt: This wouldn't work either, because the `dataset._conn` is outside the context of the list comprehension which defines dataset.

I think what I just pushed should work.

client/verta/tests/conftest.py
nhatsmrt: Fixed.

client/verta/tests/conftest.py
```diff
@@ -389,6 +393,16 @@ def model_version(registered_model):
     yield registered_model.get_or_create_version()
 
 
+@pytest.fixture
+def created_endpoints():
```
conradoverta, line 397: Duplicated?

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -601,3 +601,71 @@ def test_update_twice(self, client, registered_model, created_endpoints):
         endpoint.update(new_model_version, DirectUpdateStrategy(), wait=True)
         test_data = np.random.random((4, 12))
         assert np.array_equal(endpoint.get_deployed_model().predict(test_data), new_classifier.predict(test_data))
+
+    def test_update_from_run_diff_workspace(self, client, organization, created_endpoints, experiment_run, model_for_deployment):
```
conradoverta, line 605: I'm going to assume that `client` wasn't used to create the other objects in this argument. Let me know if I'm wrong.

conradoverta (commented):
LGTM but I don't have enough context to evaluate https://github.com/VertaAI/modeldb/pull/1332#discussion_r485270195

client/verta/tests/test_endpoint/test_endpoint.py
nhatsmrt: Actually both are created by the same client, but has different workspaces: one in a personal workspace, the other in an organizational workspace.

nhatsmrt: There's a test for two objects created by two different clients below.

client/verta/tests/conftest.py
nhatsmrt: Yep. Probably merging mishap. De-duplicated.

convoliution (approved):

PR #1333, opened 2020-08-18
VR-5848: Log verta_model_artifacts for class models

convoliution:
I've filed VR-6031 (1hr) to implement `fetch_artifacts()` tomorrow; I hadn't accounted for it in this task.

Since we have a base class, we should also really move `log_model()` to `_DeployableEntity`; I've filed VR-6030 (3hr) next sprint for that refactor.

### DIFF
--- a/client/verta/verta/_registry/modelversion.py
+++ b/client/verta/verta/_registry/modelversion.py
@@ -279,7 +279,7 @@ class RegisteredModelVersion(_ModelDBRegistryEntity, _DeployableEntity):
         self._update(self._msg, method="PUT")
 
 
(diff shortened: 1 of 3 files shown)


client/verta/verta/_registry/modelversion.py
```diff
@@ -279,7 +294,7 @@ def del_model(self):
         self._update(self._msg, method="PUT")
 
 
-    def log_artifact(self, key, artifact, overwrite=False, extension=None):
+    def log_artifact(self, key, artifact, overwrite=False, _extension=None):
```
convoliution, line 297: I'm thinking this shouldn't have been added as a public param in the first place 😞

conradoverta (approved):

PR #1336, opened 2020-08-18
VR-6022 add user to org

evgeniilezhnin:


convoliution (changes requested):
Is `share_with` meant to be a new member's email, or username (or both)?
Could you add a comment/docstring to explain what it is? 🙏

nhatsmrt:
Small request: can you add "name" and "id" property to Organization? Previously you returned the message itself, so org.id and org.name are valid and were used in tests + fixtures. It's probably easier to still provide these properties than to fix the tests.

evgeniilezhnin:
> share_with

both. Added

evgeniilezhnin:
> Small request: can you add "name" and "id" property to Organization? Previously you returned the message itself, so org.id and org.name are valid and were used in tests + fixtures. It's probably easier to still provide these properties than to fix the tests.

added

convoliution (approved):

PR #1337, opened 2020-08-18
VR-6017 add attributes overwrite

evgeniilezhnin:


client/verta/verta/_registry/modelversion.py
```diff
@@ -752,11 +756,16 @@ def add_attributes(self, attrs):
 
         # build KeyValues
         attribute_keyvals = []
+        if overwrite:
+            existing_attrs = {}
+        else:
+            existing_attrs = self.get_attributes()
```
convoliution: Because you have `if not key in existing_attrs or overwrite:` later, this `if overwrite` I think is unnecessary, right?
```suggestion
        existing_attrs = self.get_attributes()
```

convoliution (changes requested):

client/verta/verta/_registry/modelversion.py
evgeniilezhnin: Yes, it is unnecessary. It is used just to not call get_attributes if overwrite=True.

convoliution (approved):

PR #1338, opened 2020-08-18
VR-5969 set mask for update env

evgeniilezhnin:


client/verta/verta/_registry/modelversion.py
```diff
@@ -433,9 +433,9 @@ def log_environment(self, env):
 
         self._fetch_with_no_cache()
         self._msg.environment.CopyFrom(env._msg)
```
convoliution: Since we are using a separate message object, this line
```python
self._msg.environment.CopyFrom(env._msg)
```
I think can be removed

convoliution (approved):
One small suggestion

client/verta/verta/_registry/modelversion.py
evgeniilezhnin: yes, thanks

evgeniilezhnin:
Needs proxy changes deployed to dev before merging current PR: https://github.com/VertaAI/protos-all/pull/63

convoliution:
Patch mask is _not_ needed for the other things (artifacts, attributes), right?

evgeniilezhnin:
> Patch mask is _not_ needed for the other things (artifacts, attributes), right?

It works alright for attributes, but I'm not sure how to use PATCH for artifacts. Because there is a complex logic in client to calculate resulting artifacts, so only PUT can be used, right?

convoliution:
Ah yes that's right, artifacts are currently using PUT. That is fine, then. 👍

PR #1351, opened 2020-08-19
Add update configuration tutorial

convoliution:


client/verta/docs/tutorials/deployment/endpoint_config.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    verta deployment update endpoint /some-path --model-version-id "<id>" \
+        --strategy direct \
+        --autoscaling '{"max_replicas": 4, "min_scale": 0.5}' \
+        --autoscaling-metric '{"metric": "cpu_utilization", "parameters": [{"name": "target", "value": "0.75"}]}' \
+        --resources '{"cpu_millis": 250, "memory": "512Mi"}' \
+        --env-vars '{"CUDA_VISIBLE_DEVICES": "1,2"}'
+
+Each argument takes a JSON string representing their respective values. For a reference of accepted
+values, see here. To set multiple metrics, ``--autoscaling-metric`` can be provided more than once.
```
conradoverta: Missing link in `here`?

client/verta/docs/tutorials/deployment/endpoint_config.rst
```diff
@@ -0,0 +1,89 @@
...
+deployed.
+
+.. code-block:: python
+
+    from verta.deployment.resources import (CpuMillis, Memory)
+
+    resources = [
+        CpuMillis(250),
+        Memory("512Mi"),
+    ]
```
conradoverta: This is confusing, right? What happens if I pass CPU twice?

Also, we need to document what `Mi` is, since that is _very_ implementation specific.

client/verta/docs/tutorials/deployment/endpoint_config.rst
```diff
@@ -0,0 +1,89 @@
+Endpoint configuration
```
conradoverta: This tutorial will become massive, with sub sub sections for everything that is configurable. It seems less like a tutorial and more like a section-level thing, right?

client/verta/docs/tutorials/deployment/endpoint_config.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Environment variables
+^^^^^^^^^^^^^^^^^^^^^
+
+``env_vars`` takes a dictionary of string environment variable names to string values, and will be
+made available to the model when it is deployed.
+
+.. code-block:: python
+
+    env_vars = {'CUDA_​VISIBLE_​DEVICES': "1,2"}
```
conradoverta: I would advise against providing examples with `CUDA_​VISIBLE_​DEVICES` as that has a very strong implicit meaning. I would recommend picking an example with less loaded meaning.

convoliution: @conradoverta 
Could you recommend another example? This was the only env var I could think of that has some relevant to ML.

client/verta/docs/tutorials/deployment/endpoint_config.rst
convoliution: >It seems less like a tutorial and more like a section-level thing, right?

I'm trouble understanding what "section-level thing" means.

Would you suggest making each of these subsubsections (autoscaling, resources) its own page? If so, that makes sense to me.

client/verta/docs/tutorials/deployment/endpoint_config.rst
convoliution: >This is confusing, right? What happens if I pass CPU twice?

That's a great question! Would you suggest changing the Client API for resources? I think there's still time to do that. I'm imagining something like
```python
Resources(cpu_millis=250, memory="512Mi")
```
---
>Also, we need to document what Mi is, since that is very implementation specific.

ack; I had asked Evgenii to accept units based on https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory. I can copy the explanation from there.

client/verta/docs/tutorials/deployment/endpoint_config.rst
convoliution: It's TODO, right below.

client/verta/docs/tutorials/deployment/endpoint_config.rst
conradoverta: `LOG_LEVEL=debug`?

client/verta/docs/tutorials/deployment/endpoint_config.rst
conradoverta: Yes. We'll soon add more and more knobs.

client/verta/docs/tutorials/deployment/endpoint_config.rst
conradoverta: Something like that would work fine. I don't imagine we'll have too many resource options and we can split up if we do. But right now I don't know what happens if I violate what I imagine are constraints in the input.

Works for me. Probably we should have documented in the API itself.

client/verta/docs/tutorials/deployment/endpoint_config.rst
conradoverta: Oh, ok. I'm not used leaving a placeholder bc I lose track of them =) But if it works for you, go for it.

client/verta/docs/tutorials/deployment/endpoint_config.rst
convoliution: I've split endpoint_config.rst into knob-focused tutorials, and then deleted endpoint_config.rst because all that remained of it was "`endpoint.update()` exists! See other tutorials." which imo is redundant and doesn't address a particular problem

client/verta/docs/tutorials/deployment/deployment.rst
```diff
@@ -3,12 +3,32 @@ Deployment
...
+Endpoint creation
+-----------------
+
+* `Creating and retrieving an endpoint <endpoint_creation.html>`__.
+
+Endpoint update
+---------------
+
+* `Updating an endpoint <endpoint_update.html>`__
+* `Updating an endpoint with canary strategy <endpoint_canary_update.html>`__.
```
conradoverta:
```suggestion
* `Updating an endpoint with canary strategy <endpoint_canary_update.html>`__
```

client/verta/docs/tutorials/deployment/deployment.rst
```diff
@@ -3,12 +3,32 @@ Deployment
...
+The following tutorials will walk you through the entire process of using an endpoint, Verta's configurable API for deploying machine learning models.
 
-* `Creating and retrieving an endpoint <endpoint_creation.html>`_.
-* `Updating an endpoint <endpoint_update.html>`_
-* `Updating an endpoint with canary strategy <endpoint_canary_update.html>`_.
-* `Querying an endpoint <endpoint_query.html>`_.
+Endpoint creation
+-----------------
+
+* `Creating and retrieving an endpoint <endpoint_creation.html>`__.
```
conradoverta:
```suggestion
* `Creating and retrieving an endpoint <endpoint_creation.html>`__
```
Either dots or no dots =)

client/verta/docs/tutorials/deployment/endpoint_autoscaling.rst
```diff
@@ -0,0 +1,56 @@
+Endpoint autoscaling
+====================
+
+Through an endpoint update, you can configure the deployment's autoscaling behavior: upper and
+lower bounds for replication, scale-up rrate, and metrics to trigger it.
```
conradoverta:
```suggestion
lower bounds for replication, scale-up rate, and metrics to trigger it.
```

client/verta/docs/tutorials/deployment/endpoint_update.rst
```diff
@@ -50,7 +50,7 @@ Or using an experiment run:
 
 Advanced endpoint updates
 -------------------------
-Certain properties of the endpoint can also be `configured <endpoint_config.html>`__ during the
+Certain properties of the endpoint can also be :ref:`configured <endpoint-config>` during the
```
conradoverta, line 53: TIL

conradoverta (approved):
LGTM

client/verta/docs/tutorials/deployment/endpoint_update.rst
convoliution: Oh good catch, thanks!

client/verta/docs/tutorials/deployment/endpoint_autoscaling.rst
convoliution: This dang MacBook keyboard

PR #1352, opened 2020-08-19
VR-6031: Implement ModelVersion.fetch_artifacts()

convoliution:
- `experimentrun._CACHE_DIR` –> `deployable_entity._CACHE_DIR`
- `ExperimentRun._cache()` –> `_DeployableEntity._cache_file()`
- `ExperimentRun._get_cached()` –> `_DeployableEntity._get_cached_file()`
- `ExperimentRun.fetch_artifacts()` –> `_DeployableEntity.fetch_artifacts()` with some tinkering; see comments below


client/verta/verta/_tracking/deployable_entity.py
```diff
@@ -35,6 +47,167 @@ def _histogram_endpoint(self):
...
+        # validate that `keys` are actually logged
+        self._refresh_cache()
+        existing_artifact_keys = {artifact.key for artifact in self._msg.artifacts}
+        unlogged_artifact_keys = set(keys) - existing_artifact_keys
+        if unlogged_artifact_keys:
+            raise ValueError("`keys` contains keys that have not been logged: {}".format(sorted(unlogged_artifact_keys)))
+
+        # get artifact checksums
+        paths = {artifact.key: artifact.path
+                 for artifact in self._msg.artifacts}
```
convoliution, line 189: This was changed from
```python
response = _utils.make_request("GET",
                               "{}://{}/api/v1/modeldb/experiment-run/getArtifacts".format(self._conn.scheme, self._conn.socket),
                               self._conn, params={'id': self.id})
paths = {artifact['key']: artifact['path']
         for artifact in _utils.body_to_json(response)['artifacts']}
```
because thanks to caching on `self._msg`, we actually don't need to call an endpoint here at all.

client/verta/verta/_tracking/deployable_entity.py
```diff
@@ -35,6 +47,167 @@ def _histogram_endpoint(self):
...
+
+            # check cache, otherwise write to cache
+            #     "try-get-then-create" can lead multiple threads trying to write to the cache
+            #     simultaneously, but artifacts being cached at a particular location should be
+            #     identical, so multiple writes would be idempotent.
+            path = self._get_cached_file(filename)
+            if path is None:
+                contents = self._get_artifact(key)
+                if isinstance(contents, tuple):  # ExperimentRun._get_artifact() returns (contents, path_only)
+                    contents, _ = contents  # TODO: raise error if path_only
```
convoliution: This was changed from 
```python
contents, _ = self._get_artifact(key)  # TODO: raise error if path_only
```
Because `ModelVersion._get_artifact()` only returns one value instead of `ExperimentRun`'s two, which is rather annoying and needs to be unified when these child classes are more thoroughly cleaned up.

client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -462,3 +463,27 @@ def test_patch(self, registered_model):
         assert version.get_labels() == LABELS
 
         assert version.get_attributes() == ATTRIBUTES
+
+# TODO: move related tests from TestModelVersion before merging
```
conradoverta: Why hasn't it been done?

client/verta/verta/_tracking/deployable_entity.py
conradoverta: Can you note that in the code instead of a PR comment? This is very explanatory and will be 100% be lost (I see the comment, but that was more of an FYI than a FIXME explaining the issue.

conradoverta (approved):
Couple of small questions. LGTM otherwise.

client/verta/tests/test_model_registry/test_model_version.py
convoliution: It would've made the diff a lot more verbose for the reviewer.

conradoverta: Cool

PR #1354, opened 2020-08-19
VR-5963 add attribute

evgeniilezhnin:


client/verta/verta/_cli/registry/update.py
```diff
@@ -103,4 +92,16 @@ def update_model_version(model_name, version_name, label, model, custom_module,
         model_version.log_environment(Python(requirements=reqs))
 
     if description:
-        registered_model.set_description(description)
+        model_version.set_description(description)
```
evgeniilezhnin, line 94: found my mistake. This should update model version, but not model

client/verta/verta/_cli/multiple_arguments.py
```diff
@@ -0,0 +1,28 @@
+import click
+
+
+class MultipleArguments:
```
convoliution: It looks like every time this is used, it is created and then immediately `for_each()` is called. Could this whole class be turned into one function, instead?

client/verta/verta/_registry/modelversion.py
```diff
@@ -794,6 +794,9 @@ def get_attributes(self):
         self._refresh_cache()
         return _utils.unravel_key_values(self._msg.attributes)
 
+    def _get_attribute_keys(self):
+        return map(lambda attribute: attribute.key, self.get_attributes())
```
convoliution: To match `get_artifact_keys()`
```suggestion
        return list(map(lambda attribute: attribute.key, self.get_attributes()))
```

client/verta/verta/_cli/registry/update.py
```diff
@@ -103,4 +92,17 @@ def update_model_version(model_name, version_name, label, model, custom_module,
...
+
+def add_attribute(model_version, key, value):
+    model_version.add_attribute(key, json.loads(value))
+
+
+def add_attributes(model_version, attribute, overwrite):
+    attributes = MultipleArguments(attribute, "attribute")
+    attributes.for_each(
+        lambda key, value: add_attribute(model_version, key, value),
+        lambda: model_version._get_attribute_keys(), overwrite)
```
convoliution: It is confusing for me that there are util functions called `add_attribute()`, when `ModelVersion` _also_ has `add_attribute()` methods. And these don't seem to be doing a lot. Could we remove these functions and just put this code in `update_model_version`?

convoliution (changes requested):
This is very very confusing to me 😕 I have read through it a few times, and would probably still have difficulty using the new utilities. Could you simplify them, if you can?

client/verta/verta/_cli/registry/update.py
evgeniilezhnin: I agree about add_attribute. I just created it because I thought it will be doing more parsing like in these cases: '3.6'. Because json doesn't support single quotes. But it should be alright with the plain json like: "3.6", right? So I inlined it now.

But add_attributes is needed because it is used to test the feature without calling any additional things from model version update.

client/verta/verta/_cli/multiple_arguments.py
evgeniilezhnin: Yes, you are right. Replaced to single function and moved to utils.

client/verta/verta/utils.py
```diff
@@ -312,3 +314,27 @@ def predict(self, **kwargs):
...
+            (key, _) = pair
+            if key == "model":
+                raise click.BadParameter("the key \"model\" is reserved for model")
+
+            if not overwrite and key in argument_keys:
+                raise click.BadParameter(
+                    "key \"{}\" already exists; consider using --overwrite flag".format(key))
+
+        for (key, path) in argument:
+            action(key, path)
```
convoliution: Please move this to [_internal_utils/_utils.py](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_internal_utils/_utils.py) instead

convoliution (approved):
**One more request** then okay to merge.

Still looks confusing 😅 It may need adjustments it in the future, but it is good that we have a util for this

client/verta/verta/utils.py
evgeniilezhnin: moved

PR #1358, opened 2020-08-19
VR-6047 Client should get error msg from build status endpoint during endpoint.update(wait)

nhatsmrt:
This is more complicated than I thought initially. Please review carefully.


convoliution:
I'll need to test this manually, because this diff is wild!

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -600,3 +600,23 @@ def test_update_twice(self, client, registered_model, created_endpoints):
...
+        created_endpoints.append(endpoint)
+
+        update_status = endpoint.update(model_version, DirectUpdateStrategy(), True) # this should work
+        assert update_status["status"] == "active"
+
+        with pytest.raises(RuntimeError) as excinfo:
+            endpoint.update(experiment_run, DirectUpdateStrategy(), True) # this should fail, and not take forever!
+
+        excinfo_value = str(excinfo.value).strip()
+        assert "Could not find a version that satisfies the requirement blahblahblah==3.6.0" in excinfo_value
```
convoliution: - No need to test the successful case; other tests already have that covered (and the Client test suite is getting strapped for runtime 😆)
- As you noted, a bug in the Client would result in this test taking forever, rather than failing. We should use a loop that checks the endpoint status every ~5 seconds instead of `wait=True`, and fail the test if it exceeds a time limit.

convoliution (changes requested):

client/verta/tests/test_endpoint/test_endpoint.py
nhatsmrt: Hmm not sure if this is what you want, but take a look at the (fixed) test!

convoliution: @nhatsmrt I had forgotten about "taking forever" only occurring after a successful initial update.

Would you know if that's still the case? I feel like I've encountered "taking forever" recently even without a successful initial update, but I could be mistaken

nhatsmrt: Hmm, for me it only fails when there is a successful initial update.

nhatsmrt: @convoliution I rmb replying to this hmmm...
Anyway for me the "taking forever" bug only happens with a successful initial update.

convoliution: @nhatsmrt Alright. In that case, do you think we could graft this "taking forever" case to the end of an existing endpoint update test?

nhatsmrt: @convoliution and remove this test?

convoliution: Yes.

nhatsmrt: @convoliution Hmm about your other suggestion:

"We should use a loop that checks the endpoint status every ~5 seconds instead of wait=True, and fail the test if it exceeds a time limit."

The point of this test is to check if endpoint.update(..., wait=True) does not take forever, right? Now say we don't use wait=True:
- If we check the endpoint's status (which gets the stage's status), it will never be ready (hence the need for this PR). 
- If we check the actual build status, then it just defeats the point of the test, because we basically just repeat the implementation. Also if the client is buggy and it does take forever in the case of "wait=True", we won't detect it by checking build anyway (since that works properly all the time).

nhatsmrt: I wonder if pytest has a mechanism to just time out a block of code/a test if it takes too long 🤔

convoliution: @nhatsmrt
>it just defeats the point of the test

Hah! You're right. My suggestions so far have all been misses. But ultimately we have these key constraints:

- this test *must* not literally take forever to execute, because it'll block the rest of the suite
- this test *should* not require additional pytest plugins (e.g. time limit on a test), because plugins caused various issues in the past, but if the suite still works then sure
- this test *should* be attached to the end of an existing one, because the suite is approaching the limits of its runtime

I'm not sure what the solution will be (I'm tied up at the moment with a major bug) so I'll leave it to you to try to find a solution! 🙇

convoliution:
@nhatsmrt A sincere apology for going backwards on this, but I think this timeout mechanism is more complex than I imagined (I'd be hesitant to use it because I'm not sure if it has wider side effects). Could we move back to the potentially-taking-forever test? 🙇

nhatsmrt:
> @nhatsmrt A sincere apology for going backwards on this, but I think this timeout mechanism is more complex than I imagined (I'd be hesitant to use it because I'm not sure if it has wider side effects). Could we move back to the potentially-taking-forever test? 🙇

Done.

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -213,17 +213,32 @@ def _update_from_build(self, update_body, wait=False):
...
+                    (status_dict['status'] == "active" and len(status_dict['components']) > 1):
                 print(".", end='')
                 sys.stdout.flush()
                 time.sleep(5)
                 status_dict = self.get_status()
 
             print()
-            if self.get_status()['status'] == "error":
-                failure_msg = self.get_status()['components'][0].get('message', "no error message available")
+            # is this check necessary?
```
conradoverta: Why wouldn't it be?

conradoverta (approved):
LGTM

client/verta/verta/endpoint/_endpoint.py
nhatsmrt: A wrong assumption I used to have. Removed.

convoliution (approved):

PR #1360, opened 2020-08-19
Refactor Endpoint._form_update_body()

convoliution:
Enables #1361

Actually also refactors `Endpoint.update()` a little so that `_form_update_body()` (now `_create_update_body`) is more re-usable.

I'd recommend reviewing this commit-by-commit.

- rename `_form_update_body()` to `_create_update_body`, because "form" has been quite confusing for me (I kept interpreting it as a noun)
- removes `build_id` from Strategy's JSON conversion method, and instead assigns it in `Endpoint.update()`
- moves typechecks from `update()` to `_create_update_body()` so they're re-usable

PR #1361, opened 2020-08-19
VR-6049: Add Endpoint.download_manifest()

convoliution:
Actually `Client.download_endpoint_manifest()`


client/verta/verta/_deployment/endpoint.py
```diff
@@ -470,3 +476,48 @@ def get_update_status(self):
...
+            resources=None, autoscaling=None, env_vars=None):
+        """
+        Downloads this Endpoint's Kubernetes manifest YAML.
+
+        Parameters
+        ----------
+        download_to_path : str
+            Path to download manifest YAML to.
+        name : str
+            Name of the Endpoint.
```
convoliution: @conradoverta Is this an accurate description for `metadata.name`?

conradoverta: yep

client/verta/verta/_deployment/endpoint.py
```diff
@@ -470,3 +476,48 @@ def get_update_status(self):
...
+            Environment variables.
+
+        Returns
+        -------
+        downloaded_to_path : str
+            Absolute path where deployment YAML was downloaded to. Matches `download_to_path`.
+
+        """
+        data = {
+            'endpoint': {'path': self.path},
```
conradoverta: Downloading the manifest should not require the endpoint to already exist before.

convoliution: @conradoverta 
Shoot, really? Hrm the code's not going to work as a static method because it needs the `._conn` of an object instance. The best I can do right now would be a `client.download_endpoint_manifest()`. How does that sound?

conradoverta: Makes sense to me

convoliution:
Ugh I thought GitHub would be able to ignore the diffs already merged into `master` in https://github.com/VertaAI/modeldb/pull/1360, but I guess not.

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -118,6 +121,35 @@ def test_repr(self, client, created_endpoints, experiment_run, model_for_deploym
...
+            resources=resources,
+            env_vars=env_vars,
+        )
+
+        # can be loaded as YAML
+        with open(filepath, 'rb') as f:
+            manifest = yaml.safe_load(f)
+
+        assert manifest['kind'] == "Endpoint"
+        assert manifest['spec']['endpoint']['spec']['ingress']['path'] == path
```
conradoverta: This will fail very soon when we add workspace to be part of the path =) I'd recommend checking something that is directly used by k8s, like env vars.

conradoverta (commented):
One suggestion. LGTM otherwise.

PR #1362, opened 2020-08-19
VR-6058 Mulitpart upload loop sometimes raises UnboundLocalError

nhatsmrt:
Seems to work (I tried disconnecting the network in the middle of uploading to test this). Not sure if there's a more pythonic/elegant way to achieve this effect though.

### DIFF
--- a/client/verta/verta/_registry/modelversion.py
+++ b/client/verta/verta/_registry/modelversion.py
@@ -511,11 +511,15 @@ class RegisteredModelVersion(_ModelDBRegistryEntity, _DeployableEntity):
 
                 # upload part
                 #     Retry connection errors, to make large multipart uploads more robust.
-                for _ in range(3):
+                for i in range(3):
                     try:
                         response = _utils.make_request("PUT", url, self._conn, data=part_stream)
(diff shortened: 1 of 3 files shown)


client/verta/verta/_registry/modelversion.py
```diff
@@ -511,11 +511,15 @@ def _upload_artifact(self, key, file_handle, artifact_type, part_size=64*(10**6)
 
                 # upload part
                 #     Retry connection errors, to make large multipart uploads more robust.
-                for _ in range(3):
+                for i in range(3):
```
convoliution: Ooh, let's make the number of tries a variable so if we want to change this `3`, we don't need to update the `i == ` later.
Also check that `MAX_TRIES` isn't already being used.
```suggestion
                MAX_TRIES = 3
                for i in range(MAX_TRIES):
```

convoliution (changes requested):

client/verta/verta/_registry/modelversion.py
nhatsmrt: You got it!

convoliution (approved):

PR #1363, opened 2020-08-19
VR-6070 Add config JSON values to endpoint config classes

nhatsmrt:
I think that's all of them 🤔

### DIFF
--- a/client/verta/verta/deployment/autoscaling/metrics.py
+++ b/client/verta/verta/deployment/autoscaling/metrics.py
@@ -47,13 +47,11 @@ class CpuUtilizationTarget(_AutoscalingMetric):
     """
     CPU utilization target to trigger autoscaling.
 
-    The JSON equivalent for this is:
+    The JSON equivalent for this (as an element of autoscaling metrics JSON) is:
 
     .. code-block:: json
 
-        {
-            "metrics": [{"metric": "cpu_utilization", "parameters": [{"name": "target", "value": 0.5}]}]
-        }
+        {"metric": "cpu_utilization", "parameters": [{"name": "target", "value": 0.5}]}
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


client/verta/verta/deployment/update/rules.py
```diff
@@ -54,6 +54,24 @@ class MaximumAverageLatencyThresholdRule(_UpdateRule):
...
+                    "rule": "latency_avg_max",
+                    "rule_parameters": [
+                        {
+                            "name": "threshold",
+                            "value": "0.1"
+                        }
+                    ]
+                }
+            ]
+        }
```
convoliution: Similar to https://github.com/VertaAI/modeldb/pull/1329#discussion_r474103106, I think we could reduce update rules to fewer lines like we have for autoscaling metrics

client/verta/verta/deployment/autoscaling/metrics.py
```diff
@@ -47,13 +47,20 @@ class CpuUtilizationTarget(_AutoscalingMetric):
     """
     CPU utilization target to trigger autoscaling.
 
+    The JSON equivalent for this is:
+
+    .. code-block:: json
+
+        {
+            "metrics": [{"metric": "cpu_utilization", "parameters": [{"name": "target", "value": 0.5}]}]
+        }
```
convoliution: 1. I think this should just be the single metric itself, so this can also be copy&paste as a CLI argument
1. Could you also mention (as part of the line above, "The JSON equivalent for this is:") that this is meant to be an element of autoscaling metrics JSON?
    - same thing for update rules

```suggestion
        {"metric": "cpu_utilization", "parameters": [{"name": "target", "value": 0.5}]}
```

convoliution (changes requested):

client/verta/verta/deployment/autoscaling/metrics.py
nhatsmrt: Done.

client/verta/verta/deployment/update/rules.py
nhatsmrt: Done.

convoliution (approved):
image

PR #1367, opened 2020-08-20
VR-6013 find version with workspace

evgeniilezhnin:


convoliution (changes requested):
Could we add some extra logic: when user uses `with_model()` then workspace is reset, and when user uses `with_workspace()` then registered_model_id is reset. I think that makes sense 🤔

evgeniilezhnin:
> Could we add some extra logic: when user uses `with_model()` then workspace is reset, and when user uses `with_workspace()` then registered_model_id is reset. I think that makes sense

That's a good idea. That should help to not send both id and workspace.

evgeniilezhnin:
> Could we add some extra logic: when user uses `with_model()` then workspace is reset, and when user uses `with_workspace()` then registered_model_id is reset. I think that makes sense

There is 1 problem. When I set empty string to workspace name and also some model id and then try to send it to backend, it returns validation error. That is why I need to create new FindModelVersionRequest message and copy fields one by one without workspace

evgeniilezhnin:
> > Could we add some extra logic: when user uses `with_model()` then workspace is reset, and when user uses `with_workspace()` then registered_model_id is reset. I think that makes sense
> 
> There is 1 problem. When I set empty string to workspace name and also some model id and then try to send it to backend, it returns validation error. That is why I need to create new FindModelVersionRequest message and copy fields one by one without workspace

@ad-47 maybe it is something that can be changed in backend validation? When I send some model id and an empty workspace name shouldn't it be alright?

anandjakhaniya:
Ok, so on that case we are ignoring workspace and just get a registered model based on ReadOnly access of the current user and then find ModelVersion for that model. it should work https://github.com/VertaAI/registry/blob/master/src/main/java/ai/verta/registry/services/registry/RegistryDAORdbImpl.java#L1271-L1282

client/verta/verta/_registry/modelversions.py
```diff
@@ -45,10 +48,15 @@ def _create_element(self, msg):
...
         new_list = copy.deepcopy(self)
+        new_list._msg = _RegistryService.FindModelVersionRequest()
         if registered_model:
             new_list._msg.id.registered_model_id = registered_model.id
         else:
             new_list._msg.id.registered_model_id = 0
+        new_list._msg.predicates.extend(self._msg.predicates)
+        new_list._msg.pagination.CopyFrom(self._msg.pagination)
+        new_list._msg.ascending = self._msg.ascending
+        new_list._msg.sort_key = self._msg.sort_key
```
evgeniilezhnin: So @convoliution these lines is needed because named_id should be removed from a request to use model id: https://vertaai.slack.com/archives/CS7UJ399B/p1599144371104200?thread_ts=1599134352.102400&cid=CS7UJ399B

convoliution: @evgeniilezhnin I _think_ there may actually be a simpler solution. I believe adding
```python
new_list._msg.id.ClearField('named_id')
```
to the `if registered_model:` block might work, according to the [protobuf docs](https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.ClearField). In that case, we will not need any of these new lines.

client/verta/verta/_registry/modelversions.py
```diff
@@ -64,3 +72,9 @@ def page_limit(self, msg):
 
     def page_number(self, msg):
         return msg.pagination.page_number
+
+    def with_workspace(self, workspace_name=None):
+        new_list = copy.deepcopy(self)
+        new_list._msg.id.registered_model_id = 0
```
convoliution: Like my comment above, I wonder if
```suggestion
        new_list._msg.id.ClearField('registered_model_id')
```
would work, for consistency with above?

convoliution (changes requested):

client/verta/verta/_registry/modelversions.py
evgeniilezhnin: It works. Thank you

client/verta/verta/_registry/modelversions.py
evgeniilezhnin: works too

client/verta/verta/_registry/modelversions.py
evgeniilezhnin: @convoliution

convoliution (approved):
Thank you thank you!

PR #1372, opened 2020-08-20
VR-5914 Add delete() method to Client entities

nhatsmrt:
Review might be easier if going commit by commit.


convoliution (changes requested):
I don't think `client.delete_datasets()` (ff46450cb30af967edb514e2811e4e8430170472 and a08ee4d2bbc2ff719e21d5b1573f74163ed7a50a) is a good approach for now; it violates the consistency in our UX (`.delete(entities)` instead of `entity.delete()`).

We could implement `Datasets.delete()` (and `.delete()` for all the `LazyList`s), but that's outside of the scope of this ticket. I'd request reverting those two commits, and I've filed VR-6369 for `.delete()` on the collection objects.

nhatsmrt:
> I don't think `client.delete_datasets()` ([ff46450](https://github.com/VertaAI/modeldb/commit/ff46450cb30af967edb514e2811e4e8430170472) and [a08ee4d](https://github.com/VertaAI/modeldb/commit/a08ee4d2bbc2ff719e21d5b1573f74163ed7a50a)) is a good approach for now; it violates the consistency in our UX (`.delete(entities)` instead of `entity.delete()`).
> 
> We could implement `Datasets.delete()` (and `.delete()` for all the `LazyList`s), but that's outside of the scope of this ticket. I'd request reverting those two commits, and I've filed VR-6369 for `.delete()` on the collection objects.

Reverted

convoliution (approved):

PR #1377, opened 2020-08-20
Redesign endpoint resources util

convoliution:
Also @nhatsmrt to review

Changes
```python
from verta.deployment.resources import (CpuMillis, Memory)
resources = [
    CpuMillis(250),
    Memory("512Mi"),
]
```
to
```python
from verta.deployment.resources import Resources
resources = Resources(cpu_millis=250, memory="512Mi")
```

---

The old way had a couple problems:

- What happens if user has multiple `CpuMillis` or `Memory` in the list?
- `_Resource._from_dict()` was confusing because it returns a **list** of its subclass instances, and `resources = _Resource._from_dict()` doesn't make grammatical (and therefore intuitive) sense


client/verta/verta/deployment/resources.py
```diff
@@ -6,86 +6,77 @@
 from ..external import six
```
convoliution: This file's diff is a real disaster; I'd recommend just viewing the new file and reading that.

nhatsmrt:
Looks good to me, but I think this will have to move to verta.endpoint, given my recently merged PR, right?

convoliution:
@nhatsmrt The merge conflict was surprising painless. Git really pulls through sometimes.

evgeniilezhnin (approved):

PR #1378, opened 2020-08-20
VR-6125  Migrate verta.deployment to verta.endpoint

nhatsmrt:


client/verta/verta/endpoint/update/rules.py
```diff
@@ -3,7 +3,7 @@
 import abc
 import json
 
-from ...external import six
+from verta.external import six
```
convoliution: Was there a functional reason for this change? The relative import `...external` is strongly preferred.

convoliution (changes requested):

client/verta/verta/endpoint/update/rules.py
nhatsmrt: I have no idea why this is changed 😱 
Need more tea

nhatsmrt: Hmm I think that's pycharm file moving weird shennanigans. Let me fix all of them.

nhatsmrt: Fixed.

convoliution (approved):

convoliution:
fyi @evgeniilezhnin in case this impacts any of your PRs

PR #1399, opened 2020-08-25
Log model artifacts attribute in ModelVersion

convoliution:
Evidently, I didn't add an actual end-to-end test to validate this.

Another reason to refactor log_model() to the base class (and another reason to seriously not skimp on end-to-end test coverage).

### DIFF
--- a/client/verta/tests/conftest.py
+++ b/client/verta/tests/conftest.py
@@ -364,6 +364,15 @@ def model_version(registered_model):
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


conradoverta (approved):
Lgtm, but I don't see a cleanup step

convoliution:
It invokes the `created_endpoints` fixture, which has cleanup logic
```python
def endpoint(client, created_endpoints):
```

PR #1402, opened 2020-08-26
VR-6251: Change cpu_millis to cpu

convoliution:


client/verta/verta/endpoint/resources.py
```diff
@@ -39,30 +39,30 @@ class Resources(object):
...
         "E, P, T, G, M, K, Ei, Pi, Ti, Gi, Mi, Ki;",
         "for example: 128974848, 129e6, 129M, 123Mi",
     ])
 
-    def __init__(self, cpu_millis=None, memory=None):
-        if cpu_millis is not None:
-            self._validate_cpu(cpu_millis)
+    def __init__(self, cpu=None, memory=None):
+        if cpu is not None:
+            self._validate_cpu(cpu)
```
convoliution, line 55: @conradoverta Do you think I should warn the user if they're passing in an unreasonably high number for `cpu` (e.g. `1500`, if they've started using this when it was `cpu_millis`).

I could _also_ check for `cpu_millis` as a keyword argument.

conradoverta: We haven't exposed this yet
I generally prefer leaving validation to the backend as what is a "big number" here will change

conradoverta (approved):

PR #1404, opened 2020-08-26
make artifactstore config parameter backward compatible

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-6202

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/App.java
+++ b/backend/src/main/java/ai/verta/modeldb/App.java
@@ -189,7 +189,11 @@ public class App implements ApplicationContextAware {
 
   @Bean
   public S3Service getS3Service() throws ModelDBException {
-    return new S3Service(System.getProperty(ModelDBConstants.CLOUD_BUCKET_NAME));
+    String bucketName = System.getProperty(ModelDBConstants.CLOUD_BUCKET_NAME);
+    if (bucketName != null && !bucketName.isEmpty()) {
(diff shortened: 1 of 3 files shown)


backend/src/main/java/ai/verta/modeldb/App.java
```diff
@@ -189,7 +189,11 @@ public ServletWebServerFactory servletContainer(final GracefulShutdown gracefulS
 
   @Bean
   public S3Service getS3Service() throws ModelDBException {
-    return new S3Service(System.getProperty(ModelDBConstants.CLOUD_BUCKET_NAME));
+    String bucketName = System.getProperty(ModelDBConstants.CLOUD_BUCKET_NAME);
+    if (bucketName != null && !bucketName.isEmpty()) {
+      return new S3Service(System.getProperty(ModelDBConstants.CLOUD_BUCKET_NAME));
+    }
+    return null;
```
anandjakhaniya, line 196: In the NFS artifact spring boot server, s3 service bean is not required so it should be set it null.

anandjakhaniya: this is not part of backward compatibility. it just fixes the NFS server startup fix.

ravishetye (approved):
let us update the sample values in config and CONFGI.md and be clear of preference (which value will  be read in case it appears at the old as well as new location)

ravishetye (approved):

PR #1405, opened 2020-08-26
Restore support for version in dataset version

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-6256

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
@@ -189,6 +189,21 @@ public class CommitDAORdbImpl implements CommitDAO {
             Status.Code.INVALID_ARGUMENT);
       }
 
+      CommitPaginationDTO commitPaginationDTO =
+          findCommits(
+              session,
+              FindRepositoriesBlobs.newBuilder()
+                  .setPageNumber(1)
+                  .setPageLimit(1)
+                  .addRepoIds(repositoryEntity.getId())
+                  .build(),
+              authService.getCurrentLoginUserInfo(),
+              false,
+              false,
+              true,
+              null,
+              false);
+
       CommitEntity commitEntity =
           saveCommitEntity(session, commit, rootSha, datasetVersion.getOwner(), repositoryEntity);
       blobDAO.setBlobsAttributes(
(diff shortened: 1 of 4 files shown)


backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
```diff
@@ -211,6 +226,37 @@ public CommitDAORdbImpl(AuthService authService, RoleService roleService) {
               .setIdType(IDTypeEnum.IDType.VERSIONING_REPO_COMMIT_BLOB)
               .build(),
           datasetVersion.getTagsList());
+
+      long version = datasetVersion.getVersion();
```
ravishetye, line 230: if the user sets a version we should use it.

backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
```diff
@@ -211,6 +226,37 @@ public CommitDAORdbImpl(AuthService authService, RoleService roleService) {
...
+                    .setIdType(IDTypeEnum.IDType.VERSIONING_REPO_COMMIT_BLOB)
+                    .setStringId(parentCompositeId)
+                    .build(),
+                ModelDBConstants.VERSION);
+        if (parentVersion != null && !parentVersion.isEmpty()) {
+          version = Long.parseLong(parentVersion) + 1L;
+        }
+      }
+      if (version == 0) {
+        version = version + 1;
```
ravishetye: so version = 1 ??? 

we should only look at the version of parent commit or even fetch parent commit if there is no version set explicitly

```
if the request has a version save it , else increment the version of the parent commit and store it
```

ravishetye (changes requested):

backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
anandjakhaniya: Change it according to comment 
```
if the request has a version save it, else increment the version of the parent commit and store it
```

backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
ravishetye: I can not set a version from the request, was that flow verified.

anandjakhaniya: fixed

PR #1406, opened 2020-08-26
Fix .html => .rst broken links

elgalu:


convoliution:
Hey @elgalu, thanks for the interest and the PR!

These links are actually ".html" by intention, so that they work properly on our [generated documentation](https://docs.verta.ai/en/master/faqs.html).
Although changing them to ".rst" fixes the links on GitHub, they would break the generated HTML pages.

It looks like it's a quirk of Sphinx, and I haven't been able to find something simple that works for both the source and the generated docs 🤔

elgalu:
Aaaa, hehe

PR #1410, opened 2020-08-28
Fix findDatasetsByDatasetEndTimeTest test case

anandjakhaniya:
While migration from dataset to datasetInt, we do not support the find endpoints with TimeUpdated and mark test cases as ignore, but then after we are supporting it and now removed Ignore.

### DIFF
--- a/backend/src/test/java/ai/verta/modeldb/FindDatasetEntitiesTest.java
+++ b/backend/src/test/java/ai/verta/modeldb/FindDatasetEntitiesTest.java
@@ -656,14 +656,10 @@ public class FindDatasetEntitiesTest {
 
   /** Find dataset with value of endTime */
   @Test
-  @Ignore
-  public void findDatasetsByDatasetEndTimeTest() {
-    LOGGER.info("FindDatasets By Dataset EndTime test start................................");
+  public void findDatasetsByDatasetTimeUpdatedTest() {
+    LOGGER.info("FindDatasets By Dataset TimeUpdated test start................................");
 
-    DatasetTest datasetTest = new DatasetTest();
-
-    Value stringValue =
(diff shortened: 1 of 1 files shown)


backend/src/test/java/ai/verta/modeldb/FindDatasetEntitiesTest.java
```diff
@@ -656,14 +656,10 @@ public void findDatasetsByMetricsAndTagsTest() {
...
-  public void findDatasetsByDatasetEndTimeTest() {
-    LOGGER.info("FindDatasets By Dataset EndTime test start................................");
-
-    DatasetTest datasetTest = new DatasetTest();
+  public void findDatasetsByDatasetTimeUpdatedTest() {
+    LOGGER.info("FindDatasets By Dataset TimeUpdated test start................................");
 
-    Value stringValue =
-        Value.newBuilder().setStringValue(String.valueOf(dataset4.getTimeUpdated())).build();
+    Value stringValue = Value.newBuilder().setNumberValue(dataset4.getTimeUpdated()).build();
```
ravishetye: please add PR descriptions.
 probably change the var name to numberValue ?

in the PR description please mention when did we start breaking the test , and why do we end up updating the test instead of making code change.

anandjakhaniya: TimeUpdated is not a string but backend supports the string as a value and finds endpoints work fine. but I have just updated it correctly.

ravishetye (approved):
can merge after variable name change

backend/src/test/java/ai/verta/modeldb/FindDatasetEntitiesTest.java
anandjakhaniya: Fix variable name

PR #1412, opened 2020-08-28
Fix updateDatasetVersionDescriptionNegativeTest test case

anandjakhaniya:


backend/src/test/java/ai/verta/modeldb/DatasetVersionTest.java
```diff
@@ -668,7 +668,6 @@ public void updateDatasetVersionDescription() {
   }
 
   @Test
-  @Ignore
   public void updateDatasetVersionDescriptionNegativeTest() {
```
ravishetye, line 671: PR description on why the ignore was added and why it is removed

ravishetye (approved):

backend/src/test/java/ai/verta/modeldb/DatasetVersionTest.java
anandjakhaniya: While migration from dataset to datasetInt we have removed support of description from DatasetVersion but then after we have again added support for description that's why this is again useful and removed @Ignore.

PR #1422, opened 2020-09-01
VR-6386 Enable -h as shortcut for --help in CLI

nhatsmrt:


convoliution (approved):
Thanks! Simple enough to revert in the future if we need to, too

PR #1423, opened 2020-09-01
VR-6385 fix/add help docstrings for CLI commands

nhatsmrt:


client/verta/verta/_cli/deployment/download.py
```diff
@@ -8,11 +8,11 @@
 
 @deployment.group(name="download")
 def download():
-    """Create a new entry.
+    """Download resources relating to deployment.
 
     For example, to download the Docker context for a model version, run
 
-    `verta registry download dockercontext BERT latest`
+    `verta registry download dockercontext -m <model version id>`
```
convoliution: Examples with angle brackets should be wrapped in quotes, otherwise a copy&paste will result in shell redirection issues.

Ditto for the other examples, please!
```suggestion
    `verta registry download dockercontext -m "<model version id>"`
```

convoliution (changes requested):

client/verta/verta/_cli/deployment/download.py
nhatsmrt: Fixed.

convoliution (approved):

PR #1430, opened 2020-09-02
VR-6085 for CLI registry attributes, raise more informative error on JSON parse failure

nhatsmrt:


client/verta/verta/_cli/registry/update.py
```diff
@@ -106,6 +106,14 @@ def update_model_version(model_name, version_name, label, model, custom_module,
 
 
 def add_attributes(model_version, attribute, overwrite):
-    _multiple_arguments_for_each(attribute, "attribute",
-                                 lambda key, value: model_version.add_attribute(key, json.loads(value), overwrite=overwrite),
+    def action(key, value):
```
convoliution: We should definitely have a more descriptive name for this than "action"!

nhatsmrt: "action" is the name of the parameter in the method =) Maybe call it "logger"/"log" or sth?

client/verta/verta/_cli/registry/update.py
```diff
@@ -106,6 +106,14 @@ def update_model_version(model_name, version_name, label, model, custom_module,
...
+        try:
+            model_version.add_attribute(key, json.loads(value), overwrite=overwrite)
+        except ValueError as e:
+            if not (value.startswith('\"')  and value.endswith('\"')):
+                raise click.BadParameter("if the attribute's value is a string, consider wrapping it in quotes.")
+            else:
+                raise e
```
convoliution: The code in a `try` block should be as minimal as possible (so that the `except` actually catches what we want it to catch). We're trying to catch an error loading JSON, so the `try` block should only consist of loading JSON.
```suggestion
        try:
            json_value = json.loads(value)
        except ValueError as e:
            if not (value.startswith('\"')  and value.endswith('\"')):
                raise click.BadParameter("if the attribute's value is a string, consider wrapping it in quotes.")
            else:
                raise e
        else:
            model_version.add_attribute(key, json_value, overwrite=overwrite)
```

convoliution (changes requested):
See comments

client/verta/verta/_cli/registry/update.py
convoliution: I'm aware that the parameter is named `action` because Evgenii wrote it to accept more general operations, but I don't think that means the function we define here also has to be named `action`, unless I'm missing some quirk in Python.

client/verta/verta/_cli/registry/update.py
nhatsmrt: Fixed.

client/verta/verta/_cli/registry/update.py
nhatsmrt: Fixed.

convoliution (approved):
Thanks a bunch!

PR #1433, opened 2020-09-03
VR-6030 Move log_model() to _DeployableEntity

nhatsmrt:
Apparently, ModelVersion has add_attribute, whereas ExperimentRun has log_attribute 🤦. I propose that we should fix this, but in a separate ticket/PR.

Since this is of lower priority we should take our time reviewing it carefully.


client/verta/verta/_registry/modelversion.py
```diff
@@ -722,6 +655,13 @@ def download_docker_context(self, download_to_path, self_contained=False):
 
     #     self._update(self.ModelVersionMessage(archived=_CommonCommonService.TernaryEnum.TRUE))
 
+    def log_attribute(self, key, value, overwrite=False):
+        """
+        Alias for :meth:`~verta._registry.modelversion.RegisteredModelVersion.add_attribute`
+
+        """
+        self.add_attribute(key, value, overwrite)
+
```
convoliution: I had forgotten about this!

Instead of this, let's add `add_attribute()` as an alias to `ExperimentRun`.
tl;dr "add" makes more sense that "log" for attributes because it's not training metadata that's being logged, so we're preferring "add" going forward.

client/verta/verta/_registry/modelversion.py
```diff
@@ -188,61 +188,12 @@ def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=Non
...
-            model = open(model, 'rb')
-        serialized_model, method, model_type = _artifact_utils.serialize_model(model)
-
-        try:
-            extension = _artifact_utils.get_file_ext(serialized_model)
-        except (TypeError, ValueError):
-            extension = _artifact_utils.ext_from_method(method)
+        super(RegisteredModelVersion, self).log_model(model, custom_modules, model_api, artifacts, overwrite)
 
+    def _log_model_as_artifact(self, serialized_model, extension, method, overwrite):
```
convoliution: What would you think about just calling this `_log_model()`? 
`_log_model_as_artifact()` is a bit misleading, since there's nothing else you could log a model as, anyway.

client/verta/verta/_registry/modelversion.py
```diff
@@ -318,6 +250,9 @@ def log_artifact(self, key, artifact, overwrite=False, _extension=None):
             Whether to allow overwriting an existing artifact with key `key`.
 
         """
+        self._log_artifact(key, artifact, _CommonCommonService.ArtifactTypeEnum.BLOB, overwrite=overwrite)
+
+    def _log_artifact(self, key, artifact, artifact_type, _extension=None, method=None, overwrite=False):
```
convoliution: I know this'll make the diff messier, but let's remove that underscore since this entire method is now private.
```suggestion
    def _log_artifact(self, key, artifact, artifact_type, extension=None, method=None, overwrite=False):
```

convoliution (changes requested):
I might have to take an even closer look later

client/verta/verta/_registry/modelversion.py
nhatsmrt: Done.

client/verta/verta/_registry/modelversion.py

client/verta/verta/_registry/modelversion.py

nhatsmrt:
Ready for review.

convoliution:
Hey @nhatsmrt could you do me a huge, huge favor and convert the `nhatsmrt:log-model-deployable-entity` branch from your fork to a branch on this repo? I'd like to run our test pipeline on this branch before merging it, which requires it to be on VertaAI's repo 😬 

That might also require opening a new PR, in which case I'd request closing this one with a comment referring to the new one.

PR #1434, opened 2020-09-03
sort datasetversions on creation, support sort on getLatestDatasetVersion

ravishetye:
Dataset version default sorting on creation instead of updation elsewhere
fix getLatestDatasetVersion which returns the last created datasetversion
make findDatasetVersion work across workspaces

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
@@ -337,26 +337,47 @@ public class DatasetVersionServiceImpl extends DatasetVersionServiceImplBase {
             Any.pack(GetLatestDatasetVersionByDatasetId.Response.getDefaultInstance()));
       }
 
-      // TODO: Implement sort key in future
-      /*String sortKey =
-      request.getSortKey().isEmpty() ? ModelDBConstants.TIME_LOGGED : request.getSortKey();*/
+      UserInfo currentLoginUserInfo = authService.getCurrentLoginUserInfo();
 
-      int pageNumber = request.getAscending() ? 2 : 1;
-      DatasetVersionDTO datasetVersionDTO =
-          getDatasetVersionDTOByDatasetId(
-              request.getDatasetId(), pageNumber, 1, request.getAscending());
-
-      if (datasetVersionDTO.getDatasetVersions().size() != 1) {
+      FindRepositoriesBlobs.Builder findRepositoriesBlobs =
+          FindRepositoriesBlobs.newBuilder()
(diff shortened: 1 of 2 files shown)


backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
```diff
@@ -337,26 +337,47 @@ public void getLatestDatasetVersionByDatasetId(
...
+            repositoryDAO.getProtectedRepositoryById(repositoryIdentification, false);
+      }
+
+      DatasetVersion datasetVersion =
+          convertRepoDatasetVersions(repositoryEntity, commitPaginationDTO.getCommits()).get(0);
       /*Build response*/
       responseObserver.onNext(
           GetLatestDatasetVersionByDatasetId.Response.newBuilder()
-              .setDatasetVersion(datasetVersionDTO.getDatasetVersions().get(0))
+              .setDatasetVersion(datasetVersion)
```
ad-47, line 380: If `convertRepoDatasetVersions` return list is empty then `get(0)` is returns null and protobuf not accept null value so here it will throw an exception.

anandjakhaniya (approved):

backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
```diff
@@ -337,26 +337,47 @@ public void getLatestDatasetVersionByDatasetId(
...
+        repositoryEntity =
+            repositoryDAO.getProtectedRepositoryById(repositoryIdentification, false);
+      }
+
+      DatasetVersion datasetVersion =
+          convertRepoDatasetVersions(repositoryEntity, commitPaginationDTO.getCommits()).get(0);
       /*Build response*/
       responseObserver.onNext(
           GetLatestDatasetVersionByDatasetId.Response.newBuilder()
-              .setDatasetVersion(datasetVersionDTO.getDatasetVersions().get(0))
```
conradoverta, line 359: I saw a message "If convertRepoDatasetVersions return list is empty then get(0) is returns null and protobuf not accept null value so here it will throw an exception." from Anand here but GH is not showing it to me. I think this maps to the new line 376.

conradoverta (approved):
The two core changes LGTM. I'm not familiar with the "get(0)" issue that Anand is referring to, so not sure if that's relevant here or not.

backend/src/main/java/ai/verta/modeldb/datasetVersion/DatasetVersionServiceImpl.java
ravishetye: the size check was happening on commits , but i think it is good to check on versions as well is a good idea.

PR #1443, opened 2020-09-04
VR-6355 add overwrite param

evgeniilezhnin:


nhatsmrt:
If you want to be extra certain, you can add a small check for overwriting in the log_environment test.
Otherwise, LGTM.

nhatsmrt (approved):

evgeniilezhnin:
> If you want to be extra certain, you can add a small check for overwriting in the log_environment test.
> Otherwise, LGTM.

done. Also during adding a test case found out that for overwrite=True I should use PUT and also add some parameters to update_mask

PR #1446, opened 2020-09-04
Fix those test cases which not works due to isPopulateBasedOnPrivileges

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
```diff
@@ -1243,7 +1243,6 @@ public CommitPaginationDTO findCommits(
       } else {
         if (isDatasetVersion) {
           sortKey = ModelDBConstants.DATE_CREATED;
-          ascending = false;
```
ravishetye, line 1246: we need to ensure by default dataset version are sorted descending on date _created. is that taken care of else where?

backend/src/test/java/ai/verta/modeldb/ExperimentRunTest.java
```diff
@@ -10214,25 +10234,31 @@ public void findExperimentRunsCodeConfigWithRepository()
...
             ModelDBUtils.getLocationWithSlashOperator(test1Location.getLocationList());
-        assertTrue("Code blob should not empty", exprRun.containsCodeVersionFromBlob(locationKey));
-        assertTrue(
-            "Expected code config not found in map",
-            !exprRun
-                .getCodeVersionFromBlobOrThrow(locationKey)
-                .getGitSnapshot()
-                .getFilepathsList()
-                .isEmpty());
+        if (app.isPopulateConnectionsBasedOnPrivileges()) {
```
ravishetye: shoudl this if have an else?

backend/src/test/java/ai/verta/modeldb/ExperimentRunTest.java
```diff
@@ -10241,14 +10267,16 @@ public void findExperimentRunsCodeConfigWithRepository()
...
     String locationKey = ModelDBUtils.getLocationWithSlashOperator(test1Location.getLocationList());
-    assertTrue("Code blob should not empty", exprRun.containsCodeVersionFromBlob(locationKey));
-    assertFalse(
-        "Expected code config not found in map",
-        exprRun
-            .getCodeVersionFromBlobOrThrow(locationKey)
-            .getGitSnapshot()
-            .getFilepathsList()
-            .isEmpty());
+    if (app.isPopulateConnectionsBasedOnPrivileges()) {
```
ravishetye: should this if have an else?

ravishetye (commented):
we can discuss this offline if required.

backend/src/main/java/ai/verta/modeldb/versioning/CommitDAORdbImpl.java
anandjakhaniya: I have checked if someone is not set it to true then it will always have a false.

PR #1447, opened 2020-09-04
VR-5879 log_model with overwrite doesn't work with model artifacts

nhatsmrt:


client/verta/tests/test_deployment.py
```diff
@@ -207,6 +208,32 @@ def test_unlogged_keys_artifacts_error(self, experiment_run, model_for_deploymen
...
+        experiment_run.log_artifact(bad_key, bad_val)
+        experiment_run.log_model(ModelWithDependency, custom_modules=[], artifacts=[bad_key])
+
+        # log real artifact using `overwrite`
+        experiment_run.log_artifact(key, val)
+        experiment_run.log_model(ModelWithDependency, custom_modules=[], artifacts=[key], overwrite=True)
+        experiment_run.log_requirements(requirements=[])
+
+        endpoint.update(experiment_run, DirectUpdateStrategy(), wait=True)
+        assert val == endpoint.get_deployed_model().predict(val)
```
convoliution, line 236: Nice! Very clean and organized

client/verta/tests/test_entities.py
```diff
@@ -466,6 +466,19 @@ def test_clone_into_expt(self, client):
...
+        with pytest.raises(ValueError) as excinfo:
+            experiment_run.log_attribute("str-attr", "some-attr")
+
+        assert "already exists" in str(excinfo.value)
+
+        experiment_run.log_attribute("str-attr", "new-attr", True)
+        experiment_run.log_attributes({"int-attr": 5, "float-attr": 0.3, "bool-attr": False}, True)
+
+        assert experiment_run.get_attributes() == {"str-attr": "new-attr", "int-attr": 5, "float-attr": 0.3, "bool-attr": False}
+
```
convoliution: Highly encourage storing all these attributes as variables, so that the assert is easier to read and also to help mitigate typos.

convoliution (changes requested):
Minor-ish request for one of the tests

client/verta/tests/test_entities.py
nhatsmrt: Sth like this? (just pushed a commit).

convoliution: @nhatsmrt Ah, no, I means entire dictionaries, such as `{"str-attr": "attr", "int-attr": 4, "float-attr": 0.5}` should be a single variable.

Something like
```python
initial_attrs = {"str-attr": "attr", "int-attr": 4, "float-attr": 0.5}
new_attrs = {"str-attr": "new-attr", "int-attr": 5, "float-attr": 0.3, "bool-attr": False}
single_new_attr = new_attrs.popitem()

# ...

run.log_attribute(*single_new_attr, overwrite=True)  # or whatever works
run.log_attributes(new_attrs, overwrite=True)

expected_attrs = initial_attrs.copy()
expected_attrs.update([single_new_attr])  # or whatever works
expected_attrs.update(new_attrs)
assert run.get_attrs() == expected_attrs
```

Unless I'm missing a detail, this would reduce a lot of redundancy in this test and make it easier to read.

nhatsmrt: Fixed.

convoliution (approved):

PR #1448, opened 2020-09-04
VR-6412: Enable DatasetVersion list_components() and base_path

convoliution:
Adds
```python
dataset_version.list_components()
dataset_version.base_path
```

And restores
```python
dataset_version.dataset_version.path_dataset_version_info.base_path
```

### DIFF
--- a/client/verta/tests/test_datasets.py
+++ b/client/verta/tests/test_datasets.py
@@ -284,6 +284,7 @@ class TestClientDatasetVersionFunctions:
             assert retrieved.dataset_version_info is not None
             assert retrieved.dataset_version_info == version.dataset_version_info
 
+
     @pytest.mark.skip(reason="functionality removed")
     def test_reincarnation(self, client, created_datasets):
         """Consecutive identical versions are assigned the same ID."""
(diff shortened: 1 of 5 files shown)


client/verta/verta/dataset/_dataset.py
```diff
@@ -305,29 +307,29 @@ class Component(object):
     """
     def __init__(
             self,
-            path, size=None, last_modified=None,
+            path, base_path=None,
```
conradoverta: Note that you just broke api compatibility since the order of arguments is part of the API, right? I don't think you need to. Not a big deal in this case, but just pointing out to help you practice.

client/verta/verta/_dataset.py
```diff
@@ -619,13 +625,49 @@ def make_create_message(dataset_id, dataset_type,
...
         raise NotImplementedError('this function must be implemented by subclasses')
 
+    @property
+    def base_path(self):
+        components = self.list_components()
+        base_paths = set(component.base_path for component in components)
+
+        if len(base_paths) == 1:
+            return base_paths.pop()
+        else:  # shouldn't happen
```
conradoverta:
```suggestion
        else:  # the base path is not defined if they differ for the components
```
"Shouldn't happen" makes it seem like a bug, which it isn't.

client/verta/verta/dataset/_dataset.py
convoliution: I am aware! You're right, the change wasn't needed.
Just felt base_path made more sense near the beginning.
Also felt it was acceptable because these objects aren't meant to be initialized by users.

conradoverta: It's good to build the muscle memory. Otherwise it's more likely that you'll end up doing this by mistake when you didn't intend to.

client/verta/tests/test_datasets.py
```diff
@@ -301,6 +301,86 @@ def test_reincarnation(self, client, created_datasets):
         assert version.id == version1.id
 
 
+class TestBasePath:
```
conradoverta, line 304: Shouldn't we test for the exception path too?

client/verta/verta/_dataset.py
convoliution: It is a bug, though. As far as I know, there's no pathway through the Client to end up with different base paths for DatasetVersion components (unlike dataset blobs).

I can clarify this in the comment.

convoliution:
```suggestion
        else:  # shouldn't happen: DVs don't have an interface to have different base paths
```

conradoverta (approved):

PR #1463, opened 2020-09-09
VR-6550  Enable Dataset CRD for Scala Client

nhatsmrt:
Use workspace resolution from https://github.com/VertaAI/modeldb/pull/1481

PR #1475, opened 2020-09-10
VR-6551 Enable DatasetVersion CR for Scala Client

nhatsmrt:
Basically the same as #1468, but with the stuff from #1463 removed, so that this could be independently reviewed.
Test fails in (outdated) OSS setup, but passes in dev.


conradoverta:
Replaced with https://github.com/VertaAI/modeldb/pull/1489

PR #1477, opened 2020-09-10
VR-6553 Enable DatasetVersion attributes CRU for Scala Client

nhatsmrt:
Same as #1473, but without stuff from other PRs so can be reviewed independently.
Test fails due in (outdated) OSS setup, but passes in dev.


conradoverta:
Replaced with https://github.com/VertaAI/modeldb/pull/1492

PR #1504, opened 2020-09-15
Fix bug where cpu could not be None

convoliution:
...which is problematic because that's the default


conradoverta (commented):
I'm not sure why we need to force a value. It's an optional argument, right?

convoliution:
I don't consider this forcing a value, but let me make the code a bit easier to read anyway!

conradoverta (approved):
Better! I don't know what that method did, so it's better to be safe

PR #1519, opened 2020-09-17
VR-6654 Implement caching mechanism for entities in Scala client

nhatsmrt:
One small issue with current Dataset/DatasetVersion is that it does not have the caching mechanism of the Python client for the message. So if I call getDescription 100 times, the client will make 100 backend calls, even though there is no change.

Based on Python client's ModelDBEntity, I am working on a trait of entities whose message could be cached. Some goals:
- The internal cached message and cache updating mechanism should not be exposed from the trait.
- Boilerplate code for future implementation should be minimal. Ideally, user just need to call clearCache after updating, and getMessage before getting, and provide a method to "fetch" the latest version of the message.


jkwatson-verta:

PR #1522, opened 2020-09-18
VR-6547 client connects to atlas and constructs query dataset object

nhatsmrt:


liuverta:
There's been no mention of Atlas internally since 2020, so I assume this isn't needed anymore 😬

PR #1530, opened 2020-09-22
liquebase rollback

anandjakhaniya:


backend/config.default.yaml
```diff
@@ -100,4 +102,8 @@ cron_job:
     initial_delay: 300 #delay timestamp in second, default 300second = 5min
     frequency: 60 #frequency to delete entities in second, default 60
 
-populateConnectionsBasedOnPrivileges: false
\ No newline at end of file
+populateConnectionsBasedOnPrivileges: false
+
+#If liquibase have same changes then 'LIQUIBASE_MIGRATION' should have true otherwise default is false
+LIQUIBASE_MIGRATION: false
+RUN_LIQUIBASE_SEPARATE: false
```
ravishetye: these in config are confusing these are read from the environment.

PR #1531, opened 2020-09-22
Backend return 409 when logging DatasetVersion attribute with duplicate key

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-6762


backend/src/main/java/ai/verta/modeldb/versioning/blob/container/DatasetContainer.java
```diff
@@ -254,12 +256,31 @@ public void processAttribute(
...
+                Status status =
+                    Status.newBuilder()
+                        .setCode(com.google.rpc.Code.ALREADY_EXISTS_VALUE)
+                        .setMessage(
+                            "Attribute being logged already exists. existing attribute Key : "
+                                + newAttribute.getKey())
+                        .build();
+                throw StatusProto.toStatusRuntimeException(status);
+              } else {
+                newAttributeEntities.add(newAttribute);
```
ravishetye: this should be outside the inner for

ravishetye (changes requested):

backend/src/main/java/ai/verta/modeldb/versioning/blob/container/DatasetContainer.java
anandjakhaniya: Both ways will work. I am adding it to the new list when the key is not matched with the existing key. but your suggestion also works like: `newAttributeEntities.addAll(newOrUpdatedAttributeEntities);`

anandjakhaniya: Resolved with a better way

ravishetye (approved):

PR #1532, opened 2020-09-22
Clone ExperimentRun

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-6764

Proto PR: https://github.com/VertaAI/modeldb/pull/1538


backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -2597,4 +2598,25 @@ public void getExperimentRunsByDatasetVersionId(
           responseObserver, e, GetExperimentRunsByDatasetVersionId.Response.getDefaultInstance());
     }
   }
+
+  @Override
+  public void cloneExperimentRun(
```
ravishetye, line 2603: this needs protos changes right?

ravishetye (commented):
@convoliution :  are there any requirements more for clone? 
what happens to the timestamps? are they the current timestamp or arethey carried over from the src run?
would there be a need to clone runs across projects or experiments?

convoliution:
@ravishetye @ad-47 

>  are there any requirements more for clone?

I'm guessing this implementation copies over everything, in which case the only additional ask is that the protos have an optional field for the new run's name.

> what happens to the timestamps? are they the current timestamp or arethey carried over from the src run?

In my opinion, they should all be the current timestamp, not carried over.

> would there be a need to clone runs across projects or experiments?

Yes; the code appears to have this functionality already, but I think there might be cases that need to be covered:
- if both project ID and experiment ID are provided, backend should check that the experiment is inside the project, otherwise it's a bad request.
- if no project ID and yes experiment ID, backend should fill in project ID based on the experiment
- if yes project ID and no experiment ID, backend should (in my opinion) create a new experiment to contain the run

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
anandjakhaniya: Yes, it is there. https://github.com/VertaAI/modeldb/pull/1538

anandjakhaniya:
Updated Proto PR: https://github.com/VertaAI/modeldb/pull/1542

ravishetye (approved):

PR #1541, opened 2020-09-23
VR-6359 default strategy for download_endpoint_manifest and update

evgeniilezhnin:


client/verta/verta/client.py
```diff
@@ -1238,8 +1239,9 @@ def download_endpoint_manifest(
             Path of the endpoint.
         name : str
             Name of the endpoint.
-        strategy : :ref:`update strategy <update-stategies>`
+        strategy : :ref:`update strategy <update-stategies>`, optional
             Strategy (direct or canary) for updating the endpoint.
+            Default value: DirectUpdateStrategy()
```
nhatsmrt: I think the format for default value in docstring looks like this:
https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_registry/modelversion.py#L212

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -163,16 +163,18 @@ def _get_json_by_path(cls, conn, workspace, path):
...
         Updates the Endpoint with a model logged in an Experiment Run or a Model Version.
 
         Parameters
         ----------
         model_reference : :class:`~verta._tracking.experimentrun.ExperimentRun` or :class:`~verta._registry.modelversion.RegisteredModelVersion`
             An Experiment Run or a Model Version with a model logged.
-        strategy : :ref:`update strategy <update-stategies>`
+        strategy : :ref:`update strategy <update-stategies>`, optional
             Strategy (direct or canary) for updating the Endpoint.
+            Default value: DirectUpdateStrategy()
```
nhatsmrt: Same comment as above

nhatsmrt (approved):
Very minor suggestions.

Also it would be nice if we have a more direct test of the case of Endpoint.update without passing in any update strategy:

```
endpoint.update(model_version)
```

I don't think we need to add another test for this; maybe modify one of the existing test that pass DirectUpdateStrategy would suffice. Other than these suggestions, LGTM.

client/verta/verta/client.py
evgeniilezhnin: Thank you. Changed both

evgeniilezhnin:
> Very minor suggestions.
> 
> Also it would be nice if we have a more direct test of the case of Endpoint.update without passing in any update strategy:
> 
> ```
> endpoint.update(model_version)
> ```
> 
> I don't think we need to add another test for this; maybe modify one of the existing test that pass DirectUpdateStrategy would suffice. Other than these suggestions, LGTM.

modified

client/verta/tests/test_endpoint/test_resources.py
```diff
@@ -4,14 +4,15 @@
 from verta.endpoint.update import DirectUpdateStrategy
 
 
-@pytest.mark.parametrize("data", [3, 64, 0.25])
+@pytest.mark.parametrize("data", [(3, DirectUpdateStrategy()), (64, None),
+                                  (0.25, DirectUpdateStrategy())])
 def test_cpu_milli(client, data, in_tempdir):
```
convoliution: This should work, and be clearer to read.
```suggestion
@pytest.mark.parametrize("data,strategy", [(3, DirectUpdateStrategy()), (64, None),
                                  (0.25, DirectUpdateStrategy())])
def test_cpu_milli(client, data, strategy, in_tempdir):
```

convoliution (approved):
One request for the test

client/verta/tests/test_endpoint/test_resources.py
evgeniilezhnin: thank you, yes. I've updated it

PR #1546, opened 2020-09-24
VR-6825 Perform parallel tests to validate concurrency

nhatsmrt:


client/verta/tests/test_concurrency/test_concurrency.py
```diff
@@ -0,0 +1,33 @@
...
+import multiprocessing
+import pytest
+from .run_functions import log_observation, upload_artifact
+
+from operator import itemgetter
+from functools import partial
+
+
+class TestConcurrency:
+    @pytest.mark.skipif(six.PY2, reason="multiprocessing.Pool has issues in Python 2")
```
convoliution: @nhatsmrt Did you happen to check if these actually fail in Python 2? I am hoping that putting the parallel functions in a separate module would resolve that problem, but not sure

client/verta/tests/test_concurrency/test_concurrency.py
```diff
@@ -0,0 +1,33 @@
...
+        client.set_project()
+        client.set_experiment()
+
+        with multiprocessing.Pool(36) as pool:
+            result = pool.map(partial(log_observation, client), range(10))
+
+            def extract_obs_value(obs):
+                return list(map(itemgetter(0), obs))
+
+            assert all(map(lambda res: extract_obs_value((res["run"].get_observation("obs"))) == res["obs"], result))
```
convoliution: Could you do me a favor and let me know what `pytest`'s stack trace looks like if this fails (maybe hard-code a bad value returned by the parallel function)? My suspicion is that the error message would be absolutely massive because pytest afaik tries to print the value of every single variable in the `assert` statement.

If it _is_ excessively verbose, I'd recommend first assigning the `list(map(...))` to a separate variable,
then `assert all(var)`

client/verta/tests/test_concurrency/run_functions.py
```diff
@@ -0,0 +1,30 @@
...
+        run.log_observation("obs", observations[-1])
+
+    return {"run": run, "obs": observations}
+
+
+def upload_artifact(client, i):
+    run = client.set_experiment_run()
+
+    PART_SIZE = int(5.4 * (10 ** 6))  # 5.4 MB; S3 parts must be > 5 MB
+    os.environ['VERTA_ARTIFACT_PART_SIZE'] = str(PART_SIZE)
```
convoliution, line 21: I don't think it would, but this doesn't affect the environment variables in the parent process, right? Just to make sure it doesn't affect subsequent tests.

convoliution (changes requested):

client/verta/tests/test_concurrency/run_functions.py
nhatsmrt: Doesn't seem so. I tried printing out os.environ.get("VERTA_ARTIFACT_PART_SIZE"), and it returns None. This conforms to my understanding of processes in general (i.e not sharing variables between process).

client/verta/tests/test_concurrency/test_concurrency.py
nhatsmrt: Actually, I'm able to run the tests on Python 2 with small modifications (https://stackoverflow.com/questions/7447284/how-to-troubleshoot-an-attributeerror-exit-in-multiproccesing-in-python/35991842)

See the latest commit, but basically the problem I was having is that Pool doesn't play well with "with" statement.

client/verta/tests/test_concurrency/test_concurrency.py
nhatsmrt: You're right. Fixed.

conradoverta (approved):
LGTM

PR #1550, opened 2020-09-25
GetArtifact Response headers

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Controller.java
```diff
@@ -57,45 +62,36 @@
...
-    } catch (Exception ex) {
-      LOGGER.info("Could not determine file type.");
-      return null;
+      Status status =
+          Status.newBuilder().setCode(e.getCode().value()).setMessage(e.getMessage()).build();
+      throw StatusProto.toStatusRuntimeException(status);
+    } catch (Exception e) {
+      LOGGER.warn(e.getMessage(), e);
+      ErrorCountResource.inc(e);
+      throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, e.getMessage());
```
ravishetye: I do not understand the motivation of this change

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
```diff
@@ -383,23 +382,25 @@ public String uploadFile(
...
-  public Resource loadFileAsResource(String artifactPath) throws ModelDBException {
+  public S3Object loadFileAsResource(String artifactPath) throws ModelDBException {
     LOGGER.trace("S3Service - loadFileAsResource called");
     try {
       if (doesObjectExist(bucketName, artifactPath)) {
         LOGGER.trace("S3Service - loadFileAsResource - resource exists");
         LOGGER.trace("S3Service - loadFileAsResource returned");
-        return new InputStreamResource(
-            s3Client.getObject(bucketName, artifactPath).getObjectContent());
+        return s3Client.getObject(bucketName, artifactPath);
```
ravishetye: this throws .SdkClientException as well , let us add a catch for it

ravishetye: i see a warning in the javadoc of this method 

```
Be extremely careful when using this method; the returned Amazon S3 object contains a direct stream of data from the HTTP connection. The underlying HTTP connection cannot be reused until the user finishes reading the data and closes the stream. Also note that if not all data is read from the stream then the SDK will abort the underlying connection, this may have a negative impact on performance. Therefore:

Use the data from the input stream in Amazon S3 object as soon as possible
Read all data from the stream (use GetObjectRequest.setRange(long, long) to request only the bytes you need)
Close the input stream in Amazon S3 object as soon as possible
If these rules are not followed, the client can run out of resources by allocating too many open, but unused, HTTP connections
```

can you ensure we are following the correct pattern? I do not see any close() being called here

ravishetye (changes requested):

ravishetye (approved):

PR #1551, opened 2020-09-25
VR-6903 Client tests need to skip registry tests and endpoint tests when running in OSS mode

nhatsmrt:
Skip entire module, following example in:
https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules
Model registry is not supported by OSS setup, so needs to be skipped too.

### DIFF
--- a/client/verta/tests/test_cli/test_endpoint.py
+++ b/client/verta/tests/test_cli/test_endpoint.py
@@ -14,6 +14,8 @@ from verta.endpoint.resources import Resources
 
 from ..utils import get_build_ids
 
+pytestmark = pytest.mark.not_oss  # skip if run in oss setup. Applied to entire module
+
 
 class TestList:
     def test_list_endpoint(self, created_endpoints):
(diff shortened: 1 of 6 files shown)


convoliution:
@nhatsmrt Were you able to actually verify this against the OSS version of ModelDB?
To expand on that: the whole suite should be run against the OSS build (`cd tests/ && pytest --oss`) and there shouldn't be any odd errors due to unsupported features.
Also, did you happen to verify with the backend folks that registry isn't in OSS (as opposed to being some odd bug somewhere)

nhatsmrt:
> @nhatsmrt Were you able to actually verify this against the OSS version of ModelDB?
> To expand on that: the whole suite should be run against the OSS build (`cd tests/ && pytest --oss`) and there shouldn't be any odd errors due to unsupported features.
> Also, did you happen to verify with the backend folks that registry isn't in OSS (as opposed to being some odd bug somewhere)

Yep I did run the tests on my personal OSS setup. Tests were skipped (as expected). Had to spin up the docker image just for that 😆 
On the second point, need to ask the backend team. Will come back to that.

nhatsmrt:
Verified.

convoliution (approved):

PR #1552, opened 2020-09-25
VR-6823 add features support for datasets in client

nhatsmrt:


liuverta:
It would seem this was demo-only functionality, and the ticket had been marked "Incomplete" already.

PR #1553, opened 2020-09-26
Add support for replacing dash and underscore in modules

conradoverta:
Some modules have their pip freeze and import names mismatching. More
specifically for us, tensorflow-hub and prometheus-client are examples
of common packages with this behavior.

This PR checks both with underscore and dash.

### DIFF
--- a/client/verta/verta/_internal_utils/_pip_requirements_utils.py
+++ b/client/verta/verta/_internal_utils/_pip_requirements_utils.py
@@ -236,7 +236,7 @@ def set_version_pins(requirements):
             except (ImportError, AttributeError):
                 # fall back to checking pip
                 req_with_dash = req.replace("_", "-")
-                req_with_underscore = req.replace("_", "-")
+                req_with_underscore = req.replace("-", "_")
                 if req_with_dash in pip_pkg_vers:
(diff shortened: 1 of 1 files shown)


client/verta/verta/_internal_utils/_pip_requirements_utils.py
```diff
@@ -237,9 +235,13 @@ def set_version_pins(requirements):
                 ver = mod.__version__
             except (ImportError, AttributeError):
                 # fall back to checking pip
-                try:
-                    ver = pip_pkg_vers[req]
-                except KeyError:
+                req_with_dash = req.replace("_", "-")
+                req_with_underscore = req.replace("_", "-")
```
nhatsmrt: What's the difference between req_with_dash and req_with_underscore 🤔 The right-hand side of these two statements look the same to me.

conradoverta: Bad copy and paste!

nhatsmrt: Oh I think I understand the logic now.

nhatsmrt (approved):
LGTM.

PR #1554, opened 2020-09-27
Allow custom modules to handle pip-installed custom modules

conradoverta:
Before this patch, we couldn't specify modules installed via pip but that are not available in pypi. The main fix for this is to ignore venv only if the specific desired module is not inside a venv. If it is, then we _must_ include that folder in the custom modules and force the search path to be able to find it.

To facilitate the definition of such custom modules, as they can be pretty deep and different based on how people's envs are setup, we allow the user to pass the name of the module and we resolve to the right path.


nhatsmrt:
Just took a look. Everything seems solid to me, but is there a test case that we can run to verify the expected behavior?

conradoverta:
> Just took a look. Everything seems solid to me, but is there a test case that we can run to verify the expected behavior?

What I had in mind is to add a package that we pip-install, add a model that depends on it and then try to deploy. It might require more coordination than this repo allows. We have another repo with end to end tests for that purpose.

nhatsmrt (approved):

PR #1565, opened 2020-10-05
aws temporary credentials refresh logic

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-6999

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
+++ b/backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
@@ -64,6 +64,7 @@ public class S3Service implements ArtifactStoreService {
   private String bucketName;
   private Regions awsRegion;
   private static Credentials temporarySessionCredentials;
+  private static TimerTask task;
   private App app = App.getInstance();
 
   public S3Service(String cloudBucketName) throws ModelDBException {
@@ -124,7 +125,7 @@ public class S3Service implements ArtifactStoreService {
     try {
       RefreshCredentialsAndSchedule(clientRegion.toString());
 
-      initializeS3ClientWithTemporaryCredentials(clientRegion);
+      initializeS3ClientWithTemporaryCredentials(clientRegion, null);
 
     } catch (AmazonServiceException e) {
       // The call was transmitted successfully, but Amazon S3 couldn't process
@@ -137,7 +138,13 @@ public class S3Service implements ArtifactStoreService {
     }
   }
 
-  private void initializeS3ClientWithTemporaryCredentials(Regions clientRegion) {
+  private void initializeS3ClientWithTemporaryCredentials(
+      Regions clientRegion, AmazonServiceException ex) {
+
+    if (ex != null && ex.getErrorCode().equals(ModelDBConstants.EXPIRED_TOKEN)) {
+      RefreshCredentialsAndSchedule(clientRegion.getName());
+    }
+
     // Create a BasicSessionCredentials object that contains the credentials you just retrieved.
     BasicSessionCredentials awsCredentials =
         new BasicSessionCredentials(
@@ -163,8 +170,18 @@ public class S3Service implements ArtifactStoreService {
         "credentials before refresh {}",
         temporarySessionCredentials != null ? temporarySessionCredentials.hashCode() : null);
 
-    TimerTask task = new FetchTemporaryS3Token(region);
-    task.run();
+    scheduleTimer(region);
+  }
+
+  public static void scheduleTimer(String region) {
+    if (task == null) {
+      task = new FetchTemporaryS3Token(region);
+      task.run();
+      return;
+    }
+
+    task.cancel();
+    task = new FetchTemporaryS3Token(region);
 
     LOGGER.debug(
         "credentials after refresh {}",
@@ -175,7 +192,7 @@ public class S3Service implements ArtifactStoreService {
     Date now = new Date();
     long diffInSec = Math.abs(expiration.getTime() - now.getTime()) / 1000;
     long delay = diffInSec / 2;
-    LOGGER.info("sceduled cron for credentail refresh in {} seconds", delay);
+    LOGGER.info("scheduled cron for credential refresh in {} seconds", delay);
     ModelDBUtils.scheduleTask(task, delay, delay, TimeUnit.SECONDS);
     LOGGER.debug("scheduled periodic task to fetch token for s3 access");
   }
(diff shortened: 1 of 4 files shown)


backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
```diff
@@ -137,7 +138,25 @@ private void fetchCredentialsAndInitializeS3ClientWithTemporaryCredentials(Regio
     }
   }
 
-  private void initializeS3ClientWithTemporaryCredentials(Regions clientRegion) {
+  private synchronized void initializeS3ClientWithTemporaryCredentials(
+      Regions clientRegion, AmazonServiceException ex) {
+
+    if (ex == null && s3Client != null) {
```
ravishetye, line 152: comment for conrado's review: before we initializethe s3 client we check if we already have a valid s3 client

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
```diff
@@ -137,7 +138,25 @@ private void fetchCredentialsAndInitializeS3ClientWithTemporaryCredentials(Regio
...
+        return;
+      } catch (AmazonServiceException e) {
+        logAmazonServiceExceptionErrorCodes(e);
+        ex = e;
+      } catch (SdkClientException e) {
+        LOGGER.warn(e.getMessage());
+      }
+    }
+
+    if (ex != null && ex.getErrorCode().equals(ModelDBConstants.EXPIRED_TOKEN)) {
```
ravishetye, line 164: comment for conrado's review: if we get an exception and it indicates an expired token , then we fetch the new credentials. [we override the scheduled job to fetch the token because things have gone wrong some where]

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
```diff
@@ -158,13 +177,27 @@ private void initializeS3ClientWithTemporaryCredentials(Regions clientRegion) {
 
   private void RefreshCredentialsAndSchedule(String region) {
     LOGGER.debug("fetching token for s3 access");
+    if (task != null) {
```
ravishetye, line 198: comment for conrado's review: if there is a scheduled task cancel it since this refresh is off-cycle and will start a refresh cycle of its own

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
```diff
@@ -137,7 +138,25 @@ private void fetchCredentialsAndInitializeS3ClientWithTemporaryCredentials(Regio
     }
   }
 
-  private void initializeS3ClientWithTemporaryCredentials(Regions clientRegion) {
+  private synchronized void initializeS3ClientWithTemporaryCredentials(
```
ravishetye: comment for conrado's review: synchronized to avoid multiple thread doing the operations in parallel

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
```diff
@@ -158,13 +177,27 @@ private void initializeS3ClientWithTemporaryCredentials(Regions clientRegion) {
...
     LOGGER.debug(
         "credentials before refresh {}",
         temporarySessionCredentials != null ? temporarySessionCredentials.hashCode() : null);
 
-    TimerTask task = new FetchTemporaryS3Token(region);
-    task.run();
+    scheduleTimer(region);
+  }
+
+  public static synchronized void scheduleTimer(String region) {
```
ravishetye: comment for conrado's review: synchronized to avoid multiple thread doing the operations in parallel

ravishetye (commented):
the negative case of bad configuration was still leading infinite loop and high CPU and leading to a slow recovery of the system. I have requested @conradoverta for a once over hence the comments.

Follow up PR on top of this branch at https://github.com/VertaAI/modeldb/pull/1570

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
conradoverta: At the start, `ex` is null, right? So how do we know `temporarySessionCredentials` has valid credentials?

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
```diff
@@ -158,13 +177,27 @@ private void initializeS3ClientWithTemporaryCredentials(Regions clientRegion) {
...
+    if (task == null) {
+      task = new FetchTemporaryS3Token(region);
+      task.run();
+      return;
```
conradoverta: Does `.run()` perform immediately? If not, this will return without having finished running, right?

ravishetye: yes it calls https://github.com/VertaAI/modeldb/blob/master/backend/src/main/java/ai/verta/modeldb/cron_jobs/FetchTemporaryS3Token.java#L34 synchronously

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
conradoverta: Also, what happens if the exception it not null and not expired token?

conradoverta (commented):
Honestly, there is a lot of branching in this change, which makes it really hard for me to follow by myself all the possible scenarios it will encounter, which reduces significantly my ability to say if it looks good or not. I personally would recommend a more linear path because I find it easier to follow.

For example, instead of adding a lot of checks in the middle of requests to try to catch this scenario, could we just ensure that the job refreshing the token detects failures and tries again? If we place the responsibility of making sure we have a valid token on the refresh functionality instead of the request functionality, I think the logic might be simpler since we don't have the two of them crossing paths (e.g. we don't have to worry about concurrency or if the task already exists or what is the error on the basic operation).

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/s3/S3Service.java
ravishetye: on a closer look this does look weird , @ad-47 why are we not scheduling the next run here?

ravishetye (approved):

PR #1573, opened 2020-10-08
VR-6809 remove folders from s3 get dataset version part info

evgeniilezhnin:
Couldn't reproduce the part with " S3 sometimes returns folders from its ListObjects request, and sometimes it doesn’t." It always returns folders for me. This change is just for filtering files without folders for get dataset version


convoliution (approved):
Thanks, @evgeniilezhnin!

nhatsmrt (approved):

PR #1577, opened 2020-10-13
VR-6879 use not random file contents for test

evgeniilezhnin:
there is a possibility that during this test run a pickle serialized object with an unexpected type will be created and deserialized. The case is described here: https://vertaai.atlassian.net/browse/VR-6879

### DIFF
--- a/client/verta/tests/test_artifacts.py
+++ b/client/verta/tests/test_artifacts.py
@@ -163,7 +163,7 @@ class TestArtifacts:
     def test_clientside_storage(self, experiment_run, strs, in_tempdir):
         key = strs[0]
         filename = strs[1]
-        FILE_CONTENTS = os.urandom(2**16)
+        FILE_CONTENTS = bytes([3, 4, 5])
 
         # TODO: be able to use existing env var for debugging
         # NOTE: there is an assertion of `== 1` artifact that would need to be changed
--- a/client/verta/tests/test_artifacts.py
+++ b/client/verta/tests/test_artifacts.py
@@ -160,10 +160,20 @@ class TestArtifacts:
             with pytest.raises(ValueError):
                 experiment_run.log_artifact_path(key, artifact)
 
+    @staticmethod
+    def generate_random_data():
+        while True:
+            data = os.urandom(2 ** 16)
+            bytestream = six.BytesIO(data)
+            try:
+                pickle.load(bytestream)
+            except:
+                return data
+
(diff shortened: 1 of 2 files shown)


nhatsmrt:
What a small file =) Looks good to me, but is the ticket specifically for this test only, or does it apply to all tests that use randomized artifacts? cc @convoliution.

evgeniilezhnin:
> What a small file =) Looks good to me, but is the ticket specifically for this test only, or does it apply to all tests that use randomized artifacts? cc @convoliution.

I see that get_artifact calls pickle.load and this get_artifact is called in this test with random data. May be it can be called somewhere else. It may be hard to find all cases there pickle.load is called with random data.

Another solution may be to replace os.urandom(2**16) with the code that checks if this data is ok for pickle.load and then regenerate.

evgeniilezhnin:
> os.urandom(2**16)

added it for test_artifacts.py

Not sure if I should do that for other files too

nhatsmrt:
https://github.com/VertaAI/modeldb/search?q=FILE_CONTENTS+%3D+os.urandom%282**16%29
A quick GitHub search shows that it is used in a few other places. See if it requires the same fix as here 🤔

evgeniilezhnin:
> https://github.com/VertaAI/modeldb/search?q=FILE_CONTENTS+%3D+os.urandom%282**16%29
> A quick GitHub search shows that it is used in a few other places. See if it requires the same fix as here

I've added it also to test_model_version.py because it have get_artifacts call too. For another tests with FILE_CONTENTS test_registry.py and test_dataset.py there seem to be no pickle.load calls

client/verta/tests/test_artifacts.py
```diff
@@ -160,10 +160,20 @@ def test_blacklisted_key_error(self, experiment_run, all_values):
...
 
+    @staticmethod
+    def generate_random_data():
+        while True:
+            data = os.urandom(2 ** 16)
+            bytestream = six.BytesIO(data)
+            try:
+                pickle.load(bytestream)
+            except:
+                return data
```
nhatsmrt, line 171: So the logic here is that if it happens to be a data type that pickle recognize (e.g float), then we don't want that )? Sounds/looks good to me.

evgeniilezhnin: yes

nhatsmrt (approved):
LGTM, assuming tests still pass as normal.

PR #1580, opened 2020-10-14
VR-7127 fix key error

evgeniilezhnin:


client/verta/verta/_dataset.py
```diff
@@ -855,10 +855,13 @@ def get_dataset_part_infos(self):
         return dataset_part_infos
 
     def _append_s3_object_info(self, dataset_part_infos, object_info, key=None):
-        if object_info['Key'].endswith('/'):  # folder, not object
+        if self._is_folder(key or object_info['Key']):  # folder, not object
             return
         dataset_part_infos.append(self.get_s3_object_info(object_info, key))
 
+    def _is_folder(self, key):
```
convoliution: Doesn't use `self`, so it can be a static method
```suggestion
    @staticmethod
    def _is_folder(key):
```

convoliution (approved):
Minor suggestion. Does that test pass with this fix?

evgeniilezhnin:
> Minor suggestion. Does that test pass with this fix?

yes

convoliution (approved):

PR #1582, opened 2020-10-16
unnecessary params during get_or_create warning

evgeniilezhnin:


client/verta/verta/client.py
```diff
@@ -627,7 +654,7 @@ def set_dataset(self, name=None, type="local",
 
         If an accessible Dataset with name `name` does not already exist, it will be created
         and initialized with specified metadata parameters. If such a Dataset does already exist,
-        it will be retrieved; specifying metadata parameters in this case will raise an exception.
+        it will be retrieved; specifying metadata parameters in this case will raise a warning.
```
nhatsmrt, line 657: Hmm where is this enforced? I can't seem to find the use of check_unnecessary_params_warning for set_dataset.

nhatsmrt (commented):
This might seem tedious, but can you add a small test to verify that the warnings is thrown correctly? Maybe write a test for something that can be created quickly (e.g Project), or tag this along in one of the existing test_create_*'s. pytest has this construct that should allow you to check if a warning is thrown:

https://docs.pytest.org/en/stable/warnings.html#warns

client/verta/verta/client.py
evgeniilezhnin: It was there before the current PR here: https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_dataset.py#L60

evgeniilezhnin:
> This might seem tedious, but can you add a small test to verify that the warnings is thrown correctly? Maybe write a test for something that can be created quickly (e.g Project), or tag this along in one of the existing test_create_*'s. pytest has this construct that should allow you to check if a warning is thrown:
> 
> https://docs.pytest.org/en/stable/warnings.html#warns

good idea, I've added test cases

client/verta/verta/client.py
nhatsmrt: Oh I missed that. Nice.

nhatsmrt (approved):
LGTM, assuming tests passed.

client/verta/tests/test_dataset_versioning/test_dataset.py
```diff
@@ -31,6 +31,8 @@ def test_create(self, client, created_datasets):
         excinfo_value = str(excinfo.value).strip()
         assert "409" in excinfo_value
         assert "already exists" in excinfo_value
+        with pytest.warns(UserWarning, match='.*already exists.*'):
```
convoliution, line 34: Oh wow, I didn't know there  is a `match` parameter. That's useful!

PR #1585, opened 2020-10-19
VR-4471 Enable continuing previously-interrupted multipart upload

evgeniilezhnin:
Tested locally and on dev.verta, made minor changes: fixes in requests code and logic of what should be done when response_msg.artifact_parts is empty


client/verta/verta/_tracking/experimentrun.py
```diff
@@ -226,31 +226,61 @@ def _log_artifact(self, key, artifact, artifact_type, extension=None, method=Non
...
         """
         Uploads `artifact_stream` to ModelDB artifact store.
 
         Parameters
         ----------
         key : str
         artifact_stream : file-like
-        part_size : int, default 64 MB
+        start_part_num : int, default 1
             If using multipart upload, number of bytes to upload per part.
```
convoliution:
```suggestion
            If using multipart upload, what part number to start uploading with.
```

client/verta/verta/_tracking/experimentrun.py
```diff
@@ -226,32 +226,62 @@ def _log_artifact(self, key, artifact, artifact_type, extension=None, method=Non
...
+                data = _utils.proto_to_json(msg)
+                response = _utils.make_request("GET", url, self._conn, params=data)
+                if response.status_code == 404:  # old backend without multipart support
+                    # TODO: check if this also happens when artifact was not multipart
+                    raise conflict_error
+                elif not response.ok:
+                    _utils.raise_for_http_error(response)
+
+                # get last part number
+                response_msg = _utils.json_to_proto(response.json(), _CommonService.GetCommittedArtifactParts.Response)
```
convoliution: Minor: I think we should use the new `self._conn.must_proto_response()` for this

client/verta/verta/_tracking/experimentrun.py
```diff
@@ -226,32 +226,62 @@ def _log_artifact(self, key, artifact, artifact_type, extension=None, method=Non
...
+                if not response_msg.artifact_parts:
+                    url_for_artifact = self._get_url_for_artifact(key, "PUT", part_num=1)
+
+                    if url_for_artifact.multipart_upload_ok:  # there should be parts when completed
```
convoliution: Minor: This comment makes more sense to me for understanding this. Feel free to adjust/ignore!
```suggestion
                    if url_for_artifact.multipart_upload_ok:  # multipart artifact logged to MDB, but no parts uploaded
```

client/verta/verta/_tracking/experimentrun.py
```diff
@@ -272,9 +302,16 @@ def _upload_artifact(self, key, artifact_stream, part_size=64*(10**6)):
         url_for_artifact = self._get_url_for_artifact(key, "PUT", part_num=1)
 
         if url_for_artifact.multipart_upload_ok:
+            file_parts = iter(
+                lambda: artifact_stream.read(_artifact_utils.MULTIPART_UPLOAD_PART_SIZE), b'')
```
convoliution: Very minor: End `)` on new line is preferred
```suggestion
            file_parts = iter(
                lambda: artifact_stream.read(_artifact_utils.MULTIPART_UPLOAD_PART_SIZE), b'',
            )
```

client/verta/verta/_tracking/experimentrun.py
```diff
@@ -226,32 +226,62 @@ def _log_artifact(self, key, artifact, artifact_type, extension=None, method=Non
...
         ----------
         key : str
         artifact_stream : file-like
-        part_size : int, default 64 MB
-            If using multipart upload, number of bytes to upload per part.
+        start_part_num : int, default 1
+            If using multipart upload, what part number to start uploading with.
 
         """
         # TODO: add to Client config
```
convoliution, line 279: 🚨 See the next seven lines: we actually change `part_size` through an environment variable to accommodate a test ([this one](https://github.com/VertaAI/modeldb/blob/8831bd9/client/verta/tests/test_artifacts.py#L106-L113), makes `part_size` smaller so the test runs faster)

Could you implement a way that checks both `_artifact_utils.MULTIPART_UPLOAD_PART_SIZE` and `os.environ.get('VERTA_ARTIFACT_PART_SIZE', "")`? 🙇 

This may be a bigger problem because part size can change, but this env var is undocumented so it shouldn't be a problem for now (please also add a comment about this, if it makes sense to you!).

convoliution (changes requested):
Some minor comments, one major comment

client/verta/verta/_tracking/experimentrun.py
evgeniilezhnin: Oh, I realized now that part_size never used in this code. I've fixed it now. What do you mean by checks both of them? You mean that VERTA_ARTIFACT_PART_SIZE env wasn't really used here or something else?

I added a comment that it is used in test

client/verta/verta/_tracking/experimentrun.py
evgeniilezhnin: the line became shorter

client/verta/verta/_tracking/experimentrun.py
evgeniilezhnin: ok, replaced

client/verta/verta/_tracking/experimentrun.py
convoliution: >What do you mean by checks both of them?

I mean: please implement something like
```python
if os.environ.get('VERTA_ARTIFACT_PART_SIZE', ""):
    part_size = 
else:
    part_size = _artifact_utils.MULTIPART_UPLOAD_PART_SIZE
```
but you've already done that! 👍

convoliution (approved):
Looks good to me, thanks @evgeniilezhnin!

PR #1586, opened 2020-10-19
name & description both are empty then MDB throws NullPointerException

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-7183


ravishetye (commented):
this is just a minor improvement replacing an internal error with a invalid argument error, I think we should remove support for the updateNameOrDescription and switch over to using updateName and updateDescription

ravishetye (approved):
lets get this in since it is atleast fixing the NPE

PR #1589, opened 2020-10-20
VR-5950 model from torch file

evgeniilezhnin:


nhatsmrt:
Looks pretty solid in general, but is there any test that we can run to verify this? There might be some existing tests using PyTorch (probably how this issue was discovered in the first place). My suggestion is the CLI one:

https://github.com/VertaAI/modeldb/blob/70f100430d070c29d96531a1ce3b7ea16f73e6bf/client/verta/tests/test_cli/test_registry.py#L207

The nature of the CLI requires the model to first be serialized. That was how we discovered our log_model didn't play well with torch.save's output file. Replacing the pickling step in this test with torch.save and see if the test still passed.

Also running several other tests, especially the ones using PyTorch, to make sure we didn't break anything in the process is also a good idea, e.g:
https://github.com/VertaAI/modeldb/blob/8831bd96a9ffd587d63c892b56384320060648e7/client/verta/tests/test_artifacts.py#L277
https://github.com/VertaAI/modeldb/blob/9e9a0589b2fc8a374119e7d10c97ba09bb70da28/client/verta/tests/test_endpoint/test_endpoint.py#L499

client/verta/verta/_internal_utils/_artifact_utils.py
```diff
@@ -330,6 +342,13 @@ def deserialize_model(bytestring):
     except:  # not a pickled object
         bytestream.seek(0)
 
+    torch = maybe_dependency("torch")
+    if torch is not None:
+        try:
+            return torch.load(bytestream)
+        except:  # not something torch can deserialize
+            bytestream.seek(0)
```
convoliution, line 332: Let's actually try `torch` _before_ `cloudpickle` (put this block before line 338)

client/verta/verta/_internal_utils/_artifact_utils.py
```diff
@@ -200,6 +202,16 @@ def ensure_bytestream(obj):
...
+        torch = maybe_dependency("torch")
+        if torch is not None:
+            try:
+                torch.save(obj, bytestream)
+            except:  # not something torch can serialize
+                pass
+            else:
+                bytestream.seek(0)
+                return bytestream, "torch"
+
```
nhatsmrt, line 214: Is this ever going to be reached? From a cursory looks, it seems like the cloudpickling will be attempted first, correct? Which should succeed for torch model (that's how our existing passed)? Could you verify that this will be called during log_model (e.g by adding a print statement here, then try to run some of the tests I mentioned above).

nhatsmrt: I don't quite understand the logic of this part... This should be for the case where user pass a torch model object (not deserialized file) to log_model, correct? I think that case is already covered in the cloudpickling step above though, so this might not be reached at all 🤔I think that's how our existing test passed? 

Anyway, I could be very wrong; you should double check by adding a print statement here, then passed a torch model to log_model and see if this code block is reached. Running one of the existing tests that I mentioned above should suffice 🤔

nhatsmrt: For your reference, the tests that I was talking about:
https://github.com/VertaAI/modeldb/blob/8831bd96a9ffd587d63c892b56384320060648e7/client/verta/tests/test_artifacts.py#L277
https://github.com/VertaAI/modeldb/blob/9e9a0589b2fc8a374119e7d10c97ba09bb70da28/client/verta/tests/test_endpoint/test_endpoint.py#L499

client/verta/verta/_internal_utils/_artifact_utils.py
```diff
@@ -108,6 +108,8 @@ def ext_from_method(method):
         return 'hdf5'
     elif method in ("joblib", "cloudpickle", "pickle"):
         return 'pkl'
+    elif method == "torch":
+        return "pt"
```
convoliution: Because the Client actually uploads using `cloudpickle` anyway (see thread in #client), let's remove this.
Let me know if it causes a problem.

client/verta/verta/_internal_utils/_artifact_utils.py
convoliution: @evgeniilezhnin 
Because the Client actually uploads using cloudpickle anyway (see thread in #client), let's remove this.
Let me know if it causes a problem.

convoliution (changes requested):
I mentioned in our Slack thread that because of how `_artifact_utils` was written, the Client can deserialize a `torch.save()`ed model, but then will re-serialize it with `cloudpickle.dump()` and end up uploading that instead. But this PR would at least enable logging `torch.save()`ed models, instead of throwing an error.

---
I also expect this to pass:
```python
run.log_model(model)
retrieved_model = run.get_model()
model.__dict__ == retrieved_model.__dict__
```
Let me know if it doesn't

client/verta/verta/_internal_utils/_artifact_utils.py
evgeniilezhnin: removed

client/verta/verta/_internal_utils/_artifact_utils.py
evgeniilezhnin: I've tried it, seems to work alright

client/verta/verta/_internal_utils/_artifact_utils.py
evgeniilezhnin: removed

evgeniilezhnin:
> Looks pretty solid in general, but is there any test that we can run to verify this? There might be some existing tests using PyTorch (probably how this issue was discovered in the first place). My suggestion is the CLI one:
> 
> https://github.com/VertaAI/modeldb/blob/70f100430d070c29d96531a1ce3b7ea16f73e6bf/client/verta/tests/test_cli/test_registry.py#L207
> 
> The nature of the CLI requires the model to first be serialized. That was how we discovered our log_model didn't play well with torch.save's output file. Replacing the pickling step in this test with torch.save and see if the test still passed.
> 
> Also running several other tests, especially the ones using PyTorch, to make sure we didn't break anything in the process is also a good idea, e.g:
> https://github.com/VertaAI/modeldb/blob/8831bd96a9ffd587d63c892b56384320060648e7/client/verta/tests/test_artifacts.py#L277
> 
> 
> https://github.com/VertaAI/modeldb/blob/9e9a0589b2fc8a374119e7d10c97ba09bb70da28/client/verta/tests/test_endpoint/test_endpoint.py#L499

I've added a test test_create_version_with_custom_modules_torch which is a copy of test_create_version_with_custom_modules, but with torch.save

evgeniilezhnin:
> I mentioned in our Slack thread that because of how `_artifact_utils` was written, the Client can deserialize a `torch.save()`ed model, but then will re-serialize it with `cloudpickle.dump()` and end up uploading that instead. But this PR would at least enable logging `torch.save()`ed models, instead of throwing an error.
> 
> I also expect this to pass:
> 
> ```python
> run.log_model(model)
> retrieved_model = run.get_model()
> model.__dict__ == retrieved_model.__dict__
> ```
> 
> Let me know if it doesn't

It doesn't pass, but they seem to be equal. I think that is probably because comparison doesn't work for some objects from these dicts. I saw that assert torch.allclose(classifier(train_data), retrieved_model(train_data)) is sometimes used in tests, so I used this one.

nhatsmrt:
The new test looks solid to me. I'm not sure if the old CLI test is still needed, though; maybe you can just replace the old test with this new test? Defer to @convoliution for this and further review.

convoliution (changes requested):
>I'm not sure if the old CLI test is still needed, though; maybe you can just replace the old test with this new test?

I agree. Copy & pasting the whole test and only changing two lines may be confusing in the future (hard to see/remember what the difference is)

@evgeniilezhnin Let's just use the old `test_create_version_with_custom_modules`, and swap in the `model_path = "classifier.pt"` and `torch.save(classifier, model_path)` lines.

evgeniilezhnin:
> swap in

done, I've removed the test copy and replaced lines for torch in the test

convoliution (approved):
Thanks!

PR #1591, opened 2020-10-21
VR-2232 use diskfiles instead of bytestreams (start)

evgeniilezhnin:


liuverta:
This effort has been marked "Incomplete"—we should revisit, but for now it's not a blocker

PR #1592, opened 2020-10-21
Fix multipart upload race condition

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-6878

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
@@ -2635,7 +2635,7 @@ public class ExperimentRunDAORdbImpl implements ExperimentRunDAO {
         } else if (e instanceof StatusRuntimeException) {
           throw (StatusRuntimeException) e;
         } else {
-          throw new ModelDBException(e.getMessage(), Code.INTERNAL);
+          throw new ModelDBException(e);
         }
       }
     }
@@ -2654,7 +2654,7 @@ public class ExperimentRunDAORdbImpl implements ExperimentRunDAO {
       } else if (e instanceof StatusRuntimeException) {
         throw (StatusRuntimeException) e;
       } else {
-        throw new ModelDBException(e.getMessage(), Code.INTERNAL);
+        throw new ModelDBException(e);
       }
     }
   }
@@ -2731,7 +2731,7 @@ public class ExperimentRunDAORdbImpl implements ExperimentRunDAO {
         } else if (e instanceof StatusRuntimeException) {
           throw (StatusRuntimeException) e;
         } else {
-          throw new ModelDBException(e.getMessage(), Code.INTERNAL);
+          throw new ModelDBException(e);
         }
       }
     }
@@ -2741,13 +2741,24 @@ public class ExperimentRunDAORdbImpl implements ExperimentRunDAO {
   public GetCommittedArtifactParts.Response getCommittedArtifactParts(
       GetCommittedArtifactParts request) throws ModelDBException {
     try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
-      Set<ArtifactPartEntity> artifactPartEntities =
-          getArtifactPartEntities(session, request.getId(), request.getKey());
-      GetCommittedArtifactParts.Response.Builder response =
-          GetCommittedArtifactParts.Response.newBuilder();
-      artifactPartEntities.forEach(
-          artifactPartEntity -> response.addArtifactParts(artifactPartEntity.toProto()));
-      return response.build();
+      try (AutoCloseable ignored =
+          acquireReadLock(buildArtifactLockKey(request.getId(), request.getKey()))) {
+        Set<ArtifactPartEntity> artifactPartEntities =
+            getArtifactPartEntities(session, request.getId(), request.getKey());
+        GetCommittedArtifactParts.Response.Builder response =
+            GetCommittedArtifactParts.Response.newBuilder();
+        artifactPartEntities.forEach(
+            artifactPartEntity -> response.addArtifactParts(artifactPartEntity.toProto()));
+        return response.build();
+      } catch (Exception e) {
+        if (e instanceof ModelDBException) {
+          throw (ModelDBException) e;
+        } else if (e instanceof StatusRuntimeException) {
+          throw (StatusRuntimeException) e;
+        } else {
+          throw new ModelDBException(e);
+        }
+      }
     }
   }
 
@@ -2787,9 +2798,6 @@ public class ExperimentRunDAORdbImpl implements ExperimentRunDAO {
         session.beginTransaction();
         artifactEntity.setUploadCompleted(true);
         artifactEntity.setUploadId(null);
-        artifactPartEntities.forEach(session::delete);
-        artifactPartEntities.clear();
-        session.saveOrUpdate(artifactEntity);
         session.getTransaction().commit();
       } catch (Exception e) {
         if (e instanceof ModelDBException) {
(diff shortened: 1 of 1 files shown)


backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2580,23 +2616,47 @@ public void deleteLogVersionedInputs(Session session, List<Long> repoIds) {
...
+        return getS3PathAndMultipartUploadId(
+            session, artifactEntity, partNumber != 0, initializeMultipart);
+      } catch (Exception e) {
+        if (e instanceof ModelDBException) {
+          throw (ModelDBException) e;
+        } else if (e instanceof StatusRuntimeException) {
+          throw (StatusRuntimeException) e;
+        } else {
+          throw new ModelDBException(e.getMessage(), Code.INTERNAL);
+        }
```
evgeniilezhnin: I don't think that these lines are necessary, you can just use "throws Exception" in this function.  When you do "throw new ModelDBException(e.getMessage(), Code.INTERNAL);" you will lose stacktrace information that could be useful for local debugging. Exceptions are handled the same way by the observeError function in the service code (look ai.verta.modeldb.experimentRun.ExperimentRunServiceImpl#getUrlForArtifact)

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2647,16 +2707,33 @@ public ArtifactEntity getArtifactEntity(Session session, String experimentRunId,
...
-      VersioningUtils.saveOrUpdateArtifactPartEntity(
-          request.getArtifactPart(),
-          session,
-          String.valueOf(artifactEntity.getId()),
-          ArtifactPartEntity.EXP_RUN_ARTIFACT);
-      return Response.newBuilder().build();
+      try (AutoCloseable ignored =
+          acquireWriteLock(
+              buildArtifactPartLockKey(
+                  artifactEntity.getId(), request.getArtifactPart().getPartNumber()))) {
```
evgeniilezhnin, line 2713: shouldn't read locks also be used in getCommittedArtifactParts?

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2687,24 +2764,42 @@ public Response commitArtifactPart(CommitArtifactPart request) throws ModelDBExc
...
+            artifactPartEntities.stream()
+                .map(ArtifactPartEntity::toPartETag)
+                .collect(Collectors.toList());
+        commitMultipartFunction.apply(
+            artifactEntity.getPath(), artifactEntity.getUploadId(), partETags);
+        session.beginTransaction();
+        artifactEntity.setUploadCompleted(true);
+        artifactEntity.setUploadId(null);
+        artifactPartEntities.forEach(session::delete);
+        artifactPartEntities.clear();
```
evgeniilezhnin: There was a change needed for python client to keep artfact parts after upload completion: https://vertaai.atlassian.net/browse/VR-4760 to remove these lines from here.

Are you sure that these lines are necessary? There is a feature on client to continue cancelled upload. I think these lines may break it.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -2687,24 +2764,42 @@ public Response commitArtifactPart(CommitArtifactPart request) throws ModelDBExc
...
+        session.saveOrUpdate(artifactEntity);
```
evgeniilezhnin: update for existing entity should be automatic without this call

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: Also could you check this for other similar cases please?

evgeniilezhnin (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: instead of `throws Exception` I am thinking to use `throw new ModelDBException(e);` it will have the stack trace information

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: I thought the reading operation should not an issue for race condition but yes I think it will lead the client to fetch the same part then it will lead to the problem so let me add it for the safe side.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: Not necessary, but I see stale data in the artifact part table so I thought to do a cleanup but if the client has a feature to continue canceled upload then I can revert this change.

zlv (approved):

evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
zlv: Ok, @ad-47 could you also extract this code block to some method?

evgeniilezhnin: Ok, @ad-47 could you also extract this code block to some method?

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -79,11 +79,15 @@
 import ai.verta.uac.Role;
 import ai.verta.uac.UserInfo;
 import com.amazonaws.services.s3.model.PartETag;
```
ravishetye, line 81: low priority : i see an unused import on line 68, I am not sure why the buid did not remove it.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -233,6 +244,31 @@
...
+          .expireAfterWrite(DURATION, TimeUnit.MINUTES)
+          .build(
+              new CacheLoader<String, ReadWriteLock>() {
+                public ReadWriteLock load(String lockKey) {
+                  return new ReentrantReadWriteLock() {};
+                }
+              });
+
+  protected AutoCloseable acquireReadLock(String lockKey) throws ExecutionException {
+    ReadWriteLock lock = locks.get(lockKey);
```
ravishetye, line 260: do you think a log statement here and the function below will help us in the future to debug a particular slow call which probably was waiting on a lock?

ravishetye (approved):
minor comment ,  but can consider this approved

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: Sure I can add it.

PR #1593, opened 2020-10-22
Improve Repository test class with self deletion logic for created entities

anandjakhaniya:


ravishetye (approved):
Extract code to main code in separate PR , and you can merge it as well

PR #1606, opened 2020-10-27
Improve CollaboratorTest class with self deletion logic for created entities

anandjakhaniya:


ravishetye (commented):
since we shipped the collaborator endpoints to uac , no one is invoking them through modeldb, should we remove the tests instead?

ravishetye (approved):

PR #1608, opened 2020-10-28
Add default scheme if not provided for VERTA_HOST.

nhatsmrt:


client/scala/src/main/scala/ai/verta/client/ClientConnection.scala
```diff
@@ -1,7 +1,16 @@
...
 case class ClientConnection(host: String, ignoreConnErr: Boolean = false, maxRetries: Int = 5, auth: ClientAuth = null)
 
 object ClientConnection {
-  def fromEnvironment(): ClientConnection = new ClientConnection(sys.env("VERTA_HOST"), auth = ClientAuth.fromEnvironment())
+  private val DefaultScheme = "https"
+
+  private def hasScheme(url: String) = (new URI(url)).getScheme() != null
+
+  private def urlWithScheme(url: String) =
+    if (hasScheme(url)) url else (new URI(DefaultScheme, url, null, null, null)).toString()
```
conradoverta: Could we just replace the scheme? This is making hte assumption that 3 parts of the URI are `null`, which seems like a stronger assumption than we have to make here.

conradoverta (approved):
Added one comment. LGTM otherwise.

client/scala/src/main/scala/ai/verta/client/ClientConnection.scala
nhatsmrt: I think the most general solution is to pass the URL to the URI constructor, then use URI's method to extract the various components. Something like this:
```scala
val uri = new URI(url)

(new URI(Option(uri.getScheme()).getOrElse(DefaultScheme), uri.getAuthority(), uri.getPath(), uri.getQuery(), uri.getFragment())
```
Or one of the other constructors. Not sure which one is the best 🤔 
Screen Shot 2020-10-27 at 9 53 35 PM

nhatsmrt: UPDATE: ignore latest commit. That doesn't work for localhost:3000.

nhatsmrt: Actual neither is my original attempt =) It probably picks up localhost as scheme =)

nhatsmrt:
OK third attempt: pretty much resort to hard-coding =)) This is too hacky for my taste.

client/scala/src/main/scala/ai/verta/client/ClientConnection.scala
conradoverta: Localhost as scheme is crazy!

conradoverta:
> OK third attempt: pretty much resort to hard-coding =)) This is too hacky for my taste.

I'm not sure we have a better option with the URI class behaving like it does.

conradoverta (approved):

PR #1614, opened 2020-10-29
Trial support

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-7368


backend/config.default.yaml
```diff
@@ -109,3 +109,10 @@ cron_job:
     frequency: 60 #frequency to delete entities in second, default 60
 
 populateConnectionsBasedOnPrivileges: false
+
+trial:
+  enable: true
+  restrictions:
+    max_artifact_size_MB: #Integer value
+    max_artifact_per_run: #Integer value
+    max_experiment_run_per_workspace: #Integer value
```
ravishetye: newline

ravishetye (changes requested):

ravishetye (commented):
add support for values of individual threshold to be missing, we can either set then to null or -1, and then skip the restriction if it is null or -1

PR #1618, opened 2020-10-30
VR-7510 Return the Failure from the uploading step itself for Scala Client.

nhatsmrt:
I wonder if this is more useful to users than a generic "Uploading components blah blah blah fails."

I will also make number of upload attempts configurable in https://github.com/VertaAI/modeldb/pull/1619.


conradoverta (approved):
Not sure why we need to split into 2 functions but sure.

PR #1621, opened 2020-11-02
VR-7508 Make sure the autoscaling quantities in Python client test is correct.

nhatsmrt:


conradoverta (commented):
Were this test passsing?

nhatsmrt:
> Were this test passsing?

These tests passed previously (since we had no guard against setting invalid autoscaling quantities, backend or client side). They have been failing since, because we no longer allowed these invalid values.

evgeniilezhnin (approved):

PR #1622, opened 2020-11-03
Fix test for same endpoint, different workspace

convoliution:
Deployment backend currently returns
```
requests.exceptions.HTTPError: 409 Client Error: "Path already in use by endpoint ######"
```
when creating endpoint w/ same name in different workspace.

I'm adjusting the test to check for that error, until the ticket for enabling creation is implemented.

### DIFF
--- a/client/verta/tests/test_organization.py
+++ b/client/verta/tests/test_organization.py
@@ -43,7 +43,8 @@ class TestOrganization:
         repository_name = _utils.generate_default_name()
         model_name = _utils.generate_default_name()
         version_name = _utils.generate_default_name()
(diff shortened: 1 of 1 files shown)


nhatsmrt:
Oh crap. I forgot about this test failing. I saw it yesterday when I was trying to test the log_model refactoring 🤦

nhatsmrt (approved):
LGTM. I assume there's a follow-up ticket?

convoliution:
>I assume there's a follow-up ticket?

Great point!
VR-6053 is the deployment backend ticket, and I've filed VR-7697 to adjust the Client test after that's done.

PR #1624, opened 2020-11-04
VR-7699 custom modules tests fail: captured fewer files than expected

nhatsmrt:
Basically, we change the logic of skipping venv files in
https://github.com/VertaAI/modeldb/pull/1554/
to accommodate the case of pip-installed custom module, but did not update the test to reflect the new logic 🤔 
Let me know what you think about the change. Tests do pass now.


convoliution (approved):
Makes sense to me!

PR #1625, opened 2020-11-04
VR-7701: Recommend user downgrade h5py on Keras save/load error

convoliution:
Due to https://github.com/h5py/h5py/issues/1732, `h5py==3.0.0` raises unhelpful string-handling errors Keras's save/load.
We catch the error, verify that `h5py==3.0.0` is installed, and recommend a resolution for the user.

### DIFF
--- a/client/verta/verta/_internal_utils/_artifact_utils.py
+++ b/client/verta/verta/_internal_utils/_artifact_utils.py
@@ -290,6 +290,9 @@ def serialize_model(model):
                 h5py = maybe_dependency("h5py")
                 if (str(e) == "a bytes-like object is required, not 'str'"
                         and h5py is not None and h5py.__version__ == "3.0.0"):
+                    # h5py v3.0.0 improperly checks if a `bytes` contains a `str`.
+                    # Encountering this generic error message here plus the fact
+                    # that h5py==3.0.0 suggests that this is the problem.
                     six.raise_from(KERAS_H5PY_ERROR, e)
                 else:
                     six.raise_from(e, None)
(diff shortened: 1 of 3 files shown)


client/verta/verta/_internal_utils/_artifact_utils.py
```diff
@@ -30,6 +30,12 @@
 }
 
 
+KERAS_H5PY_ERROR = RuntimeError(  # https://github.com/h5py/h5py/issues/1732
+    "Keras encountered an error saving/loading the model due to a bug in h5py v3.0.0;"
+    " consider downgrading with `pip install \"h5py!=3.0.0\"`"
```
conradoverta: This is not necessarily a downgrade. Are we *sure* that 3.0.1 or whatever comes after has a fix?

client/verta/verta/_internal_utils/_artifact_utils.py
```diff
@@ -275,10 +281,18 @@ def serialize_model(model):
...
-            else:
-                model.save(tempf.name)
+            try:
+                if get_tensorflow_major_version() == 2:  # save_format param may not exist in TF 1.X
+                    model.save(tempf.name, save_format='h5')  # TF 2.X uses SavedModel by default
+                else:
+                    model.save(tempf.name)
+            except TypeError as e:
+                h5py = maybe_dependency("h5py")
+                if (str(e) == "a bytes-like object is required, not 'str'"
```
conradoverta, line 291: Could you add a note to this string comparison? Seem pretty brittle and misterious.

client/verta/verta/_internal_utils/_artifact_utils.py
convoliution: >This is not necessarily a downgrade.

Good point!

>Are we sure that 3.0.1 or whatever comes after has a fix?

"Sure" insofar as the h5py team have it [tracked as a bug](https://github.com/h5py/h5py/issues/1732) and are [preparing a fix](https://github.com/h5py/h5py/pull/1734) for their coming release.

@conradoverta Would it make more sense to just suggest `pip install "h5py<3.0.0"`, and _maybe_ update the message later with `!=` if the bug is fixed?

conradoverta: Yep.

conradoverta (approved):

PR #1637, opened 2020-11-09
PESSIMISTIC_WRITE locking

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-7712

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
@@ -52,6 +52,8 @@ import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.hibernate.LockMode;
+import org.hibernate.LockOptions;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
 import org.hibernate.query.Query;
@@ -264,7 +266,8 @@ public class ExperimentDAORdbImpl implements ExperimentDAO {
   public Experiment updateExperimentName(String experimentId, String experimentName)
       throws InvalidProtocolBufferException {
     try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
-      ExperimentEntity experimentEntity = session.load(ExperimentEntity.class, experimentId);
+      ExperimentEntity experimentEntity =
+          session.load(ExperimentEntity.class, experimentId, LockMode.PESSIMISTIC_WRITE);
       experimentEntity.setName(experimentName);
       long currentTimestamp = Calendar.getInstance().getTimeInMillis();
       experimentEntity.setDate_updated(currentTimestamp);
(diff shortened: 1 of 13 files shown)


coryverta (approved):

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -899,8 +910,9 @@ public ExperimentRun deleteExperimentRunTags(
   public void logObservations(String experimentRunId, List<Observation> observations)
       throws InvalidProtocolBufferException {
     try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
+      Transaction transaction = session.beginTransaction();
```
coryverta, line 913: Does this Transaction automatically rollback on an exception?  I see the explicit commit, but no explicit rollback.

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
```diff
@@ -373,7 +377,8 @@ public ExperimentPaginationDTO getExperimentsInProject(
   public Experiment addExperimentTags(String experimentId, List<String> tagsList)
       throws InvalidProtocolBufferException {
     try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
-      ExperimentEntity experimentObj = session.load(ExperimentEntity.class, experimentId);
+      ExperimentEntity experimentObj =
+          session.load(ExperimentEntity.class, experimentId, LockMode.PESSIMISTIC_WRITE);
```
conradoverta, line 381: We should not lock on read only operations or that will make the system super slow. We shouldn't even have a transaction.

Is it possible to only have pessimistic locks to write operations? It's fine even if we have read-then-write, but it would be ideal not to have on read only db queries.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: when we are called `transaction.commit()` then it will reflect on DB and before that exception throw then no need to do a rollback. In our entire project, we have not used rollback anywhere.

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
anandjakhaniya: I have added Lock on only on writable entities' operation. here it fetches but after that, it will update on https://github.com/VertaAI/modeldb/pull/1637/files/0423bed3779f3b3edc3d36d1858bd4106eb18fae#diff-5874488cfbcfa089079b8b4bc42688c13066312280f478d66f37d58b80898cc7R402 that's why I have to take a lock on it.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
coryverta: 👍

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
conradoverta: sg

conradoverta:
Merge it away with Cory's approval.

coryverta (approved):

coryverta:
LGTM

PR #1644, opened 2020-11-11
Audit logs

anandjakhaniya:
- Added audit logs logic
- Added local audit logs for the project, experiment, experiment run


coryverta (approved):

conradoverta:
@coryverta did you actually review the 1.2k LOC?

@ad-47 this PR is massive. Can we break it apart on the different components? Also, it seems _very_ prone to a bad copy and paste that will be impossible to catch (human eyes are not good at that). Is there a better way for us to enforce correctness? Maybe some auxiliary code that is more specific for each kind of operation?

anandjakhaniya:
Code separated into many PRs and created base feature branch https://github.com/VertaAI/modeldb/pull/1645 for it.

PR #1645, opened 2020-11-12
Initialize AuditLogLocalEntity and liquibase support

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/audit_log/AuditLogLocalDAORdbImpl.java
```diff
@@ -0,0 +1,35 @@
...
+      if (ModelDBUtils.needToRetry(ex)) {
+        saveAuditLogs(auditLogEntities);
+      } else {
+        throw ex;
+      }
+    }
+  }
+
+  public void saveAuditLogs(Session session, List<AuditLogLocalEntity> auditLogEntities) {
+    auditLogEntities.forEach(session::save);
```
coryverta, line 33: Do we need to wrap this in a transaction as well to ensure the entire data set is committed as a unit of work?  The above code does this.  We should attempt to keep them the same unless there is a reason for them to differ.

anandjakhaniya: this is called from https://github.com/VertaAI/modeldb/pull/1645/files#diff-b2cc154256d0c1d31090d7d5c768bd613135f469dac3e4ae7d08b7f00220451bR20 so it is already wrapped on the transaction
```
      saveAuditLogs(session, auditLogEntities);
      transaction.commit();
      LOGGER.debug("Audit logged successfully");
```

coryverta: 👍

coryverta (approved):

PR #1650, opened 2020-11-12
VR-7747 change client to only show "message" field from http error response if it has one

nhatsmrt:
Error msg is now:
```python
429 Client Error: LIMIT_RUN_NUMBER: “Number of experiment runs exceeded”: Your trial account allows you to log upto 10 experiment runs. Try deleting prior experiment runs in order to proceed. for url: <url> at 2020-11-12 19:00:31.865000 UTC
```
(I removed the url for privacy reasons).
Basically re-introduce the logic changed in https://github.com/VertaAI/modeldb/pull/1083, with some slight modifications.

### DIFF
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -531,10 +531,14 @@ def raise_for_http_error(response):
         try:
             reason = body_to_json(response)
         except ValueError:
-            reason = response.text.strip()
+            reason = response.text.strip()  # response is not json
 
-        if isinstance(reason, dict) and 'message' in reason:
-            reason = reason['message']
+        if isinstance(reason, dict):
+            if 'message' in reason:
+                reason = reason['message']
+            else:
+                # fall back to entire text
+                reason = response.text.strip()
 
         reason = six.ensure_str(reason)
 
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -528,7 +528,12 @@ def raise_for_http_error(response):
         curr_time = timestamp_to_str(now(), utc=True)
         time_str = " at {} UTC".format(curr_time)
 
-        reason = six.ensure_str(response.text.strip())
+        try:
+            reason = body_to_json(response)['message']
+        except (ValueError,  # not JSON response
+                 KeyError):  # no 'message' from back end
+            reason = six.ensure_str(response.text.strip())
+
         if not reason:
             e.args = (e.args[0] + time_str,) + e.args[1:]  # attach time to error message
             six.raise_from(e, None)  # use default reason
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -529,10 +529,12 @@ def raise_for_http_error(response):
         time_str = " at {} UTC".format(curr_time)
 
         try:
-            reason = six.ensure_str(body_to_json(response)['message'])
+            reason = body_to_json(response)['message']
         except (ValueError,  # not JSON response
                  KeyError):  # no 'message' from back end
-            reason = six.ensure_str(response.text.strip())
+            reason = response.text.strip()
+
+        reason = six.ensure_str(reason)
 
         if not reason:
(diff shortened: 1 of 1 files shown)


nhatsmrt:
cc @conradoverta this is what you want right?

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -528,7 +528,12 @@ def raise_for_http_error(response):
         curr_time = timestamp_to_str(now(), utc=True)
         time_str = " at {} UTC".format(curr_time)
 
-        reason = six.ensure_str(response.text.strip())
+        try:
+            reason = six.ensure_str(body_to_json(response)['message'])
```
nhatsmrt: Need to call ensure str, otherwise would get `'ascii' codec can't encode character u'\u201c' in position 18: ordinal not in range(128)` error.

convoliution: Out of curiosity, did you test this in Python 2? (Could you try it out in both 2 and 3, if you can?)

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -528,7 +528,12 @@ def raise_for_http_error(response):
...
+        except (ValueError,  # not JSON response
+                 KeyError):  # no 'message' from back end
+            reason = six.ensure_str(response.text.strip())
```
convoliution: I think the call to `ensure_str()` should be pulled out of the `try-except`

1. both branches use it, so we might as well just call it once, after the fact
2. the `try` block is only testing for a JSON error and a key error—the call to `ensure_str()` is unrelated and so should be pulled out if possible

```suggestion
        try:
            reason = body_to_json(response)['message']
        except (ValueError,  # not JSON response
                 KeyError):  # no 'message' from back end
            reason = response.text.strip()
            
        reason = six.ensure_str(reason)
```

convoliution (changes requested):

client/verta/verta/_internal_utils/_utils.py
nhatsmrt: Fixed

client/verta/verta/_internal_utils/_utils.py
nhatsmrt: Just tried for both. LGTM.

convoliution (approved):
Looks good to me

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -528,7 +528,15 @@ def raise_for_http_error(response):
         curr_time = timestamp_to_str(now(), utc=True)
         time_str = " at {} UTC".format(curr_time)
 
-        reason = six.ensure_str(response.text.strip())
+        try:
+            reason = body_to_json(response)['message']
+        except (ValueError,  # not JSON response
+                TypeError,  # body_to_json return a single string
```
nhatsmrt: @convoliution apparently body_to_json could return a string 🤷Specifically, response.json() returns `LIMIT_ENDPOINT_NUMBER: currently at maximum number of endpoints` when we try to create a second endpoint 🤔

convoliution: Oh, that complicates things.
In that case, could you do me a favor and split this `try-except` into multiple checks?
It was already a shaky idea when I tried catching two error types simultaneously—trying to catch 3 is really prone to issues down the line

Something like (pseudocode)
```
try:
    reason = body_to_json()
except ValueError:
    raise

if reason is dict and 'message' in reason:
    reason['message']

# etc...

convoliution (changes requested):
See comment

client/verta/verta/_internal_utils/_utils.py
nhatsmrt: Fixed. See [latest](https://github.com/VertaAI/modeldb/pull/1650/commits/5ad1c0319dba02d7a492e8cfffe50e0babfbdf14) [commits](https://github.com/VertaAI/modeldb/pull/1650/commits/d597c23b607a9544317d402b0cff1338613b1001)

convoliution (approved):
Makes sense to me! Thanks ✨

nhatsmrt:
🚢

PR #1658, opened 2020-11-17
Random name generation endpoint on the metadata service

anandjakhaniya:


conradoverta (approved):

anandjakhaniya:
@conradoverta could you generate the auto-generated files and update the proxy for these changes here in the same PR then it would be helpful.

coryverta (approved):

PR #1660, opened 2020-11-17
Remove the logging of the YAML properties map

coryverta:
This map contains DB creds and other secrets, and should not be logged.


conradoverta (approved):

conradoverta (commented):
This seems pretty bad

PR #1666, opened 2020-11-18
VR-7716 unable to search runs by tags

nhatsmrt:
This is all it takes, apparently 🤷 Lmk if I missed sth.

### DIFF
--- a/client/verta/tests/test_entities.py
+++ b/client/verta/tests/test_entities.py
@@ -587,6 +587,24 @@ class TestExperimentRuns:
             assert row['metric.'+metric1] == run.get_metric(metric1)
             assert row['metric.'+metric2] == run.get_metric(metric2)
 
+    def test_find_by_tag(self, client):
+        client.set_project()
+        expt = client.set_experiment()
+        tag = "some-tag"
+        diff_tag = "diff-tag"
+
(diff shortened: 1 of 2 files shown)


convoliution (changes requested):
It's been a while so I don't remember, myself:
Looks like the syntax being used here is `tags ~= some-tag`; does `==` work?
Either way, let's also add this to the `Examples` section in the docstring

convoliution:
@conradoverta Should this also be enabled for Projects & Experiments (if it's not, already)?

nhatsmrt:
@convoliution `~=` is contains (according to LazyList's implementation), so it makes more sense for me, but I just double checked and `==` works too 🤷. Lmk which do you prefer.

nhatsmrt:
@convoliution Looks like searching by tags is already allowed in [projects](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_tracking/projects.py#L24) and [experiments](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_tracking/experiments.py#L24). I think we just missed this one key in runs' implementation :P

convoliution (approved):
>Looks like searching by tags is already allowed in projects and experiments. I think we just missed this one key in runs' implementation :P

😅 

>I just double checked and == works too 🤷. Lmk which do you prefer.

`~=` will do just fine

PR #1667, opened 2020-11-18
VR-7870 Support passing a dataframe in the client predict method

nhatsmrt:
I'm thinking of sth like this, in order to allow the user to define their model in terms of data frame (modulo some wrapper).
Need to actually verify that it works 😬 But let me know if sth like this would be problematic, design-wise 🤔


client/verta/verta/deployment/__init__.py
```diff
@@ -4,4 +4,5 @@
     DeployedModel,
     prediction_input_unpack,
     prediction_io_cleanup,
+    prediction_input_df
```
convoliution: Trailing commas preferred!
```suggestion
    prediction_input_df,
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -255,6 +255,14 @@ def predict(self, x, compress=False, max_retries=5, always_retry_404=True, alway
         _utils.raise_for_http_error(response)
 
 
+def prediction_input_df(func):
+    import pandas as pd
+
+    def prediction(self, records):
+        return func(self, pd.DataFrame.from_records(records))
```
convoliution: I _think_ the import would be better in the inner function. Not sure though

convoliution:
I think it would be important to maintain backwards compatibility (users who are passing a `DataFrame` to `predict()`, but are handling it as a plain array of values). @conradoverta What do you think?

conradoverta:
> I think it would be important to maintain backwards compatibility (users who are passing a `DataFrame` to `predict()`, but are handling it as a plain array of values). @conradoverta What do you think?

We need to ship this asap. I'm ok breaking compatibility on this and adding a note.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -784,7 +784,8 @@ def to_builtin(obj):
     if obj_class == "Series":
         return obj.values.tolist()
     if obj_class == "DataFrame":
-        return obj.values.tolist()
+        return list(map(to_builtin, obj.to_dict("records")))
```
conradoverta: This seems different from what we discussed, no?

conradoverta (commented):
Make sure that we provide the conversion back too.

client/verta/verta/_internal_utils/_utils.py
nhatsmrt: So what we want is `[{"feature-1": 1, "feature-2": 2}, {"feature-1": 3, "feature-2": 4}]`, right? 🤔

nhatsmrt:
> Make sure that we provide the conversion back too.

@conradoverta what do you mean by conversion back 🤔

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -255,6 +255,19 @@ def predict(self, x, compress=False, max_retries=5, always_retry_404=True, alway
...
+def prediction_input_df(func):
+    def prediction(self, input):
+        import pandas as pd
+
+        if isinstance(input, pd.DataFrame):
+            # input is dataframe.
+            return func(self, input)
+        else:
+            # input is list of records
+            return func(self, pd.DataFrame.from_records(input))
```
nhatsmrt: This is to allow user to pass in data frame to model.predict as usual (instead of having to convert it into records). The conversion into list of records (and back) will be done behind the scene by our client.

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -683,3 +683,34 @@ def test_update_from_version_diff_workspace_no_access_error(self, client_2, mode
...
+
+        with sys_path_manager() as sys_path:
+            sys_path.append(".")
+
+            from dataframe_custom_model.dataframe_model import DataFrameModel  # pylint: disable=import-error
+
+            df_model = DataFrameModel()
+            model_version.log_model(df_model, custom_modules=[
+                "dataframe_custom_model/",
+                "../verta/"
```
nhatsmrt: I think this is required so that the `prediction_input_df` decorator is found.

client/verta/verta/_internal_utils/_utils.py
nhatsmrt: lmk if you have sth else in mind.

client/verta/verta/deployment/__init__.py
nhatsmrt: fixed.

client/verta/verta/deployment/_deployedmodel.py
nhatsmrt: Pretty sure you're right. Fixed.

client/verta/verta/_internal_utils/_utils.py
nhatsmrt: Oh I think what you want is `{"feature-1": [1, 3], "feature-2": [3, 4]}`. Let me fix that.

conradoverta: The dictionary is more intuitive for me

conradoverta:
> > Make sure that we provide the conversion back too.
> 
> @conradoverta what do you mean by conversion back 🤔

How do I get my dataframe back inside my prediction model?

nhatsmrt:
> > > Make sure that we provide the conversion back too.
> > 
> > 
> > @conradoverta what do you mean by conversion back 🤔
> 
> How do I get my dataframe back inside my prediction model?

@conradoverta I think what you meant is:
```python
class MyModel(object):
  @prediction_input_df
  def predict(self, input):
    # input here should be a dataframe
```
right? It's handled by the `prediction_input_df` decorator, which, receiving the dictionary, will convert it back to a data frame and feed it to the predict method. This is analogous to other helper decorators that I found in DeployedModel.

client/verta/verta/_internal_utils/_utils.py
nhatsmrt: Sg. Should not be a problem. Just fixed, but let me re-run the test.

nhatsmrt: Tested. Still works 🤔

client/verta/tests/test_endpoint/test_endpoint.py
convoliution: It only looks like it's required because we haven't published a Client version with that decorator.

I'd recommend

1. Increment [the version number](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/__about__.py#L9) to `0.15.9a0`
2. Run our `client/publish-pypi` job, passing in this branch for `client_branch`
3. Delete this line
4. Run this test

This is important to validate, because we shouldn't need to send over the Client as a custom module for this functionality.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -255,6 +255,41 @@ def predict(self, x, compress=False, max_retries=5, always_retry_404=True, alway
...
+    """
+    Decorator to handle dataframe inputs for ``predict()``.
+
+    Examples
+    --------
+
+    .. code-block:: python
+
+        class Model(object):
+        @prediction_input_df
```
convoliution:
```suggestion
            @prediction_input_df
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -255,6 +255,41 @@ def predict(self, x, compress=False, max_retries=5, always_retry_404=True, alway
...
+            def predict(self, input):
```
convoliution:
```suggestion
            def predict(self, input_df):
```

convoliution (changes requested):
Logic makes sense to me; one main comment about the test

client/verta/verta/deployment/_deployedmodel.py
nhatsmrt: Fixed.

client/verta/verta/deployment/_deployedmodel.py
nhatsmrt: Fixed.

client/verta/tests/test_endpoint/test_endpoint.py
nhatsmrt: Do I need to include `verta==0.15.9a0` in the requirements instead? Or does the version automatically get picked up?

nhatsmrt: Nvm, it works without it. Probably pick up the verta version information locally.

convoliution: It picks up the version the user has installed installed locally 👍

convoliution (approved):
Looks good to me. I think @conradoverta's concern is covered as well, but I'll leave it to him to check

convoliution:
@conradoverta 

> > > Make sure that we provide the conversion back too.
> > 
> > 
> > @conradoverta what do you mean by conversion back 🤔
> 
> How do I get my dataframe back inside my prediction model?

Did Jim's reply address this?

liuverta:
We're revisiting model I/O very, very soon. And that may come with changes in how the client handles predictions.

And I'm under the impression that users already find these sorts of decorators rather confusing 😬

PR #1677, opened 2020-11-19
Support environment blob in ER

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-7980
**Proto PR:** https://github.com/VertaAI/modeldb/pull/1676


backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -3056,6 +3079,185 @@ public ExperimentRun cloneExperimentRun(
...
+      ArtifactStoreDAO artifactStoreDAO,
+      PythonEnvironmentBlob.Builder pythonEnvironmentBuilder,
+      Artifact artifact)
+      throws ModelDBException {
+    InputStream inputStream = artifactStoreDAO.downloadArtifact(artifact.getPath());
+
+    try (BufferedReader br = new BufferedReader(new InputStreamReader(inputStream))) {
+      String line;
+      while ((line = br.readLine()) != null) {
+        Pattern pattern = Pattern.compile(ModelDBConstants.VER_SPEC_PATTERN);
```
coryverta: We should not need to recompile the pattern every iteration.  This should only need to be done once.

coryverta (approved):

conradoverta (approved):
Tested the change in a dev env and it works great!

PR #1682, opened 2020-11-20
Add workspace_id to ProjectEntity and create Liquibase migration

coryverta:
Migration add `workspace_id` column to `project` table.

PR #1683, opened 2020-11-20
Add workspace id to dataset entity

coryverta:
Merge https://github.com/VertaAI/modeldb/pull/1682 first to preserve database migration order


conradoverta:
> Merge #1682 first to preserve database migration order

What I recommend in these cases is to build a branch on top of your original branch. That makes harder to make mistakes when consolidating things.

conradoverta:
Like I just did =) Now the base is cj/VR-7942_add_workspace_id_to_modeldb so that we see only the dataset changes and, if we merge (by mistake or intentionally), then we merge on top of the other branch instead of master.

coryverta:
I have been using github for years and had no idea you could do that

backend/src/main/resources/liquibase/db-changelog-2.0.xml
```diff
@@ -137,7 +137,7 @@
         <tagDatabase tag="db_version_2.4"/>
     </changeSet>
 
-    <changeSet id="add-workspace-id-to-project" author="anandJ">
+    <changeSet id="add-workspace-id-to-project" author="cory">
```
conradoverta: This will change the hash of the change set, I believe.

coryverta: I did have to clean out my local DB and rebuild it from scratch.  That should not happen on master when we merge this though since only the final change should get applied (if I understand Liquibase properly...)

coryverta: fixed this in the base branch to avoid the hash changing

conradoverta (approved):
We'll probably need this for repository too. Since dataset is just a repository, we might not need it in the dataset table but I could be wrong.

coryverta:
Repository additions coming in the next PR.  There is no class hierarchy that connects Repository and DataSet so I think we will need the workspace ID in both tables, but I will look into it.

PR #1685, opened 2020-11-20
Cj/vr 7942 add workspace id to repository entity

coryverta:


coryverta:
@conradoverta should I split this up?  Most of it the the changes from the proto update

conradoverta:
> @conradoverta should I split this up? Most of it the the changes from the proto update

I'd recommend splitting the proto update from the code update.

PR #1680, opened 2020-11-19
VR-8091 LazyList.find should receive *args instead of a list of parameters.

nhatsmrt:
Let me know what y'all think about this change.

PR #1688, opened 2020-11-21
Allow tuples for find()'s first argument

convoliution:
Our CLI machinery passes a tuple (rather than a list) as the first argument for `find()`, which causes an error since #1680.

It seems reasonable to support tuples for backwards compatibility,
*and also* unpack predicates in the CLI implementation because that's the preferred method going forward


nhatsmrt:
LGTM.

PR #1690, opened 2020-11-22
Proposal API for client datasets

conradoverta:
I was confused by the 3 packages related to dataset, so I commented a lot of stuff out to make sure the path I wanted works =)

Essentially DatasetVersion.content == DatasetBlob.

```
from verta import Client
client = Client("https://dev.verta.ai")
dataset = client.set_dataset("test")
print(dataset)
print('')

from verta.dataset import Path
content = Path("test.py")
print(content)
print('')

dataset_version = dataset.create_version(content)
print(dataset_version)
print('')

dataset_version = dataset.get_latest_version()
print(dataset_version)
print('')
```


convoliution:
@conradoverta This direction looks fine to me, but I should note that I was under the impression one of the goals of the Dataset/DatasetVersion system was that the user wouldn't need to directly interact with a blob at all. I think backwards compatibility is going to be a bit harder than I was imaging. 🤔 

edit:
At a glance, it looks like `create_version()` taking a blob means that when we introduce new dataset types, we don't need to implement anything on the DatasetVersion pieces (it's all implemented on the blob classes). Pretty sure there's a CompSci term for this. I'm still on the fence for whether that's a _better_ user-facing API, but it seems reasonable.

convoliution:
Also this PR introduces a circular import somewhere that makes the client unimportable in Python 2

PR #1691, opened 2020-11-22
Add support for HDFS in blob types

conradoverta:
The HDFS blob is converted to a Path after we use the client to fetch all the metadata related to HDFS itself.

This is the result of running `python hdfs-test/test.py`. It needs to be converted into a proper test.
```
['census-test.csv', 'census-train.csv']
{'directoryCount': 0, 'fileCount': 1, 'length': 9950413, 'quota': -1, 'spaceConsumed': 9950413, 'spaceQuota': -1, 'typeQuota': {}}
HDFSPath Version
    hdfs://data/census/census-test.csv
        9950413 bytes
        last modified: 2020-11-22 18:12:09.330000
        MD5 checksum: 7b0b17a5271852616ed19a72914e776d
HDFSPath Version
    hdfs://data/census/census-test.csv
        9950413 bytes
        last modified: 2020-11-22 18:12:09.330000
        MD5 checksum: 7b0b17a5271852616ed19a72914e776d
    hdfs://data/census/census-train.csv
        3271573 bytes
        last modified: 2020-11-22 18:12:01.846000
        MD5 checksum: 64af2ff44dd04acceb277d024939b619
HDFSPath Version
    hdfs://data/census/census-test.csv
        9950413 bytes
        last modified: 2020-11-22 18:12:09.330000
        MD5 checksum: 7b0b17a5271852616ed19a72914e776d
    hdfs://data/census/census-train.csv
        3271573 bytes
        last modified: 2020-11-22 18:12:01.846000
        MD5 checksum: 64af2ff44dd04acceb277d024939b619
```

Using pyspark:
```
$ python3 hdfs-test/test-pyspark.py 
20/11/22 23:05:00 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
HDFSPath Version                                                                
    hdfs://ip-10-0-147-175.ec2.internal:8020/data/census/census-test.csv
        9950413 bytes
        MD5 checksum: 7b0b17a5271852616ed19a72914e776d
    hdfs://ip-10-0-147-175.ec2.internal:8020/data/census/census-train.csv
        3271573 bytes
        MD5 checksum: 64af2ff44dd04acceb277d024939b619
```
Note: pyspark doesn't provide a *supported* way to get the timestamp from the hdfs location. There are hacks to do that, but they use unsupported APIs (private), so we don't support them either.
Note2: this pyspark functionality is actually independent of the type of data (s3 or even local), so we should probably place it somewhere else later.

### DIFF
--- a/client/verta/verta/dataset/_hdfs.py
+++ b/client/verta/verta/dataset/_hdfs.py
@@ -20,6 +20,16 @@ class HDFSPath(Path):
         if isinstance(paths, six.string_types):
             paths = [paths]
 
+        filepaths = set()
+        for path in paths:
+            for base, dirs, files in self.client.walk(path):
+                for filename in files:
+                    filepaths.add(base + filename)
+            else:
+                filepaths.add(path)
+
+        paths = sorted(list(filepaths))
+
         paths = list(map(lambda path: _HDFS_PREFIX+path if not path.startswith(_HDFS_PREFIX) else path, paths))
         if base_path and not base_path.startswith(_HDFS_PREFIX):
             base_path = _HDFS_PREFIX+base_path
--- a/client/verta/verta/dataset/_hdfs.py
+++ b/client/verta/verta/dataset/_hdfs.py
@@ -22,10 +22,13 @@ class HDFSPath(Path):
 
         filepaths = set()
         for path in paths:
+            previous_length = len(filepaths)
             for base, dirs, files in self.client.walk(path):
                 for filename in files:
                     filepaths.add(base + "/" + filename)
-            else:
+
+            # If the path was a file, then we didn't add any new entry and we should add the file
+            if len(filepaths) == previous_length:
                 filepaths.add(path)
 
         paths = sorted(list(filepaths))
--- a/client/verta/verta/dataset/_hdfs.py
+++ b/client/verta/verta/dataset/_hdfs.py
@@ -13,7 +13,7 @@ from . import _dataset
 
 _HDFS_PREFIX = "hdfs://"
 
-class HDFSPath(Path):
+class HDFSPath(Path):  # TODO: docstrings, add to docs
     # TODO: support mdb versioning
     def __init__(self, hdfs_client, paths, base_path=None):
         self.client = hdfs_client
--- a/client/verta/verta/dataset/_hdfs.py
+++ b/client/verta/verta/dataset/_hdfs.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+
+import hashlib
+
+from ._path import Path
+
+from . import _dataset
+
+class HDFSPath(Path):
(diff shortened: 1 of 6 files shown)


client/verta/verta/dataset/_hdfs.py
```diff
@@ -0,0 +1,96 @@
...
+
+        paths = list(map(lambda path: _HDFS_PREFIX+path if not path.startswith(_HDFS_PREFIX) else path, paths))
+        if base_path and not base_path.startswith(_HDFS_PREFIX):
+            base_path = _HDFS_PREFIX+base_path
+
+        super(HDFSPath, self).__init__(paths, base_path)
+
+    def _file_to_component(self, filepath):
+        original_filepath = filepath
+        filepath = filepath[len(_HDFS_PREFIX):]
```
convoliution:
```suggestion
        filepath = filepath[len(_HDFS_PREFIX):]  # prefix prepended in init
```

client/verta/verta/dataset/_hdfs.py
```diff
@@ -0,0 +1,96 @@
...
+
+from ._path import Path
+
+from ..external import six
+
+from . import _dataset
+
+_HDFS_PREFIX = "hdfs://"
+
+class HDFSPath(Path):
```
convoliution:
```suggestion
class HDFSPath(Path):  # TODO: docstrings, add to docs
```

client/verta/verta/dataset/_hdfs.py
```diff
@@ -0,0 +1,96 @@
...
+            return _dataset.Component(
+                path=filepath,
+                size=len(content),
+                # last_modified=metadata['modificationTime'], # handle timezone?
+                md5=hashlib.md5(content).hexdigest(),
+            )
+
+        result = rdd.map(get_component)
+        result = result.collect()
+        obj = HDFSPath(None, [])
```
convoliution, line 90: Out of curiosity, does this look hacky because it _is_ hacky, or is this actually a sensible way to handle Spark?

convoliution (approved):
Makes sense to me. I'm trusting that the tests work as expected

client/verta/verta/dataset/_hdfs.py
conradoverta: Both. I need to create the object and fill the content later. We don't have an API for that (and not sure if we want one!). I could call __new__ and work around init, but I found it cleaned to just initialize it with nothing.

conradoverta:
> Makes sense to me. I'm trusting that the tests work as expected

Me too =) The logs seem to make sense and to be compatible with what I get in my local machine.

PR #1692, opened 2020-11-23
Implement get/log tags for experiment and project

nhatsmrt:
Eventually, we will need to refactor the get/log (and potentially delete) tags of these entities into a TaggableEntity superclass.


liuverta:
I wish I'd had the chance to get this through before we got these merge conflicts 😕 But it should be trivial to do if it's needed in the future.

PR #1693, opened 2020-11-23
Add workspace id to registry service proto

coryverta:


protos/protos/public/registry/RegistryService.proto
```diff
@@ -32,9 +32,8 @@ message RegisteredModel {
     // Visibility of Registered Models
     common.VisibilityEnum.Visibility visibility = 7;
 
-    // id of the workspace this belongs to
-    //not storing workspace_type uac should support https://github.com/VertaAI/protos-all/pull/19/files
-    string workspace_id = 8;
+    // Legacy id of the workspace this belongs to
+    string legacy_workspace_id = 8;
```
conradoverta: Changing names breaks backwards compatibility with the client. Please check with Michael if the field is used.

protos/protos/public/registry/RegistryService.proto
```diff
@@ -44,6 +43,9 @@ message RegisteredModel {
 
     // Other attributes
     repeated common.KeyValue attributes = 11;
+
+    // id of the workspace that this ``registered_model`` belongs to
+    uint64 parent_workspace_id = 12;
```
conradoverta, line 48: What does "parent" mean? It seems a bit confusing to me.

conradoverta:
@coryverta this was replaced with https://github.com/VertaAI/modeldb/pull/1716 since I changed how we update protos (I'll send info later today)

PR #1640, opened 2020-11-09
VR-6722 Client needs to propagate the original exception for verifyConnection errors

nhatsmrt:
Error format: 
```python3
Traceback (most recent call last):
  File "../connection_err.py", line 2, in <module>
    Client()
  File "/Users/nhatpham/Documents/modeldb/client/verta/verta/client.py", line 194, in __init__
    _utils.raise_for_http_error(response)
  File "/Users/nhatpham/Documents/modeldb/client/verta/verta/_internal_utils/_utils.py", line 534, in raise_for_http_error
    six.raise_from(e, None)  # use default reason
  File "/Users/nhatpham/Documents/modeldb/client/verta/verta/external/six/__init__.py", line 737, in raise_from
    raise value
requests.exceptions.HTTPError: 401 Client Error: authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`; Unauthorized for url: ...
```

PR #1694, opened 2020-11-23
VR-8120 test_entities.py::TestClient::test_wrong_credentials failing; unexpected error msg

nhatsmrt:
Low priority, but it's one line so 🤷‍♂️

### DIFF
--- a/client/verta/verta/client.py
+++ b/client/verta/verta/client.py
@@ -188,8 +188,13 @@ class Client(object):
 
             if is_unauthorized(response):
                 # response.reason was "Unauthorized"
-                new_reason = "authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`\n\n{}".format(response.reason)
-                response.reason = new_reason
+                try:
+                    response.raise_for_status()
+                except requests.HTTPError as e:
+                    e.args = ("authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`\n\n{}".format(
+                        e.args[0]),) + e.args[1:]
+
+                raise e
 
             _utils.raise_for_http_error(response)
             print("connection successfully established")
--- a/client/verta/verta/client.py
+++ b/client/verta/verta/client.py
@@ -194,7 +194,7 @@ class Client(object):
                     e.args = ("authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`\n\n{}".format(
                         e.args[0]),) + e.args[1:]
 
-                raise e
+                    raise e
 
             _utils.raise_for_http_error(response)
             print("connection successfully established")
--- a/client/verta/tests/test_entities.py
+++ b/client/verta/tests/test_entities.py
@@ -182,13 +182,10 @@ class TestClient:
         try:
             os.environ[EMAIL_KEY] = "abc@email.com"
             os.environ[DEV_KEY_KEY] = "def"
+            err_msg = "401 Client Error: User not authorized. Please check the developer_key."
 
-            with pytest.raises(requests.exceptions.HTTPError) as excinfo:
+            with pytest.raises(requests.exceptions.HTTPError, match=err_msg) as excinfo:
                 verta.Client()
-
-            excinfo_value = str(excinfo.value).strip()
-            assert "401 Client Error" in excinfo_value
-            assert "User not authorized. Please check the developer_key" in excinfo_value
         finally:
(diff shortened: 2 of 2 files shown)


convoliution:
@nhatsmrt I'm more interested in _why_ the error message _doesn't_ contain "authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`" because the client [supposedly attaches that to the exception](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py#L189-L194).

nhatsmrt:
> @nhatsmrt I'm more interested in _why_ the error message _doesn't_ contain "authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`" because the client [supposedly attaches that to the exception](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py#L189-L194).

@convoliution It seems that previously we go through this [if branch](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_internal_utils/_utils.py#L552), and the args field of the exception here contains [our implanted reason](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/client.py#L191) (i.e `response.reason`).

Now we go to the else branch. This means that the [`reason`](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_internal_utils/_utils.py#L538-L550) local variable within `raise_for_http_error` (not to be confused with `response.reason`) is non-empty, and does not contain `response.reason`. This could happen because either: 
- [We recently updated `reason` computation logic](https://github.com/VertaAI/modeldb/commit/e1d856b09e306b3f232336977b4f14feae8b7183#diff-96985fbf866d7e24247b00e971fa09472ba6d75baa9fd05b79eebc66140eb3ce), but this is unlikely to affect this, because I just checked and `reason` now still computed via `response.text.strip()`, which _should_ be similar to before this PR.
- The response is updated recently in the backend. I'm not too familiar with UAC to know this.

convoliution:
@nhatsmrt I might've lost track of recent changes on these pieces, but my understanding is that the client is _supposed_ to inject `"authentication failed..."` into the exception message (that's the purpose of [this line](https://github.com/VertaAI/modeldb/blob/31d308d/client/verta/verta/client.py#L191)), but that's not happening, so this test is failing. Changing what the test checks for still leaves this odd behavior in the client logic.

It looks like your investigation found some good potential leads, but ultimately I suspect the client code needs a fix to make sure our error message makes it into the stack trace (and the test code should be correct as it was). We might even need to replace the whole [`is_unauthorized()`](https://github.com/VertaAI/modeldb/blob/31d308d/client/verta/verta/client.py#L187-L194) check with something like what we do [here for `_get_histogram()`](https://github.com/VertaAI/modeldb/pull/1684/files#diff-2f65f6d96823a02929ddfed61dbe257ef1839e8879a67e494d9f8bf83a45b622R374-R380).

nhatsmrt:
Yeah my theory is that something has changed with the response, which makes `response.text` no longer empty. In fact, the error message we are currently receiving (`401 Client Error: User not authorized. Please check the developer_key `) is not even included in the error message that I reported back when I merged the [old PR](https://github.com/VertaAI/modeldb/pull/1640):
```
requests.exceptions.HTTPError: 401 Client Error: authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`; Unauthorized for url: ...
```
Anyway I can probably just re-attach the `authentication failed; please check VERTA_EMAIL and VERTA_DEV_KEY ` part somewhere else. As you suggested, we can just throw the exception right here, then tag this custom message `get_histogram`-style.

nhatsmrt:
Done. See latest commit.

client/verta/verta/client.py
```diff
@@ -188,8 +188,13 @@ def is_unauthorized(response): return response.status_code == 401
...
                 # response.reason was "Unauthorized"
-                new_reason = "authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`\n\n{}".format(response.reason)
-                response.reason = new_reason
+                try:
+                    response.raise_for_status()
+                except requests.HTTPError as e:
+                    e.args = ("authentication failed; please check `VERTA_EMAIL` and `VERTA_DEV_KEY`\n\n{}".format(
+                        e.args[0]),) + e.args[1:]
+
+                raise e
```
```suggestion
                    raise e
```

convoliution (approved):
Thanks!

PR #1697, opened 2020-11-24
Suppress linting errors for protobuf attribute access

convoliution:
Didn't even know this was an option!

Protobufs dynamically generate their classes, or something, so pylint raises attribute errors for all of them.
And I've been basically blocklisting the classes one by one. Until now!

### DIFF
--- a/client/verta/.pylintrc
+++ b/client/verta/.pylintrc
@@ -11,7 +11,7 @@ ignore=CVS,
     # everything below was added by @convoliution
        _swagger,
        _protos,
-       external
+       external,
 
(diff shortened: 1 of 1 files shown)


convoliution:
I've discovered post hoc that pylint in Python 2 pays no heed to `ignored-modules`. 😕 

So I've set my active linter in VS Code to use Python 3, and also periodically run
```bash
python -c "from verta import Client" && pylint --disable=R,C,W,E1101 verta
```
in a Python 2 environment to catch any syntax gotchas.

PR #1699, opened 2020-11-24
Switch over Dataset Versioning

convoliution:
If you are reading this as a user of the ModelDB client, please see [our documentation](https://docs.verta.ai/en/master/tutorials/migrate_dataset.html) for pointers and tips for the new dataset versioning API.

---
Might be more digestible to review commit-by-commit.

Switches the Client's old `Dataset`/`DatasetVersion` APIs to the new ones.

Incorporates several ideas from https://github.com/VertaAI/modeldb/pull/1690

```python
from verta import Client
client = Client()

dataset = client.set_dataset("test")
print(dataset, end='\n\n')

# create version with content
from verta.dataset import Path
content = Path("sandbox.ipynb")
version = dataset.create_version(content)
print(version, end='\n\n')

# dataset latest version
latest_version = dataset.get_latest_version()
assert latest_version.id == version.id
assert latest_version.get_content().list_components() == version.get_content().list_components()

# log to / get from ExperimentRun
key = "test_data"
run = client.set_experiment_run()
run.log_dataset_version(key, version)
retrieved_version = run.get_dataset_version(key)
assert retrieved_version.id == version.id
assert retrieved_version.get_content().list_components() == version.get_content().list_components()
```


client/verta/verta/_tracking/experimentrun.py
```diff
@@ -1012,38 +1015,6 @@ def log_dataset_version(self, key, dataset_version, overwrite=False):
...
-        _utils.validate_flat_key(key)
-
-        warnings.warn("`log_dataset_path()` is deprecated and will removed in a later version;"
-                      " consider using `client.set_dataset(..., type=\"local\")`"
-                      " and `run.log_dataset_version()` instead",
-                      category=FutureWarning)
-
-        # create impromptu DatasetVersion
-        dataset = _dataset.LocalDataset(self._conn, self._conf, name=key)
-        dataset_version = dataset.create_version(path=path)
```
convoliution, line 1043: This is extremely hacky to implement in the new system, and this method has been deprecated for a while anyway, so 👋

client/verta/verta/client.py
```diff
@@ -644,160 +643,160 @@ def get_or_create_repository(self, name=None, workspace=None, id=None, public_wi
         else:
             raise ValueError("must specify either `name` or `id`")
 
-    # NOTE: dataset visibility cannot be set via a client
-    def set_dataset(self, name=None, type="local",
```
convoliution, line 648: I'm intending to later uncomment & rename these old `client` dataset methods, as a temporary option for users who are using the old API but don't have immediate time to migrate their workflows over

convoliution: That means I'll also need to bring back the deleted `_dataset.py`...but it's Just Temporary™

nhatsmrt: Holy Molly that's a lot of commenting! I might be missing it, but I think you'll also have to comment out or remove [the tests for this old dataset API](https://github.com/VertaAI/modeldb/blob/master/client/verta/tests/test_datasets.py) right?

nhatsmrt: Actually maybe it still works, if the new implementation keeps the same client interface? Idk 🤔

convoliution: The old tests will need to be adjusted/removed, but that will come later.

conradoverta: What would it take to support the old API in the new style? Not re-using a lot of the old logic (i.e. pretend those fields have been removed). Maybe it's not worth doing, but just considering the options.

conradoverta (commented):
Some nice cleanup!

Could we merge `client/verta/verta/_dataset_versioning` and `client/verta/verta/dataset`? I was very confused by the existence of both.

evgeniilezhnin (approved):

convoliution:
@conradoverta 

>Could we merge client/verta/verta/_dataset_versioning and client/verta/verta/dataset? I was very confused by the existence of both.

I'd actually be opposed to combining them at the moment, because I see verta/dataset/ and verta/_dataset_versioning as two separate tracks of functionality. Also organizationally, verta/dataset has blob substypes and verta/_dataset_versioning has mdb eneity subtypes.

I could be open to revisiting it later; another reason I wouldn't want to do it now is that this change is kind of complex as-is, and shifting around imported submodules is a bit messier than my mind can probably handle for now.

>What would it take to support the old API in the new style?

[response moved to thread]

client/verta/verta/client.py
convoliution: My plan is to explore this today—the main puzzle piece I'm considering is for `Dataset.create_version()` to check if the first arg is a `str`/`list<str>` (which means it's being used old-style, instead of a blob), and make some `isfile()`/`boto3` checks to generate the content blob internally.

conradoverta (approved):

PR #1700, opened 2020-11-24
Revert "Revert "PESSIMISTIC_WRITE locking""

coryverta:
Reverts VertaAI/modeldb#1698

Fixes the bug causing the the StaleObjectException reported in https://vertaai.atlassian.net/browse/VR-8223 which was due to attempting to obtain a pessimistic write lock on a new object that had not yet been written.

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/versioning/VersioningUtils.java
+++ b/backend/src/main/java/ai/verta/modeldb/versioning/VersioningUtils.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.hibernate.LockMode;
 import org.hibernate.Session;
 import org.hibernate.query.Query;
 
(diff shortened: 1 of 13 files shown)


backend/src/main/java/ai/verta/modeldb/versioning/VersioningUtils.java
```diff
@@ -220,7 +221,7 @@ public static String getVersioningCompositeId(
     }
   }
 
-  public static void saveOrUpdateArtifactPartEntity(
+  public static void saveArtifactPartEntity(
       ArtifactPart artifactPart, Session session, String artifactId, int artifactType) {
     ArtifactPartEntity artifactPartEntity =
         new ArtifactPartEntity(
```
coryverta, line 227: @conradoverta the bug fix is in this method.  There was originally at line 229 a call to 
```
session.lock(artifactEntityPart, LockMode.PESSIMISTIC_WRITE);
```
But since the object is always new this would fail.  Removing the attempt to lock the object succeeds.

conradoverta (approved):

PR #1701, opened 2020-11-24
VR-8224 Write class docstring for HDFSPath blob

nhatsmrt:
It's been awhile since I last wrote docstrings... Could you double check that I'm using the right format?
Also I based the parameters type and description on test files + documentation of those classes, so lmk if I misunderstood any.

### DIFF
--- a/client/verta/verta/dataset/_hdfs.py
+++ b/client/verta/verta/dataset/_hdfs.py
@@ -21,8 +21,8 @@ class HDFSPath(Path):
     ----------
     hdfs_client : hdfs.client.Client
         Instance of HDFS client.
-    paths : str or list of strs
-        A path, or a list of paths, to files from HDFS.
+    paths : list of strs
+        List of paths to files from HDFS.
     base_path : str, optional
         Directory path to be removed from the beginning of `paths` before saving to ModelDB.
 
@@ -89,8 +89,8 @@ class HDFSPath(Path):
         ----------
         sc : pyspark.SparkContext
             SparkContext instance.
-        paths : str or list of strs
-            Path, or list of paths, to binary input data file(s) from HDFS.
+        paths : list of strs
+            List of paths to binary input data file(s) from HDFS.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


client/verta/verta/dataset/_hdfs.py
```diff
@@ -13,7 +13,20 @@
...
+class HDFSPath(Path):
+    """
+    Captures metadata about files from HDFS.
+
+    Parameters
+    ----------
+    hdfs_client : hdfs.client.Client
+        Instance of HDFS client.
+    paths : str or list of strs
+        A path, or a list of paths, to files from HDFS.
```
convoliution: List is preferred!
```suggestion
    paths : list of strs
        List of paths to files from HDFS.
```

client/verta/verta/dataset/_hdfs.py
```diff
@@ -69,6 +82,22 @@ def _hash_file(self, filepath):
...
     def with_spark(sc, paths):
+        """
+        Creates an HDFSPath blob with a SparkContext instance.
+
+        Parameters
+        ----------
+        sc : pyspark.SparkContext
+            SparkContext instance.
+        paths : str or list of strs
+            Path, or list of paths, to binary input data file(s) from HDFS.
```
convoliution: List is preferred!
```suggestion
        paths : list of strs
            List of paths to binary input data file(s) from HDFS.
```

convoliution (changes requested):
Minor suggestions, otherwise these sound about right to me.

Could you also do me a huge, huge favor and add an entry to the dataset blob section of the [docs page](https://github.com/VertaAI/modeldb/blob/master/client/verta/docs/api/api/versioning.rst) as well?

client/verta/verta/dataset/_hdfs.py
nhatsmrt: Fixed.

client/verta/verta/dataset/_hdfs.py
nhatsmrt: Fixed.

nhatsmrt:
@convoliution a new entry is added to the docs page! See latest commit.

convoliution (approved):
Thanks!

PR #1706, opened 2020-11-26
Install spacy<2.3.3 for Python 2

convoliution:
Building wheel for `blis==0.7.3` fails on Debian (https://github.com/explosion/cython-blis/issues/42).

### DIFF
--- a/client/verta/requirements.txt
+++ b/client/verta/requirements.txt
@@ -12,6 +12,7 @@ six>=1.11
 pytest>=4.3
 
 # unit testing dependencies
+blis<0.7.3; python_version == '2.7'  # needed for spaCy, https://github.com/explosion/cython-blis/issues/42
 boto3
 google-cloud-bigquery
 h5py<3.0.0  # https://github.com/tensorflow/tensorflow/issues/44467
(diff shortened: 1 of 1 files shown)


convoliution:
Wow! This doesn't even work, because even though `blis==0.7.2` gets downloaded first during `pip install -r requirements.txt`, `blis<0.8.0,>=0.4.0` is a spaCy build requirement, which means pip ends up fetching `blis==0.7.3` anyway.

Thanks, Python.

convoliution:
Okay, I see that `spacy==2.3.4` was released today and doesn't have its 2.7 wheels up yet, so maybe forcing `<=2.3.4` will avoid this issue until then.

convoliution:
Holy toast, spaCy hasn't had 2.7 wheels for its _last six+ versions_. It looks like I'll have to modify our actual test pipeline to install `blis==0.7.2` beforehand.

convoliution:
My gods, installing `blis` (`<0.5` is what I needed) beforehand _still_ doesn't work because pip absolutely insists on fetching whatever build dependencies it pleases regardless of what's already present on the system.

convoliution:
Downgrading spaCy not one but _two_ versions was the solution, here.

https://github.com/explosion/spaCy/issues/6454 is a thing now.

PR #1713, opened 2020-11-29
Upgrade setuptools during HDFS install

convoliution:
Without this, I encountered
```
ValueError: bad marshal data (unknown type code)
```
while installing `pyspark`


conradoverta (approved):
Interesting

PR #1718, opened 2020-11-30
VR-8143 VR-8144 VR-8145 Send empty names/paths instead of creating custom

conradoverta:


convoliution (approved):
I hope this doesn't break anything!
(To my memory, I don't expect it to)

conradoverta:
> I hope this doesn't break anything!
> (To my memory, I don't expect it to)

We're changing the backend to create random names instead of failing on empty.

PR #1723, opened 2020-12-01
Use grpc interceptor to gather metrics instead of explict

conradoverta:
The goal is to automatically capture metrics in the correct way, without depending on each person to implement the functionality correctly (e.g. forget to copy and paste).


coryverta (approved):
That's a big delta!  This cleans everything up so much.

conradoverta:
> That's a big delta! This cleans everything up so much.

Sorry for the big delta =)

PR #1724, opened 2020-12-01
Remove logAndThrowError

conradoverta:
Right now we have logAndThrowError and observeError to handle errors.
This PR deletes logAndThrowError to have a single path.


coryverta (approved):
That is a massive change, but thank you 10000x @conradoverta for doing this.

PR #1725, opened 2020-12-01
Fix HQL queries broken when the column and property names changed

coryverta:


conradoverta:
@coryverta should we close this since we have reverted things?

PR #1737, opened 2020-12-01
Replace transaction as it is necessary, but not the lock

coryverta:


conradoverta (commented):
Not sure why we need the transaction and not the lock. Could you expand?

coryverta:
The Transaction was there prior to the locking: https://github.com/VertaAI/modeldb/commit/ba6328a626583a8a12a181f42c7b3212b999b0d9#diff-c6ab79e1aacaefd8dd66115f73d85ff498fa54a407a62be7ce2a38f86bd88859R1208
It is necessary for the CommitEntity to get stored and have its ID populated.  
I should not have removed the transaction, only the write lock that will always fail since the entity is new.

conradoverta (approved):

PR #1739, opened 2020-12-02
Remove write lock in addOrUpdateKeyValueProperties

coryverta:
https://vertaai.atlassian.net/browse/VR-8467

This lock will always fail to acquire since this code path is only ever used to add new properties, never for update.  

The method name was changed to reflect this.

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/metadata/MetadataDAORdbImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/metadata/MetadataDAORdbImpl.java
@@ -369,11 +369,12 @@ public class MetadataDAORdbImpl implements MetadataDAO {
             KeyValuePropertyMappingEntity.createId(
                 request.getId(), keyValue.getKey(), request.getPropertyName());
         KeyValuePropertyMappingEntity existingEntity =
-            session.get(KeyValuePropertyMappingEntity.class, id0, LockMode.PESSIMISTIC_WRITE);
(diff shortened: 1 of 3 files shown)


backend/src/main/java/ai/verta/modeldb/metadata/MetadataDAORdbImpl.java
```diff
@@ -361,15 +361,15 @@ public boolean deleteProperty(IdentificationType id, String key) {
...
+  public void addKeyValueProperties(AddKeyValuePropertiesRequest request) {
     try (Session session = ModelDBHibernateUtil.getSessionFactory().openSession()) {
       Transaction transaction = session.beginTransaction();
       for (KeyValueStringProperty keyValue : request.getKeyValuePropertyList()) {
         KeyValuePropertyMappingEntity.KeyValuePropertyMappingId id0 =
             KeyValuePropertyMappingEntity.createId(
                 request.getId(), keyValue.getKey(), request.getPropertyName());
         KeyValuePropertyMappingEntity existingEntity =
-            session.get(KeyValuePropertyMappingEntity.class, id0, LockMode.PESSIMISTIC_WRITE);
+            session.get(KeyValuePropertyMappingEntity.class, id0);
```
coryverta, line 372: Since this is an add the properties do not yet exist.  Any attempt to acquire a lock is guaranteed to fail.

conradoverta (approved):

PR #1740, opened 2020-12-02
VR-8474 add created flag to ProjectEntity and database

coryverta:
https://vertaai.atlassian.net/browse/VR-8475

### DIFF
--- a/backend/src/main/resources/liquibase/db-changelog-2.0.xml
+++ b/backend/src/main/resources/liquibase/db-changelog-2.0.xml
@@ -325,4 +325,22 @@
         <tagDatabase tag="db_version_2.15"/>
     </changeSet>
 
+    <changeSet id="add-created-on-project" author="cory">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="project" columnName="created"/>
+            </not>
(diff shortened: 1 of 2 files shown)


backend/src/main/resources/liquibase/db-changelog-2.0.xml
```diff
@@ -325,4 +325,22 @@
...
     </changeSet>
 
+    <changeSet id="add-created-on-project" author="cory">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="project" columnName="created"/>
+            </not>
+        </preConditions>
+        <addColumn tableName="project">
+            <column name="created" type="boolean"/>
```
conradoverta: Is there a way we can default it to 1 or `true` in the liquibase change? This will allow all previously existing ones to work correctly.

evgeniilezhnin (approved):

backend/src/main/resources/liquibase/db-changelog-2.0.xml
coryverta: :+1:

conradoverta (approved):

PR #1742, opened 2020-12-03
Set the created column value to true

coryverta:
The original value migration did not apply due to liquibase syntax.  This migration will properly update all existing projects to have `created=true`


conradoverta (commented):
Why do we need this? We already had set it to true, no?

conradoverta (approved):

PR #1744, opened 2020-12-03
VR-7698: Keep but deprecate old dataset versioning attributes/methods

convoliution:
Not complete backwards-compatibility as suggested by the ticket, but gives pointers toward updating code that are more informative than "Attribute does not exist! Scour the docs to find its replacement"

### DIFF
--- a/client/verta/verta/_dataset_versioning/dataset_version.py
+++ b/client/verta/verta/_dataset_versioning/dataset_version.py
@@ -339,6 +339,9 @@ class DatasetVersion(entity._ModelDBEntity):
         )
         self._conn.must_response(response)
 
+    def list_components(self):  # from legacy DatasetVersion
+        return self.get_content().list_components()
+
     # The following properties are holdovers for backwards compatibility
     @property
     def desc(self):
@@ -399,5 +402,18 @@ class DatasetVersion(entity._ModelDBEntity):
         version_info_oneof = self._msg.WhichOneof('dataset_version_info')
         return getattr(self._msg, version_info_oneof)
 
-    # TODO: base_path
-    # TODO: list_components
+    @property
+    def base_path(self):  # copied from legacy DatasetVersion
+        warnings.warn(
+            "this attribute is deprecated and will removed in an upcoming version;"
+            " consider checking individual files' base paths"
+            " with `get_content().list_components()` instead",
+            category=FutureWarning,
+        )
+        components = self.get_content().list_components()
+        base_paths = set(component.base_path for component in components)
+
+        if len(base_paths) == 1:
+            return base_paths.pop()
+        else:  # shouldn't happen: DVs don't have an interface to have different base paths
+            raise AttributeError("multiple base paths among components: {}".format(base_paths))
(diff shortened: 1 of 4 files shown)


client/verta/verta/client.py
```diff
@@ -1370,6 +1370,41 @@ def get_dataset_version(self, id):
...
+                      dataset_ids=None, name=None,
+                      tags=None,
+                      sort_key=None, ascending=False,
+                      workspace=None):
+        warnings.warn(
+            "this method is deprecated and will removed in an upcoming version;"
+            " consider using `client.datasets.find()` instead",
+            category=FutureWarning,
+        )
+        datasets = self.datasets
```
convoliution, line 1384: This method is copied over from `0.15.9`; this line is the only functional difference, used to be
```
datasets = Datasets(self._conn, self._conf)
```

client/verta/verta/_dataset_versioning/dataset_version.py
```diff
@@ -336,3 +345,71 @@ def delete(self):
...
+        )
+        # enum value for PATH, which is currently the only supported type
+        # https://github.com/VertaAI/modeldb/blob/af5e3a7/protos/protos/public/modeldb/DatasetService.proto#L32
+        return 1
+
+    @property
+    def dataset_version(self):
+        warnings.warn(
+            "this attribute is deprecated and will removed in an upcoming version;"
+            " interfacing directly with the internal protobuf structure is not recommended",
```
conradoverta:
```suggestion
            " interfacing directly with the internal protobuf structure is not supported",
```

client/verta/verta/client.py
```diff
@@ -863,8 +863,8 @@ def get_or_create_endpoint(self, path=None, description=None, workspace=None, pu
             return Endpoint._get_by_id(self._conn, self._conf, workspace, id)
         else:
             return Endpoint._get_or_create_by_name(self._conn, path,
-                                            lambda path: Endpoint._get_by_path(self._conn, self._conf, workspace, path),
-                                            lambda path: Endpoint._create(self._conn, self._conf, workspace, public_within_org, path, description),
+                                            lambda name: Endpoint._get_by_path(self._conn, self._conf, workspace, path),
```
conradoverta: Keep it for consistency

conradoverta (approved):

client/verta/verta/client.py
convoliution: Huh, I don't even know how that change made it in here; it's totally unrelated 🤔

client/verta/verta/_dataset_versioning/dataset_version.py
```diff
@@ -336,3 +345,71 @@ def delete(self):
...
+            "this attribute is deprecated and will removed in an upcoming version;"
+            " consider using `get_content()` instead",
+            category=FutureWarning,
+        )
+        self._refresh_cache()
+        version_info_oneof = self._msg.WhichOneof('dataset_version_info')
+        return getattr(self._msg, version_info_oneof)
+
+    @property
+    def base_path(self):  # copied from legacy DatasetVersion
```
convoliution:
```suggestion
    def base_path(self):
```

PR #1753, opened 2020-12-08
VR-8152 backfill existing project resources

anandjakhaniya:
Rebase PR https://github.com/VertaAI/modeldb/pull/1751 and created this.


backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
```diff
@@ -1187,15 +1187,18 @@ private boolean createWorkspacePermissions(
...
       ResourceVisibility resourceVisibility = getResourceVisibility(workspaceType, visibility);
       SetResources.Builder setResourcesBuilder =
-          SetResources.newBuilder()
-              .setResources(resources)
-              .setVisibility(resourceVisibility)
-              .setCollaboratorType(collaboratorType);
+          SetResources.newBuilder().setResources(resources).setVisibility(resourceVisibility);
+
+      if (!(resourceVisibility.equals(ResourceVisibility.ORG_DEFAULT)
+          || resourceVisibility.equals(ResourceVisibility.PRIVATE))) {
```
conradoverta: Nit: let's just equal for the condition that we want instead of not equal all the other ones.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,204 @@
...
+
+          // Fetch the project owners userInfo
+          Map<String, UserInfo> userInfoMap =
+              authService.getUserInfoFromAuthServer(userIds, null, null);
+          for (ProjectEntity project : projectEntities) {
+            if (userInfoMap.containsKey(project.getOwner())) {
+              try {
+                UserInfo ownerInfo = userInfoMap.get(project.getOwner());
+                SetEntityVisibility(authServiceChannel, project, ownerInfo);
+                setWorkspacePermissionsForOwner(project, ownerInfo);
```
conradoverta: This should be a single call to UAC. If we do 2, we actually make our lives much harder because we have to ensure that whatever we `set` is equal the second time or it will override with the wrong values.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,204 @@
...
+          LOGGER.debug("Projects userId list : " + userIds);
+          if (userIds.size() == 0) {
+            LOGGER.warn("userIds not found for Projects on page lower boundary {}", lowerBound);
+            lowerBound += pagesize;
+            continue;
+          }
+
+          // Fetch the project owners userInfo
+          Map<String, UserInfo> userInfoMap =
+              authService.getUserInfoFromAuthServer(userIds, null, null);
```
conradoverta, line 102: Why do we need to fetch the user information? We just need to send the owner id, which we already have from the db, right?

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+      }
+    }
+
+    LOGGER.debug("Projects migration finished");
+  }
+
+  private void setWorkspacePermissionsForOwner(ProjectEntity project, UserInfo ownerInfo) {
+    final Optional<Long> ownerId =
+        ownerInfo != null
```
conradoverta: When would owner info be null? This seems like a bug to me, no?

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,204 @@
...
+        typedQuery.setMaxResults(pagesize);
+        List<ProjectEntity> projectEntities = typedQuery.getResultList();
+
+        if (projectEntities.size() > 0) {
+          Set<String> userIds = new HashSet<>();
+          for (ProjectEntity project : projectEntities) {
+            userIds.add(project.getOwner());
+          }
+          LOGGER.debug("Projects userId list : " + userIds);
+          if (userIds.size() == 0) {
```
conradoverta: When would this be zero? I believe only when projects is zero, right?

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,204 @@
...
+        Long.parseLong(project.getWorkspace()),
+        Optional.ofNullable(WorkspaceTypeEnum.WorkspaceType.forNumber(project.getWorkspace_type())),
+        project.getId(),
+        ownerId,
+        ModelDBServiceResourceTypes.PROJECT,
+        CollaboratorTypeEnum.CollaboratorType.READ_WRITE,
+        project.getProjectVisibility());
+  }
+
+  private void SetEntityVisibility(
```
conradoverta: This seems to already set all the fields, right? I'm wondering why we need `setWorkspacePermissionsForOwner` too.

conradoverta (commented):
If the checks on user info being null or owner id being zero is about oss, then I recommend we just completely skip the migration logic in oss since it doesn't make sense there.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: OSS case. I can skip the migration for OSS.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Yes, that will also fine. I can set the owner id which exists on the project entity.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: OSS case. I can skip the migration for OSS.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
conradoverta: Yeah, I suggest we skip it for oss. It also simplifies things considerably =)

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: I do not have an idea UAC doing it on set resource call so I have made the call again for the owner.

conradoverta: The UAC `set` call is equivalent to a `PUT` in http. So whatever fields you send, it sets. I recommend comparing this with `setWorkspacePermissionsForOwner` in the ModelDB codebase.

coryverta (approved):

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,166 @@
...
+              // ignore the CollaboratorType
+              roleService.createWorkspacePermissions(
+                  workspaceDTO.getWorkspaceName(),
+                  Optional.ofNullable(
+                      WorkspaceTypeEnum.WorkspaceType.forNumber(project.getWorkspace_type())),
+                  project.getId(),
+                  Optional.of(Long.parseLong(project.getOwner())),
+                  ModelDBServiceResourceTypes.PROJECT,
+                  CollaboratorTypeEnum.CollaboratorType.READ_WRITE,
+                  project.getProjectVisibility());
```
evgeniilezhnin: I think here old visibility ORG_SCOPED_PUBLIC should become new ORG_DEFAULT. cc: @conradoverta to check if I'm wrong here

evgeniilezhnin: And CollaboratorType and canDeploy parameters should have default value for setResource in case of ORG_DEFAULT

evgeniilezhnin: I was thinking of something like https://github.com/VertaAI/registry/pull/102/files#diff-00d851bf9cc1b4dfa8203d45a8ac9fed69662de44f46e9872a412e8f270e2844R125

backend/config.default.yaml
```diff
@@ -91,6 +91,9 @@ migration:
   POPULATE_VERSION_MIGRATION:
     enable: true
     record_update_limit: 100
+  COLLABORATOR_RESOURCE_MIGRATION:
+    enable: true
```
conradoverta, line 95: This should be done automatically on every bringup. Then when we want to disable, we should delete the code. The days of manual migration need to be gone.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,236 @@
...
+        if (projectEntities.size() > 0) {
+          Set<String> userIds = new HashSet<>();
+          for (ProjectEntity projectEntity : projectEntities) {
+            userIds.add(projectEntity.getOwner());
+          }
+          LOGGER.debug("Project userId list : " + userIds);
+
+          // Fetch the experiment owners userInfo
+          Map<String, UserInfo> userInfoMap =
+              authService.getUserInfoFromAuthServer(userIds, null, null);
```
conradoverta, line 102: All of this can be removed.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,236 @@
...
+              // if projectVisibility is not equals to ResourceVisibility.ORG_SCOPED_PUBLIC then
+              // ignore the CollaboratorType
+              createWorkspacePermissions(
+                  Optional.of(workspaceDTO.getWorkspaceName()),
+                  Optional.ofNullable(
+                      WorkspaceTypeEnum.WorkspaceType.forNumber(project.getWorkspace_type())),
+                  project.getId(),
+                  Optional.of(Long.parseLong(project.getOwner())),
+                  ModelDBServiceResourceTypes.PROJECT,
+                  CollaboratorTypeEnum.CollaboratorType.READ_WRITE,
```
conradoverta: Why are we defaulting to this type of collaborator?

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,236 @@
...
+    } catch (Exception ex) {
+      if (ModelDBUtils.needToRetry(ex)) {
+        return getEntityCount(klass);
+      } else {
+        throw ex;
+      }
+    }
+  }
+
+  private void createWorkspacePermissions(
```
conradoverta: There's another method in another currently open PR with exactly the same name. cc @coryverta is this a copy?

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,236 @@
...
+          authServiceChannel.getCollaboratorServiceBlockingStub().setResources(setResources);
+
+      LOGGER.info("SetResources message sent.  Response: " + setResourcesResponse);
+    } catch (StatusRuntimeException ex) {
+      LOGGER.error(ex);
+      ModelDBUtils.retryOrThrowException(ex, false, retry -> null);
+    }
+  }
+
+  private ResourceVisibility getResourceVisibility(
```
conradoverta: Not used if we delete `createWorkspacePermissions`.

backend/config.default.yaml
conradoverta: Actually, this doesn't even seem to be used. So that's fine, but it's confusing.

evgeniilezhnin: This is used here: https://github.com/VertaAI/modeldb/blob/aj/feature/VR-8152_backfill_existing_project_resources/backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java#L762

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
anandjakhaniya: Changed

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Changed

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: No, it is used to fetch the workspace name so we need userInfo of the project owner. so this is required

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: So what is expected here as a default? this is adding default when the project has the ORG_DEFAULT workspace.

evgeniilezhnin: READ_ONLY CollaboratorType is default because READ_ONLY = 0

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Yes it is a copy-paste method but there is one change which is https://github.com/VertaAI/modeldb/blob/fe66220573e9ebb12c2781b45257a825db853f0a/backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java#L228 so when we are migrating if the project visibility has `Visibility.ORG_SCOPED_PUBLIC` then we are setting `ResourceVisibility.ORG_DEFAULT;` at UAC. but when we create project and getting visibility `Visibility.ORG_SCOPED_PUBLIC` then we are setting `ResourceVisibility.ORG_SCOPED_PUBLIC;` at tge UAC from RoleService. thats why define second method here.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: We can not remove this https://github.com/VertaAI/modeldb/pull/1753/files/fe66220573e9ebb12c2781b45257a825db853f0a#r542103742

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
evgeniilezhnin: so createWorkspacePermissions looks the same as in roleService, but it uses another getResourceVisibility functionality? Can't we just pass the right visibility to roleService.createWorkspacePermissions?

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Ok, then i am updating it.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
ad-47: Ok, let me try if its possible.

anandjakhaniya: Yup, let me check if possible. i have duplicated it because createWorkspacePermissions called from multiple places and i don't want to touch those all for converting `Visivility.*` to `ResourceVisibility.*` logic multiple places.

conradoverta: Let's not duplicate. I prefer to have this implemented well in the first pass than with considerable technical debt before it's even functional.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
conradoverta: Since ORG_CUSTOM is new, we'll never encounter a situation where this field matters in doing the migration, so let's just make sure it doesn't raise an error.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
conradoverta: Let's resolve the duplicated code and see if this is needed

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
conradoverta: Could you expand on why we need to load user information? I don't get it

backend/config.default.yaml
conradoverta: k

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: We have a common method for fetch the workpaceName from workspaceId `roleService.getWorkspaceDTOByWorkspaceId(` so for that i need a userInfo as an argument to fetch workspaceName.
```
WorkspaceDTO workspaceDTO =
                  roleService.getWorkspaceDTOByWorkspaceId(
                      userInfoMap.get(project.getOwner()),
                      project.getWorkspace(),
                      project.getWorkspace_type());
```

anandjakhaniya: https://github.com/VertaAI/modeldb/blob/fe66220573e9ebb12c2781b45257a825db853f0a/backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java#L124

evgeniilezhnin: I think getWorkspaceDTOByWorkspaceId looks like a code for getting old workspaces information. Here is what I tried to write for new (in registry): https://github.com/VertaAI/registry/pull/102/files#diff-eb351bf96b15680d857baab17a91a6f29aecacc59d81271080602ed248773a16R1226

evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,186 @@
...
+              roleService.createWorkspacePermissions(
+                  workspaceDTO.getWorkspaceName(),
+                  project.getId(),
+                  Optional.of(Long.parseLong(project.getOwner())),
+                  ModelDBServiceResourceTypes.PROJECT,
+                  CollaboratorTypeEnum.CollaboratorType.READ_ONLY,
+                  getResourceVisibility(
+                      Optional.ofNullable(
+                          WorkspaceTypeEnum.WorkspaceType.forNumber(project.getWorkspace_type())),
+                      VisibilityEnum.Visibility.forNumber(project.getProject_visibility())));
```
evgeniilezhnin: I think this getter project.getProject_visibility() will return a transient value (in this case always PRIVATE), but expected is a value from the database

evgeniilezhnin (changes requested):
https://github.com/VertaAI/modeldb/pull/1753#discussion_r543038887

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: No @evgeniilezhnin , `project.getProject_visibility()` return initeger which exists on database. the transient value returns by the `project.getProjectVisibility()`

evgeniilezhnin: got it

evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Updated like your suggestion

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,170 @@
...
+
+        typedQuery.setFirstResult(lowerBound);
+        typedQuery.setMaxResults(pagesize);
+        List<ProjectEntity> projectEntities = typedQuery.getResultList();
+
+        if (projectEntities.size() > 0) {
+          for (ProjectEntity project : projectEntities) {
+            try {
+              WorkspaceDTO workspaceDTO =
+                  roleService.getWorkspaceDTOByWorkspaceId(project.getWorkspace());
```
conradoverta: Is `getWorkspace` populated? I'm pretty sure we never ended up populating it. Only the legacy ID is present today.

conradoverta (commented):
Has this been tested? I see use of `getWorkspace` without the associated type, which I think it the wrong behavior.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: we are always populating `workspace` field in the database https://github.com/VertaAI/modeldb/blob/fad1cd78aacc52c395179d649fb601534a142d7f/backend/src/main/java/ai/verta/modeldb/entities/ProjectEntity.java#L56. the legacy ID is not populated on the database

anandjakhaniya:
Running on testing and I realize that at Modeldb we are keeping vertaId as a workspace so we can not use the same logic like Registry does. cc @evgeniilezhnin. So I am reverting the last commit. the old code with fetching userInfo and get VertaId as workspace is the right logic. You are right @conradoverta for this https://github.com/VertaAI/modeldb/pull/1753#pullrequestreview-552144929

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Revert this. we are used vertaId as a workspace on the project so your logic is not working on MDB.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Revert changes

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,186 @@
...
+                  workspaceDTO.getWorkspaceName(),
+                  project.getId(),
+                  Optional.of(Long.parseLong(project.getOwner())),
+                  ModelDBServiceResourceTypes.PROJECT,
+                  CollaboratorTypeEnum.CollaboratorType.READ_ONLY,
+                  getResourceVisibility(
+                      Optional.ofNullable(
+                          WorkspaceTypeEnum.WorkspaceType.forNumber(project.getWorkspace_type())),
+                      VisibilityEnum.Visibility.forNumber(project.getProject_visibility())));
+            } catch (Exception ex) {
```
conradoverta: We cannot just ignore that an exception happened and move on with the code. If we do, we'll end up running code on unmigrated data and it will cause the system to be inconsistent. Let's be **extremely** noisy if something fails that will prevent us from operating correctly.

conradoverta (changes requested):

evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Ok. we are running this migration on a separate thread so it is not extremely noisy right now. but I will update it too noisy. :)

anandjakhaniya: Changed the migration to noisy

evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -0,0 +1,169 @@
...
+          }
+        } else {
+          LOGGER.debug("Total projects count 0");
+        }
+        lowerBound += pagesize;
+      } catch (Exception ex) {
+        if (ModelDBUtils.needToRetry(ex)) {
+          migrateProjects();
+        } else {
+          throw ex;
```
conradoverta, line 130: I'm assuming this will make the system crash, right? If yes, approved.

conradoverta (approved):
Approved depending on the answer of the qeustion.

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
anandjakhaniya: Yes, we are crashing the system

coryverta (approved):

PR #1756, opened 2020-12-09
Fix the bad syntax in add-create-on-project and update the checksum

coryverta:
@ad-47 Gave me the correct syntax to use so I have modified the existing migration to use that syntax and. apply a corrected checksum.


evgeniilezhnin (approved):

anandjakhaniya (approved):

conradoverta (approved):
Please check if this will brick applications that already had the update performed on them.

PR #1758, opened 2020-12-10
Fix setProjectWorkspace endpoint & clean up code

anandjakhaniya:
- Remove setVisibility DAO impl
- Fix setProjectWorkspace endpoint


coryverta (commented):
LGTM

PR #1759, opened 2020-12-10
prototype basename support for reverse proxy case

CaptEmulation:


webapp/.env.default
```diff
@@ -1,2 +1,3 @@
 BACKEND_API_PROTOCOL='https'
 BACKEND_API_DOMAIN='dev.verta.ai'
+BACKEND_API_BASE_PATH='/'
```
CaptEmulation, line 3: If modeldb is being a reverse proxy at /my/modeldb then set this to `/my/modeldb/`

webapp/client/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "verta",
   "version": "0.1.0",
   "private": true,
+  "homepage": "/",
```
CaptEmulation, line 5: This should also match the base path in server.js, e.g. `/my/modeldb/`

webapp/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN yarn install
 WORKDIR /usr/src/app/client
 COPY client /usr/src/app/client
 ENV REACT_APP_OSS 1
+ENV REACT_APP_PUBLIC_URL /
```
CaptEmulation, line 17: I think this CRA in OSS is old so I needed to add this as well

PR #1762, opened 2020-12-11
Implement create cleanup and delete cleanup flow for Project in MDB

anandjakhaniya:
**Jira Tickets:** https://vertaai.atlassian.net/browse/VR-9343, https://vertaai.atlassian.net/browse/VR-9344


coryverta (approved):

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -121,7 +120,7 @@ private void deleteProjects(Session session) {
         projectIds.add(projectEntity.getId());
       }
       try {
-        deleteRoleBindingsForProjects(projectEntities);
+        roleService.deleteProjectResources(projectIds);
       } catch (StatusRuntimeException ex) {
```
conradoverta: We need to ignore just chosen exceptions instead of all.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
```diff
@@ -0,0 +1,121 @@
...
+            .append(" AND ")
+            .append(alias)
+            .append(".")
+            .append(ModelDBConstants.DATE_CREATED)
+            .append(" < :created_date ")
+            .toString();
+
+    // Time less then a minute because possible to have create project request running when cron job
+    // running
+    long time = Calendar.getInstance().getTimeInMillis() - 60000; // 1 minute lesser time
```
conradoverta:
```suggestion
    long time = Calendar.getInstance().getTimeInMillis() - 300000; // 5 minutes lesser time
```

conradoverta: Just in case

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
```diff
@@ -0,0 +1,121 @@
...
+    Query projectDeleteQuery = session.createQuery(deleteProjectsQueryString);
+    projectDeleteQuery.setParameter("created", false);
+    projectDeleteQuery.setParameter("created_date", time);
+    projectDeleteQuery.setMaxResults(this.recordUpdateLimit);
+    List<ProjectEntity> projectEntities = projectDeleteQuery.list();
+
+    List<String> projectIds = new ArrayList<>();
```
conradoverta, line 86: Everything below here seems to be the same as the deletion cronjob, right?

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: https://vertaai.slack.com/archives/CJG0NL86B/p1607920510000500?thread_ts=1607681257.124500&cid=CJG0NL86B

anandjakhaniya: So this code is fine here.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
anandjakhaniya: Yup. but just added new line  `roleService.deleteProjectResources(projectIds);` new before deleting project entities from DB

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
anandjakhaniya: updated

evgeniilezhnin (approved):

coryverta (approved):

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
conradoverta: Let's avoid code duplication. Is it possible for us to re-use the code?

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
conradoverta: The link says that UAC will return 200 in case of success. This means that we should _stop_ in case of an error, instead of going ahead and deleting things from the db. Once we delete from the db, we'll never call this deletion again, which will cause a leak.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
anandjakhaniya: No, we can not re-use it because there is some code to update the status of the sub-entities like experiments while deleting projects.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: `roleService.deleteProjectResources(projectIds);` bulk deletion call to UAC if we stop in case error then all the projects are skipped for deletion due to the one project. so do you think we call UAC for each project instead of bulk??

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
conradoverta: Could you expand? The code really seems like it's mostly duplicate, so I don't understand what's the issue.

anandjakhaniya: https://github.com/VertaAI/modeldb/blob/d291a71eee766bed8381a61b28535395bb7efce7/backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java#L136

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
conradoverta: What we cannot have it fail and then we proceed like it didn't fail. That is clearly a bug that will leave to inconsistent behavior. We should fix so that it's not buggy anymore, whatever the solution happens to be.

Please check the other similar pieces of code for this same bug and let's fix them in another PR.

backend/src/main/java/ai/verta/modeldb/cron_jobs/CleanUpEntitiesCron.java
conradoverta: I still don't understand the issue, so pointing to a piece of code that I also am not familiar with doesn't really help me to understand. But sure, we can add the code duplication.

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: Started a thread for this on slack https://vertaai.slack.com/archives/CJG0NL86B/p1608020346023100
No need to create a new PR for this. it is not a complex logic. it just needs to move `roleService.deleteProjectResources(projectIds);` in the same try-catch to stop going for the further process.

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
```diff
@@ -511,18 +511,14 @@ private void setRoleBindingOnAuthService(boolean retry, RoleBinding roleBinding)
               GetCollaboratorResponseItem.newBuilder()
                   .setAuthzEntityType(collaborator.getAuthzEntityType())
                   .setVertaId(collaborator.getId());
-          getCollaboratorResponseBuilder
-              .getPermissionBuilder()
-              .setCollaboratorType(CollaboratorType.READ_ONLY);
+          CollaboratorPermissions.Builder collPermBuilder = CollaboratorPermissions.newBuilder();
```
conradoverta, line 514: I'm pretty sure I have already reviewed all these changes in another PR.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -732,6 +732,8 @@ public Project deepCopyProjectForUser(String srcProjectID, UserInfo newOwner)
       LOGGER.debug("Mark Projects as deleted : {}, count : {}", allowedProjectIds, updatedCount);
       transaction.commit();
       LOGGER.debug("Project deleted successfully");
+
+      roleService.deleteProjectResources(allowedProjectIds);
```
conradoverta: This must be done in the cronjob instead of the API call itself as it might fail. Since we already do it there, we should remove from here or we can end up sending an error for something that was as success. We should only do things once =)

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
anandjakhaniya: Yes, it is the same.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: it is fine to remove this line from here.

anandjakhaniya: Fixed.

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
conradoverta: Did we copy code then? I couldn't review this file because it was too close to the other one, so I wouldn't be able to perceive the differences.

conradoverta (approved):

PR #1766, opened 2020-12-14
Enable managed versioning for DatasetVersion

convoliution:
Summary of changes:

- copy artifact upload methods from `Commit` to `DatasetVersion`
- overwrite `_create()` in `DatasetVersion` to perform the file preparation and upload steps (simplified from `Commit`)
- add helper methods to dataset blob so that `download()` can handle both `Commit` and `DatasetVersion` sources

### DIFF
--- a/client/verta/verta/_dataset_versioning/dataset_version.py
+++ b/client/verta/verta/_dataset_versioning/dataset_version.py
@@ -105,27 +105,40 @@ class DatasetVersion(entity._ModelDBEntity):
         raise NotImplementedError("cannot get DatasetVersion by name")
 
     @classmethod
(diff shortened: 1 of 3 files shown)


client/verta/verta/_dataset_versioning/dataset_version.py
```diff
@@ -413,3 +442,146 @@ def base_path(self):
             " considering accessing the paths of specific components"
             " using `list_components()[i].path` instead",
         )
+
+    # The following properties are for managed
+    # TODO: consolidate this with similar method in `_ModelDBEntity`
+    def _get_url_for_artifact(self, dataset_component_path, method, part_num=0):
```
convoliution, line 448: The rest of this file (`_get_url_for_artifact` and `_upload_artifact`) are copied directly from `Commit`; the only changes are for the URLs and the protos.

conradoverta (approved):
LGTM

PR #1772, opened 2020-12-15
Dataset, repository entities and liquibase changes

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/entities/DatasetEntity.java
```diff
@@ -216,21 +226,30 @@ public void setDeleted(Boolean deleted) {
...
+                RdbmsUtils.convertAttributeEntityListFromAttributes(getAttributeMapping()))
+            .setTimeCreated(getTime_created())
+            .setTimeUpdated(getTime_updated())
+            .setWorkspaceId(getWorkspace())
+            .setWorkspaceTypeValue(getWorkspace_type());
+
+    GetResourcesResponseItem repositoryResource =
+        roleService.getEntityResource(
+            String.valueOf(this.id), ModelDBResourceEnum.ModelDBServiceResourceTypes.DATASET);
+    datasetBuilder.setVisibility(repositoryResource.getVisibility());
```
conradoverta, line 245: Owner, visibility, workspace and custom permissions all must come from UAC and _not_ the local db. It should be safe for me to drop the columns from the local db that have been migrated.

backend/src/main/java/ai/verta/modeldb/entities/versioning/RepositoryEntity.java
```diff
@@ -208,9 +226,12 @@ public Repository toProto() throws InvalidProtocolBufferException {
...
         .addAllAttributes(
             RdbmsUtils.convertAttributeEntityListFromAttributes(getAttributeMapping()));
-    if (repository_visibility != null) {
-      builder.setRepositoryVisibilityValue(repository_visibility);
-    }
+
+    GetResourcesResponseItem repositoryResource =
+        roleService.getEntityResource(
+            String.valueOf(this.id), ModelDBResourceEnum.ModelDBServiceResourceTypes.REPOSITORY);
+    builder.setVisibility(repositoryResource.getVisibility());
```
conradoverta, line 233: Same

conradoverta (changes requested):
Also, let's make sure that we have a follow-up PR for all the places where we don't use hibernate directly and instead write HQL so that we can filter on the `created` state like we do with `deleted`.

evgeniilezhnin (approved):

evgeniilezhnin (changes requested):
build failed

backend/src/main/java/ai/verta/modeldb/entities/DatasetEntity.java
anandjakhaniya: Currently, visibility comes from UAC (line 250) but the owner and workspace come from the local DB. i will update it as well.

anandjakhaniya: Updated owner and workspace as well now.

backend/src/main/java/ai/verta/modeldb/entities/versioning/RepositoryEntity.java
anandjakhaniya: Updated

backend/src/main/java/ai/verta/modeldb/entities/DatasetEntity.java
```diff
@@ -108,9 +108,6 @@ public DatasetEntity(Dataset dataset) throws InvalidProtocolBufferException {
   @Column(name = "deleted")
   private Boolean deleted = false;
 
-  @Column(name = "created")
```
conradoverta: Why are we deleting this? I'm confused.

anandjakhaniya: Because now we are used RepositoryEntity for the dataset so DatasetEntity and DatasetVersionEntity are not in use. also the dataset and dataset version Dao implementation is not in use because we have replaced it with Repository and commit.

conradoverta: Oh, cool. That makes sense to me

conradoverta (approved):

coryverta (changes requested):
@ad-47 the build is failing due to compilation errors and a merge conflict

coryverta (approved):

evgeniilezhnin (approved):

PR #1774, opened 2020-12-15
Dataset and repository delete cron fixes

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
```diff
@@ -771,7 +629,9 @@ private void deleteRepositories(Session session) {
         } catch (Exception ex) {
           LOGGER.warn("DeleteEntitiesCron : deleteRepositories : Exception: ", ex);
         } finally {
-          transaction.commit();
+          if (transaction != null) {
+            transaction.commit();
```
conradoverta: We should not commit something that had an exception during execution.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/cron_jobs/DeleteEntitiesCron.java
anandjakhaniya: FIxed

conradoverta (approved):
Approved assuming the build passes.

Could you also double check the other cron jobs to make sure we're not committing something that has failed? I imagine you copied this skeleton from somewhere else.

coryverta (changes requested):
@ad-47 the build is failing due to compilation errors

coryverta (approved):

PR #1781, opened 2020-12-16
Fix compilation for remaining changes and test class

anandjakhaniya:


coryverta (changes requested):
@ad-47  The build is failing with compiler errors.

coryverta (approved):

PR #1784, opened 2020-12-17
Fix visibility logic and test cases

anandjakhaniya:


backend/src/main/resources/liquibase/db-changelog-2.0.xml
```diff
@@ -362,7 +362,7 @@
             </not>
         </preConditions>
         <addColumn tableName="repository">
-            <column name="created" type="boolean" value="true" defaultValue="false"/>
+            <column name="created" type="boolean" defaultValue="false"/>
```
coryverta, line 365: This will likely change the SHA.  Please test in the dev environments to make sure the migration will execute.

coryverta (approved):

evgeniilezhnin (approved):

backend/src/main/resources/liquibase/db-changelog-2.0.xml
anandjakhaniya: this is the new changeset so it will not break anything. i have tested it with local setup. but i will check it with dev as well.

PR #1785, opened 2020-12-18
Expand org-sharing tests

convoliution:
Adds tests for `Dataset` and `RegisteredModel`.

They succeed with `verta==0.16.0` but have proto-related failures with `0.16.2`.

- `TestProject.test_org_public_project` fails because `ORG_SCOPED_PUBLIC` was removed from an enum in `ProjectService`
- `TestProject.test_share_org_project` fails because the message field `AddCollaboratorRequest.collaborator_type` was removed (though this is not from a public client function)
- `TestDataset.test_org_public_dataset` fails because `DatasetService.DatasetVisibilityEnum` was removed
- `TestRegisteredModel.test_org_public_registered_model` fails because `ORG_SCOPED_PUBLIC` was removed from an enum somewhere
- `TestRepository.test_org_public_repository` fails because `VersioningService.RepositoryVisibilityEnum` was removed


conradoverta (commented):
Please merge master and double check with the reverted protos

convoliution:
@conradoverta They all pass, after merging master

conradoverta:
> @conradoverta They all pass, after merging master

Including the ones that are disabled and relate to collaboration? Like the one they encountered?

convoliution:
@conradoverta Yes, the collab tests pass (when I run them on my system), including the case encountered by the user.

PR #1793, opened 2020-12-22
resources integration test

evgeniilezhnin:


backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
```diff
@@ -0,0 +1,451 @@
...
+    // Delete entities by cron job
+    deleteEntitiesCron.run();
+
+    // shutdown test server
+    serverBuilder.build().shutdownNow();
+  }
+
+  private static final Collection<Object[]> requestData =
+      Arrays.asList(
+          new Object[][] {
```
conradoverta: Could we define an actual type? I'm not sure what each field means and I really like relevant names for things.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
```diff
@@ -0,0 +1,451 @@
...
+  @Test
+  public void createProjectWithGlobalSharingOrganization() {
+    LOGGER.info("Global organization Project test start................................");
+
+    if (app.getAuthServerHost() == null || app.getAuthServerPort() == null) {
+      Assert.assertTrue(true);
+      return;
+    }
+
+    String orgName = "Org-test-verta";
```
conradoverta, line 126: Does the organization already exist? If so, should this be a global constant?

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
```diff
@@ -0,0 +1,451 @@
...
+        GetRoleByName.newBuilder()
+            .setName(orgRoleName)
+            .setScope(RoleScope.newBuilder().setOrgId(organization.getId()).build())
+            .build();
+    GetRoleByName.Response getRoleByNameResponse =
+        roleServiceBlockingStub.getRoleByName(getRoleByName);
+    assertEquals(
+        "Expected role name not found in DB",
+        orgRoleName,
+        getRoleByNameResponse.getRole().getName());
```
conradoverta, line 154: Why are we testing role name from modeldb tests?

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
```diff
@@ -0,0 +1,451 @@
...
+    String orgResourceId = null;
+    Repository repository = null;
+    try {
+      for (Object[] data : requestData) {
+        ResourceVisibility resourceVisibility = (ResourceVisibility) data[0];
+        CollaboratorPermissions.Builder customPermission =
+            (CollaboratorPermissions.Builder) data[1];
+        boolean isReadAllowed = (boolean) data[2];
+        boolean isWriteAllowed = (boolean) data[3];
+        if (orgResourceId == null) {
```
conradoverta: This will change as part of the loop and it's pretty confusing to me how things relate in this test because of this.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
```diff
@@ -0,0 +1,451 @@
...
+                          .addResourceIds(orgResourceId)
+                          .setResourceType(
+                              ResourceType.newBuilder()
+                                  .setModeldbServiceResourceType(
+                                      resourceType
+                                              == ModelDBResourceEnum.ModelDBServiceResourceTypes
+                                                  .DATASET
+                                          ? ModelDBResourceEnum.ModelDBServiceResourceTypes
+                                              .REPOSITORY
+                                          : resourceType)
```
conradoverta: We seem to be ignoring the type `DATASET`. I'm not sure why. Let's make sure everything except for the internals of modeldb treat dataset and repository as two completely independent types. The fact that datasets are implemented as repositories is an internal implementation detail and should not be exposed outside of the internal logic.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
```diff
@@ -0,0 +1,451 @@
...
+              client2RepositoryServiceStub.updateRepository(
+                  SetRepository.newBuilder()
+                      .setId(
+                          RepositoryIdentification.newBuilder()
+                              .setRepoId(Long.parseLong(orgResourceId)))
+                      .setRepository(
+                          repository
+                              .toBuilder()
+                              .setDescription("new_description" + new Date().getTime()))
+                      .build());
```
conradoverta: Why are these 3 paths testing setting different things? I'm confused.

conradoverta (changes requested):
Could we break `createProjectWithGlobalSharingOrganization` apart with a few helper methods? I see 4 switches on the same variable, which indicates to me that this should be at least one function per type in the switch.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: added type for that

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: it doesn't exist

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: I moved the test with this code from ProjectTest. I guess because it is test that checks integration between MDB and UAC

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: it checks if it should create or update resource. If orgResourceId is null it means that it is not created

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: Internally dataset is repository so all bindings and resources are created in UAC for REPOSITORY. I'm not sure how to use DATASET here. When in Webapp you open a dataset, there is collaborator service used for REPOSITORY so bindings are created for REPOSITORY.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: No reason. Why? I have no preferred way to do it. I've updated the code to use only description.

evgeniilezhnin:
> Could we break `createProjectWithGlobalSharingOrganization` apart with a few helper methods? I see 4 switches on the same variable, which indicates to me that this should be at least one function per type in the switch.

added helper methods

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: And MDB don't check any DATASET bindings. What should be done to do as you suggest? All UAC endpoints for Dataset should start working with repositories when called? Can it be done?

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
conradoverta: Ok. I think mdb shouldn't worry about roles anymore, but we can revisit this later.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
conradoverta: Thanks!

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
conradoverta: >all bindings and resources are created in UAC for REPOSITORY

This is very surprising to me. I'd say this is definitely a bug, but not one that we need to resolve right now. We should have UAC consider repositories and datasets as completely separate entity types.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
conradoverta: Thanks! I was wondering if there was any gain or issue with testing a single path.

conradoverta (commented):
I think this looks correct, but I'd appreciate if @ad-47 could take a look too.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
anandjakhaniya: this is the case which i have asked when we have stated working on visibility @conradoverta Slack thread: https://vertaai.slack.com/archives/CJG0NL86B/p1608212405074500

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
anandjakhaniya: @evgeniilezhnin `orgResourceId` is not null then as per code logic it has failed some where and not clear the created resources right? and you ignore all other next text scenario from for loop. Also if `orgResourceId` is nut null then you are not throwing error and mark test case fail.

anandjakhaniya (changes requested):
Could you change or simplify loop for https://github.com/VertaAI/modeldb/pull/1793/files#r550082890 then it would be helpful.

backend/src/test/java/ai/verta/modeldb/GlobalSharingTest.java
evgeniilezhnin: The logic is: there is a loop on different TestParameters. On a first iteration it creates a resource. On next iterations it updates an existing resource with current TestParameters. When the test ends there is a resource deletion.

evgeniilezhnin:
> Could you change or simplify loop for https://github.com/VertaAI/modeldb/pull/1793/files#r550082890 then it would be helpful.

simplified

anandjakhaniya (approved):
LGTM

PR #1794, opened 2020-12-22
Fix visibility logic bugs

anandjakhaniya:


evgeniilezhnin (approved):

conradoverta (approved):
I'd appreciate if we could be more specific on what are the bugs, how we identified them and how have we verified that they are resolved.

PR #1795, opened 2020-12-23
Fix implementation without owner, workspace and workspace_type in project entity

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -482,6 +482,7 @@ public RepositoryEntity setRepository(
         LOGGER.debug("Project role bindings created successfully");
         Transaction transaction = session.beginTransaction();
         repositoryEntity.setCreated(true);
+        repositoryEntity.setVisibility_migration(true);
```
conradoverta: This is a completely unrelated fix. Please do not mix unrelated changes in the same PR. We've had a ton of issues with that in the past and it's super confusing for the reviewer.

backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
```diff
@@ -205,11 +137,12 @@ public void updateProjectName(
...
           ModelDBServiceResourceTypes.PROJECT, request.getId(), ModelDBServiceActions.UPDATE);
 
+      UserInfo userInfo = authService.getCurrentLoginUserInfo();
       Project updatedProject =
           projectDAO.updateProjectName(
-              request.getId(), ModelDBUtils.checkEntityNameLength(request.getName()));
+              userInfo, request.getId(), ModelDBUtils.checkEntityNameLength(request.getName()));
       saveAuditLogs(
-          null,
+          userInfo,
```
conradoverta, line 145: This seems unrelated.

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
```diff
@@ -1184,18 +1192,31 @@ public GetResourcesResponseItem getEntityResource(
...
               .getCollaboratorServiceBlockingStub()
-              .getResources(getResourcesBuilder.build());
-      return response.getItemList();
+              .getResources(GetResources.newBuilder().setResources(resources).build());
+      return response.getItemList().stream()
+          .filter(
+              item ->
+                  item.getResourceType().getModeldbServiceResourceType()
+                          == modelDBServiceResourceTypes
+                      && item.getService() == ServiceEnum.Service.MODELDB_SERVICE)
```
conradoverta: If this is not equal, then it's a bug on UAC, right? Why did you feel the need to add this check?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -165,25 +157,83 @@ public ProjectDAORdbImpl(
...
-        "Project",
-        "projectName",
-        project.getName(),
-        ModelDBConstants.WORKSPACE,
-        project.getWorkspaceId(),
-        project.getWorkspaceType(),
-        LOGGER);
+  private void checkIfEntityAlreadyExists(
+      Session session, Workspace workspace, String projectName) {
+    List<String> projectEntityIds = getProjectIdsByName(session, projectName);
```
conradoverta, line 188: Project names only have to be unique in the workspace. We can have duplicated names in different workspaces. I believe this code is checking for globally unique, right?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -209,34 +259,29 @@ public Project insertProject(Project project, UserInfo userInfo)
       LOGGER.debug("Project role bindings created successfully");
       transaction = session.beginTransaction();
       projectEntity.setCreated(true);
+      projectEntity.setVisibility_migration(true);
```
conradoverta: Unrelated

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -662,7 +707,8 @@ public Project deepCopyProjectForUser(String srcProjectID, UserInfo newOwner)
 
     // cloned project
     Project newProject = copyProjectAndUpdateDetails(srcProject, newOwner);
-    insertProject(newProject, newOwner);
+    // TODO: FIX ME
+    // insertProject(newProject, newOwner);
```
conradoverta: Please add more detail on what needs to be fixed

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -798,22 +844,33 @@ public Project getProjectByID(String id) throws InvalidProtocolBufferException {
...
-                .setCode(Code.ALREADY_EXISTS_VALUE)
-                .setMessage("Project already exist with given short name")
-                .build();
-        throw StatusProto.toStatusRuntimeException(status);
+
+      List<String> accessibleProjectIds =
+          roleService.getAccessibleResourceIds(
+              null,
+              new CollaboratorUser(authService, userInfo),
+              ResourceVisibility.PRIVATE,
```
conradoverta: Why do we need visibility in this listing?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -798,22 +844,33 @@ public Project getProjectByID(String id) throws InvalidProtocolBufferException {
...
+              ModelDBServiceResourceTypes.PROJECT,
+              Collections.emptyList());
+
+      if (!accessibleProjectIds.isEmpty()) {
+        Query query = session.createQuery(GET_PROJECT_BY_SHORT_NAME_HQL);
+        query.setParameter("projectShortName", projectShortName);
+        query.setParameterList(ModelDBConstants.PROJECT_IDS, accessibleProjectIds);
```
conradoverta: I'm unsure why we need to get all project ids to check name collision. We just need to verify within a workspace. Maybe we should move the collision checking to the collaborator service?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -949,29 +970,35 @@ public ProjectPaginationDTO findProjects(
...
-                      builder.equal(
-                          projectRoot.get(ModelDBConstants.WORKSPACE_TYPE),
-                          WorkspaceType.ORGANIZATION_VALUE))));
+
+        List<GetResourcesResponseItem> items =
+            roleService.getResourceItems(accessibleProjectIds, ModelDBServiceResourceTypes.PROJECT);
+        for (GetResourcesResponseItem item : items) {
+          if (orgWorkspaceIds.contains(String.valueOf(item.getWorkspaceId()))) {
+            accessibleProjectIds.remove(item.getResourceId());
+          }
```
conradoverta, line 978: What is this doing? I'm confused.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -949,29 +970,35 @@ public ProjectPaginationDTO findProjects(
...
+            Workspace workspace = roleService.getWorkspaceByWorkspaceName(userInfo, workspaceName);
+            List<GetResourcesResponseItem> items =
+                roleService.getResourceItems(
+                    accessibleProjectIds, ModelDBServiceResourceTypes.PROJECT);
+            for (GetResourcesResponseItem item : items) {
+              if (workspace.getId() != item.getWorkspaceId()) {
+                accessibleProjectIds.remove(item.getResourceId());
+              }
+            }
+          }
```
conradoverta: I don't follow this logic.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -1222,30 +1225,38 @@ public Project deleteArtifacts(String projectId, String artifactKey)
...
-      WorkspaceDTO workspaceDTO =
-          roleService.getWorkspaceDTOByWorkspaceName(currentLoginUserInfo, workspaceName);
+      Workspace workspace =
+          roleService.getWorkspaceByWorkspaceName(currentLoginUserInfo, workspaceName);
+      List<GetResourcesResponseItem> items =
+          roleService.getResourceItems(accessibleProjectIds, ModelDBServiceResourceTypes.PROJECT);
+      for (GetResourcesResponseItem item : items) {
+        if (workspace.getId() != item.getWorkspaceId()) {
+          accessibleProjectIds.remove(item.getResourceId());
+        }
```
conradoverta: If we're getting projects not in the workspace, that looks like a bug, right?

conradoverta (changes requested):
What is the difference between the short name and regular name?

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
anandjakhaniya: Yes. But I have added it for the same side.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: As per the old logic, we have a unique project name in the workspace so currently I am checking uniqueness at the workspace level. We can have duplicate names in different workspaces.https://github.com/VertaAI/modeldb/pull/1795/files#diff-19df052e74b46634ca35a406234e4539b5f4e5ea7a7c18f743fc8162ba1847c2R615-R635

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: The New visibility project does not need to migrate again so It will protect to fetch the new visibility project created while running the migration. 
Unrelated creating separate PR for this.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: I will add  description, but this change for the `deepCopyProject` endpoint. so Could you confirm that we used `deepCopyProject` endpoint or not?? if it is not used then we can remove this logic.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: it is not used any where we can remove it. i will remove it.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: i forget to add workspace only check here, will add it.

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
anandjakhaniya: I will remove this.

anandjakhaniya:
Last commit changes are
 **Moved to other PR**
 - `*.setVisibility_migration(true);` 
 - Removed visibility from `getAccessibleResourceIds` method

 **Modify as per review**
 - Removed ServiceEnum.Service.MODELDB_SERVICE check from getResource
 - Fix the // TODO: FIX ME for inserProject in the deepCopyProject
 - Call getResourceItems with the workspace name and change logic according to it

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
```diff
@@ -1034,6 +1034,22 @@ public WorkspaceDTO getWorkspaceDTOByWorkspaceName(
...
 
+  @Override
+  public Workspace getWorkspaceByWorkspaceName(
+      UserInfo currentLoginUserInfo, String workspaceName) {
+    /*from the name for workspace, get the workspace id and type.
+    if no workspace is present assume user's personal workspace*/
+    if (workspaceName == null
+        || workspaceName.isEmpty()
+        || workspaceName.equalsIgnoreCase(
+            authService.getUsernameFromUserInfo(currentLoginUserInfo))) {
```
conradoverta: I'm pretty sure we don't need to do this check. If we have a workspace name, we can just use it. It doesn't matter if it's the current user or not.

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
```diff
@@ -1184,17 +1192,31 @@ public GetResourcesResponseItem getEntityResource(
...
+  public List<GetResourcesResponseItem> getResourceItems(
+      Workspace workspace,
+      Set<String> resourceIds,
+      ModelDBServiceResourceTypes modelDBServiceResourceTypes) {
     try (AuthServiceChannel authServiceChannel = new AuthServiceChannel()) {
-      final GetResources.Builder getResourcesBuilder = GetResources.newBuilder();
-      if (filterTo.isPresent()) {
-        getResourcesBuilder.setResources(filterTo.get());
+      if (resourceIds == null || resourceIds.isEmpty()) {
+        return Collections.emptyList();
```
conradoverta: This looks like a bug waiting to happen. I'd imagine that, if I don't send resource IDs, it fetches everything instead of nothing. Do we have this behavior documented anywhere to avoid issues?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -165,25 +185,86 @@ public ProjectDAORdbImpl(
...
+
+  /**
+   * Method to convert createProject request to Project object. This method generates the project Id
+   * using UUID and puts it in Project object.
+   *
+   * @param request : CreateProject
+   * @param userInfo : UserInfo
+   * @return Project
+   */
+  private Project getProjectFromRequest(CreateProject request, UserInfo userInfo) {
```
conradoverta, line 213: Is this function exactly the same that existed in the other file and just got moved here? I cannot see what's the difference between them

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -796,11 +871,22 @@ public Project getProjectByID(String id) throws InvalidProtocolBufferException {
...
-      Query query = session.createQuery(GET_PROJECT_BY_SHORT_NAME_AND_OWNER_HQL);
+      Set<String> accessibleProjectIds =
+          ModelDBUtils.filterWorkspaceOnlyAccessibleIds(
+              roleService,
+              new HashSet<>(Collections.singletonList(projectId)),
+              null,
+              userInfo,
+              ModelDBServiceResourceTypes.PROJECT);
+      if (accessibleProjectIds.isEmpty()) {
+        throw new ModelDBException("You can not set project short_name in others workspace");
```
conradoverta: This seems like a big leap. Why the list being empty always means it's another workspace?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
conradoverta: Please add a TODO that we'll use the workspace argument of the UAC call. If there's a follow up, let's make sure to have a TODO (if small or we plan to do right after) or a ticket.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -1173,83 +1233,55 @@ public Project deleteArtifacts(String projectId, String artifactKey)
...
         @SuppressWarnings("unchecked")
-        Query<String> query = session.createQuery(IDS_FILTERED_BY_WORKSPACE);
+        Query<String> query = session.createQuery(NON_DELETED_PROJECT_IDS_BY_IDS);
         query.setParameterList(ModelDBConstants.PROJECT_IDS, accessibleProjectIds);
-        query.setParameter(ModelDBConstants.WORKSPACE, workspaceDTO.getWorkspaceId());
-        query.setParameter(
-            ModelDBConstants.WORKSPACE_TYPE, workspaceDTO.getWorkspaceType().getNumber());
         resultProjects = query.list();
-
-        // in personal workspace show projects directly shared
```
conradoverta, line 1238: How id this being handled after the change? I'm not clear on it.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
```diff
@@ -592,6 +595,60 @@ public static void checkPersonalWorkspace(
...
+              .setCode(Code.PERMISSION_DENIED_VALUE)
+              .setMessage(
+                  "Creation of "
+                      + resourceNameString
+                      + " in other user's workspace is not permitted")
+              .addDetails(Any.pack(UpdateProjectName.Response.getDefaultInstance()))
+              .build();
+      throw StatusProto.toStatusRuntimeException(status);
+    }
+  }
```
conradoverta: Can't we just let UAC handle this?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Yes this is the same method which is existed on other file just removed part from that method is 
```
if (userInfo != null) {
      String vertaId = authService.getVertaIdFromUserInfo(userInfo);
      projectBuilder.setOwner(vertaId);
      String workspaceName = request.getWorkspaceName();
      WorkspaceDTO workspaceDTO =
          roleService.getWorkspaceDTOByWorkspaceName(userInfo, workspaceName);
      if (workspaceDTO.getWorkspaceId() != null) {
        projectBuilder.setWorkspaceId(workspaceDTO.getWorkspaceId());
        projectBuilder.setWorkspaceType(workspaceDTO.getWorkspaceType());
        if (workspaceDTO.getWorkspaceServiceId() != null) {
          projectBuilder.setWorkspaceServiceId(workspaceDTO.getWorkspaceServiceId());
        }
      }
    }
```

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
anandjakhaniya: Ok, got it.  i will remove  that check.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: So here i am setting given project id `new HashSet<>(Collections.singletonList(projectId)),` so i am calling to the UAC with the worksapceName and the projectID. So if that projectId is not from my workspace then it should not be editable to me so UAC return empty resource list.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Here we can not use the workspace name with UAC call because we have multiple org_workspace and if i used that call with  workspace then it will call UAC in for loop for each org_worksapce. So here i have call UAC once without org_worksapce and applying logic to remove those ids.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: This logic is moved above with workspace https://github.com/VertaAI/modeldb/pull/1795/files/ebc9a94d3c55394424e1ff1c739bf16b477c3986#diff-de14dc78b43342a659f26880706f23c24558f63961a0d39bffad3a86125908e2R1247-R1268

Old logic called two time to database so now i have managed single list based on workspace and call DB to once.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
anandjakhaniya: No, this logic is just check the requested workspace is matched with current login users workspace and we have enough info to check that instead of calling UAC.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Yes

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
conradoverta: Oh, wow. How did I miss that!?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
conradoverta: But `filterWorkspaceOnlyAccessibleIds` is not being passed any info about workspace, right? It's also possible that I can edit project in other workspaces (e.g. collaborator or in organization). Shouldn't we leave decisions of whether the user can access a resource or not to UAC? Spreading the logic around seems error prone, so I'm not sure we have a good reason to do it. Here it could just be a call to "isSelfAllowed" and it would be done (and it would be clearer what is the permission check). Maybe I'm missing something?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
conradoverta: That link actually just points me to the top of the page. Not sure why. What are the file and lines that you are referring to?

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
conradoverta: It's not a matter of whether we can or not, but whether we should or not. Every time we make a permission decision in a component vs UAC, it needs to be duplicated in exactly the same way in all components, which is much more error prone. It would be much more robust to have this permission verified in a central place that every system can rely on, wouldn't you agree?

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Ok, got it your point now. `filterWorkspaceOnlyAccessibleIds` is not work here. i will revert the logic based on the `isSelfAllowed`

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Yes, it is not pointing there but after redirecting go to the file backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java and open the `load diff` it will show you selected block (line 1247-1268)

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
anandjakhaniya: Yes, removing from MDB and i hope the UAC handle this thing on the `serResource` endpoint

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
conradoverta: Cool. We've built a lot of machinery in MDB that might not work the same and might be simplified by a quick call to UAC =) We'll catch them as we find them during code changes.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
conradoverta: If it doesn't, I'd say it's a UAC bug that we have to fix anyway. cc @evgeniilezhnin to double check

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
conradoverta: Got it! Checking

conradoverta: Thanks! Makes sense. I'm not sure how I missed it before.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: updated code with isSelfAllowed endpoint

conradoverta (approved):
Let's get this in and follow up removing the project listing (O(N)) with point-wise operations in UAC.

PR #1801, opened 2020-12-28
Fix fuzzy search on owner field

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1865,22 +1870,34 @@ public static Order getOrderBasedOnSortKey(
...
-        List<String> vertaIds =
-            userInfoList.stream()
-                .map(authService::getVertaIdFromUserInfo)
-                .collect(Collectors.toList());
+        Set<String> projectIdSet = new HashSet<>();
+        for (UserInfo userInfo : userInfoList) {
+          List<GetResourcesResponseItem> accessibleAllWorkspaceItems =
+              roleService.getResourceItems(
+                  Workspace.newBuilder()
+                      .setId(authService.getWorkspaceIdFromUserInfo(userInfo))
```
conradoverta, line 1886: This is already available and doesn't need an API call, right? It's part of the protos, iirc

conradoverta (approved):
Nice catch.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
anandjakhaniya: Yes. worksapceId available in UserInfo proto obj so we are just extracted it in authService.

PR #1804, opened 2020-12-29
Implemented repository, dataset without owner, workspace, workspaceType

anandjakhaniya:

PR #1807, opened 2020-12-30
Create Repository and Dataset without owner, workspace, workspaceType

anandjakhaniya:
Splitted PR: https://github.com/VertaAI/modeldb/pull/1804


anandjakhaniya:
Splitted as much as. Now this PR only content the create flow repository and dataset.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
```diff
@@ -679,11 +678,11 @@ public static void logAmazonServiceExceptionErrorCodes(Logger LOGGER, AmazonServ
   }
 
   public static ResourceVisibility getResourceVisibility(
-      Optional<WorkspaceType> workspaceType, ProtocolMessageEnum visibility) {
-    if (!workspaceType.isPresent()) {
+      Workspace workspace, ProtocolMessageEnum visibility) {
```
conradoverta: I'd recommend leaving Optional in. It's _much_ safer than something that might be null. It's also much more obvious in terms of api that something is optional vs required.

conradoverta (approved):
LGTM

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
anandjakhaniya: Change it to optional now.

PR #1812, opened 2020-12-31
Use config for database information

conradoverta:
This is the first commit focused on changing our config to a more structured format.

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
+++ b/backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
@@ -13,7 +13,6 @@ public interface ModelDBConstants {
   String CLOUD_BUCKET_NAME = "cloudBucketName";
   String MINIO_ENDPOINT = "minioEndpoint";
   String AWS_REGION = "aws_region";
-  String DATABASE = "database";
   String DB_TYPE = "DBType";
   String MIN_CONNECTION_POOL_SIZE = "minConnectionPoolSize";
   String MAX_CONNECTION_POOL_SIZE = "maxConnectionPoolSize";
--- a/backend/src/main/java/ai/verta/modeldb/config/Config.java
+++ b/backend/src/main/java/ai/verta/modeldb/config/Config.java
@@ -1,0 +1,55 @@
+package ai.verta.modeldb.config;
+
+import static ai.verta.modeldb.utils.ModelDBUtils.appendOptionalTelepresencePath;
+
+import ai.verta.modeldb.ModelDBConstants;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.Constructor;
+import java.util.Optional;
+
+public class Config {
+  public static String MISSING_REQUIRED = "required field is missing";
+
+  private static Config config = null;
+
+  public DatabaseConfig database;
+  public TestConfig test;
+
+  // FIXME
+  public Object artifactStoreConfig;
+  public Object artifactStore_grpcServer;
+  public Object authService;
+  public Object cron_job;
+  public Object enableTrace;
+  public Object grpcServer;
+  public Object mdb_service_user;
+  public Object populateConnectionsBasedOnPrivileges;
+  public Object springServer;
+  public Object telemetry;
+
+  public static Config getInstance() throws InvalidConfigException, FileNotFoundException {
+    if (config == null) {
+      Yaml yaml = new Yaml(new Constructor(Config.class));
+      String filePath = System.getenv(ModelDBConstants.VERTA_MODELDB_CONFIG);
+      filePath = appendOptionalTelepresencePath(filePath);
+      InputStream inputStream = new FileInputStream(new File(filePath));
+      config = yaml.load(inputStream);
+      config.Validate();
+    }
+    return config;
+  }
+
+  public void Validate() throws InvalidConfigException {
+    if (database == null) throw new InvalidConfigException("database", MISSING_REQUIRED);
+    database.Validate("database");
+
+    if (test != null) {
+      test.Validate("test");
+    }
+  }
+}
--- a/backend/src/main/java/ai/verta/modeldb/config/TestConfig.java
+++ b/backend/src/main/java/ai/verta/modeldb/config/TestConfig.java
@@ -1,0 +1,11 @@
+package ai.verta.modeldb.config;
+
+public class TestConfig {
+  public DatabaseConfig database;
+
+  public void Validate(String base) throws InvalidConfigException {
+    if (database == null)
+      throw new InvalidConfigException(base + ".database", Config.MISSING_REQUIRED);
+    database.Validate(base + ".database");
+  }
+}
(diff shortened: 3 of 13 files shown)


backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
```diff
@@ -13,7 +13,6 @@
   String CLOUD_BUCKET_NAME = "cloudBucketName";
   String MINIO_ENDPOINT = "minioEndpoint";
   String AWS_REGION = "aws_region";
-  String DATABASE = "database";
```
anandjakhaniya, line 16: We should have to remove other DB config parameters which is not used now like 
```
  String DB_TYPE = "DBType";
  String MIN_CONNECTION_POOL_SIZE = "minConnectionPoolSize";
  String MAX_CONNECTION_POOL_SIZE = "maxConnectionPoolSize";
  String CONNECTION_TIMEOUT = "connectionTimeout";
```

backend/src/main/java/ai/verta/modeldb/config/Config.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class Config {
+  public static String MISSING_REQUIRED = "required field is missing";
+
+  private static Config config = null;
+
+  public DatabaseConfig database;
+  public TestConfig test;
+
+  // FIXME
```
anandjakhaniya, line 23: What is still pending to fix?? i can't get it.

conradoverta: We need to support the other parts of the config with concrete types instead of the placeholder `Object`. Then we need to replace their use across the code.

backend/src/main/java/ai/verta/modeldb/config/TestConfig.java
```diff
@@ -0,0 +1,11 @@
+package ai.verta.modeldb.config;
+
+public class TestConfig {
+  public DatabaseConfig database;
+
```
anandjakhaniya, line 5: Test user config is missing here
```
testUsers:
    primaryUser:
      email: 
      devKey: 
    secondaryUser:
      email: 
      devKey: 
```

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
```diff
@@ -24,7 +24,8 @@ public ParentTimestampUpdateCron(int recordUpdateLimit) {
   }
 
   public void initExperimentUpdateQueryString(int recordUpdateLimit) {
-    if (ModelDBHibernateUtil.rDBDialect.equals(ModelDBConstants.POSTGRES_DB_DIALECT)) {
+    if (ModelDBHibernateUtil.config.RdbConfiguration.RdbDialect.equals(
```
anandjakhaniya, line 27: instead of this if check should we add just 
```
isPostgresDB(){
  return RdbConfiguration.RdbDialect.equals(
        ModelDBConstants.POSTGRES_DB_DIALECT);
}
```
and will used just `isPostgresDB()` at all places

conradoverta: Check one of the follow-up PRs ;)

backend/src/main/java/ai/verta/modeldb/config/TestConfig.java
conradoverta: I'll add more items as I see them missing. The other ones were added because the library complains about fields in the yaml not being assigned to vars (which is good!)

backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
conradoverta: I'll do that in a follow-up. I'm not a big fan of centralized constants list for things that are local. That's why we missed that some feature flags were not being used at all, I believe.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -889,68 +806,55 @@ private static void lockedMigration(
...
     ResultSet resultSet = connection.getMetaData().getCatalogs();
 
-    boolean dbExists = false;
     while (resultSet.next()) {
       String databaseNameRes = resultSet.getString(1);
-      if (databaseName.equals("`" + databaseNameRes + "`")) {
-        dbExists = true;
-        break;
+      if (rdb.RdbDatabaseName.equals(databaseNameRes)) {
+        System.out.println("the database " + rdb.RdbDatabaseName + " exists");
```
anandjakhaniya, line 848: LOGGER.info() ?

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -889,68 +806,55 @@ private static void lockedMigration(
...
+        DriverManager.getConnection(rdb.RdbUrl, rdb.RdbUsername, rdb.RdbPassword);
     ResultSet resultSet = connection.getMetaData().getCatalogs();
 
-    boolean dbExists = false;
     while (resultSet.next()) {
       String databaseNameRes = resultSet.getString(1);
-      if (databaseName.equals("`" + databaseNameRes + "`")) {
-        dbExists = true;
-        break;
+      if (rdb.RdbDatabaseName.equals(databaseNameRes)) {
```
anandjakhaniya, line 847:
```suggestion
      if (rdb.RdbDatabaseName.equals("`" + databaseNameRes + "`")) {
```
to support database name with `-`

anandjakhaniya (approved):
Small suggestions are added If you agree then required to change other wise LGTM

anandjakhaniya (changes requested):
Need to fix some review comments.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
conradoverta: I believe ` is only necessary when performing a query, right? This is comparing a string provided by the user with a string returned by the db. Neither will have the backtick, afaik.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
conradoverta: Sure. I was just copying a command that was a few lines below. The main goal of the PR is to change config behavior. Everything else is extra.

backend/src/main/java/ai/verta/modeldb/ModelDBConstants.java
anandjakhaniya: Ok. agree

backend/src/main/java/ai/verta/modeldb/config/Config.java
anandjakhaniya: Ok, got it.

backend/src/main/java/ai/verta/modeldb/cron_jobs/ParentTimestampUpdateCron.java
anandjakhaniya: Sorry about this comment. i does not show it. https://github.com/VertaAI/modeldb/pull/1814

conradoverta: Don't worry. This PR is also useful for you to see what I mean of 1 PR for 1 thing changing. It has ~5 small follow-ups that are not part of the core change.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
anandjakhaniya: i am not sure about it but if you have tested lets take it in now. i will test it later

anandjakhaniya (approved):

PR #1815, opened 2020-12-31
Run migrations together with liquibase

conradoverta:
Right now the two migrations happen in completely separate places. They both should happen together
and potentially at the init container.


backend/src/main/java/ai/verta/modeldb/App.java
```diff
@@ -450,32 +456,8 @@ public static void initializeServicesBaseOnDataBase(
...
-
-    switch (database.DBType) {
-      case ModelDBConstants.RELATIONAL:
-
-        // --------------- Start Initialize relational Database base on configuration
-        // ---------------
-        app.propertiesMap = propertiesMap;
-        ModelDBHibernateUtil.createOrGetSessionFactory(database);
-
-        LOGGER.trace("RDBMS configured with server");
```
anandjakhaniya, line 459: what about this check? currently we do not have other database type so you can move migration up but when second database will come up we will need this switch case.

conradoverta: We'll hit that problem when we get there. The config is already checking the type and ensuring it's only relational for now.

anandjakhaniya: Ok, then it will fine for now with assuming we are supporting just RDBMS.

anandjakhaniya (approved):

PR #1817, opened 2020-12-31
Use client timestamps if provided

conradoverta:


anandjakhaniya (approved):
Great to remove it. now code is looks better. :)

PR #1832, opened 2021-01-01
Add config for migrations

conradoverta:


backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -561,109 +562,37 @@ public static Query getWorkspaceEntityQuery(
...
-            CommonUtils.unregisteredBackgroundUtilsCount();
-          }
+            break;
         }
       }
-
-      CollaboratorResourceMigration.execute();
     }
+
+    CollaboratorResourceMigration.execute();
```
conradoverta, line 595: This was inside the dataset version migration, I believe, which is wrong

backend/src/main/java/ai/verta/modeldb/config/DatabaseConfig.java
```diff
@@ -12,6 +14,7 @@
   public Integer connectionTimeout = ModelDBConstants.CONNECTION_TIMEOUT_DEFAULT;
 
   public RdbConfig RdbConfiguration;
+  public List<MigrationConfig> migrations;
```
anandjakhaniya, line 16: Migration format is like Map structure in old flow
```
migration:
  DATASET_VERSIONING_MIGRATION:
    enable: false
    record_update_limit: 100
  SUB_ENTITIES_OWNERS_RBAC_MIGRATION:
    enable: false
  POPULATE_VERSION_MIGRATION:
    enable: false
    record_update_limit: 100
  COLLABORATOR_RESOURCE_MIGRATION:
    enable: true
    record_update_limit: 100
```

but you have implemented it as list

conradoverta: Yes, because order of migrations matter. By using a list, we enforce a specific order. A map can ran them in any order

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -561,109 +562,37 @@ public static Query getWorkspaceEntityQuery(
...
   public static void runMigration(DatabaseConfig config)
-      throws ClassNotFoundException, ModelDBException {
+      throws ClassNotFoundException, ModelDBException, DatabaseException, SQLException {
     RdbConfig rdb = config.RdbConfiguration;
 
-    App app = App.getInstance();
-    Map<String, Map<String, Object>> migrationTypeMap =
-        (Map<String, Map<String, Object>>) app.getPropertiesMap().get(ModelDBConstants.MIGRATION);
-    if (migrationTypeMap != null && migrationTypeMap.size() > 0) {
-      new Thread(
```
anandjakhaniya, line 571: this is added because we are running migration on the background to service liveness for the kubernetes so added second thread for that which is not added by you. are you expecting all the migration are run on main thread and do not service up without finish?

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -561,109 +562,37 @@ public static Query getWorkspaceEntityQuery(
...
-              () -> {
-                CommonUtils.registeredBackgroundUtilsCount();
-                int index = 0;
-                try {
-                  CompletableFuture<Boolean>[] completableFutures =
-                      new CompletableFuture[migrationTypeMap.size()];
-                  for (String migrationName : migrationTypeMap.keySet()) {
-                    Map<String, Object> migrationDetailMap = migrationTypeMap.get(migrationName);
-                    if ((boolean) migrationDetailMap.get(ModelDBConstants.ENABLE)) {
```
anandjakhaniya, line 580: this `enable` flag is not available in new config.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
conradoverta: Yes. The migration is now run with the liquibase changes. We shouldn't run the migration in the background to avoid database inconsistencies

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
conradoverta: I'll add it back. Ideally, we only have safe migrations and we control them by code, but that hasn't always been the case.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
anandjakhaniya: I think it is already out side the dataset version migration. i have checked on based PR.

anandjakhaniya (changes requested):
Config changes seems to be wrong format

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
anandjakhaniya: Yes there is some code changes which we have to `enable` and `disable` then it would helps to manage by just config parameter.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
anandjakhaniya: Ok, then it would be fine.

backend/src/main/java/ai/verta/modeldb/config/DatabaseConfig.java
anandjakhaniya: Ok then list is fine for but we need to add `enable: true` in list

anandjakhaniya (approved):

PR #1834, opened 2021-01-01
Improve initialization

conradoverta:


backend/src/main/java/ai/verta/modeldb/App.java
```diff
@@ -174,38 +175,37 @@ public static void main(String[] args) {
...
+
+      // Initialize data access
+      DAOSet daos = DAOSet.fromServices(services);
+
+      // Initialize telemetry
+      initializeTelemetryBasedOnConfig(config);
+
+      // Initialize cron jobs
+      CronJobUtils.initializeBasedOnConfig(
+          config, services.authService, services.roleService, services.artifactStoreService);
```
anandjakhaniya: it should be `initializeCronJobs` instead of `initializeBasedOnConfig`

anandjakhaniya (approved):
Approved after minor fix

backend/src/main/java/ai/verta/modeldb/App.java
conradoverta: Could you suggest the change via github? It's faster to fix that way =) I can open the IDE if too hard

anandjakhaniya:
```suggestion
      CronJobUtils.initializeCronJobs(
          config, services.authService, services.roleService, services.artifactStoreService);
```

anandjakhaniya: Done, Sorry for that.

PR #1835, opened 2021-01-04
Fix test classes and code according to new Config pattern

anandjakhaniya:


backend/src/test/java/ai/verta/modeldb/ArtifactStoreTest.java
```diff
@@ -61,43 +56,22 @@
...
-    App.initializeServicesBaseOnDataBase(
-        serverBuilder, Config.getInstance().test.database, propertiesMap, authService, roleService);
+    Config config = Config.getInstance();
+    // Initialize services that we depend on
+    ServiceSet services = ServiceSet.fromConfig(config);
+    // Initialize data access
+    DAOSet daos = DAOSet.fromServices(services);
+    App.migrate(config);
+
+    App.initializeBackendServices(serverBuilder, services, daos);
```
coryverta: This code seems to be repeated a number of times across the test.  Please consolidate this to a utility method that all the tests can share so it's easier to maintain when we need to make changes.

coryverta (changes requested):
One refactor for cleanup then this is ready for merge.

backend/src/main/java/ai/verta/modeldb/config/Config.java
```diff
@@ -30,6 +31,7 @@
   public TelemetryConfig telemetry;
   public TestConfig test;
   public TrialConfig trial;
+  public List<MigrationConfig> migrations;
```
conradoverta, line 34: I placed this inside the database config, if I remember correctly. Could you check?

conradoverta: Actually, I just saw you removed it. Sure, but I don't see any changes to the migrations file itself for some reason

anandjakhaniya: Yes you have added it in database config but it should be a part of the main config because migration is the separate thing so we had added it in root of config in old config pattern flow. that's why I have moved migration into the root config.

backend/src/test/java/ai/verta/modeldb/ArtifactStoreTest.java
anandjakhaniya: Extracted all the  possible common code into the parent class https://github.com/VertaAI/modeldb/pull/1835/files#diff-a2e1970c71662f894d55dc778bb9128804612527c6534c3c68826c4fcc394828

coryverta (approved):

PR #1838, opened 2021-01-05
Fix find repositories logic using new visibility and workspace

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1720,23 +1720,38 @@ public static Order getOrderBasedOnSortKey(
...
+              if (key.equalsIgnoreCase(ModelDBConstants.OWNER)) {
+                if ((operator.equals(Operator.CONTAIN) || operator.equals(Operator.NOT_CONTAIN))) {
+                  Predicate fuzzySearchPredicate =
+                      getFuzzyUsersQueryPredicate(
+                          authService,
+                          roleService,
+                          builder,
+                          entityRootPath,
+                          predicate,
+                          ModelDBResourceEnum.ModelDBServiceResourceTypes.PROJECT);
```
conradoverta: Project only?

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1720,23 +1720,38 @@ public static Order getOrderBasedOnSortKey(
...
+                  Map<String, UserInfo> userInfoMap =
+                      authService.getUserInfoFromAuthServer(
+                          new HashSet<>(Collections.singleton(ownerId)),
+                          Collections.emptySet(),
+                          Collections.emptyList());
+                  Set<String> resourceIdSet =
+                      getResourceIdsBasedOnUserInfo(
+                          authService,
+                          roleService,
+                          ModelDBResourceEnum.ModelDBServiceResourceTypes.PROJECT,
```
conradoverta: Project only?

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1811,6 +1815,28 @@ private static Predicate getFuzzyUsersQueryPredicate(
     return null;
   }
 
+  public static Set<String> getResourceIdsBasedOnUserInfo(
```
conradoverta:
```suggestion
  public static Set<String> getResourceIdsFromUserWorkspaces(
```
Let's give things very explicit names

anandjakhaniya: Renamed as per the suggestion

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -1426,7 +1426,8 @@ public ExperimentRunPaginationDTO findExperimentRuns(
                 criteriaQuery,
                 experimentRunRoot,
                 authService,
-                roleService);
+                roleService,
+                ModelDBServiceResourceTypes.EXPERIMENT_RUN);
```
conradoverta: Hm. What is this doing? We don't use experiment run as a resource type for most things.

conradoverta (commented):
Let's 1) add the TODO, 2) use the new DTO for pagination.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1720,23 +1722,38 @@ public static Order getOrderBasedOnSortKey(
...
+                          roleService,
+                          builder,
+                          entityRootPath,
+                          predicate,
+                          modelDBServiceResourceTypes);
+                  if (fuzzySearchPredicate != null) {
+                    keyValuePredicates.add(fuzzySearchPredicate);
+                  } else {
+                    throw new ModelDBException(
+                        ModelDBConstants.INTERNAL_MSG_USERS_NOT_FOUND, Code.FAILED_PRECONDITION);
```
conradoverta: Bad request?

conradoverta (commented):
We seem to have a lot of duplicated code in these search functionalities, right?

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: this is used for fuzzy search. the MDB call `getResources` with this type with workspace to getting entity owners for filter.

anandjakhaniya:
> Let's 1) add the TODO, 2) use the new DTO for pagination.

I do not understand this comment.  for which code should i add TODO??

anandjakhaniya:
> We seem to have a lot of duplicated code in these search functionalities, right?

For all entities we have a common Hibernate query code for search functionalities excluding repository. For repository we have a SQL based filtering.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
anandjakhaniya: this is for internal purpose. we are throwing from here and it will return empty response after catch it from service https://github.com/VertaAI/modeldb/blob/40c3035419b03bbda87cea130d4b9fab95f51923/backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java#L1435

conradoverta:
> > Let's 1) add the TODO, 2) use the new DTO for pagination.
> 
> I do not understand this comment. for which code should i add TODO??

I think that was another PR, actually. My bad!

conradoverta:
> > We seem to have a lot of duplicated code in these search functionalities, right?
> 
> For all entities we have a common Hibernate query code for search functionalities excluding repository. For repository we have a SQL based filtering.

Why? Just trying to follow because it seems like it's giving very similar functionality.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
conradoverta: Oh man, this looks very brittle. We can fix it later.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
conradoverta: As discussed, we cannot call getResources with those types. We should leave the logic intact.

Also, this PR is called "Fix find repositories logic using new visibility and workspace", so this change is clearly unrelated because it's changing the behavior for experiment run. Please don't mix multiple changes together in a single PR.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: @conradoverta i have did it on separate PR https://github.com/VertaAI/modeldb/pull/1839 but after merging PR in this that's why it is here.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
anandjakhaniya: Ok.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
anandjakhaniya: Removed unrelated changes to repository entity

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1774,34 +1773,22 @@ public static Order getOrderBasedOnSortKey(
...
-
-        Expression<String> exp = entityRootPath.get(ModelDBConstants.ID);
+        Expression<String> exp = entityRootPath.get(requestedPredicate.getKey());
+        List<String> vertaIds =
+            userInfoList.stream()
+                .map(authService::getVertaIdFromUserInfo)
+                .collect(Collectors.toList());
         if (operator.equals(Operator.NOT_CONTAIN) || operator.equals(Operator.NE)) {
-          return builder.not(exp.in(projectIdSet));
+          return builder.not(exp.in(vertaIds));
```
conradoverta, line 1801: `vertaIds` is user ids, right? I'm confused on why we're adding it to the query since MDB doesn't have information about owner anymore 🤔  before we were placing the projects ids, which seems correct to me. Could you clarify?

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -1140,38 +1137,34 @@ public void deleteBranchByCommit(Session session, Long repoId, String commitHash
...
-                    .collect(Collectors.toList()));
+        Set<String> accessibleResourceIds =
+            new HashSet<>(
+                roleService.getAccessibleResourceIds(
+                    null,
+                    new CollaboratorUser(authService, currentLoginUserInfo),
+                    ModelDBServiceResourceTypes.REPOSITORY,
+                    request.getRepoIdsList().stream()
+                        .map(String::valueOf)
+                        .collect(Collectors.toList())));
```
conradoverta, line 1158:
```suggestion
                        .collect(Collectors.toSet())));
```
?

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -1140,38 +1137,34 @@ public void deleteBranchByCommit(Session session, Long repoId, String commitHash
...
-        WorkspaceDTO workspaceDTO = null;
-        List<String> accessibleResourceIds =
-            roleService.getAccessibleResourceIds(
-                null,
-                new CollaboratorUser(authService, currentLoginUserInfo),
-                ModelDBServiceResourceTypes.REPOSITORY,
-                request.getRepoIdsList().stream()
-                    .map(String::valueOf)
-                    .collect(Collectors.toList()));
+        Set<String> accessibleResourceIds =
```
conradoverta: Are these Collaborator or Authorization level resources? I can't tell fromt eh name

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -1140,38 +1137,34 @@ public void deleteBranchByCommit(Session session, Long repoId, String commitHash
...
+              accessibleAllWorkspaceItems.stream()
+                  .map(GetResourcesResponseItem::getResourceId)
+                  .collect(Collectors.toSet());
         } else {
-          LOGGER.debug("Workspace not found or not match with current login username");
-          workspaceDTO =
-              roleService.getWorkspaceDTOByWorkspaceName(
-                  currentLoginUserInfo, request.getWorkspaceName());
+          accessibleResourceIds =
+              ModelDBUtils.filterWorkspaceOnlyAccessibleIds(
```
conradoverta: This should be directly available from CollaboratorService. No need to filter it down (i.e. overfetch and then remove some).

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -1140,38 +1137,34 @@ public void deleteBranchByCommit(Session session, Long repoId, String commitHash
...
+                        .collect(Collectors.toList())));
 
         String workspaceName = request.getWorkspaceName();
-        if (workspaceName != null
-            && !workspaceName.isEmpty()
+        if (!workspaceName.isEmpty()
             && workspaceName.equals(authService.getUsernameFromUserInfo(currentLoginUserInfo))) {
-          LOGGER.debug("Workspace found and match with current login username");
+          List<GetResourcesResponseItem> accessibleAllWorkspaceItems =
+              roleService.getResourceItems(
```
conradoverta, line 1164: What's the difference between this and the other getResource methods?

conradoverta (commented):
I suggest we start commenting the code with what different blocks are doing since it's not obvious from reading, specially around the getResources area. We have 3 methods that I'm not sure what relationship they have with each other and I'm fairly confident we're fetching more data than we need to.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
anandjakhaniya: This logic is for the other entities excluding project, repository logic where still owner is at the MDB. above removed code with projectIds is added with this logic in subsequence PR. where added two function one for old logic with owner Id for other entities excluding project, repository

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: No, this is the argument of function which except the `List<String>` and i have `request.getRepoIdsList()` which has `List<Long>` so i am converting it to string using stream here.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: it is accessible resources with Collaborator. i can update the name.

anandjakhaniya: Rename it to `accessibleResourceIdsWithCollaborator`

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: No, from `ModelDBUtils.filterWorkspaceOnlyAccessibleIds(` method i am calling the UAC with accessibleResourceIdsWithCollaborator and workspace so the UAC return ids with filtering only my workspace ids.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: `getResourceItems` method is the main roleService method at MDB which actually call to the UAC using endpoint `getResources`

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: internally it is used  `roleService.getResourceItems` with the workspace.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
conradoverta: Could we have that as a comment?

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
conradoverta: Oh, a comment about this anywhere nearby would be super useful. Even better if we give the method a meaningful name that highlights that.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
conradoverta: Nevermind. I didn't noticed this was for an argument. I thought it was for the return of `getAccessibleResourceIds`.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
conradoverta: I believe you are overfetching since you called `getAccessibleResourceIds`, which returns everything regardless of what workspace they belong to, and then call `filterWorkspaceOnlyAccessibleIds` which is filtering, as the name suggests. What am I missing here?

anandjakhaniya: Thread: https://vertaai.slack.com/archives/C01JVHUHKR6/p1610415593088700

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
anandjakhaniya: Added comment on the method

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: Added comment on the method

conradoverta (approved):
We can optimize the ids later

PR #1843, opened 2021-01-06
Consolidate client tests' "created entities" fixtures

convoliution:
The client tests currently have three distinct fixtures to collect and clean up datasets, registered models, endpoints.
This used to be necessary because deletion used to require directly making DELETE requests.

Now that MDB and registry entities all expose a `delete()` method, these fixtures can be combined into a single `created_entities()`

### DIFF
--- a/client/verta/tests/conftest.py
+++ b/client/verta/tests/conftest.py
@@ -364,7 +364,7 @@ def registered_model(client):
 
 @pytest.fixture
 def created_entities():
-    """Container to track and clean up `RegisteredModel`s created during tests."""
+    """Container to track and clean up ModelDB, Registry, etc. entities created during tests."""
     to_delete = []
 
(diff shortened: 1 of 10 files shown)


client/verta/tests/conftest.py
```diff
@@ -374,14 +363,14 @@ def registered_model(client):
...
+def created_entities():
     """Container to track and clean up `RegisteredModel`s created during tests."""
     to_delete = []
 
     yield to_delete
 
-    for registered_model in to_delete:
-        registered_model.delete()
+    for entity in to_delete:
+        entity.delete()
```
convoliution: This is basically the gist of the entire PR

conradoverta (commented):
APIs FTW

conradoverta (approved):

PR #1845, opened 2021-01-07
Removed Repository owner role binding migration

anandjakhaniya:


conradoverta:
Why should we do this right now? Let's leave cleaning the migrations to later as they should be pretty easy to do. We need to do at least one OSS release before removing them.

coryverta (approved):

conradoverta:
@ad-47 should we close this to reduce the number of open PRs? We can remove later.

anandjakhaniya:
@jkwatson-verta , @conradoverta Can we merge this as a part of the cleanup? it is open for a long time.

jkwatson-verta (approved):

PR #1847, opened 2021-01-07
Improve constructor parameters

anandjakhaniya:


conradoverta (approved):
This is great! You have read my mind as this was the next step. Every time we have to extend the arguments and propagate across gave me pain =)

PR #1850, opened 2021-01-08
Fix fuzzy search logic for all entities and bug for test cases

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1771,6 +1786,77 @@ public static Order getOrderBasedOnSortKey(
     return finalPredicatesList;
   }
 
+  private static void setFuzzyOwnerPredicateBasedOnUACWorkspace(
```
conradoverta, line 1789: I am confused. I'd assume this code already existed before and you're just moving, but it seems to have a lot of new logic added to it. Am I missing something here?

anandjakhaniya: https://github.com/VertaAI/modeldb/pull/1838/files#r554756537
this method is for Project and repository based on UAC workspace which returns me resourcesIds.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
```diff
@@ -1771,6 +1786,77 @@ public static Order getOrderBasedOnSortKey(
...
+              authService,
+              roleService,
+              modelDBServiceResourceTypes,
+              Collections.singletonList(userInfoMap.get(ownerId)));
+      Expression<String> exp = entityRootPath.get(ModelDBConstants.ID);
+      keyValuePredicates.add(exp.in(resourceIdSet));
+    }
+  }
+
+  private static void setFuzzySearchPredicateBasedOnMDBOwner(
```
anandjakhaniya, line 1831: https://github.com/VertaAI/modeldb/pull/1838/files#r554756537
This method used for other entities excluding project and repository which still used the owner column at the MDB.

backend/src/main/java/ai/verta/modeldb/utils/RdbmsUtils.java
conradoverta: Correct, but I don't understand how _both_ paths setFuzzyOwnerPredicateBasedOnUACWorkspace and setFuzzySearchPredicateBasedOnMDBOwner are new, since they're considerably bigger than any other code being deleted in this PR, given that this predicate construction existed before. Are we forgetting to delete some code?

conradoverta (approved):

PR #1852, opened 2021-01-08
Fix migration code for created=0

anandjakhaniya:


conradoverta (approved):
Nice and short PR. I like it!

PR #1856, opened 2021-01-11
Fix health check and added error log

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/health/HealthServiceImpl.java
```diff
@@ -105,7 +109,7 @@ void clearStatus(String service) {
       HealthCheckRequest request = HealthCheckRequest.newBuilder().setService(service).build();
       ServingStatus status = getServingStatus(request);
       if (status == null) {
-        LOGGER.trace("Spring custom health check returned with unknown service");
+        LOGGER.error("Spring custom health check returned with unknown service");
         throw new ResponseStatusException(
```
anandjakhaniya, line 116:
```
 public HealthServiceImpl() {
    setStatus("", HealthCheckResponse.ServingStatus.SERVING);
  }
```
was not initialize so `ServingStatus status = getServingStatus(request);` was return the null so from here it was throwing 500.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBHibernateUtil.java
```diff
@@ -434,10 +434,15 @@ public static boolean ping() {
...
+      if (isReady && ping()) {
+        return HealthCheckResponse.ServingStatus.SERVING;
+      }
+      return HealthCheckResponse.ServingStatus.NOT_SERVING;
+    } catch (Exception ex) {
+      LOGGER.error("Getting error on health checkReady: " + ex.getMessage(), ex);
+      return HealthCheckResponse.ServingStatus.NOT_SERVING;
     }
-    return HealthCheckResponse.ServingStatus.NOT_SERVING;
   }
```
anandjakhaniya, line 446: Added try catch to just identify log message

CaptEmulation (approved):
yes please

coryverta (approved):

backend/src/main/java/ai/verta/modeldb/health/HealthServiceImpl.java
```diff
@@ -52,6 +52,10 @@
   private final Map<String, ServingStatus> statusMap = new ConcurrentHashMap<>();
   private static final Logger LOGGER = LogManager.getLogger(HealthServiceImpl.class);
 
+  public HealthServiceImpl() {
+    setStatus("", HealthCheckResponse.ServingStatus.SERVING);
```
conradoverta: Will this automatically set the service as running even when it hasn't finished initialization? We have
```
healthStatusManager.setStatus("", HealthCheckResponse.ServingStatus.SERVING);
```
in the main already, so is this a problem of healthStatusManager not propagating to HealthServiceImpl? I want to make sure we're solving the base problem here as sometimes we'll set it as not serving (when there's an actual failure) and we need to make sure the system behaves appropriately in such scenarios.

conradoverta (changes requested):
Marking as request changes as precaution to avoid merging before the concern is considered.

conradoverta:
After some thought, I think the error log will be the main thing here. The container did come up well and worked for at least 30min before and then it stopped. I'm not sure if the initialization change is necessary (not a big deal), but the log is definitely missing.

backend/src/main/java/ai/verta/modeldb/health/HealthServiceImpl.java
anandjakhaniya: > Will this automatically set the service as running even when it hasn't finished initialization?

Yes, but it is never excess before it is up because spring boot do not allowed client request before it is completely start.

Before Main method restructured, we initialized `healthStatusManager.setStatus("", HealthCheckResponse.ServingStatus.SERVING);` in main method for rest endpoint as well as GRPC health check also and then after that same reference of  `healthStatusManager` is used while spring boot server will run. but now the spring boot server started before its initialization in main so we should have to initialize it again from its constructor.

I am sure it will not creating cause when actually backend is not in serving state.

anandjakhaniya:
> After some thought, I think the error log will be the main thing here. The container did come up well and worked for at least 30min before and then it stopped. I'm not sure if the initialization change is necessary (not a big deal), but the log is definitely missing.

It was printing TRACE from https://github.com/VertaAI/modeldb/pull/1856#discussion_r554751816 that why i have replace it with ERROR now.

conradoverta:
Holding this off until we have verification that we are not saying that the system is ready until all of its components (spring, grpc and whatever else needs to be initialize) are classified as ready.

backend/src/main/java/ai/verta/modeldb/ServiceSet.java
```diff
@@ -20,12 +20,16 @@
   public ArtifactStoreService artifactStoreService = null;
   public AuthService authService;
   public RoleService roleService;
+  public App app;
```
conradoverta, line 23: This is not used except to be set once.

backend/src/main/java/ai/verta/modeldb/ServiceSet.java
```diff
@@ -20,12 +20,16 @@
...
 
   public static ServiceSet fromConfig(Config config) throws IOException {
     ServiceSet set = new ServiceSet();
     set.authService = AuthServiceUtils.FromConfig(config);
     set.roleService = RoleServiceUtils.FromConfig(config, set.authService);
 
+    // Initialize App.java singleton instance
+    set.app = App.getInstance();
```
conradoverta, line 31: I might be missing something, but why do we need to do this dummy initialization?

anandjakhaniya: it was already there when we are initializing artifact store https://github.com/VertaAI/modeldb/pull/1856/files#diff-1dbb0555b7d8fbec12866ae57bb60e96c5a457126dab4cb3445c620450b66e9eL41 but when artifact store is disabled the `App.java` main class was not initialized and it will broken while starting spring boot server from main class https://github.com/VertaAI/modeldb/blob/52fc539481559a541289cbe3d247859e27401312/backend/src/main/java/ai/verta/modeldb/App.java#L94. So i have just moved that line to up to initialize Main class.

conradoverta: Let's add a comment whenever we have a non-obvious reason to do something (e.g. it will trigger a side effect).

conradoverta (approved):

PR #1857, opened 2021-01-11
Added created=true check on the fetch calls

anandjakhaniya:


conradoverta (approved):
We should have a consistent way to ensure this is always correct.

PR #1865, opened 2021-01-13
Send auth headers without underscore

convoliution:
[NGINX silently drops HTTP headers with underscores in them by default](https://www.nginx.com/nginx-wiki/build/dirhtml/start/topics/tutorials/config_pitfalls/#missing-disappearing-http-headers), meaning that an infrastructure setup using NGINX as an intermediary may drop our platform's authentication headers.

...until now.

### DIFF
--- a/client/verta/verta/deployment/_deployedmodel.py
+++ b/client/verta/verta/deployment/_deployedmodel.py
@@ -75,7 +75,10 @@ class DeployedModel:
             except KeyError:
                 six.raise_from(EnvironmentError("${} not found in environment".format('VERTA_EMAIL')), None)
             try:
-                self._session.headers.update({_utils._GRPC_PREFIX+'developer_key': os.environ['VERTA_DEV_KEY']})
+                self._session.headers.update({
+                    _utils._GRPC_PREFIX+'developer_key': os.environ['VERTA_DEV_KEY'],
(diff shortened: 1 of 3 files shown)


conradoverta (approved):

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -75,7 +75,10 @@ def __init__(self, _host=None, _run_id=None, _from_url=False, **kwargs):
             except KeyError:
                 six.raise_from(EnvironmentError("${} not found in environment".format('VERTA_EMAIL')), None)
             try:
-                self._session.headers.update({_utils._GRPC_PREFIX+'developer_key': os.environ['VERTA_DEV_KEY']})
+                self._session.headers.update({
+                    _utils._GRPC_PREFIX+'developer_key': os.environ['VERTA_DEV_KEY'],
+                    _utils._GRPC_PREFIX+'developer-key': os.environ['VERTA_DEV_KEY'],  # see Client.__init__()
+                })
```
convoliution, line 81: Note that this code path in `DeployedModel` only runs if a user instantiates the object directly.
It is _not_ executed when using our `get_deployed_model()` methods, which has been preferred for a while.

PR #1870, opened 2021-01-14
VR-8155 delete bindings for projects

evgeniilezhnin:

PR #1874, opened 2021-01-15
VR-8155 delete bindings for repositories during migration

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -297,6 +301,7 @@ private static void migrateRepositories() {
               }
             }
           }
+          deleteRoleBindingsOfRepositories(migratedRepositoryEntities);
```
conradoverta: This needs to be inside the migration step. We cannot set `repository.setVisibility_migration(true);` and then have this. It's possible to have a failure between these steps and the role binding will never be removed, leading to inconsistencies. We need to make our migrations safe.

conradoverta (changes requested):
We also need to do this for projects, right?

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
evgeniilezhnin: updated

evgeniilezhnin:
> We also need to do this for projects, right?

It was done here: https://github.com/VertaAI/modeldb/pull/1870/files

Here is a fix after your recent comment: https://github.com/VertaAI/modeldb/pull/1876

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
conradoverta: This is not done before `repository.setVisibility_migration(true);`. The code has the same issue as before: I can mark the migration as completed before it's completed. We can only set the migration as done if every single operation we'll do for that given element is done and successful. If they are not, we *cannot* set the migration as done.

evgeniilezhnin: Sorry, my bad. Updated again

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
conradoverta: Thanks! It's a small change, but it's good for us to develop this muscle memory.

PR #1877, opened 2021-01-18
Fix Internal error with invalid_argument error

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-9582


CaptEmulation (approved):

conradoverta (approved):
Can we log the type that is unknown? Let's make sure we do that for all types that have enum. Otherwise the logs are not very meaningful.

anandjakhaniya:
> Can we log the type that is unknown? Let's make sure we do that for all types that have enum. Otherwise the logs are not very meaningful.

added  blob type name in error message

conradoverta:
Let's make sure we do this (error code and log the value) for all of these oneof instead of just this particular one or we'll be back here pretty soon

PR #1885, opened 2021-01-20
Migration of dataset from REPOSITORY resource type into DATASET resource type

anandjakhaniya:


coryverta (approved):

conradoverta:
I think this migration is unnecessary since we have not done any migration for the permission effort yet in any namespace that matters.

PR #1889, opened 2021-01-21
VR-9627: Use new visibility values during entity creation

convoliution:


convoliution:
Rebased to https://github.com/VertaAI/modeldb/pull/1896

PR #1890, opened 2021-01-21
Use hibernate as the cache for objects (#1886)

conradoverta:
* Use hibernate as the cache for objects

* https://jenkins.dev.verta.ai/job/build/job/autoformat/job/modeldb-backend/583/


conradoverta:
Against ref this time

PR #1891, opened 2021-01-21
Fix fetch repositories in personal workspace excluding organization repositories

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-9646

PR #1892, opened 2021-01-21
Fix get repository by name while same name exists in organization and personal workspace

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-9639

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
@@ -350,22 +350,32 @@ public class RepositoryDAORdbImpl implements RepositoryDAO {
 
   private Optional<RepositoryEntity> getRepositoryByName(
       Session session, String name, Workspace workspace, RepositoryTypeEnum repositoryType) {
-    List<Long> repositoryIds = getRepositoryEntityIdsByName(session, name, repositoryType);
+    ModelDBServiceResourceTypes modelDBServiceResourceTypes =
+        ModelDBServiceResourceTypes.REPOSITORY;
+    if (repositoryType.equals(RepositoryTypeEnum.DATASET)) {
+      modelDBServiceResourceTypes = ModelDBServiceResourceTypes.DATASET;
+    }
     List<GetResourcesResponseItem> accessibleAllWorkspaceItems =
         roleService.getResourceItems(
-            workspace,
-            !repositoryIds.isEmpty()
-                ? repositoryIds.stream().map(String::valueOf).collect(Collectors.toSet())
-                : Collections.emptySet(),
-            ModelDBServiceResourceTypes.REPOSITORY);
-    Optional<Long> repoId =
+            workspace, Collections.emptySet(), modelDBServiceResourceTypes);
+    Set<Long> repoIds =
         accessibleAllWorkspaceItems.stream()
+            .filter(
+                getResourcesResponseItem -> getResourcesResponseItem.getResourceName().equals(name))
             .map(
                 getResourcesResponseItem ->
                     Long.parseLong(getResourcesResponseItem.getResourceId()))
-            .findFirst();
-    if (repoId.isPresent()) {
-      return getRepositoryById(session, repoId.get());
+            .collect(Collectors.toSet());
+    if (!repoIds.isEmpty()) {
+      if (repoIds.size() > 1) {
+        throw new InternalErrorException(
+            "Multiple "
+                + modelDBServiceResourceTypes
+                + " found for the name '"
+                + name
+                + "' in the workspace");
+      }
+      return getRepositoryById(session, new ArrayList<>(repoIds).get(0));
     } else {
       return Optional.empty();
     }


evgeniilezhnin (approved):

evgeniilezhnin:
tested this case https://vertaai.slack.com/archives/C01JVHUHKR6/p1611155203094300?thread_ts=1611122448.087300&cid=C01JVHUHKR6 locally

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -350,22 +350,32 @@ public RepositoryEntity getProtectedRepositoryById(
 
   private Optional<RepositoryEntity> getRepositoryByName(
       Session session, String name, Workspace workspace, RepositoryTypeEnum repositoryType) {
-    List<Long> repositoryIds = getRepositoryEntityIdsByName(session, name, repositoryType);
+    ModelDBServiceResourceTypes modelDBServiceResourceTypes =
+        ModelDBServiceResourceTypes.REPOSITORY;
+    if (repositoryType.equals(RepositoryTypeEnum.DATASET)) {
+      modelDBServiceResourceTypes = ModelDBServiceResourceTypes.DATASET;
+    }
```
coryverta: Please extract this to a utility method.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -350,22 +350,32 @@ public RepositoryEntity getProtectedRepositoryById(
...
+    if (!repoIds.isEmpty()) {
+      if (repoIds.size() > 1) {
+        throw new InternalErrorException(
+            "Multiple "
+                + modelDBServiceResourceTypes
+                + " found for the name '"
+                + name
+                + "' in the workspace");
+      }
+      return getRepositoryById(session, new ArrayList<>(repoIds).get(0));
```
coryverta, line 411: Is creating a new list with `new ArrayList<>(repoIds)` necessary?

conradoverta:
>Fix get repository by name while same name exists in organization and personal workspace

Note that this isn't the right description to the issue I found, as it listed repositories that don't have equivalent names in my personal workspace. Just checking to make sure we're actually solving the problem.

Edit: I just realized this is a fix to a thing that was already worked around and not to the new one that was fixed in https://github.com/VertaAI/modeldb/pull/1891. Ignore my comment above.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
conradoverta: I already asked for this in a different pr =)

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -350,22 +350,32 @@ public RepositoryEntity getProtectedRepositoryById(
...
+    }
     List<GetResourcesResponseItem> accessibleAllWorkspaceItems =
         roleService.getResourceItems(
-            workspace,
-            !repositoryIds.isEmpty()
-                ? repositoryIds.stream().map(String::valueOf).collect(Collectors.toSet())
-                : Collections.emptySet(),
-            ModelDBServiceResourceTypes.REPOSITORY);
-    Optional<Long> repoId =
+            workspace, Collections.emptySet(), modelDBServiceResourceTypes);
```
conradoverta: Why do we list everything if we're getting a single element? It's wasteful. Don't we have the option to fetch by name? Or even fetching by ids subset like before.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -350,22 +350,32 @@ public RepositoryEntity getProtectedRepositoryById(
...
+    Set<Long> repoIds =
         accessibleAllWorkspaceItems.stream()
+            .filter(
+                getResourcesResponseItem -> getResourcesResponseItem.getResourceName().equals(name))
```
conradoverta, line 397: If you do an RPC and then do a filter, it's wasteful. Can the filter be pushed to the remote system?

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: `repoIds` is a Set so i have to get data from it which exists on 0 index so I used `ArrayList`. it is necessary.

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: I have implemented overfetch because `getResource` with name is not implemented OR there is not have any other option to fetch resources based on the `name`.  Could you mine to implement it at UAC?

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
anandjakhaniya: Yes but still remote system (UAC) do not have any endpoint which is give me a filter data based on the name.

conradoverta (approved):

PR #1895, opened 2021-01-21
Constrain pyyaml to <5.4

convoliution:
Since https://github.com/yaml/pyyaml/pull/407, `pyyaml` has `Cython` (and therefore gcc) as a build dependency.

We'd like to make our client fully build-from-source-able on systems without gcc, and `pyyaml` is only being used for simple config file reading, so we're constraining the version.

### DIFF
--- a/client/verta/verta/client.py
+++ b/client/verta/verta/client.py
@@ -19,7 +19,6 @@ import warnings
 import zipfile
 
 import requests
-import yaml
 from verta._tracking.organization import Organization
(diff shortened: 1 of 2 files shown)


client/verta/verta/client.py
```diff
@@ -19,7 +19,6 @@
 import zipfile
 
 import requests
-import yaml
```
convoliution, line 22: Clean up: not being used in client.py

conradoverta (approved):

PR #1898, opened 2021-01-22
Fix find datasets and find projects

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-9669

**Issue:** While fetching entities we fetched all workspace datasets excluding organization  and return to the WebApp which is wrong. 

**Solution:** we should have to fetch all entities with collaborators and then exclude the organization entities and return to the WebApp is the right way like we have did it for other entities

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
@@ -919,18 +919,21 @@ public class ProjectDAORdbImpl implements ProjectDAO {
       projectRoot.alias("pr");
       List<Predicate> finalPredicatesList = new ArrayList<>();
 
-      Set<String> accessibleProjectIds = new HashSet<>();
+      Set<String> accessibleResourceIdsWithCollaborator =
+          new HashSet<>(
+              roleService.getAccessibleResourceIds(
+                  null,
+                  new CollaboratorUser(authService, currentLoginUserInfo),
+                  ModelDBServiceResourceTypes.PROJECT,
+                  queryParameters.getProjectIdsList()));
+
       String workspaceName = queryParameters.getWorkspaceName();
       if (!workspaceName.isEmpty()
           && workspaceName.equals(authService.getUsernameFromUserInfo(currentLoginUserInfo))) {
         List<GetResourcesResponseItem> accessibleAllWorkspaceItems =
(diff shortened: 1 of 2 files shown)


backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -919,18 +919,21 @@ public ProjectPaginationDTO findProjects(
       projectRoot.alias("pr");
       List<Predicate> finalPredicatesList = new ArrayList<>();
 
-      Set<String> accessibleProjectIds = new HashSet<>();
+      Set<String> accessibleResourceIdsWithCollaborator =
```
conradoverta, line 922: I'm assuming owner counts as collaborator, right?

conradoverta (commented):
I only see a fix for dataset. Don't we need one for repository?

anandjakhaniya:
> I only see a fix for dataset. Don't we need one for repository?

For Repository it is already there. I have seen all the entities but just two dataset and project are left so it is done here in this PR.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: I do not understand your question.  `accessibleResourceIdsWithCollaborator ` is a list of  the accessible resources ids with collaborators.

conradoverta: If I own a project, will the project be included in the list?

anandjakhaniya: Yes. it will add it. that list has my own project ids as well as those ids where i am a collaborator

conradoverta: The name is super confusing to me. Because `accessibleResourceIdsWithCollaborator` reads as "accessible ids of resources with collaborators", which I don't think it what it's doing from the description (does it have to have a collaborator?)

conradoverta:
@ad-47 is this still an issue?

anandjakhaniya:
This is too old PR and we have a much more improved implementation. Now, this does not make sense so closing this.

PR #1900, opened 2021-01-23
VR-9680 [1/2]: Add is_dirty param to Git()

convoliution:


client/verta/verta/code/_git.py
```diff
@@ -58,7 +61,7 @@ class Git(_code._Code):
         )
 
     """
-    def __init__(self, repo_url=None, branch=None, tag=None, commit_hash=None, autocapture=True, _autocapture=True):
+    def __init__(self, repo_url=None, branch=None, tag=None, commit_hash=None, is_dirty=None, autocapture=True, _autocapture=True):
```
convoliution, line 64: This doesn't break backwards compatibility, so long as it's merged before the next client release (since `autocapture` was only added earlier today)

conradoverta (approved):

PR #1901, opened 2021-01-25
VR-9680 [2/2]: Add is_dirty param to log_code()

convoliution:
Follow-up to https://github.com/VertaAI/modeldb/pull/1900

### DIFF
--- a/client/verta/verta/_tracking/entity.py
+++ b/client/verta/verta/_tracking/entity.py
@@ -316,7 +316,9 @@ class _ModelDBEntity(object):
             )
             msg.code_version.git_snapshot.repo = code_ver.repo_url or ""
             msg.code_version.git_snapshot.hash = code_ver.commit_hash or ""
-            if code_ver.is_dirty:
+            if not autocapture and is_dirty is None:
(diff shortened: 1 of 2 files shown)


client/verta/verta/_tracking/entity.py
```diff
@@ -164,7 +164,7 @@ def _create_proto(cls, conn, *args, **kwargs):
     def _create_proto_internal(cls, conn, ctx, name, desc=None, tags=None, attrs=None, date_created=None, **kwargs):  # recommended params
         raise NotImplementedError
 
-    def log_code(self, exec_path=None, repo_url=None, commit_hash=None, overwrite=False, autocapture=True):
+    def log_code(self, exec_path=None, repo_url=None, commit_hash=None, overwrite=False, is_dirty=None, autocapture=True):
```
convoliution, line 167: I'm not sure about this positioning, but I think it's the best option right now.

1. It's after after `overwrite` to maintain API backwards compatibility (thanks Python 2, for not having keyword-only args).
2. It's before `autocapture` because I want "informational" parameters like `is_dirty` to be tend left, and "behavioral" parameters like `autocapture` to tend right.

conradoverta (approved):

PR #1896, opened 2021-01-21
VR-9627: Use new visibility values during entity creation

convoliution:
On top of https://github.com/VertaAI/modeldb/pull/1888

Remaining items:

- [ ] ~implement for endpoint~ –> moved to a separate ticket
- [x] finish writing tests

PR #1904, opened 2021-01-27
VR-9692: Use new visibility values during endpoint creation

convoliution:
Split from https://github.com/VertaAI/modeldb/pull/1896

### DIFF
--- a/client/verta/verta/endpoint/_endpoint.py
+++ b/client/verta/verta/endpoint/_endpoint.py
@@ -88,8 +88,8 @@ class Endpoint(object):
         return data['date_updated']
 
     @classmethod
-    def _create(cls, conn, conf, workspace, path, description=None, public_within_org=None):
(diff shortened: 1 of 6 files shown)


client/verta/verta/endpoint/_endpoint.py
```diff
@@ -88,8 +89,8 @@ def _date_updated(self, data):
         return data['date_updated']
 
     @classmethod
-    def _create(cls, conn, conf, workspace, public_within_org, path, description=None):
-        endpoint_json = cls._create_json(conn, workspace, public_within_org, path, description)
+    def _create(cls, conn, conf, workspace, path, description=None, public_within_org=None, visibility=None):
+        endpoint_json = cls._create_json(conn, workspace, path, description, public_within_org, visibility)
```
convoliution, line 93: Reordering the params so they make more sense. I double-checked that it's updated everywhere this is called (which isn't that many places)

conradoverta (approved):

PR #1905, opened 2021-01-27
Fix Repository for visibility tests

convoliution:
`Repository` still needs a full refactor to subclass `_ModelDBEntity` (VR-5542), which means it doesn't have `._msg`

This PR impls `_get_proto_by_id()` and `_get_proto_by_name()` so that the new tests have a way of checking the visibility value.

Versioning/repository tests still pass.

### DIFF
--- a/client/verta/verta/_repository/repository.py
+++ b/client/verta/verta/_repository/repository.py
@@ -5,6 +5,7 @@ from __future__ import print_function
 from .._protos.public.modeldb.versioning import VersioningService_pb2 as _VersioningService
 
 from .._internal_utils import _utils
+from ..visibility import _visibility
 from . import commit
 
 import requests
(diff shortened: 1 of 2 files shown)


client/verta/verta/_repository/repository.py
```diff
@@ -56,6 +57,8 @@ def workspace(self):
 
     @classmethod
     def _create(cls, conn, name, workspace, public_within_org, visibility):
+        visibility, public_within_org = _visibility._Visibility.translate_public_within_org(visibility, public_within_org)
```
convoliution, line 60: https://github.com/VertaAI/modeldb/blob/2582b52/client/verta/verta/visibility/_visibility.py#L33-L56

conradoverta (approved):
Yay cleanup

PR #1907, opened 2021-01-27
500 when commits have the same hash

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-9637


coryverta (approved):
Just a note: there are some other options we can explore besides using sessions.contains if the problem continues to occur in the future.

anandjakhaniya:
> Just a note: there are some other options we can explore besides using sessions.contains if the problem continues to occur in the future.

The problem is not resolved using `sessions.contains` so i have implemented commit hash lock mechanism.

conradoverta:
Let's verify the behavior before closing the ticket.

PR #1908, opened 2021-01-27
Fix missing user during migration

conradoverta:
Needs to be repeated for non-project things

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
+++ b/backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
@@ -34,6 +34,7 @@ import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.Root;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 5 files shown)


backend/src/main/java/ai/verta/modeldb/batchProcess/CollaboratorResourceMigration.java
```diff
@@ -124,11 +127,20 @@ private static void migrateProjects() {
...
+              try {
+                workspaceDTO =
+                    roleService.getWorkspaceDTOByWorkspaceId(
+                        userInfoMap.get(project.getOwner()),
+                        project.getWorkspace(),
+                        project.getWorkspace_type());
+              } catch (StatusRuntimeException ex) {
+                if (ex.getStatus().getCode() == Status.Code.NOT_FOUND) {
+                  LOGGER.warn("Failed to get workspace (skipping) : " + ex.toString());
+                  continue;
```
conradoverta: Need to think if we should mark this as migrated to avoid trying to migrate all over again.

conradoverta: We should probably tag this as deleted.

PR #1914, opened 2021-01-28
VR-9624: Consolidate get_workspace_name_by_id() into conn

convoliution:


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -198,7 +224,7 @@ class LazyList(object):
     _VALID_QUERY_KEYS = None  # NOTE: must be overridden by subclasses
 
     def __init__(self, conn, conf, msg):
-        self._conn = conn
+        self = conn
```
convoliution: oops
```suggestion
        self._conn = conn
```

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -155,6 +158,29 @@ def _request_to_curl(request):
 
         print(curl)
 
+    def get_workspace_name_from_legacy_id(self, workspace_id):
```
conradoverta, line 161: I _think_ this might not be necessary anymore as things now return the new id too. Need to double check and not a blocker

conradoverta (approved):

client/verta/verta/_internal_utils/_utils.py
convoliution: I'm erring on the side of keeping it, in the interest of "the logic ain't broke", but could be revisited and cleaned up in the future

PR #1928, opened 2021-02-01
Extracted common HibernateUtils logic to CommonHibernateUtil

anandjakhaniya:


coryverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/CommonConstants.java
```diff
@@ -10,6 +10,10 @@
   Integer MAX_CONNECTION_SIZE_DEFAULT = 20;
   Integer CONNECTION_TIMEOUT_DEFAULT = 300;
   String INTERNAL_ERROR = "Internal server error";
+  // Set to true to export the liquibase schema as sql statements
+  Boolean EXPORT_SCHEMA = false;
+  String userDir = "user.dir";
+  String DELETED = "deleted";
```
conradoverta, line 16: Can we start splitting these generic "common constants" to more meaningful locations instead of being a centralization of completely independent pieces? Probably close to where they are used would make things easier.

conradoverta (approved):
One suggestion.

backend/common/src/main/java/ai/verta/modeldb/common/CommonConstants.java
anandjakhaniya: Sure. will do it in other PR.

PR #1933, opened 2021-02-03
Fix GetLatestDatasetVersionByDatasetId endpoint with/without workspace

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-9779, https://vertaai.atlassian.net/browse/VR-9778


conradoverta (commented):
Are we sure this doesn't have unintended consequences?

anandjakhaniya:
> Are we sure this doesn't have unintended consequences?

Yes, I have covered following scenarios
- with workspace: Returns only the workspace accessible ids
- without workspace: All workspace accessible ids
- with requested ids with workspace: Returns only those requested ids which is accessible to his/her workspace
- with requested ids without workspace: Returns those requested ids which is accessible from all his/her workspaces

conradoverta (approved):

PR #1935, opened 2021-02-04
VR-9734: Enable passing PySpark model to run.log_model()

convoliution:
It works for `model_version.log_model()` too!
But _not_ for `log_artifact()` (the specialized model-object-handling logic is only in the `log_model()` path).

### DIFF
--- a/client/verta/tests/test_artifacts.py
+++ b/client/verta/tests/test_artifacts.py
@@ -434,6 +434,37 @@ class TestModels:
         assert experiment_run.get_model().__dict__ == custom.__dict__
         assert experiment_run.get_model().predict(strs) == custom.predict(strs)
 
+    def test_pyspark(self, experiment_run, in_tempdir):
+        data_filename = "census-train.csv"
+        spark_model_dir = "spark-model"
+
+        pytest.importorskip("boto3").client("s3").download_file("verta-starter", data_filename, data_filename)
+        SparkSession = pytest.importorskip("pyspark.sql").SparkSession
+        col = pytest.importorskip("pyspark.sql.functions").col
+        LogisticRegression = pytest.importorskip("pyspark.ml.classification").LogisticRegression
+        LogisticRegressionModel = pytest.importorskip("pyspark.ml.classification").LogisticRegressionModel
+        VectorAssembler = pytest.importorskip("pyspark.ml.feature").VectorAssembler
+
+        spark = SparkSession.builder.master("local").appName("parquet_example").getOrCreate()
+
+        df = spark.read.csv(data_filename, header=True, inferSchema=True)
+        df.repartition(5).write.mode("overwrite").parquet("datasets/census-train-parquet")
+        df = VectorAssembler(
+            inputCols=[c for c in df.columns if c != ">50k"],
+            outputCol="features",
+        ).transform(df)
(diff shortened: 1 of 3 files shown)


client/verta/tests/test_artifacts.py
```diff
@@ -434,6 +434,37 @@ def predict(self, data):
         assert experiment_run.get_model().__dict__ == custom.__dict__
         assert experiment_run.get_model().predict(strs) == custom.predict(strs)
 
+    def test_pyspark(self, experiment_run, in_tempdir):
```
convoliution, line 437: I suspect it's going to be difficult to get this running nightly (due to requiring Spark).

client/verta/verta/_internal_utils/_artifact_utils.py
```diff
@@ -248,8 +252,8 @@ def serialize_model(model):
         try:  # attempt to deserialize
             reset_stream(model)  # reset cursor to beginning in case user forgot
             model = deserialize_model(model.read())
-        except pickle.UnpicklingError:  # unrecognized model
-            bytestream, _ = ensure_bytestream(model)  # pass along file-like
```
convoliution, line 252: This `ensure_bytestream()` was

- problematic, because PySpark models have a `read()`/`write()` interface and bungle up our file-handling logic
- unnecessary, because all appropriate code paths in this method end with an `ensure_bytestream()` further down

client/verta/verta/_internal_utils/_artifact_utils.py
```diff
@@ -262,6 +266,23 @@ def serialize_model(model):
...
+        )
+        if isinstance(model, pyspark_base_classes):
+            temp_dir = tempfile.mkdtemp()
+            try:
+                spark_model_dir = os.path.join(temp_dir, "spark-model")
+                model.save(spark_model_dir)
+                bytestream = zip_dir(spark_model_dir)
+            finally:
+                shutil.rmtree(temp_dir)
+            return bytestream, "zip", "pyspark"
```
convoliution, line 286: Model API and model service (as I understand it) check for a "serialization method" and a "model type", so this PR introduces "zip" and "pyspark", respectively.

conradoverta:
> It works for `model_version.log_model()` too!
> But _not_ for `log_artifact()` (the specialized model-object-handling logic is only in the `log_model()` path).

What would take to support in log_artifact too?

client/verta/tests/test_artifacts.py
conradoverta: We can install in the container maybe =)

client/verta/verta/_internal_utils/_artifact_utils.py
conradoverta: I _think_ we need more info than `pyspark` if we want to deserialize, but we can handle that later. let's just add a TODO for now

conradoverta (approved):
LGTM

convoliution:
@conradoverta 
> What would take to support in log_artifact too?

...Hmm actually, I think I'd just need to add one method call. Might be super simple.
I'll merge this for now so it can get a round in the nightly tests, and add a follow-up ticket for `log_artifact()`

convoliution:
Darn it, I forgot to merge this before going to bed!

PR #1936, opened 2021-02-04
Fix null pointer due to empty metadata for service user on entity deletion flow

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-9798

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
@@ -4,6 +4,7 @@ import ai.verta.common.*;
 import ai.verta.common.ModelDBResourceEnum.ModelDBServiceResourceTypes;
 import ai.verta.modeldb.*;
 import ai.verta.modeldb.authservice.RoleService;
+import ai.verta.modeldb.common.CommonUtils;
 import ai.verta.modeldb.common.authservice.AuthService;
 import ai.verta.modeldb.common.collaborator.CollaboratorBase;
 import ai.verta.modeldb.common.collaborator.CollaboratorUser;
@@ -260,8 +261,13 @@ public class ProjectDAORdbImpl implements ProjectDAO {
       deletedEntitiesQuery.setParameter("projectName", project.getName());
       List<String> deletedEntityIds = deletedEntitiesQuery.list();
       if (!deletedEntityIds.isEmpty()) {
-        roleService.deleteEntityResourcesWithServiceUser(
-            deletedEntityIds, ModelDBServiceResourceTypes.PROJECT);
+        try {
+          CommonUtils.registeredBackgroundUtilsCount();
+          roleService.deleteEntityResourcesWithServiceUser(
+              deletedEntityIds, ModelDBServiceResourceTypes.PROJECT);
+        } finally {
+          CommonUtils.unregisteredBackgroundUtilsCount();
+        }
       }
 
       Transaction transaction = session.beginTransaction();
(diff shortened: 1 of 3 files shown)


evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -260,8 +261,13 @@ private Project insertProject(Project project, String workspaceName, UserInfo us
...
       if (!deletedEntityIds.isEmpty()) {
-        roleService.deleteEntityResourcesWithServiceUser(
-            deletedEntityIds, ModelDBServiceResourceTypes.PROJECT);
+        try {
+          CommonUtils.registeredBackgroundUtilsCount();
+          roleService.deleteEntityResourcesWithServiceUser(
+              deletedEntityIds, ModelDBServiceResourceTypes.PROJECT);
+        } finally {
+          CommonUtils.unregisteredBackgroundUtilsCount();
+        }
```
coryverta, line 270: I don't understand how this solves a null pointer exception.  It appears to swallow the exception.  Please explain why an NPE is occurring.

anandjakhaniya: We have set the null value for the metadata at https://github.com/VertaAI/modeldb/blob/b32413a82f84c10c354517128f9d08e7ef6a30fa/backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java#L106 for the service user so it will broken at https://github.com/VertaAI/modeldb/blob/b32413a82f84c10c354517128f9d08e7ef6a30fa/backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthServiceChannel.java#L59 because metadata was null.

here added `CommonUtils.registeredBackgroundUtilsCount();` because whenever we are making call to other service from utils or any other places, we are increase the background util count to protect the API calls with empty headers.

So now you had question on mind, why we have used `CommonUtils.registeredBackgroundUtilsCount();`? right. So let say if we have just check the metadata null then there is possibility to user make a api call without metadata and that case MDB talking service user for that request. so to protect that scenario we had introduce that count mechanism.

conradoverta: I agree with Cory. Even with the explanation, I don't know what is causing the NPE. What does `registeredBackgroundUtilsCount` do?

>So let say if we have just check the metadata null then there is possibility to user make a api call without metadata and that case MDB talking service user for that request. so to protect that scenario we had introduce that count mechanism.

This sounds like a hack instead of solving the actual problem, since you add an extra signal to infer what is the desired behavior indirectly. It also seems like we do that change regardless of whether the caller was the one to set the count or not. Seems like a global flag, which I believe I had reviewed and asked not to do at the time.

anandjakhaniya: We can improve and simplify this logic with new service and channel initialization which we have recently done but for now for the old design we should have to consider that background utils count logic.

conradoverta: Ok. Let's make sure to remove any implicit signal between things as it makes bugs like these very, very easy =)

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Simply the bug was occurred as per following comments. forget about the `registeredBackgroundUtilsCount` as of now.

We have set the null value for the metadata at https://github.com/VertaAI/modeldb/blob/b32413a82f84c10c354517128f9d08e7ef6a30fa/backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java#L106 for the service user so it will broken at https://github.com/VertaAI/modeldb/blob/b32413a82f84c10c354517128f9d08e7ef6a30fa/backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthServiceChannel.java#L59 because metadata was null.

PR #1940, opened 2021-02-04
VR-9732: Remove "file:" prefix from path datasets

convoliution:
There's actually an [entire Wikipedia article](https://en.wikipedia.org/wiki/File_URI_scheme) dedicated to explaining what the number of slashes after `file:` means.

I spent way too long trying to find a smart or stdlib way to do this, and there doesn't seem to be one! I think this works.

### DIFF
--- a/client/verta/tests/test_versioning/test_dataset.py
+++ b/client/verta/tests/test_versioning/test_dataset.py
@@ -1,7 +1,7 @@
+import itertools
 import os
 import shutil
 import pathlib2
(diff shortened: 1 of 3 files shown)


conradoverta (approved):

convoliution:
Ooh, this doesn't work with `with_spark()` because it bypasses the constructor.

I'll need to think about whether to resolve that here, or by implementing `with_spark()` differently.

PR #1942, opened 2021-02-05
Fix find project by name using getResources

anandjakhaniya:
**Change logs:**
- find project by name using getResources


backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
```diff
@@ -140,56 +141,51 @@ public boolean deleteEntityResourcesWithServiceUser(
...
+      }
+      Optional<GetResourcesResponseItem> responseItem = responseItems.stream().findFirst();
+      if (responseItem.isPresent()){
+        return responseItem.get();
+      } else {
+        StringBuilder errorMessage =
+                new StringBuilder("Failed to locate ")
+                        .append(modelDBServiceResourceTypes.name())
+                        .append(" resources in UAC for ")
+                        .append(modelDBServiceResourceTypes.name());
```
conradoverta, line 193: I don't think you meant to use the same `modelDBServiceResourceTypes.name()` in both places.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
```diff
@@ -140,56 +141,51 @@ public boolean deleteEntityResourcesWithServiceUser(
...
-    throw new IllegalArgumentException(
-        "Failed to locate "
-            + modelDBServiceResourceTypes.name()
-            + " resources in UAC for "
-            + modelDBServiceResourceTypes.name()
-            + " ID "
-            + entityId);
-  }
-
-  public List<GetResourcesResponseItem> getResourceItems(Optional<Resources> filterTo) {
```
conradoverta, line 195: Yay for removing public methods we don't need.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -309,16 +264,41 @@ private Project insertProject(Project project, String workspaceName, UserInfo us
...
 
       ProjectEntity projectEntity =
           session.load(ProjectEntity.class, projectId, LockMode.PESSIMISTIC_WRITE);
+      Project project = projectEntity.getProtoObject(roleService, authService);
+      roleService.createWorkspacePermissions(
+          workspace.getId(),
+          Optional.empty(),
+          projectId,
+          projectName,
+          Optional.empty(), // UAC will populate the owner ID
```
conradoverta: This means we might end up changing the owner! Let's not do this as it's a bug. First you get, then update the fields, then save the updated object. Please make sure you _only_ change the field you wanted to change.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -309,16 +264,41 @@ private Project insertProject(Project project, String workspaceName, UserInfo us
...
+        roleService.deleteEntityResourcesWithServiceUser(
+            deletedEntityIds, ModelDBServiceResourceTypes.PROJECT);
+      } finally {
+        CommonUtils.unregisteredBackgroundUtilsCount();
+      }
+    }
+  }
+
   @Override
   public Project updateProjectName(UserInfo userInfo, String projectId, String projectName)
```
conradoverta: This method can lead to inconsistency where MDB thinks the name is one thing and UAC thinks it's a different name.

conradoverta (changes requested):
Some bugs, but please split 1 PR per functionality/fix. One of the fixes looks good, but now it's blocked because the other seems buggy, so we can't benefit from the fix.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
anandjakhaniya: No, it is the expected and added intentionally. the error message generated by it will be like 
`Failed to locate PROJECT resources in UAC for PROJECT Name projectName` 
OR
`Failed to locate PROJECT resources in UAC for PROJECT Id abcd-abcd-abcd-abcd`

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Yes i am sure i am updating only the two fields.
```
projectEntity.setName(projectName);
projectEntity.setDate_updated(Calendar.getInstance().getTimeInMillis());
```

anandjakhaniya: `Optional.empty(), // UAC will populate the owner ID` was changed at first phase of visibility changes where we have moved owner field to UAC so we have used `Optional.empty()` everywhere. we will remove this unused parameter later.

conradoverta: You're probably assuming I know the inner workings on `roleService.createWorkspacePermissions`, which I do not and the name `create` indicates to me that it might behave wrong on an `update`. I am basing on the API. If you call the `setResources` API with **any** field different from the value you get on a `getResources` API (including empty), you'll do an update. Are you _absolutely_ sure this will _never_ happen in that method? If yes, let's fix the comment in this line because it's describing the wrong behavior.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
conradoverta: let's just skip the second one then

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: This will remove in different PR.

anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/1945

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Removed `updateProjectName` method and additionally added java doc comment on the method `roleService.createWorkspacePermissions` to understand the method parameters   https://github.com/VertaAI/modeldb/pull/1946/files

backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
anandjakhaniya: No we can not skip anyone because we have that both get by name and get by Id So both the case will works. look the line diff bellow this comment on file changes tab.

conradoverta (approved):

PR #1958, opened 2021-02-11
Fix dataset service as per new audit log design

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/dataset/DatasetServiceImpl.java
```diff
@@ -209,8 +224,16 @@ public void deleteDataset(
...
-      responseObserver.onNext(DeleteDataset.Response.newBuilder().setStatus(true).build());
+      DeleteDataset.Response response = DeleteDataset.Response.newBuilder().setStatus(true).build();
+      UserInfo userInfo = authService.getCurrentLoginUserInfo();
+      saveAuditLogs(
+          Optional.of(userInfo),
+          ModelDBServiceActions.DELETE,
+          Collections.singletonList(request.getId()),
+          ModelDBUtils.getStringFromProtoObject(request),
+          ModelDBUtils.getStringFromProtoObject(response),
+          authService.getWorkspaceIdFromUserInfo(userInfo));
```
conradoverta: This seems wrong to me. The dataset has a workspace it came from. Could we just use that workspace information? Otherwise this log will be incorrect.

@ad-47 @evgeniilezhnin let's make sure we fix this everywhere. I believe we might have done a similar mistake in other places.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/dataset/DatasetServiceImpl.java
evgeniilezhnin: https://github.com/VertaAI/modeldb/pull/1980

coryverta (approved):

conradoverta (approved):

PR #1963, opened 2021-02-15
Fix AuditLogsCron as per new design

anandjakhaniya:


conradoverta (approved):
This is so much simpler =)

PR #1964, opened 2021-02-15
Fix versioning service as per new audit log design

anandjakhaniya:
- @conradoverta should we add audit logs for commit entity?? if yes then we should have to implement it. cc @evgeniilezhnin


conradoverta:
> * @conradoverta should we add audit logs for commit entity?? if yes then we should have to implement it. cc @evgeniilezhnin

Every API call needs to be logged. It's incorrect for any API call to happen and not be logged.

conradoverta (approved):

PR #1966, opened 2021-02-16
VR-9843 audit log interceptor move to common

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/App.java
```diff
@@ -215,7 +215,7 @@ public static void main(String[] args) {
       serverBuilder.intercept(new ExceptionInterceptor());
       serverBuilder.intercept(new MonitoringInterceptor());
       serverBuilder.intercept(new AuthInterceptor());
-      serverBuilder.intercept(new AuditLogInterceptor());
+      serverBuilder.intercept(new AuditLogInterceptor(true));
```
conradoverta: We don't want it always true, right? I recall some discussion around that and putting in a config.

conradoverta (approved):
One ask to place the flag in config. LGTM otherwise

backend/src/main/java/ai/verta/modeldb/App.java
evgeniilezhnin: added config. Also added call count = 1 check

evgeniilezhnin: added change for this comment https://github.com/VertaAI/verta-uacservice/pull/561#discussion_r577395634

coryverta (approved):

anandjakhaniya (approved):

PR #1971, opened 2021-02-18
VR-9891: Allow arbitrary artifacts in log_model()

convoliution:
The model-logging code path involves determining a "model type" and a "serialization method", with the intent that they would be used by our model service to deserialize the model. To allow arbitrary artifacts, these checks are removed.

This PR:

- removes the check for model type in `_artifact_utils.serialize_model()`
- removes the check for serialization method in `ExperimentRun.log_model()`
- handles passing a directory path in `ExperimentRun.log_model()`


convoliution:
~I'm looking into how straightforward it would be to allow the same for `RegisteredModelVersion`~

✅

client/verta/verta/_tracking/experimentrun.py
```diff
@@ -1239,14 +1239,15 @@ def log_model(self, model, custom_modules=None, model_api=None, artifacts=None,
         except (TypeError, ValueError):
             extension = None
         _utils.THREAD_LOCALS.active_experiment_run = self
-        if isinstance(model, six.string_types):  # filepath
-            model = open(model, 'rb')
+        if isinstance(model, six.string_types):
+            if os.path.isdir(model):
+                model = _artifact_utils.zip_dir(model)
+            else:  # filepath
+                model = open(model, 'rb')
```
daniel-verta: Here and in other places in this module I think we're neglecting to close the file object. Supposedly it's something which some Python implementations handle automatically, but the behavior might be implementation specific.

daniel-verta (commented):
Currently there's a deprecated method `log_model_for_deployment` which has very similar behavior to `log_model`. Until that deprecated method is removed we should probably keep it consistent with `log_model`.

client/verta/verta/_tracking/experimentrun.py
convoliution: Yes! There are quite a few places where file objects are created, and not explicitly closed. It's potetially problematic for everything other than CPython.

Fixing it is complicated by the fact that some of these methods, e.g. for artifact upload, were written a bit naïvely by as less-experienced me to accommodate file objects passed in by the user. I wanted to avoid closing a caller's file object, but didn't have a good way to distinguish when it was one created internally (since they're all bound to the same vars), so ended up not closing any of them.

convoliution: I think it's worth a follow-up, but I'm not sure that I can resolve it even just within the scope of this ticket.

daniel-verta (approved):

convoliution:
Sorry about the sudden additional changes—they're copying this behavior to `ModelVersion.log_model()`, as well.

client/verta/verta/_registry/modelversion.py
```diff
@@ -234,8 +234,11 @@ def log_model(self, model, custom_modules=None, model_api=None, artifacts=None,
         if artifacts:
             self.add_attribute(_MODEL_ARTIFACTS_ATTR_KEY, artifacts, overwrite=overwrite)
 
-        if isinstance(model, six.string_types):  # filepath
-            model = open(model, 'rb')
+        if isinstance(model, six.string_types):
+            if os.path.isdir(model):
+                model = _artifact_utils.zip_dir(model)
```
houqp: curious how should we expect clients to consume a model built from a directory. it looks like the get model method will return bytes representing the zip file in this case? do we expect the clients to unzip that as a follow up step?

convoliution: @houqp https://github.com/VertaAI/modeldb/pull/1973 will allow `run.download_artifact()` and `run.download_model()` to automatically unzip the artifact into a directory on download, which should fit this usecase.

`get_artifact()` and `get_model()` are usually instead used for direct Python objects, rather than files/directories.

client/verta/verta/_registry/modelversion.py
```diff
@@ -234,8 +234,11 @@ def log_model(self, model, custom_modules=None, model_api=None, artifacts=None,
         if artifacts:
             self.add_attribute(_MODEL_ARTIFACTS_ATTR_KEY, artifacts, overwrite=overwrite)
 
-        if isinstance(model, six.string_types):  # filepath
-            model = open(model, 'rb')
+        if isinstance(model, six.string_types):
```
daniel-verta: Now it looks to me like this handling of file or directory should happen internally to `_artifact_utils.serialize_model` because the same logic is getting copied in `modelversion.py` and in both `log_model` methods in `experimentrun.py`. Do you think that small refactor is small enough to do in the scope of this PR?

daniel-verta (commented):
Small refactor suggested

client/verta/verta/_registry/modelversion.py
houqp: Ha, got it, sorry i missed that PR, thanks for the clarification 👍

convoliution: No worries—thank you for asking! Helps me make sure common workflows are fully supported.

daniel-verta (approved):

PR #1973, opened 2021-02-18
VR-9892: Unzip directory artifacts on download

convoliution:
- split artifact download helper function
- upload client-zipped directories with `.dir.zip` extension
- unzip downloaded artifacts if they have that `.dir.zip` extension
- add `run.download_model()`


client/verta/verta/_tracking/experimentrun.py
```diff
@@ -1678,12 +1674,31 @@ def download_artifact(self, key, download_to_path):
...
+        download_to_path : str
+            Path to download to.
+
+        Returns
+        -------
+        downloaded_to_path : str
+            Absolute path where artifact was downloaded to. Matches `download_to_path`.
+
+        """
+        self.download_artifact("model.pkl", download_to_path)
```
daniel-verta: suggest: moving string constants "model.pkl" in this file to a named constant. This default filename should always be identical to the string used for `log_model`.

client/verta/verta/_internal_utils/_request_utils.py
```diff
@@ -3,53 +3,116 @@
...
+        if not overwrite_ok:
+            # prevent overwrite in case `dirpath` was taken during download
+            dirpath = _file_utils.without_collision(dirpath)
+
+        # extract request contents to `dirpath`
+        with zipfile.ZipFile(temp_filepath, 'r') as zipf:
+            zipf.extractall(dirpath)
+    finally:
+        os.remove(temp_filepath)
+    print("download complete; directory extracted to {}".format(dirpath))
```
daniel-verta, line 116: With your approval @convoliution I'll make a tech-debt ticket to evaluate logging practices in our client library and plan for improvement. I'm pretty sure that in the longer term we want to steer away from using `print` statements so that our users can do things like choose their own log level. There's a good starting point [here in the Hitchhiker's Guide to Python](https://docs.python-guide.org/writing/logging/) which we can use as a jumping off point.

client/verta/verta/_internal_utils/_request_utils.py
```diff
@@ -3,53 +3,116 @@
...
+import zipfile
 
 from . import _file_utils
 
 
 # TODO: migrate request utils from _utils
 
 
-def download(response, filepath, chunk_size=32*(10**6), overwrite_ok=False):
+def download_response(response, chunk_size=32*(10**6)):
```
daniel-verta, line 14: Can this default value become an informative named constant?

client/verta/tests/test_artifacts.py
```diff
@@ -5,16 +5,14 @@
...
-
-from . import utils
+from verta._internal_utils import (
+    _artifact_utils,
+    _file_utils,
+    _utils,
+)
 
 
 class TestUtils:
```
daniel-verta, line 22: I'd like to see some test of the collision avoidance behavior of `download_file` and `download_zipped_dir`.

daniel-verta (changes requested):

client/verta/verta/_internal_utils/_request_utils.py
convoliution: @daniel-verta Certainly! I'd be more than thrilled to move nearly the entirety of the client's print statements to a logger, provided it's done consistently.

client/verta/verta/_internal_utils/_request_utils.py
convoliution: Would `_32MB` be informative, or would you be thinking of something that explains why that number was chosen (there wasn't a particular reason 😅).

client/verta/verta/_tracking/experimentrun.py
convoliution: be2e7e6888c381ce71eeaf3673f03e1ba3477ec5

client/verta/tests/test_artifacts.py
convoliution: Is this what you'd have in mind? 95bd655c257bd5b6cf2ecd8ace64e7c617b05b00

client/verta/verta/_internal_utils/_request_utils.py
conradoverta: I believe we have received such request, actually.

conradoverta (approved):
LGTM, but need to address Daniel's feedback.

client/verta/verta/_internal_utils/_request_utils.py
daniel-verta: _32MB is fine. If there was a specific motivation for that number then a comment would be helpful. Just a rename in this case is helpful.

daniel-verta (approved):

client/verta/verta/_internal_utils/_request_utils.py
convoliution: I think this idea makes sense, and there are also different places (and different values...) where `N*(10**6)` is used. I'm going to consolidate them and make that a separate PR shortly.

PR #1982, opened 2021-02-20
versioning audit logs

evgeniilezhnin:


backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
```diff
@@ -696,4 +696,12 @@ public static ModelDBServiceResourceTypes getModelDBServiceResourceTypesFromRepo
...
     return modelDBServiceResourceTypes;
   }
+
+  public static String getStringFromProtoObjectSilent(MessageOrBuilder object) {
+    try {
+      return JsonFormat.printer().preservingProtoFieldNames().print(object);
+    } catch (InvalidProtocolBufferException e) {
+      throw new ModelDBException(e);
+    }
+  }
```
anandjakhaniya: this is already available at the ModelDBUtils.java. we can simply use it for silent purpose as well.

backend/src/main/java/ai/verta/modeldb/versioning/VersioningServiceImpl.java
```diff
@@ -242,11 +253,23 @@ public void deleteRepository(
   public void listCommits(
       ListCommitsRequest request, StreamObserver<ListCommitsRequest.Response> responseObserver) {
     try {
+      Repository repository =
+          repositoryDAO
+              .getRepository(
+                  GetRepositoryRequest.newBuilder().setId(request.getRepositoryId()).build())
+              .getRepository();
```
ad-47, line 260: Instead of fetching whole repository we should have to fetch entity resource from UAC and get workspace_id from it.

ad-47: while getting repository using `getRepository` it will also call UAC for entity resource. so instead calling DB and UAC both just get entity resource from UAC is better.

anandjakhaniya (changes requested):
Needs to be fix logic using get entity resource from UAC.

backend/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
anandjakhaniya: Fixed

anandjakhaniya (approved):

coryverta (approved):

coryverta (approved):

PR #1987, opened 2021-02-23
VR-9939: Make log_model() output less confusing for a user when logging arbitrary model

convoliution:
Might help to review with "Hide whitespace changes" on.

## Changes
- rearrange internals of `log_model()` methods to only upload `custom_modules` and `model_api.json` if the model is Verta-deployable, or if provided by the user.

## Testing
- verifies these artifacts are not uploaded if the model is not Verta-deployable.

### DIFF
--- a/client/verta/verta/_tracking/experimentrun.py
+++ b/client/verta/verta/_tracking/experimentrun.py
@@ -365,7 +365,10 @@ class ExperimentRun(_DeployableEntity):
                     "PUT", url_for_artifact.url, self._conn, data=artifact_stream)
             _utils.raise_for_http_error(response)
 
-        print("upload complete ({})".format(key))
+        if key is _artifact_utils.MODEL_KEY:
+            print("model upload complete")
+        else:
+            print("upload complete ({})".format(key))
 
     def _log_artifact_path(self, key, artifact_path, artifact_type, overwrite=False):
         """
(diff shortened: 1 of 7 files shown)


convoliution:
~The arbitrary file handle is causing problems during the model serialization path in Python 2~
Okay, the good news is that it's not caused by this PR.

client/verta/verta/_tracking/experimentrun.py
```diff
@@ -1332,36 +1335,37 @@ def log_model(self, model, custom_modules=None, model_api=None, artifacts=None,
...
+                model_api['model_packaging'] = {
+                    'python_version': _utils.get_python_version(),
+                    'type': model_type,
+                    'deserialization': method,
+                }
+            if self._conf.debug:
+                print("[DEBUG] model API is:")
+                pprint.pprint(model_api.to_dict())
+
+            self._log_artifact("model_api.json", model_api,
```
daniel-verta: suggest naming this constant string

daniel-verta (approved):
LGTM

client/verta/verta/_tracking/experimentrun.py
convoliution: 4a8de7719bc9503f34f2f20f407fe7db5c04dec1
2a62500b1514c331b02411bc503be407ac8dcb98

convoliution:
Running tests to double-check the named constant swap-in didn't break anything.

PR #1990, opened 2021-02-23
Set version pin in environment test case

convoliution:
`Python()` raises an exception for the user if it's passed a library that isn't installed, and without a version pin. This is because a version number is needed for environment reproducibility.
```python
ValueError: unable to determine a version number for requirement 'tensorflow';
it might not be installed; please manually specify it as 'tensorflow==x.y.z'
```

The nightly client tests involve a runthrough where only the client is installed (to validate core functionality without external dependencies), so passing `"tensorflow"` raises this exception.


convoliution:
Wait I missed something, sorry.

convoliution:
Got it.

daniel-verta (approved):

PR #1994, opened 2021-02-24
Use teardown for entity deletion in test_create_same_name_diff_workspace

convoliution:
Since yesterday, this test encounters an error in teardown: a `403` when it tries to delete `new_project()`.
Which makes complete sense, since it's deleted at the end of the test, before teardown.
This PR uses the test suites's `created_entities` fixture to properly clean up these entities without such conflicts.

The root cause of this seems to be that before yesterday, repeatedly calling `project.delete()` would not raise an error (while `delete()` on other entities has always been a `4XX` error from the backend)


daniel-verta (approved):
LGTM

PR #2016, opened 2021-03-04
VR-10117, 10118: Set and get lock level on model versions

convoliution:
Feature-complete; test-writing in progress

### DIFF
--- a/client/verta/docs/api/api/lock.rst
+++ b/client/verta/docs/api/api/lock.rst
@@ -6,14 +6,14 @@ Lock
 .. note::
     .. versionadded:: 0.17.3
 
-.. autoclass:: verta.lock.Closed
+.. autoclass:: verta.registry.lock.Closed
     :members:
     :inherited-members:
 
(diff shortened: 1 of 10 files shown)


client/verta/docs/api/api/lock.rst
```diff
@@ -0,0 +1,19 @@
+.. lock-api:
+
+Lock
+====
+
+.. note::
+    .. versionadded:: 0.17.3
+
+.. autoclass:: verta.lock.Closed
```
conradoverta: `lock` as a top level package is very weird to me. I was expecting `verta.registry.lock` or something like that. Sorry if I missed this in the original proposal

convoliution: Oops that's a typo! It is supposed to be `verta.registry.lock`
Thanks!

conradoverta (approved):
LGTM

convoliution:
I shall write the tests separately!

PR #2024, opened 2021-03-08
Fix async ExecutionException, InterruptedException exception throwing strategy for less code part and file effect

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/entities/DatasetEntity.java
```diff
@@ -239,23 +237,27 @@ public Dataset getProtoObject(RoleService roleService)
...
+      datasetBuilder.setCustomPermission(repositoryResource.getCustomPermission());
+
+      DatasetVisibilityEnum.DatasetVisibility visibility =
+          (DatasetVisibilityEnum.DatasetVisibility)
+              ModelDBUtils.getOldVisibility(
+                  ModelDBServiceResourceTypes.DATASET, repositoryResource.getVisibility());
+      datasetBuilder.setDatasetVisibility(visibility);
+    } catch (InterruptedException | ExecutionException e) {
+      throw new ModelDBException(e);
+    }
```
conradoverta, line 260: I'm not sure it's a good idea to replicate this try/catch every single place where we use async calls, which will become more and more places. I personally would rather tuck it into the method signature than spread in the code. Thoughts?

anandjakhaniya: but this individual `InterruptedException | ExecutionException` exception try/catch we are sure about it is thrown from Listenable Future tasks. Just adding exception on method signature will not give surety about that. but i am agree with `which will become more and more places` line. so if you want to reduce that then it will fine to have in method signature.

conradoverta: I think leaving it in the method signature is ugly, but it's more sustainable than spreading the copied try/catch across the code bc the compiler handles that for us.

anandjakhaniya:
We have implemented latest Future class so now this is not required

PR #2025, opened 2021-03-08
multiple resourceIds for audit log in single request

anandjakhaniya:
**Proto PR:** https://github.com/VertaAI/protos-all/pull/152/files

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
@@ -80,6 +80,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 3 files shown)


backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
```diff
@@ -720,7 +721,9 @@ public void getProjects(
       saveAuditLog(
           Optional.of(userInfo),
           ModelDBServiceActions.READ,
-          ModelDBConstants.EMPTY_STRING,
+          projects.stream()
+              .map(Project::getId)
+              .collect(Collectors.joining(ModelDBConstants.COMMA_DELIMITER)),
```
conradoverta: Can you add a TODO of "don't save every single ID on fetch"? We should just move on from that right now, but this will become massive later on with us listing every single id we touch.

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
anandjakhaniya: Added TODO

PR #2028, opened 2021-03-09
Added catch for the optimistic lock exception on reconcile deletion

anandjakhaniya:


conradoverta (commented):
We should not be masking exceptions. It's important for the system to know that the operation it tried to do didn't go through vs just ignoring it.

anandjakhaniya:
> We should not be masking exceptions. It's important for the system to know that the operation it tried to do didn't go through vs just ignoring it.

but due to single failed entry the other one not delete. we are logging exception as error properly so we knew the error. so masking is good to have there. we are following this pattern on the old DeleteEntitiesCron job.

conradoverta:
> > We should not be masking exceptions. It's important for the system to know that the operation it tried to do didn't go through vs just ignoring it.
> 
> but due to single failed entry the other one not delete. we are logging exception as error properly so we knew the error. so masking is good to have there. we are following this pattern on the old DeleteEntitiesCron job.

Masking is not good to have there, actually. The reconciler machinery is more complex than just a cron job, so letting the wrapper handle failures is better for you (e.g. do a retry). It's ok that the others are not deleted. It's a trade-off when we want to do batch things. We ignored the error in the cronjob bc it was driven by a clock alone, so we wanted to delete as much as possible in that clock tick, but that's now always the best behavior.

conradoverta (commented):
Marking as reviewed in my notifications due to previous comment.

anandjakhaniya:
This is too old and now we have reconciler implemented so not need this anymore.

PR #2033, opened 2021-03-12
Fix audit logs with empty resource ids

anandjakhaniya:


anandjakhaniya:
@conradoverta this PR is related to slack thread https://vertaai.slack.com/archives/C01M8E8DUF3/p1615528195017500?thread_ts=1615525256.006100&cid=C01M8E8DUF3

conradoverta (changes requested):
Let's not hack and inject fake resources in the list. Let's just add a workspace_id to the protos

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentServiceImpl.java
```diff
@@ -162,7 +164,8 @@ public void createExperiment(
           ModelDBServiceActions.CREATE,
           Collections.singletonMap(experiment.getId(), entityResource.getWorkspaceId()),
           ModelDBUtils.getStringFromProtoObject(request),
-          ModelDBUtils.getStringFromProtoObject(response));
+          ModelDBUtils.getStringFromProtoObject(response),
+          entityResource.getWorkspaceId());
```
conradoverta, line 168: Workspace service id? Please check everywhere as I think some other cases also use this old id instead of the new one.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
```diff
@@ -1779,28 +1817,30 @@ public void findExperimentRuns(
...
-      experimentRuns.forEach(
-          experimentRun ->
-              auditResourceMap.put(experimentRun.getId(), entityResource.getWorkspaceId()));
-      saveAuditLog(
-          Optional.empty(),
-          ModelDBServiceActions.READ,
-          auditResourceMap,
-          ModelDBUtils.getStringFromProtoObject(request),
-          ModelDBUtils.getStringFromProtoObject(response));
+      if (experimentRuns.isEmpty()) {
```
conradoverta, line 1829: The workspace of the request doesn't care if you return zero or more. So I'm not sure why we added this conditional.

Similar comment for other locations.

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentServiceImpl.java
anandjakhaniya: No, if we are getting from entity then it should be `workspace_service_id`  like `project.getWorkspaceServiceId()` already i have used. but when we are getting it from `GetResourcesResponseItem` then there only single field which is `workspace_id` like `entityResource.getWorkspaceId()`

conradoverta: Oh, got it.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunServiceImpl.java
anandjakhaniya: We have added those condition because if those entities will be zero then we need the workspace otherwise we will take workspace id from return entities so on that case we do not want to fetch workspace again because it will already filter and returns by workspace.

conradoverta (approved):

PR #2041, opened 2021-03-19
Add getters that do NOT block for the UAC and Workspace services

ghost:


conradoverta (approved):
Nice. We should figure out how to make the formatting changes standard to avoid extra changes =/ cc @coryverta for thoughts

coryverta:
I think setting a pre-commit hook that invokes the code formatter via maven should help out a lot with formatting changes.  It is worth an experiment.

conradoverta:
But this runs jenkins auto-formatter and even that didn't change the formatting

PR #2042, opened 2021-03-22
BasicConnectionPool in java

anandjakhaniya:


anandjakhaniya:
We have choose JDBI instead of custom pool so we can close this now.

PR #2045, opened 2021-03-22
GCS Artifact Service

jaiganeshp:
Added a GCSService class that would enable us to store data to GCS using the GCS java library instead of Minio endpoint.


jaiganeshp:
Need to merge this into the 2.0.8.1 tag and not master code.

PR #2047, opened 2021-03-23
Added a GCSService class that would enable us to store data to GCS us…

### LINKED ISSUE #2008: Artifact Store Support for GCS
### Background
The default configuration of ModelDB is working well with AWS Access and AWS Secret Access Keys but does not work with GCS HMAC Keys. 

### Solution
Extend ArtifactStoreService Interface implementation for GCS using Google JDK. I am right now working on this implementation for the 2.0.8.1 release and will submit a merge request soon.

coryverta:
…ing GCS java library instead of minio endpoint.

https://github.com/VertaAI/modeldb/issues/2008

### DIFF
--- a/backend/config/config.yaml
+++ b/backend/config/config.yaml
@@ -8,7 +8,7 @@ springServer:
   shutdownTimeout: 30 #time in second
 
 artifactStoreConfig:
-  artifactStoreType: NFS #S3, GCP, NFS
+  artifactStoreType: NFS #S3, GCS, NFS
   NFS:
     nfsUrlProtocol: http
     nfsRootPath: /artifact-store/
--- a/backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/GCSService.java
+++ b/backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/GCSService.java
@@ -1,0 +1,183 @@
+package ai.verta.modeldb.artifactStore.storageservice;
+
+import ai.verta.modeldb.App;
+import ai.verta.modeldb.ModelDBConstants;
+import ai.verta.modeldb.ModelDBException;
+import com.amazonaws.services.s3.model.PartETag;
+import com.google.auth.Credentials;
+import com.google.auth.ServiceAccountSigner;
+import com.google.cloud.storage.*;
+import com.google.cloud.storage.BlobInfo;
+import com.google.rpc.Code;
+import com.google.rpc.Status;
+import io.grpc.protobuf.StatusProto;
+import java.io.BufferedReader;
+import java.io.DataOutputStream;
+import java.io.InputStreamReader;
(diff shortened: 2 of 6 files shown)


backend/config/config.yaml
```diff
@@ -8,7 +8,7 @@ springServer:
   shutdownTimeout: 30 #time in second
 
 artifactStoreConfig:
-  artifactStoreType: NFS #S3, GCP, NFS
+  artifactStoreType: NFS #S3, GCS, NFS
```
evgeniilezhnin, line 11: should be added to config.default.yaml too

evgeniilezhnin (approved):

backend/config/config.yaml
jaiganeshp: Done and also added config for GCS.

evgeniilezhnin (approved):

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/GCSService.java
```diff
@@ -0,0 +1,183 @@
...
+      throw StatusProto.toStatusRuntimeException(status);
+    }
+  }
+
+  @Override
+  public Optional<String> initiateMultipart(String s3Key) {
+    // Returning empty would let the underlying caller know that this artifact store method doesn't
+    // support multipart
+    // uploads and will only call generatePresignedUrl once to get the upload URL.
+    return Optional.empty();
```
anandjakhaniya:
```suggestion
    String errorMessage = "Multipart not supported by GCS";
    Status status =
            Status.newBuilder().setCode(Code.FAILED_PRECONDITION_VALUE).setMessage(errorMessage).build();
    LOGGER.error(errorMessage);
    throw StatusProto.toStatusRuntimeException(status);
```
As per your code at https://github.com/VertaAI/modeldb/pull/2047/files#diff-cd7c39f9e673d88a84de50bef8afce24fdd1f0a8d293a477c25b88ad4b8937d2R181, we are not supporting the multipart on GCS so could we throw exception here to protect initialization of multipart on getUrlForArtifact call instead of commit multipart like i have suggested??

anandjakhaniya:
I have tested the code all works fine excluding https://github.com/VertaAI/modeldb/pull/2047/files#r609348957. due to that i am miss leading on multipart upload. it is allowed me to upload artifact based on multipart and at the last when i am trying to commit that multipart it will throw `Not supported by GCS`.

backend/src/main/java/ai/verta/modeldb/artifactStore/storageservice/GCSService.java
jaiganeshp: @ad-47 : Thank you for the suggestion. I have accepted the commit suggestion.

PR #2050, opened 2021-03-24
Add more flexibility for specifying account for uac

conradoverta:


backend/common/src/main/java/ai/verta/modeldb/common/connections/UAC.java
```diff
@@ -73,18 +73,22 @@ private UAC(UAC other) {
   }
 
   public UAC withServiceAccount() {
+    return this.withServiceAccount(this.serviceUserEmail, this.serviceUserDevKey);
+  }
+
+  public UAC withServiceAccount(String serviceUserEmail, String serviceUserDevKey) {
     UAC c = new UAC(this);
```
ghost, line 80: why would you name this `c`

ghost (approved):

backend/common/src/main/java/ai/verta/modeldb/common/connections/UAC.java
conradoverta: Probably the type was named something else before.

PR #2052, opened 2021-03-24
fix config example

evgeniilezhnin:


anandjakhaniya (approved):

backend/config.example.yaml
```diff
@@ -8,16 +8,15 @@ springServer:
...
   devKey:
 
 artifactStoreConfig:
   artifactStoreType: NFS #S3, GCP, NFS
   pickArtifactStoreHostFromConfig: false
-  artifactStoreServerHost: localhost:8086
-  artifactStoreUrlProtocol: http
-  artifactStoreServerPort: 8086
+  host: localhost:8086
+  protocol: http
```
anandjakhaniya, line 21: there is one missing field in config here : `enabled`,

anandjakhaniya: there are many other fields are missing in config file like in s3 config `minioEndpoint:` is missing etc. need to verify entire config file.

anandjakhaniya (changes requested):
Needs to be check whole config file for missing parameters

conradoverta:
Is this still relevant?

evgeniilezhnin:
> Is this still relevant?

yes, but I forgot to add other missing fields, sorry

anandjakhaniya (approved):

evgeniilezhnin:
Anands changes LGTM

PR #2067, opened 2021-03-30
[VR-10479] Cache calls to getWorkspaceById in findProjects

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10479


conradoverta (approved):
LGTM. Let's do this for datasets too bc I imagine they have the same issue, but I just didn't benchmark yet.

PR #2068, opened 2021-03-31
Fix caching map initialization for projects

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -268,7 +268,7 @@ private Project insertProject(Project project, String workspaceName, UserInfo us
       transaction.commit();
       LOGGER.debug("Project created successfully");
       TelemetryUtils.insertModelDBDeploymentInfo();
-      return projectEntity.getProtoObject(roleService, authService, Collections.emptyMap());
```
anandjakhaniya, line 271: `Collections.emptyMap()` is immutable so while putting value in map it will break the flow so replaced it with new map initialization.

conradoverta (approved):

PR #2069, opened 2021-03-31
[VR-10506] Cache workspace for datasets, repositories

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10506


anandjakhaniya:
@conradoverta Here problem is we are calling all the `toProto` process in FutureTasks and it will execute in parallel so it might be  leading inconsistency to populate values in map. need your thoughts here????

conradoverta:
> @conradoverta Here problem is we are calling all the `toProto` process in FutureTasks and it will execute in parallel so it might be leading inconsistency to populate values in map. need your thoughts here????

Just synchronize on the operations on the map, no?

anandjakhaniya:
> > @conradoverta Here problem is we are calling all the `toProto` process in FutureTasks and it will execute in parallel so it might be leading inconsistency to populate values in map. need your thoughts here????
> 
> Just synchronize on the operations on the map, no?

No We can not use syncronize OR concurrent map. using that still issue is there because the parallel thread of ListenableFuture check the map has entry or not and for both the thread it will empty so both are called `workspace = authService.workspaceById(false, repository.getWorkspaceId());` and then populate in map.

conradoverta (approved):

PR #2070, opened 2021-03-31
[VR-10478] Cache calls to getResources in findProjects

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10478


backend/src/main/java/ai/verta/modeldb/entities/ProjectEntity.java
```diff
@@ -351,11 +354,16 @@ public Project getProtoObject(
...
     }
 
-    GetResourcesResponseItem projectResource =
-        roleService
-            .getEntityResource(
-                Optional.of(this.id), Optional.empty(), ModelDBServiceResourceTypes.PROJECT)
-            .get();
+    GetResourcesResponseItem projectResource;
+    if (getResourcesMap.isPresent() && !getResourcesMap.get().isEmpty()) {
+      projectResource = getResourcesMap.get().get(this.id);
```
conradoverta: `projectResource` might still not be in the map. I'd rather be safer here just in case.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -1052,9 +1075,11 @@ public Project logArtifacts(String projectId, List<Artifact> newArtifacts)
       }
 
       Map<Long, Workspace> cacheWorkspaceMap = new HashMap<>();
+      Map<String, GetResourcesResponseItem> getResourcesMap = new HashMap<>();
```
conradoverta, line 1073: For example, this is a map without the project id as key.

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
```diff
@@ -914,15 +930,21 @@ public ProjectPaginationDTO findProjects(
                 ? (UserInfo) host.getCollaboratorMessage()
                 : currentLoginUserInfo;
         if (userInfo != null) {
-          accessibleProjectIds =
-              ModelDBUtils.filterWorkspaceOnlyAccessibleIds(
```
conradoverta, line 916: I don't understand why we had to change the method.

backend/src/main/java/ai/verta/modeldb/entities/ProjectEntity.java
```diff
@@ -351,11 +354,16 @@ public Project getProtoObject(
...
+    } else {
+      projectResource =
```
conradoverta, line 363: We should add to the map if it exists so that we don't have to fetch this again, just in case.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: Internally `filterWorkspaceOnlyAccessibleIds`  method called the GetResourcesItem method and it is returning list of accessibleProjectIds so we can not build the GetResources Map using only ids so i have extracted that UAC endpoint here.

backend/src/main/java/ai/verta/modeldb/entities/ProjectEntity.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java
anandjakhaniya: got it. Fixed

backend/src/main/java/ai/verta/modeldb/entities/ProjectEntity.java
anandjakhaniya: Agree with comment. added now

conradoverta (approved):

PR #2075, opened 2021-04-02
Reduce uac calls and optimize in findExperimentRuns

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10476


conradoverta (commented):
I'm not sure about the content of this PR. Could you clarify?

anandjakhaniya:
> I'm not sure about the content of this PR. Could you clarify?

So in this PR i have try to reduce various UAC calls on the findExpeprimentRuns flow via either using cache maps OR using `getResources` endpoint which we are already called and have cache map. So it will reduce various UAC interaction like `getSelfAllowedResources` , `isSelfAllowed`, `getWorkspaceProjectIDs` filter method etc.

conradoverta (approved):

PR #2077, opened 2021-04-02
[VR-10510] Add monitoring sub-client to Verta Python Client

daniel-verta:


client/verta/setup.py
```diff
@@ -33,6 +33,7 @@
         "protobuf>=3.8",
         "pyyaml>=5.1, <5.4",
         "requests>=2.21, <3.0",
+        "numpy",
```
convoliution, line 36: Looks like we're going to need to add scipy and pandas as well, otherwise
```python
class Profilers:

    DEFAULT_ENVIRONMENT = Python(requirements=["numpy", "scipy", "pandas"])
````
fails at import time (for our linters) because `scipy` and `pandas` aren't present in the environment.

**Review by convoliution — CHANGES_REQUESTED** [review]
One initial comment. I'll take a closer look tomorrow.

File: client/verta/tests/opertions/monitoring/test_profilers.py [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+        assert created_profiler.name == new_name
+        assert created_profiler.name != old_name
+        assert old_profiler_version == created_profiler.reference
+
+        delete = profilers.delete(created_profiler)
+        assert delete
+
+    # TODO: Write a test plan for this "heavy" operation
+    def test_deploy(self, client):
+        pass
```
convoliution, line 42: We might want to `@pytest.mark.skip()` this, so that it appears as a deliberate `SKIPPED` rather than a `PASSED` in the test results.

client/verta/verta/operations/monitoring/client.py
```diff
@@ -0,0 +1,74 @@
...
+from .summaries import Summaries
+from .labels import Labels
+
+
+class Client(object):
+    def __init__ (self, verta_client):
+        self._client = verta_client
+        self.profilers = Profilers(self._conn, self._conf, self._client)
+        self.summaries = Summaries(self._conn, self._conf)
+        self.labels = Labels(self._conn, self._conf)
```
convoliution, line 19: I'd prefer having these created & returned in `@property`s rather than instance attributes like these (prevent accidental mutation / overwrites)

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -0,0 +1,205 @@
...
+        success = self._conn.must_proto_response(
+            response, FindSummarySampleRequest.Response
+        )
+        samples = [SummarySample(self._conn, self._conf, record) for record in success.samples]
+        return samples
+
+    @staticmethod
+    def _labels_proto(labels):
+        return {
+            key: LabelFilterQuerySummarySample(label_value=values)
```
convoliution, line 118: There's a [`_internal_utils._utils.as_list_of_str()`](https://github.com/VertaAI/modeldb/blob/master/client/verta/verta/_internal_utils/_utils.py#L1430) that can convert `{"key": "value"}` to `{"key": ["value"]}` which I personally would find helpful (otherwise `LabelFilterQuerySummarySample` ends up iterating through `"value"` to produce `['v', 'a', 'l', 'u', 'e']`)
```suggestion
            key: LabelFilterQuerySummarySample(label_value=as_list_of_str(values))
```

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -0,0 +1,205 @@
...
+        endpoint = "/api/v1/summaries/deleteSample"
+        msg = DeleteSummarySampleRequest(ids=ids)
+        response = self._conn.make_proto_request("DELETE", endpoint, body=msg)
+        self._conn.must_proto_response(response, EmptyProto)
+        return
+
+
+class SummarySample(
+    entity._ModelDBEntity
+):  # Note: trying out record instead of sample to see how this name feels
```
convoliution:
```suggestion
class SummarySample(entity._ModelDBEntity):
```

client/verta/verta/operations/monitoring/client.py
```diff
@@ -0,0 +1,74 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+import itertools
+
+from verta._tracking import _Context
```
convoliution, line 6: Relative import is preferred (and required for Python 2 in some cases iirc)
```suggestion
from ..._tracking import _Context
```

client/verta/verta/operations/monitoring/labels.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+            Find a list of labels according to provided parameters.
+
+            :key summary_query:
+            :key sample_ids:
+            :key labels:
+            :key time_window_start:
+            :key time_window_end:
+            :return: find_keys should return a list of strings used as label keys
+            :rtype: list
```
convoliution, line 31: I'd prefer these to be NumPy-style docstrings, but open to discuss/change some other time.

client/verta/verta/operations/monitoring/monitored_entity.py
```diff
@@ -0,0 +1,102 @@
...
+
+        print(
+            "created new MonitoredEntity: {} in {}".format(
+                obj.name, WORKSPACE_PRINT_MSG
+            )
+        )
+        return obj
+
+    def _set_client(self, client):
+        self._client = client
```
convoliution, line 91: I'd prefer having this in the `__init__()` so that all attributes are defined in one place, but this is fine for now if it would break things.

convoliution (changes requested):
Some initial comments, while we're working on things

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -0,0 +1,205 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+from builtins import object
```
convoliution: This isn't needed as far as I know—`object` is defined already in both Python 2 and 3.
```suggestion
```

convoliution (changes requested):

client/verta/verta/_internal_utils/time_utils/__init__.py
```diff
@@ -0,0 +1,6 @@
+from ...external.six import PY2, PY3
+
+if PY3:
+    from ._time_utils_py3 import *
+if PY2:
+    from ._time_utils_py2 import *
```
convoliution, line 6: I'm curious whether we can just use the py2 code for both 2 and 3—might be less to maintain in the future. Is there a tangible difference between the two?

client/verta/verta/client.py
```diff
@@ -207,6 +207,8 @@ def is_unauthorized(response): return response.status_code == 401
         self._ctx = _Context(self._conn, self._conf)
         self._workspace = None
 
+        self.operations = MonitoringClient(self)
```
convoliution, line 210: Similar to https://github.com/VertaAI/modeldb/pull/2077/files#r607208703, I'd prefer this to be a `@property`, but that might just be what I'm used to.

client/verta/verta/operations/monitoring/client.py
```diff
@@ -0,0 +1,74 @@
...
+    @property
+    def _ctx(self):
+        return self._client._ctx
+
+    def get_or_create_monitored_entity(self, name=None, workspace=None, id=None):
+        if name is not None and id is not None:
+            raise ValueError("cannot specify both `name` and `id`")
+
+        name = self._client._set_from_config_if_none(name, "monitored_entity")
+        workspace = self._client._set_from_config_if_none(workspace, "workspace")
```
convoliution:
```suggestion
        if workspace is None:
            workspace = self._client.get_workspace()
```

client/verta/verta/operations/monitoring/labels.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            keys = []
+        msg = FindSampleLabelValuesRequest(filter=summary_filter, labels=keys)
+        endpoint = "/api/v1/labels/findLabelValues"
+        response = self._conn.make_proto_request("POST", endpoint, body=msg)
+        proto = self._conn.must_proto_response(
+            response, FindSampleLabelValuesRequest.Response
+        )
+        return {
+            key: {v for v in valuesItem.values}
```
convoliution: perhaps
```suggestion
            key: set(valuesItem.values)
```

client/verta/verta/operations/monitoring/labels.py
```diff
@@ -0,0 +1,92 @@
...
+        sample_ids = kwargs.get('sample_id', None)
+
+        window_start = kwargs.get('time_window_start', None)
+        window_start_at_millis = maybe(lambda t: time_utils.epoch_millis(t), window_start)
+
+        window_end = kwargs.get('time_window_end', None)
+        window_end_at_millis = maybe(lambda t: time_utils.epoch_millis(t), window_end)
+
+        labels = kwargs.get('labels', None)
+        labels_proto = maybe(lambda li: Summary._labels_proto(li), labels)
```
convoliution: perhaps
```suggestion
        labels_proto = maybe(Summary._labels_proto, labels)
```

client/verta/verta/operations/monitoring/monitored_entity.py
```diff
@@ -0,0 +1,102 @@
...
+)
+
+
+class MonitoredEntity(entity._ModelDBEntity):
+    def __init__(self, conn, conf, msg):
+        super(MonitoredEntity, self).__init__(
+            conn, conf, _DataMonitoringService, "monitored_entity", msg
+        )
+
+    def __repr__(self):
```
convoliution, line 23: This PR has 3 different formats for its `__repr__()`s 😬. The client also has multiple formats grandfathered in; I think the whole thing is due to be cleaned up to match.

I'm personally a fan of `"\n\t".join()`

client/verta/verta/operations/monitoring/monitored_entity.py
```diff
@@ -0,0 +1,102 @@
...
+    def _generate_default_name(cls):
+        return "MonitoredEntity {}".format(_utils.generate_default_name())
+
+    @classmethod
+    def _get_proto_by_id(cls, conn, id):
+        Message = _DataMonitoringService.GetMonitoredEntityRequest
+        msg = Message(id=id)
+        endpoint = "/api/v1/monitored_entity/getMonitoredEntity"
+        response = conn.make_proto_request("GET", endpoint, params=msg)
+        return conn.maybe_proto_response(response, Message.Response).monitored_entity
```
convoliution, line 58: perhaps
```suggestion
        msg = _DataMonitoringService.GetMonitoredEntityRequest(id=id)
        endpoint = "/api/v1/monitored_entity/getMonitoredEntity"
        response = conn.make_proto_request("GET", endpoint, params=msg)
        return conn.maybe_proto_response(response,  msg.Response).monitored_entity
```

client/verta/verta/operations/monitoring/profiler.py
```diff
@@ -0,0 +1,78 @@
...
+        return self.profile(df)
+
+    def _profile_columns(self, df):
+        return [self._profile_column(df, column) for column in self.columns]
+
+    def _profile_column(self, df, column):
+        raise NotImplementedError("")
+
+    def __name__(self):
+        return "Profiler"
```
convoliution, line 30: I wonder where/if this is being used? I'm concerned about messing with assigning `__name__` (conventionally, it's only defined by default on classes rather than instances)

client/verta/verta/operations/monitoring/profilers.py
```diff
@@ -0,0 +1,249 @@
...
+    ListProfilersRequest,
+    UpdateProfilerRequest,
+    DeleteProfilerRequest,
+    EnableProfilerRequest,
+    GetProfilerStatusRequest,
+    DisableProfilerRequest,
+    KeyValue,
+    ValueTypeEnum,
+    DeployStatusEnum as DeployStatus,
+    BuildStatusEnum as BuildStatus,
```
convoliution, line 22: I feel like removing the `Enum` suffix kind of obfuscates the usefulness of the naming convention. Would prefer keeping it.

client/verta/verta/operations/monitoring/profilers.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @classmethod
+    def _create_proto_internal(cls, conn, ctx, name, reference):
+        msg = CreateProfilerRequest(name=name, profiler_reference=reference)
+        endpoint = "/api/v1/monitored_entity/createProfiler"
+        response = conn.make_proto_request("POST", endpoint, body=msg)
+        profile = conn.must_proto_response(
+            response, CreateProfilerRequest.Response
+        ).profiler
+        return profile
```
convoliution:
```suggestion
        profiler = conn.must_proto_response(
            response, msg.Response
        ).profiler
        return profiler
```

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -0,0 +1,202 @@
...
+    def delete(self, summary_records):
+        try:
+            ids = [record.id for record in summary_records]
+        except:
+            ids = summary_records
+        endpoint = "/api/v1/summaries/deleteSample"
+        msg = DeleteSummarySampleRequest(ids=ids)
+        response = self._conn.make_proto_request("DELETE", endpoint, body=msg)
+        self._conn.must_proto_response(response, EmptyProto)
+        return
```
convoliution:
```suggestion
        return True
```

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -0,0 +1,202 @@
...
+        maybe_summaries = [Summary(self._conn, self._conf, msg) for msg in maybe_summaries.summaries]
+        return maybe_summaries
+
+    def delete(self, summaries):
+        summary_ids = [summary.id for summary in summaries]
+        msg = DeleteSummaryRequest(ids=summary_ids)
+        endpoint = "/api/v1/summaries/deleteSummary"
+        response = self._conn.make_proto_request("DELETE", endpoint, body=msg)
+        self._conn.must_proto_response(response, EmptyProto)
+        return
```
convoliution:
```suggestion
        return True
```

client/verta/verta/operations/monitoring/utils.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+
+def extract_ids(objects_or_ids):
+    try:
+        ids = [obj.id for obj in objects_or_ids]
+    except:
+        ids = objects_or_ids
+    return ids
+
+def maybe(fn, val):
```
convoliution, line 10: This is amazing

convoliution (changes requested):
These are in addition to prior comments!

client/verta/verta/_internal_utils/time_utils/__init__.py
daniel-verta: Unfortunately the way time zones are handled in Python 2 is different enough from Python 3 that some amount of this splitting is necessary. I think that the compatible core could be extracted from these, but I think that supporting Python `datetime` objects is relatively important for the UX of working with times.

convoliution (approved):
Looks good to merge. We'll follow up in subsequent PRs.

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -0,0 +1,202 @@
...
+        return samples
+
+    @staticmethod
+    def _labels_proto(labels):
+        return {
+            key: LabelFilterQuerySummarySample(label_value=values)
+            for key, values in labels.items()
+        }
+
+    def delete(self, summary_records):
```
convoliution, line 122: This should probably be `delete_samples()` instead—`_ModelDBEntity.delete()` should delete `self`.

PR #2080, opened 2021-04-05
VR-10556 set max message size for grpc server

evgeniilezhnin:


backend/common/src/main/java/ai/verta/modeldb/common/config/GrpcServerConfig.java
```diff
@@ -4,6 +4,7 @@
   public int port;
   public int requestTimeout = 30;
   public int metrics_port = 8087;
+  public Integer maxInboundMessageSize;
```
anandjakhaniya: Could we add some default value in case if some how we have not added it in config then we can used default one.??

evgeniilezhnin: added

conradoverta (approved):

PR #2090, opened 2021-04-06
Disable mysql connector SSL

coryverta:
Adds a configuration parameter that is used to set the SSL mode for database connections.  This is required to migrate to newer versions of the JDK to prevent database connection failure.


coryverta:
Note for later: we need to clean up and consolidate the different methods for obtaining a DB connection.  JDBI provides a clean simple method for obtaining a connection that is preconfigured, that would be a good path forward.

conradoverta (approved):

PR #2091, opened 2021-04-07
[VR-10522] Add dist() to data types where applicable

convoliution:
Ported over directly from the old `data_formats.py` implementations.

### DIFF
--- a/client/verta/verta/data_types/_confusion_matrix.py
+++ b/client/verta/verta/data_types/_confusion_matrix.py
@@ -71,4 +71,4 @@ class ConfusionMatrix(_VertaDataType):
                 "`other` must be type {}, not {}".format(type(self), type(other))
             )
 
-        return -1  # TODO
+        raise NotImplementedError  # TODO
--- a/client/verta/verta/data_types/_confusion_matrix.py
+++ b/client/verta/verta/data_types/_confusion_matrix.py
@@ -64,3 +64,11 @@ class ConfusionMatrix(_VertaDataType):
             value=data["value"],
             labels=data["labels"],
         )
+
+    def dist(self, other):
+        if not isinstance(other, type(self)):
+            raise TypeError(
+                "`other` must be type {}, not {}".format(type(self), type(other))
+            )
+
+        return -1  # TODO
--- a/client/verta/verta/data_types/_discrete_histogram.py
+++ b/client/verta/verta/data_types/_discrete_histogram.py
@@ -72,7 +72,11 @@ class DiscreteHistogram(_VertaDataType):
 
         # TODO: assuming labels are consistent
         if collections.Counter(self._buckets) != collections.Counter(other._buckets):
-            return -1  # fix error return
+            raise ValueError(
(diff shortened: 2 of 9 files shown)


client/verta/verta/data_types/_confusion_matrix.py
```diff
@@ -64,3 +64,11 @@ def _from_dict_inner(cls, d):
...
             labels=data["labels"],
         )
+
+    def dist(self, other):
+        if not isinstance(other, type(self)):
+            raise TypeError(
+                "`other` must be type {}, not {}".format(type(self), type(other))
+            )
+
+        return -1  # TODO
```
daniel-verta: Maybe `raise NotImplementedError`?

client/verta/verta/data_types/_discrete_histogram.py
```diff
@@ -59,3 +63,27 @@ def _from_dict_inner(cls, d):
...
             data=data["data"],
         )
+
+    def dist(self, other):
+        if not isinstance(other, type(self)):
+            raise TypeError(
+                "`other` must be type {}, not {}".format(type(self), type(other))
+            )
+
+        # TODO: assuming labels are consistent
```
daniel-verta: I'd like to get this right within the scope of this PR because I think it will be more work to circle back and fix it than to do it now. I can jump in and provide the implementation if you would like.

client/verta/verta/data_types/_numeric_value.py
```diff
@@ -54,3 +54,11 @@ def _from_dict_inner(cls, d):
...
             unit=data["unit"],
         )
+
+    def dist(self, other):
+        if not isinstance(other, type(self)):
+            raise TypeError(
+                "`other` must be type {}, not {}".format(type(self), type(other))
+            )
+
+        return self._value - other._value
```
daniel-verta: I'd like to suggest we use absolute percentage change here instead of raw difference.

- This difference isn't a distance function, which will be fine once we let users define the direction of comparison (above or below threshold), but right now we don't offer that and so I would assume +20% change and -20% change should be equally alert worthy.
- It's harder to set a threshold for a raw difference than for a percentage change unless you're intimately familiar with your data.

daniel-verta (changes requested):

client/verta/verta/data_types/_discrete_histogram.py
convoliution: I believe I've got this.

client/verta/verta/data_types/_numeric_value.py
convoliution: For a reference alerter, it makes sense to me to make this a percentage change (and then alert if that absolute percentage change is greater than `threshold`.

I'll need to figure out what to do for the fixed alerter, since the original plan was to use `NumericValue.dist()` for that, but that threshold comparison should be a raw difference

daniel-verta (approved):

daniel-verta (approved):

PR #2092, opened 2021-04-07
VR-10570 set max_allowed_packet cron

evgeniilezhnin:
Motivation: https://vertaai.slack.com/archives/CLTEFV78X/p1617655905050300


backend/common/src/main/java/ai/verta/modeldb/common/CommonHibernateUtil.java
```diff
@@ -331,7 +342,17 @@ public Connection getDBConnection(RdbConfig rdb) throws SQLException, ClassNotFo
...
-      return con.isValid(timeout);
+      boolean valid = con.isValid(timeout);
+      if (valid) {
+        try {
+          JdbcConnection jdbcCon = new JdbcConnection(con);
+          setMaxAllowedPacket(jdbcCon);
+        } catch (DatabaseException | SQLException e) {
+          LOGGER.error(e.getMessage(), e);
+          throw new ModelDBException(e.getMessage(), e);
+        }
```
anandjakhaniya: `max_allowed_packet` is a global level setting for the MySQL right?? then why we are doing it for each connections?? if we will set it once at boot time then it will not works??

evgeniilezhnin: it will reset every time sql server restart

conradoverta: Doing at every connection seems pretty excessive and slow. Could we use a cronjob instead? That allows us to be pretty certain it has the right value without overwhelming every connection. We don't want to add excessive logic on the critical path.

evgeniilezhnin: yes, cron can be used for that

evgeniilezhnin: The solution became rather complex because there is a problem with the max_allowed_packet option to set it dynamically: https://stackoverflow.com/a/9923059 So for the new value to take effect the code first checks if the value has changed comparing to database, and if it's different sets the new one and runs resetSessionFactory. There can be a problem if in UAC, registry and MDB there will be different values of this parameter because in that case crons will create multiple db connections at the short period of time trying to reset max_allowed_packet

conradoverta: Hm. Let's take a step back and re-think the solution then. It seems to be causing more problems than it might be worth.

evgeniilezhnin: Currently I can think of only 4 solutions:
1. Reset connection on every request.
2. Reset connection on timer.
3. Set parameter on start and hope mysql doesn't restart.
4. Set parameter in sql server.

evgeniilezhnin: I think 3 is better than nothing

conradoverta:
I believe we can drop these with the audit log changes.

PR #2095, opened 2021-04-07
[VR-10493] Alert CRUD

convoliution:
```python
from verta import Client
from verta.operations.monitoring.alert import FixedAlerter

client = Client().operations
monitored_entity = client.get_or_create_monitored_entity("ml-test")
alert = monitored_entity.alerts.create(
    "fixed",
    FixedAlerter(threshold=.7),
    # SummarySampleQuery() from PR#2097
)
```

### Config classes

- Add `_Alerter` config classes (e.g. `FixedAlert(threshold=.7)`) to `verta.operations.monitoring.alert`
- Add `_AlertStatus` config classes (e.g. `Ok()`) to `verta.operations.monitoring.alert.status`

### Entities

- Add `Alert` create & get via `monitored_entity.alerts.create()`, etc.


client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -0,0 +1,231 @@
...
+    def __init__(self, conn, conf, monitored_entity_id):
+        self._conn = conn
+        self._conf = conf
+        self._monitored_entity_id = int(monitored_entity_id)
+
+    def create(
+        self,
+        name,
+        alerter,
+        summary_sample_query=None,
```
convoliution: Should the summary sample query be required for alerts?

convoliution: The answer is "yes"

client/verta/verta/operations/monitoring/alert/status/_status.py
```diff
@@ -0,0 +1,52 @@
...
+        ----------
+        msg : int
+            Variant of ``AlertStatusEnum``.
+
+        Returns
+        -------
+        :class:`_AlertStatus` subclass
+
+        """
+        SUBCLASSES = [
```
daniel-verta: I don't think this is necessary, and in fact we should use the `__subclasses__` attribute because that will stay current with any further subclasses which we introduce.

client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -0,0 +1,274 @@
...
+        notification_channels=None,
+        created_at_millis=None,
+        updated_at_millis=None,
+        last_evaluated_at_millis=None,
+    ):
+        if notification_channels is None:
+            notification_channels = []
+
+        for channel in notification_channels:
+            if isinstance(channel, notification_channel._NotificationChannel):
```
daniel-verta: This looks funny. Should it be `if not isinstance`?

convoliution: I guess this is as confusing as I feared it'd be 😕 

So, there are these config structs that are used to specify different types of notification channels, similar to how the data type classes work
```python
channel_config = SlackNotificationChannel(
    "https://hooks.slack.com/...",
)
```

And then there's the actual entity that exists in our platform
```python
channel_entity = client.notification_channels.create(
    "test-webhook",
    channel_config,
)
```

So what this check here does is let the user know that when creating an alert, if they passed in the struct, they actually need to pass in the entity. I can add a comment explaining this `if` condition, though this may be a deeper issue with the API 😓

client/verta/verta/operations/monitoring/alert/status/_status.py
convoliution: I've been conflicted about using `__subclasses__` because based on what I've read, it seems to be an internal detail of the CPython implementation. Have you seen it used reliably elsewhere, out of curiosity?

daniel-verta: Interesting. Nothing I read seemed to indicate that it was a CPython thing. It's documented in the reference after all. I'm fine with taking this approach if we're in doubt / until we know vet that it's not an implementation specific thing.

client/verta/verta/operations/monitoring/alert/_entities/alert.py
daniel-verta: Sure, but wouldn't we still want to test against the expected/supported type and raise an error if we get something else? For example, if we had a list of animals which really needed to all be dogs, it would look funny to raise an error if any of them were cats because then an error wouldn't be raised on elements which are birds.

client/verta/verta/operations/monitoring/alert/status/_status.py
convoliution: For completeness:

My
> seems to be an internal detail of the CPython implementation

perception comes from [this frequently cited old entry in the Python mailing list](https://mail.python.org/pipermail/python-list/2003-August/210297.html):
> In short, \_\_subclasses__ exists to support efficient runtime class mutation
at the C level.

and the [Python reference](https://docs.python.org/2/library/stdtypes.html?highlight=__subclasses__#special-attributes) mentions (emphasis mine):
> **The implementation** adds a few special read-only attributes to several object types . . .

If I find some time to get PyPy running on my machine, I could verify this eventually 😅

convoliution:
Linters are failing because this PR is dependent on https://github.com/VertaAI/modeldb/pull/2096.
I'll merge that one first and see if that resolves the issues.

client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -194,16 +222,16 @@ def create(
         updated_at_millis=None,
         last_evaluated_at_millis=None,
     ):
+        if self._monitored_entity_id is None:
+            raise RuntimeError(
+                "this Alerts cannot be used to create because it was not"
```
daniel-verta:
```suggestion
                "this Alert cannot be used to create because it was not"
```

daniel-verta (approved):

PR #2106, opened 2021-04-11
Backport existing code to start ER optimizations

conradoverta:


backend/common/src/main/java/ai/verta/modeldb/common/config/GrpcServerConfig.java
```diff
@@ -2,6 +2,7 @@
 
 public class GrpcServerConfig {
   public int port;
+  public int threadCount = 8;
```
anandjakhaniya, line 5: Why here 8 is the ideal one??

conradoverta: Just a number that isn't zero for backwards compatibility with the existing config =) There's no "ideal" one for the number of threads.

ghost (approved):

PR #2112, opened 2021-04-13
Add method to convert list of futures to future of list

ghost:
Used the standard `sequence` strategy (ex http://gene-t-taylor.com/posts/2016-07-20-CompletableFuture-sequence.html)


conradoverta (approved):
A bit wasteful but sure =)

PR #2120, opened 2021-04-14
[VR-10693] Add comparisons for FixedAlerter

convoliution:
## Changes

- add `_VertaComparison` classes to `verta.common.comparison` submodule
- add unit tests for comparison classes
- incorporate comparison classes into tests for `FixedAlerter` and the `Alert` entity

### DIFF
--- a/client/verta/verta/common/comparison/_verta_comparison.py
+++ b/client/verta/verta/common/comparison/_verta_comparison.py
@@ -14,7 +14,7 @@ class _VertaComparison(object):
         self._value = value
 
     def __repr__(self):
-        return '<comparison "{} {}">'.format(
+        return '<comparison ({} {})>'.format(
             self._SYMBOL,
             self.value,
         )
--- a/client/verta/verta/common/comparison/_verta_comparison.py
+++ b/client/verta/verta/common/comparison/_verta_comparison.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+
+import abc
+
+from verta.external import six
+
+from verta._internal_utils import _utils
+
+
+@six.add_metaclass(abc.ABCMeta)
+class _VertaComparison(object):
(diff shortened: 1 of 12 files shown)


client/verta/verta/common/comparison/_verta_comparison.py
```diff
@@ -0,0 +1,84 @@
...
+
+        """
+        for subcls in cls.__subclasses__():
+            if msg == subcls._OPERATOR:
+                return subcls(value)
+
+        raise ValueError("operator {} not recognized".format(msg))
+
+    @abc.abstractmethod
+    def compare(self, other_value):
```
daniel-verta, line 69: This is a case where we might think of implementing `__call__` instead of having this named `compare` method, because instances of `_VertaComparison` are basically functions which return boolean values according to the comparison. See the example ipython session below for what I'm talking about. I like this approach but if it feels overly-clever and not fluent we don't need to do this.

```
In [1]: class Comparison(object):
   ...:     def __init__(self, threshold):
   ...:         self.threshold = threshold
   ...:     def __call__(self, value):
   ...:         return self.threshold > value
   ...:

In [2]: comp = Comparison(0.5)

In [3]: comp(0.3)
Out[3]: True

In [4]: comp(0.7)
Out[4]: False
```

daniel-verta (approved):
A non-essential suggestion. Functionality LGTM.

client/verta/verta/common/comparison/_verta_comparison.py
convoliution: I'd normally jump at opportunities to tap into Python's object model, but this case I'm wary about whether `__call__` offers a practical advantage, and `compare()` to me makes the API more explicit and therefore potentially readable.

I'd leave it open as an _option_ if we ever decide to allow, say, arbitrary callables as comparisons (which our current proto-based system wouldn't accommodate)

PR #2121, opened 2021-04-14
[VR-10716] Enable unzipping directories with download_model()

convoliution:
An artifact needs to be assigned `ZIP_EXTENSION` in order for the `download_*()` methods to recognize them as zipped by the client.

`run.log_artifact()` accomplished this because the extension is determined directly from the filename, but the `log_model()` path determines it based on the model's serialization method. This PR ensures that path recognized artifacts ZIPed by the client.


conradoverta (approved):
LGTM

PR #2126, opened 2021-04-15
[VR-10699] Capture summary samples in alert status objects

convoliution:
**before**
```python
from verta.operations.monitoring.alert.status import Alerting, Ok
alert.set_status(Alerting(), monitored_entity1)
alert.set_status(Alerting(), monitored_entity2)
alert.set_status(Ok())
```
**after**
```python
from verta.operations.monitoring.alert.status import Alerting, Ok
alert.set_status(Alerting([monitored_entity1, monitored_entity2]))
alert.set_status(Ok())
```

## Changes

- move "triggering summary samples" parameter from `alert.set_status()` to alert status objects (e.g. `Alerting()`)
- add docstrings for alert status objects, while I'm at it
- improve implementation of `alert.status` property (ignore violating summary samples if status is `Ok()`)

### DIFF
--- a/client/verta/verta/operations/monitoring/alert/_entities/alert.py
+++ b/client/verta/verta/operations/monitoring/alert/_entities/alert.py
@@ -192,10 +192,10 @@ class Alert(entity._ModelDBEntity):
 
         self._update(alert_msg)
 
-    def set_status(self, status, event_time_millis=None):
+    def set_status(self, status, event_time=None):
         msg = _AlertService.UpdateAlertStatusRequest(
             alert_id=self.id,
-            event_time_millis=event_time_millis,
+            event_time_millis=time_utils.epoch_millis(event_time) if event_time else None,
             status=status._ALERT_STATUS,
             violating_summary_sample_ids=status._sample_ids,
         )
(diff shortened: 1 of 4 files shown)


client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -182,13 +192,12 @@ def add_notification_channels(self, notification_channels):
 
         self._update(alert_msg)
 
-    # TODO: alternatively, fire() & resolve()?
-    def set_status(self, status, summary_sample=None, event_time_millis=None):
+    def set_status(self, status, event_time_millis=None):
```
daniel-verta: I think we should follow the example of `Summary` and `SummarySample` and use `time_utils` to accept `datetime` objects or epoch milliseconds, and by storing the `datetime` on objects and only encoding to millis for proto interactions, we can support better string representations of these objects.

daniel-verta (changes requested):
One change requested to continue using time objects which will be more natural than epoch millis.

client/verta/verta/operations/monitoring/alert/_entities/alert.py
convoliution: 1e0094bd5c8bfccba8b584c36599efe8bd653370

daniel-verta (approved):

PR #2128, opened 2021-04-15
Add compare() to comparisons

convoliution:
Basically re-invented Polish notation.

I found it very useful for the alerter for the actual comparison logic to be a part of the `_VertaComparison` interface.

### DIFF
--- a/client/verta/verta/common/comparison/_verta_comparison.py
+++ b/client/verta/verta/common/comparison/_verta_comparison.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import abc
+import math
 
 from verta.external import six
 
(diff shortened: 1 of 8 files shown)


client/verta/verta/common/comparison/_verta_comparison.py
```diff
@@ -24,9 +25,60 @@ def __repr__(self):
             self.value,
         )
 
+    @staticmethod
+    def isclose(a, b, rel_tol=1e-9, abs_tol=0):
```
daniel-verta, line 29: This is a candidate for one of our utils modules but NBD for it to live here until there's a second use for it in the codebase. 👍

daniel-verta (approved):

client/verta/verta/common/comparison/_verta_comparison.py
convoliution: Great point!

PR #2129, opened 2021-04-15
Removed return project from the response of CRUD attribute endpoints

anandjakhaniya:


conradoverta (approved):
We should also remove from the existing ProjectService too to avoid breaking the build later. Different PR though

anandjakhaniya:
> We should also remove from the existing ProjectService too to avoid breaking the build later. Different PR though

Sure. when removing proto i will remove it from old project service as well.

PR #2130, opened 2021-04-15
[VR-10670] Async artifacts CRUD endpoints for the experiment runs

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10670


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -274,4 +280,39 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    final var now = Calendar.getInstance().getTimeInMillis();
+
+    return checkPermission(runId, ModelDBActionEnum.ModelDBServiceActions.UPDATE)
+        .thenCompose(unused -> artifactHandler.logArtifacts(runId, artifacts), executor)
+        .thenCompose(unused -> updateModifiedTimestamp(runId, now), executor);
+  }
+
+  public InternalFuture<List<Artifact>> getArtifacts(GetArtifacts request) {
+    final var runId = request.getId();
+    final var key = request.getKey();
```
conradoverta, line 306: `key` is always present, so `Optional.of(key)` will never be empty. You probably wanted to check if the key string was empty, right?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -274,4 +280,39 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    final var keys =
+        request.getKey().isEmpty()
+            ? new ArrayList<String>()
+            : Collections.singletonList(request.getKey());
+
+    return checkPermission(runId, ModelDBActionEnum.ModelDBServiceActions.UPDATE)
+        .thenCompose(
+            unused ->
+                artifactHandler.deleteArtifacts(
+                    runId, keys.isEmpty() ? Optional.empty() : Optional.of(keys)),
```
conradoverta: Like this. Also, let's avoid inlining these logic as they make reading more complicated. Extracting to a var like `maybeKeys` will make things clear.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
conradoverta, line 306: Also, doesn't this have more than one key?

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/ArtifactHandler.java
```diff
@@ -0,0 +1,223 @@
...
+                            .bind("type", artifact.getArtifactTypeValue())
+                            .bind("path_only", artifact.getPathOnly())
+                            .bind("linked_artifact_id", artifact.getLinkedArtifactId())
+                            .bind("filename_extension", artifact.getFilenameExtension())
+                            .bind(
+                                "store_type_path",
+                                !artifact.getPathOnly()
+                                    ? Config.getInstance().artifactStoreConfig.storeTypePathPrefix()
+                                        + artifact.getPath()
+                                    : "")
```
conradoverta: Extract to avoid complex inline logic in the middle of a clean `bind` sequence.

conradoverta (commented):
Some minor asks

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya:
```
message GetArtifacts {
    string id = 1;
    string key = 2;

    message Response {
        repeated common.Artifact artifacts = 1;
    }
}
```
So i have take key as optional. If only id is there then it will returns all artifact of that run and if key is present then matching key artifacts will be return.

anandjakhaniya: Yes. it should have but currently in proto it is not 
```
message GetArtifacts {
    string id = 1;
    string key = 2;

    message Response {
        repeated common.Artifact artifacts = 1;
    }
}
```

conradoverta: I mean that `request.getKey()` will always return a valid string. That's protobuf's internal implementation

anandjakhaniya: Ok, Updating it without Optional

anandjakhaniya: Removed optional for key

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Extracted to a var

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/ArtifactHandler.java
anandjakhaniya: Extracted to a var

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
conradoverta: That's not what I meant. Fixing

conradoverta: Optional was the right choice. But its use was incorrect.

PR #2131, opened 2021-04-15
[VR-10674] Implemented async for delete experimentRuns endpoint

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10674


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -215,8 +231,19 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
                 .execute());
   }
 
+  private InternalFuture<Void> updateProjectModifiedTimestamp(List<String> runIds, Long now) {
```
conradoverta: This manipulates project. It should be in the project DAO.

We should also leave this to the parent timestamp reconciliation and soft delete reconciliation. Avoid doing db operations while the customer is waiting if you can.

conradoverta (changes requested):
One PR, two unrelated changes: implement delete and change the signature of method to a list. These should be two PRs.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: removed updateProjectModifiedTimestamp from run dao it will handle by reconcile utils

anandjakhaniya:
> One PR, two unrelated changes: implement delete and change the signature of method to a list. These should be two PRs.

extracted list changes in separate PR https://github.com/VertaAI/modeldb/pull/2139

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -283,13 +283,53 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+              return checkProjectPermission(
+                  maybeProjectIds, ModelDBActionEnum.ModelDBServiceActions.UPDATE);
             default:
               return checkProjectPermission(maybeProjectIds, action);
           }
         },
         executor);
   }
 
+  public InternalFuture<Void> deleteExperimentRuns(DeleteExperimentRuns request) {
```
conradoverta, line 297: Don't carry bad signatures forward. This shouldn't return anything.

PR #2132, opened 2021-04-15
[VR-10512] Document New Monitoring Features

daniel-verta:


client/verta/verta/operations/monitoring/profiler.py
```diff
@@ -45,25 +123,70 @@ def _profile_column(self, df, column):
...
         return (column + "_histogram", DiscreteHistogram(values, keys))
 
 
+# TODO: Consider design/interface for different bins
 class ContinuousHistogramProfiler(Profiler):
+    """Produces float histograms counting values according to bin intervals.
+
+    Attributes
+    ----------
+    columns : :obj:`list` of :obj:`str`
```
convoliution: From my past experience, the `:obj:` directive doesn't actually do anything. _But_ we can get the link to the std docs without them, using https://github.com/VertaAI/modeldb/pull/2142.
```suggestion
    columns : list of str
```

client/verta/verta/operations/monitoring/monitored_entity.py
```diff
@@ -16,7 +16,37 @@
...
+    name
+    workspace
+    alerts
+    """
+
     def __init__(self, conn, conf, msg):
+        """Intended for internal use only.
+
+        Users should obtain a monitored entity through the
+        ``get_or_create_monitored_entity`` method of the operations sub-client.
```
convoliution:
```suggestion
        :meth:`~verta.opertaions.monitoring.client.Client.get_or_create_monitored_entity`
        method of the operations sub-client.
```

client/verta/verta/operations/monitoring/client.py
```diff
@@ -33,15 +54,58 @@ def _conf(self):
...
     def notification_channels(self):
+        """Notification channel repository."""
         return NotificationChannels(self._conn, self._conf)
 
     def get_or_create_monitored_entity(self, name=None, workspace=None, id=None):
+        """Get or create a monitored entity by name.
+
+        Gets or creates a monitored entity. A name will be auto-generated if one
+        is not provided. Either ``name`` or ``id`` can be provided but not both.
+        If id is provided, this will act only as a get method and no object will
```
convoliution: I can't say whether it's sensible, but the convention I've been following (which I gleaned from reading NumPy's docstrings) is to use one backtick for parameter names (renders in italics), and two backticks for arbitrary code snippets (renders as monospace).
```suggestion
        is not provided. Either `name` or ``id` can be provided but not both.
        If `id` is provided, this will act only as a get method and no object will
```

client/verta/verta/operations/monitoring/client.py
```diff
@@ -33,15 +54,58 @@ def _conf(self):
...
+        be created.
+
+        Parameters
+        ----------
+        name : string, optional
```
convoliution: These'll need to be `str` rather than `string` to link to the std lib
```suggestion
        name : str, optional
```

client/verta/verta/operations/monitoring/client.py
```diff
@@ -33,15 +54,58 @@ def _conf(self):
...
+            A unique name for this monitored entity.
+        workspace: string, optional
+            A workspace for this entity. Defaults to the client's default workspace.
+        id : int, optional
+            This should not be provided if ``name`` is provided.
+        
+        Returns
+        -------
+        verta.operations.monitoring.monitored_entity.MonitoredEntity
```
convoliution: This will let Sphinx insert a link when doc pages actually exist.
```suggestion
        :class:`~verta.operations.monitoring.monitored_entity.MonitoredEntity`
```

client/verta/verta/operations/monitoring/labels.py
```diff
@@ -12,23 +12,46 @@
 from verta._internal_utils import time_utils
 from .utils import maybe
 
-class Labels:
+
+class Labels(object):
+    """
+    Repository object for finding labels.
```
convoliution, line 18: I wonder if "Repository object" would be confusing, since our platform has an actual standalone concept of `Repository`. I'd favor "Collection object", but I'm not sure if there's a standard nomenclature for this sort of concept.

client/verta/verta/operations/monitoring/profilers.py
```diff
@@ -227,20 +343,23 @@ def create(self, name, model_version):
...
+        """Delete the provided profiler reference.
 
-    @staticmethod
-    def enable(profiler_reference, monitored_entity):
-        return profiler_reference.enable(monitored_entity)
-
-    @staticmethod
-    def disable(profiler_reference, monitored_entity):
-        return profiler_reference.disable(monitored_entity)
+        Instructs Verta Services to delete .
```
convoliution: Extended summaries may not be necessary for some of these! Unless `pydocstyle` favors them.

client/verta/verta/operations/monitoring/profiler.py
```diff
@@ -1,41 +1,119 @@
 # -*- coding: utf-8 -*-
+
 import collections
 
+import numpy as np
```
convoliution: This'll need `maybe_dependency()`!

client/verta/verta/operations/monitoring/monitored_entity.py
```diff
@@ -100,6 +148,20 @@ def _update(self, msg, response_proto, endpoint, method):
...
+
+        Returns
+        -------
+        bool
+            True if the delete was successful.
+
+        Raises
+        ------
+        Error
+            Raises an error if the delete failed for any reason.
```
convoliution, line 176: I personally usually forgo the `Raises` section unless there's a particular exception case that's unintuitive/surprising. I'm open to cataloguing them more thoroughly if that seems to be a user-friendlier approach.

client/verta/verta/operations/monitoring/labels.py
daniel-verta: [_Repository_](https://martinfowler.com/eaaCatalog/repository.html) might regrettably also be the most appropriate term for this object's role in the library. That said, we don't yet have a consistent framework or practice which we've settled upon in our library, so an alternative term can be found until we have a framework which forces the clarification of language.

client/verta/verta/operations/monitoring/monitored_entity.py
daniel-verta: Fair enough.

convoliution: Actually, it might make sense for `delete()`. The function returns `True` if the delete succeed, but raises an exception otherwise as opposed to returning `False`.

client/verta/verta/operations/monitoring/profilers.py
```diff
@@ -190,6 +287,7 @@ def __init__(self, conn, conf, client):
         self._conf = conf
         self._client = client
 
+    # TODO: Hide this as an internal method.
```
convoliution: Are `list()` and `get()` meant to be internal? I thought they were intended to be user-facing 😮

client/verta/verta/operations/monitoring/client.py
```diff
@@ -33,15 +54,58 @@ def _conf(self):
...
     @property
     def notification_channels(self):
+        """Notification channel repository."""
         return NotificationChannels(self._conn, self._conf)
 
     def get_or_create_monitored_entity(self, name=None, workspace=None, id=None):
+        """Get or create a monitored entity by name.
+
+        Gets or creates a monitored entity. A name will be auto-generated if one
+        is not provided. Either `name` or ``id` can be provided but not both.
```
convoliution:
```suggestion
        is not provided. Either `name` or `id` can be provided but not both.
```

client/verta/verta/operations/monitoring/monitored_entity.py
```diff
@@ -16,7 +16,38 @@
...
+            A connection object to the backend service.
+        conf
+            A configuration object used by conn methods.
+        msg
+            A protobuf message ai.verta.monitoring.MonitoredEntity
+        Returns
+        -------
+        MonitoredEntity
+            A monitored entity which is persisted by Verta.
+        """
```
convoliution: I'm not sure what `pydocstyle` has in mind with `D107`, but the [NumPy style guide](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring) says `__init__()` is meant to be documented in the class docstring 🤔 I'm open to splitting them, but it's not a convention I've been following.

client/verta/verta/operations/monitoring/profilers.py
daniel-verta: No I think that comment may have moved around when it was supposed to earmark the `create` method as internal, which only creates a profiler reference.

client/verta/verta/operations/monitoring/monitored_entity.py
daniel-verta: That's probably correct and I'll adjust these things accordingly.

convoliution (approved):

PR #2134, opened 2021-04-15
Return everything from summary /find requests

convoliution:
Follow-up to #2133


client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -25,7 +25,15 @@
...
 class SummaryQuery(object):
-    def __init__(self, ids=None, names=None, data_type_classes=None, monitored_entities=None):
+    def __init__(
+        self,
+        ids=None,
+        names=None,
+        data_type_classes=None,
+        monitored_entities=None,
+        page_number=1,
+        page_limit=-1,
```
daniel-verta: Can we hide this value and use `None` in the Python client for "give me everything" and convert that to the appropriate -1 value for the proto interface?

daniel-verta (approved):
Approved with a suggestion

client/verta/verta/operations/monitoring/summaries.py
convoliution: That certainly makes a lot more sense!

PR #2140, opened 2021-04-16
[VR-10673] Async create experiment run

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10673


backend/src/main/java/ai/verta/modeldb/utils/TrialUtils.java
```diff
@@ -47,6 +50,17 @@ public static void validateMaxArtifactsForTrial(
     }
   }
 
+  public static InternalFuture<ExperimentRun> futureValidateMaxArtifactsForTrial(
+      TrialConfig config,
+      ExperimentRun newExperimentRun,
```
conradoverta: Having to pass the whole ER here is a bad sign. It usually means that you're collecting more data than needed to make the decision. For example, do you need to load the observations?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
         .thenCompose(unused -> artifactHandler.deleteArtifacts(runId, optionalKeys), executor)
         .thenCompose(unused -> updateModifiedTimestamp(runId, now), executor);
   }
+
+  private InternalFuture<UserInfo> getCurrentLoginUserInfo() {
+    return FutureGrpc.ClientRequest(
+        uac.getUACService().getCurrentUser(Empty.newBuilder().build()), executor);
+  }
```
conradoverta: Remove

https://github.com/VertaAI/docs/blob/main/engineering-practices/api-implementation/performance.md#avoid-layers-of-abstraction

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+              }
+            },
+            executor);
+
+    return futureTask
+        .thenCompose(
+            unused ->
+                checkProjectPermission(
+                    Collections.singletonList(request.getProjectId()),
+                    ModelDBActionEnum.ModelDBServiceActions.UPDATE),
```
conradoverta: We need to check if the user can update the project and create the ER. We're missing the create part.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+
+  private InternalFuture<ExperimentRun> checkIfEntityAlreadyExists(
+      ExperimentRun experimentRun, Boolean isInsert) {
+    return jdbi.useHandle(
+            handle -> {
+              try {
+                String queryStr;
+                if (isInsert) {
+                  queryStr =
+                      "SELECT id FROM experiment_run ere WHERE "
```
conradoverta: Avoid weird name bindings. It really hurts readability. In this case, it's not even necessary.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+                          + ModelDBConstants.NAME
+                          + " = :experimentRunName "
+                          + " AND ere."
+                          + ModelDBConstants.PROJECT_ID
+                          + " = :projectId "
+                          + " AND ere."
+                          + ModelDBConstants.EXPERIMENT_ID
+                          + " = :experimentId "
+                          + " AND ere."
+                          + ModelDBConstants.DELETED
```
conradoverta: Avoid this constant patching with constants. It really hurts readability. Just type the field name instead. It's really bad practice to depend on a kitchen sink central list of constants.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    return builder.build();
+  }
+
+  private InternalFuture<ExperimentRun> checkIfEntityAlreadyExists(
+      ExperimentRun experimentRun, Boolean isInsert) {
+    return jdbi.useHandle(
+            handle -> {
+              try {
+                String queryStr;
+                if (isInsert) {
```
conradoverta: They're different enough that we should split into 2 methods instead of patching all across the code to account for this flag.

https://github.com/VertaAI/docs/blob/main/engineering-practices/api-implementation/performance.md#avoid-layers-of-abstraction

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+
+                if (isInsert) {
+                  query.bind("experimentRunName", experimentRun.getName());
+                  query.bind("projectId", experimentRun.getProjectId());
+                  query.bind("experimentId", experimentRun.getExperimentId());
+                } else {
+                  query.bind(ModelDBConstants.EXPERIMENT_RUN_ID_STR, experimentRun.getId());
+                }
+
+                long count = query.mapTo(String.class).stream().count();
```
```suggestion
                Long count = query.mapTo(Long.class).one();
```
Push the count to the DB. Don't do the count in Java. It's wasteful.

https://github.com/VertaAI/docs/blob/main/engineering-practices/api-implementation/performance.md#simple-db-operations-with-indices

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+                          + " = :projectId "
+                          + " AND ere."
+                          + ModelDBConstants.EXPERIMENT_ID
+                          + " = :experimentId "
+                          + " AND ere."
+                          + ModelDBConstants.DELETED
+                          + " = false ";
+                } else {
+                  queryStr =
+                      "SELECT id FROM experiment_run ere WHERE "
```
```suggestion
                      "SELECT COUNT(id) FROM experiment_run ere WHERE "
```

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```suggestion
                      "SELECT COUNT(id) FROM experiment_run ere WHERE "
```

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+
+                // Throw error if it is an insert request and ExperimentRun with same name already
+                // exists
+                if (existStatus && isInsert) {
+                  throw new AlreadyExistsException("ExperimentRun already exists in database");
+                } else if (!existStatus && !isInsert) {
+                  // Throw error if it is an update request and ExperimentRun with given name does
+                  // not exist
+                  throw new NotFoundException("ExperimentRun does not exist in database");
+                }
```
conradoverta: Don't throw exceptions around to pass information between methods. Use exceptions for what they're designed to do: an exceptional scenario that you were not expecting.

In this case, the right approach is to return an Optional (or boolean) and let the caller do what it wants with the result.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+              } catch (Exception ex) {
+                if (ModelDBUtils.needToRetry(ex)) {
+                  checkIfEntityAlreadyExists(experimentRun, isInsert);
+                } else {
+                  throw ex;
+                }
+              }
```
conradoverta: Remove. We need to stop duplicating this code. It's hard to maintain and use.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+            })
+        .thenCompose(unused -> InternalFuture.completedInternalFuture(experimentRun), executor);
```
conradoverta: The caller is already sending the ER to you. There's no need to send it back to them. This method should return an optional.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+  }
+
+  private String buildRoleBindingName(
```
conradoverta: This is already present somewhere else as we have to delete the role binding. Re-use that instead.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+                                        .build())
+                                .build())
+                        .build()),
+            executor)
+        .thenAccept(
+            response -> {
+              LOGGER.trace(CommonMessages.ROLE_SERVICE_RES_RECEIVED_TRACE_MSG, response);
+            },
+            executor)
+        .thenCompose(unused -> InternalFuture.completedInternalFuture(experimentRun), executor);
```
conradoverta: Don't return the thing that the caller sent you.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+                errorMessage = "Experiment ID not found in CreateExperimentRun request";
+              }
+
+              if (errorMessage != null) {
+                throw new InvalidArgumentException(errorMessage);
+              }
+            },
+            executor);
+
+    return futureTask
```
conradoverta: This is a massive chain of `thenCompose` with a lot of methods that 1) return what was passed to them and 2) returning a future for things that happen immediately. This is an anti-pattern as it makes the code more complicated than it has to be. This needs to be split into actually meaningful methods/lambdas.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+                      .append(commaFields)
+                      .append(") values (");
+
+              for (int i = 0; i < fieldsArr.length; i++) {
+                queryStrBuilder.append(":").append(fieldsArr[i]);
+                if (i < fieldsArr.length - 1) {
+                  queryStrBuilder.append(",");
+                }
+              }
+              queryStrBuilder.append(" ) ");
```
conradoverta: This SQL command is very hard to read, which means it's unmaintainable and has a bug for all I care. We need to make this more readable by humans.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    }
+
+    return InternalFuture.completedInternalFuture(experimentRunBuilder.build());
+  }
+
+  private InternalFuture<ExperimentRun> insertExperimentRun(ExperimentRun newExperimentRun) {
+    final var now = Calendar.getInstance().getTimeInMillis();
+    return jdbi.withHandle(
+            handle -> {
+              Map<String, Object> runValueMap = new LinkedHashMap<>();
```
conradoverta: Only acceptable exception to the rule of not using `Object`.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+              var query = handle.createUpdate(queryStrBuilder.toString());
+              for (Map.Entry<String, Object> objectEntry : runValueMap.entrySet()) {
+                query.bind(objectEntry.getKey(), objectEntry.getValue());
+              }
+              query.execute();
+
+              return newExperimentRun;
+            })
+        .thenCompose(
+            handle -> tagsHandler.addTags(newExperimentRun.getId(), newExperimentRun.getTagsList()),
```
conradoverta: `handle` will already have been closed. Plus it's not used. Plus we're not putting this in a transaction, which is buggy since we don't have a `created` flag for ERs.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+            executor)
+        .thenCompose(
+            handle ->
```
conradoverta: It's definitely not a handle here anymore. Naming variables wrong will cause bugs.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+                observationHandler.logObservations(
+                    newExperimentRun.getId(), newExperimentRun.getObservationsList(), now),
+            executor)
+        // TODO .thenCompose(handle -> featureHandler.logFeatures(newExperimentRun.getId(),
+        // newExperimentRun.getFeaturesList()), executor)
+        // TODO .thenCompose(handle -> addCodeVersionSnapShot(), executor)
+        // TODO .thenCompose(handle -> versioned_inputs, executor)
+
+        .thenCompose(unused -> InternalFuture.completedInternalFuture(newExperimentRun), executor);
```
conradoverta: Don't return the argument to the caller.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
         .thenCompose(unused -> artifactHandler.deleteArtifacts(runId, optionalKeys), executor)
         .thenCompose(unused -> updateModifiedTimestamp(runId, now), executor);
   }
+
+  private InternalFuture<UserInfo> getCurrentLoginUserInfo() {
+    return FutureGrpc.ClientRequest(
+        uac.getUACService().getCurrentUser(Empty.newBuilder().build()), executor);
+  }
+
+  public InternalFuture<ExperimentRun> createExperimentRun(final CreateExperimentRun request) {
```
conradoverta: Let's remove this whole creation process to a helper class. This code is massive and having it mixed with other core functionality will hinder readability.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -365,4 +418,362 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    // Validate arguments
+    var futureTask =
+        InternalFuture.runAsync(
+            () -> {
+              String errorMessage = null;
+              if (request.getProjectId().isEmpty() && request.getExperimentId().isEmpty()) {
```
conradoverta: No need to check both, then check one and the other. This leads to exponential growth in checks. Just check the individual fields.

conradoverta (changes requested):
Core code is reasonable, but built on a lot of bad practices. I will no longer allow these bad practices to go in.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
```diff
@@ -0,0 +1,381 @@
...
+
+          long count = query.mapTo(Long.class).one();
+          return count > 0;
+        });
+  }
+
+  private String buildRoleBindingName(
+      String roleName, String resourceId, String vertaId, String resourceTypeName) {
+    return roleName + "_" + resourceTypeName + "_" + resourceId + "_" + "User_" + vertaId;
+  }
```
conradoverta, line 344: This is already defined somewhere, right? As we have to use it for deletion too.

backend/src/main/java/ai/verta/modeldb/utils/TrialUtils.java
anandjakhaniya: Removed whole object from here

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Updated query without binding name

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed constant fields from query and added fields directly in query  string

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed and created function only for the insert one

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: updated

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed update part for now because it is not an used

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: updated with count query

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Updated code as per suggestions

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Updated code as per suggestions

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: updated code with InternalFuture void

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Updated this massive code with help of you.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Improve for loop with inline labda and simplify it

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Code is modify with better solution

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Code is modify with better solution

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Code is modify with better solution

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Moved to helper class called `CreateExperimentRunHandler`

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed this check

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/2185

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/2185

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
```diff
@@ -0,0 +1,384 @@
...
+    this.uac = uac;
+
+    attributeHandler = new AttributeHandler(executor, jdbi, "ExperimentRunEntity");
+    hyperparametersHandler =
+        new KeyValueHandler(executor, jdbi, "hyperparameters", "ExperimentRunEntity");
+    metricsHandler = new KeyValueHandler(executor, jdbi, "metrics", "ExperimentRunEntity");
+    observationHandler = new ObservationHandler(executor, jdbi);
+    tagsHandler = new TagsHandler(executor, jdbi, "ExperimentRunEntity");
+    artifactHandler = new ArtifactHandlerBase(executor, jdbi, "artifacts", "ExperimentRunEntity");
+    featureHandler = new FeatureHandler(executor, jdbi, "ExperimentRunEntity");
```
conradoverta: Don't create new via copy and paste. Re-use the existing ones. Follow the pattern already shown in other handlers.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
```diff
@@ -0,0 +1,384 @@
...
+            .addAllTags(ModelDBUtils.checkEntityTagsLength(request.getTagsList()))
+            .addAllAttributes(request.getAttributesList())
+            .addAllHyperparameters(request.getHyperparametersList())
+            .addAllArtifacts(request.getArtifactsList())
+            .addAllDatasets(request.getDatasetsList())
+            .addAllMetrics(request.getMetricsList())
+            .addAllObservations(request.getObservationsList())
+            .addAllFeatures(request.getFeaturesList());
+
+    if (request.getDateCreated() != 0L) {
```
conradoverta, line 187: Need to check the two fields independently

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
```diff
@@ -0,0 +1,384 @@
...
+      experimentRunBuilder
+          .setDateCreated(request.getDateCreated())
+          .setDateUpdated(request.getDateCreated());
+    } else {
+      experimentRunBuilder
+          .setDateCreated(Calendar.getInstance().getTimeInMillis())
+          .setDateUpdated(Calendar.getInstance().getTimeInMillis());
```
conradoverta: Use exactly the same time as default

conradoverta (commented):
Some comments. Please don't update this PR manually anymore bc the method is gigantic. Send PRs on top of this branch instead.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: i do not follow you on this. Could you explain what is wrong here??

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/2194

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/2194

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: I have used whatever we have and if it is not anywhere in your or mine PR then i am creating new one.

conradoverta: This looks like a copy and paste of this: https://github.com/VertaAI/modeldb/blob/master/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java#L42

anandjakhaniya: Ok, got it so mean i have to use those in parameters for create handler. ok, i will do that. thanks for clarification.

anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/2196/files

conradoverta (approved):

PR #2144, opened 2021-04-16
Populate linked datasets based on privileges in runs while creating ER and Find ER response

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -259,8 +261,10 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
                 .execute());
   }
 
-  private InternalFuture<Void> checkProjectPermission(
-      String projId, ModelDBActionEnum.ModelDBServiceActions action) {
+  private InternalFuture<Void> checkEntityPermission(
```
conradoverta: Creating this generic method is confusing to the reader. For example, when should I use `checkPermission` or `checkEntityPermission`? This will lead to bugs by using the wrong method.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -384,6 +390,83 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    List<String> accessibleDatasetVersionIds = new ArrayList<>();
+    List<String> newDatasetLinkedIds = new ArrayList<>();
+    for (Artifact dataset : newDatasets) {
+      String datasetVersionId = dataset.getLinkedArtifactId();
+      if (!datasetVersionId.isEmpty()) {
+        newDatasetLinkedIds.add(datasetVersionId);
+      }
+    }
+
+    jdbi.useHandle(
```
conradoverta: You're accessing the db, so you _have_ to use the returning future. This just introduced a race condition.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -384,6 +390,83 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    return experimentRunBuilder.build();
+  }
+
+  /**
+   * @param newDatasets : new datasets for privilege check
+   * @param errorOut : Throw error while creation (true) otherwise we will keep it silent (false)
+   * @return {@link List} : accessible datasets
+   * @throws ModelDBException: modelDBException
+   */
+  private List<Artifact> getPrivilegedDatasets(List<Artifact> newDatasets, boolean errorOut) {
```
conradoverta: Please add a description to the method.

conradoverta: Passing flags to decide if silent or exception is a hack. The lower level method shouldn't implement business logic for the caller. Return `List<Optional<Artifact>>` instead and let the caller handle it however they want.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -384,6 +390,83 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+    List<Artifact> accessibleDatasets = new ArrayList<>();
+    List<String> accessibleDatasetVersionIds = new ArrayList<>();
+    List<String> newDatasetLinkedIds = new ArrayList<>();
```
conradoverta: `newDatasetLinkedIds` is a very non-descriptive name for what this is

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -384,6 +390,83 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+            if (!datasetVersionId.isEmpty()
+                && !accessibleDatasetVersionIds.contains(datasetVersionId)) {
+              try {
+                if (datasetVersionDatasetMap.containsKey(datasetVersionId)
+                    && datasetVersionDatasetMap.get(datasetVersionId) != null) {
+                  Long datasetId = datasetVersionDatasetMap.get(datasetVersionId);
+                  checkEntityPermission(
+                      String.valueOf(datasetId),
+                      ModelDBActionEnum.ModelDBServiceActions.READ,
+                      ModelDBServiceResourceTypes.DATASET);
```
conradoverta: Never do API call while you're holding a connection to the database.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -384,6 +390,83 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+
+          for (Artifact dataset : newDatasets) {
+            String datasetVersionId = dataset.getLinkedArtifactId();
+            if (!datasetVersionId.isEmpty()
+                && !accessibleDatasetVersionIds.contains(datasetVersionId)) {
+              try {
+                if (datasetVersionDatasetMap.containsKey(datasetVersionId)
+                    && datasetVersionDatasetMap.get(datasetVersionId) != null) {
+                  Long datasetId = datasetVersionDatasetMap.get(datasetVersionId);
+                  checkEntityPermission(
```
conradoverta: You're ignoring the return of `checkEntityPermission`, which just introduces a race condition.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -384,6 +390,83 @@ public FutureExperimentRunDAO(Executor executor, FutureJdbi jdbi, UAC uac) {
...
+                } else {
+                  throw new InvalidArgumentException(
+                      "Dataset not found for dataset version: " + datasetVersionId);
+                }
+              } catch (Exception ex) {
+                LOGGER.debug(ex.getMessage());
+                if (errorOut) {
+                  throw ex;
+                }
+              }
```
conradoverta: This is bad practice. So if we fail to call UAC, we just skip silently saying that the user doesn't have access? That is communicating something different from what the user expects.

conradoverta (changes requested):
Lots of old bad practices that killed performance or interpretability. Added comments on how to fix them.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
conradoverta: You seem to have forgotten about this

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -819,9 +900,22 @@ public FutureExperimentRunDAO(
...
-    return checkProjectPermission(
+    return checkEntityPermissionBasedOnResourceTypes(
             Collections.singletonList(request.getProjectId()),
-            ModelDBActionEnum.ModelDBServiceActions.UPDATE)
-        .thenCompose(unused -> createExperimentRunHandler.createExperimentRun(request), executor);
+            ModelDBActionEnum.ModelDBServiceActions.UPDATE,
+            ModelDBServiceResourceTypes.PROJECT)
+        .thenCompose(
+            unused ->
+                filterAndGetPrivilegedDatasetsOnly(request.getDatasetsList(), true)
```
conradoverta: When is this second argument `false`?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -409,6 +414,82 @@ public FutureExperimentRunDAO(
...
+
+    if (newDatasetLinkedIds.isEmpty()) {
+      return InternalFuture.completedInternalFuture(newDatasets);
+    }
+    return jdbi.withHandle(
+            handle -> {
+              Map<String, Long> datasetVersionDatasetMap = new HashMap<>();
+              handle
+                  .createQuery(
+                      " SELECT commit_hash, repository_id FROM repository_commit WHERE commit_hash IN (<datasetIds>) ")
```
conradoverta: `commit_hash` is the right filter here? Considering the name is `dataseIds`.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -409,6 +414,82 @@ public FutureExperimentRunDAO(
...
+                  .bindList("datasetIds", newDatasetLinkedIds)
+                  .map(
+                      (rs, ctx) ->
+                          datasetVersionDatasetMap.put(
+                              rs.getString("commit_hash"), rs.getLong("repository_id")))
```
conradoverta: Avoid `map` for side effects as that's bad practice. Instead, return a list of entries and then convert to a map.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -409,6 +414,82 @@ public FutureExperimentRunDAO(
...
+                          checkEntityPermissionBasedOnResourceTypes(
+                                  Collections.singletonList(String.valueOf(datasetId)),
+                                  ModelDBActionEnum.ModelDBServiceActions.READ,
+                                  ModelDBServiceResourceTypes.DATASET)
+                              .thenCompose(
+                                  unused -> {
+                                    accessibleDatasetVersionIds.add(datasetVersionId);
+                                    return InternalFuture.completedInternalFuture(dataset);
+                                  },
+                                  executor));
```
conradoverta: This code is buggy because you're doing `accessibleDatasetVersionIds.add(datasetVersionId)` as a side effect inside the future. This means that you're adding a race condition on the system, which is bad.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -409,6 +414,82 @@ public FutureExperimentRunDAO(
...
+                      return InternalFuture.failedStage(
+                          new InvalidArgumentException(
+                              "Dataset not found for dataset version: " + datasetVersionId));
+                    }
+                  } catch (Exception ex) {
+                    LOGGER.trace(ex.getMessage());
+                    if (errorOut) {
+                      return InternalFuture.failedStage(ex);
+                    }
+                  }
```
conradoverta: This is wrong as we shouldn't just ignore legit errors. For example, if UAC is unavailable, we should still fail regardless of what `Exception` says.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -409,6 +414,82 @@ public FutureExperimentRunDAO(
         .thenCompose(unused -> deleteExperimentRuns(runIds), executor);
   }
 
+  /**
+   * @param newDatasets : new datasets for privilege check
+   * @param errorOut : Throw error while creation (true) otherwise we will keep it silent (false)
+   * @return {@link List} : accessible datasets
+   */
+  private InternalFuture<List<Artifact>> filterAndGetPrivilegedDatasetsOnly(
+      List<Artifact> newDatasets, boolean errorOut) {
```
conradoverta: `newDatasets` is also a bad name. new compared to what? Is this only for insertion?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -409,6 +414,82 @@ public FutureExperimentRunDAO(
...
+              return datasetVersionDatasetMap;
+            })
+        .thenCompose(
+            datasetVersionDatasetMap -> {
+              List<String> accessibleDatasetVersionIds = new ArrayList<>();
+              List<InternalFuture<Artifact>> internalFutures = new LinkedList<>();
+              for (Artifact dataset : newDatasets) {
+                String datasetVersionId = dataset.getLinkedArtifactId();
+                if (!datasetVersionId.isEmpty()
+                    && !accessibleDatasetVersionIds.contains(datasetVersionId)) {
```
conradoverta: You're building your own cache of ids already processed, which introduces the race condition below. This is classical serial thinking, which also makes the code more confusing. Instead, here's my suggestion of low:
- create a set out of the ids of `newDatasets`. Now you have de-dup and can drop `accessibleDatasetVersionIds` completely
- for each item in the set, you 1) filter based on `datasetVersionDatasetMap`, 2) map it to `checkEntityPermissionBasedOnResourceTypes` (you'll have to transform List<Future> to Future<List>), 3) filter only the successes, 4) convert to a set
- Now you have a set of valid ids, then you can filter `newDatasets` based on whether the id is present in the set or not

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Changed name and no it is not only for insertion it will also used for read experiment run as well.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Changed code as per suggestions

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Updated name and added comment on code as well

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Remove map and used list as per suggestions

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Changed entire code as per this suggestions

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Changed entire code as per suggestions

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Catch only permission denied exception

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: While fetching experiment run and we will filter datasets based on the privileges after fetching it.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -428,6 +433,120 @@ public FutureExperimentRunDAO(
...
+    Set<String> linkedDatasetVersionIds = new HashSet<>();
+    for (Artifact dataset : datasets) {
+      String datasetVersionId = dataset.getLinkedArtifactId();
+      if (!datasetVersionId.isEmpty()) {
+        linkedDatasetVersionIds.add(datasetVersionId);
+      }
+    }
+
+    // If linked dataset version ids are not exists in dataset then no need to check dataset
+    // authorization.
```
```suggestion
    // If linked dataset version ids do not exist in dataset then no need to check dataset
    // authorization.
```

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -428,6 +433,120 @@ public FutureExperimentRunDAO(
...
+              for (var entry : datasetVersionDatasetList) {
+                datasetVersionDatasetMap.put(entry.getKey(), String.valueOf(entry.getValue()));
+              }
+              return InternalFuture.completedInternalFuture(datasetVersionDatasetMap);
+            },
+            executor)
+        .thenCompose(
+            datasetVersionDatasetMap -> {
+              // Validate linked dataset in dataset version exists in database records if not then
+              // we will throw the error
```
conradoverta: Under which scenarios would this happen? Sounds like a bug, no?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -428,6 +433,120 @@ public FutureExperimentRunDAO(
         .thenCompose(unused -> deleteExperimentRuns(runIds), executor);
   }
 
+  /**
+   * @param datasets : datasets for privilege check
+   * @param errorOut : Throw error while creation (true) otherwise we will keep it silent (false) at
+   *     fetch time
+   * @return {@link List} : accessible datasets
+   */
+  private InternalFuture<List<Artifact>> filterAndGetPrivilegedDatasetsOnly(
```
conradoverta: Move to a handler. This file is already way too big to be maintainable.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -428,6 +433,120 @@ public FutureExperimentRunDAO(
...
+                                  checkEntityPermissionBasedOnResourceTypes(
+                                          Collections.singletonList(datasetId),
+                                          ModelDBActionEnum.ModelDBServiceActions.READ,
+                                          ModelDBServiceResourceTypes.DATASET)
+                                      .thenCompose(
+                                          unused -> InternalFuture.completedInternalFuture(id),
+                                          executor),
+                              executor));
+
+                } catch (PermissionDeniedException ex) {
```
conradoverta: The command above will never throw this exception. This is left over from when you were doing the blocking call.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -428,6 +433,120 @@ public FutureExperimentRunDAO(
...
+              // If all dataset with linked dataset versions found in database then checking
+              // authorization for the current user for those datasets
+              List<InternalFuture<String>> internalFutures = new LinkedList<>();
+              for (String datasetId : datasetIds) {
+                try {
+                  internalFutures.add(
+                      InternalFuture.completedInternalFuture(datasetId)
+                          .thenCompose(
+                              id ->
+                                  checkEntityPermissionBasedOnResourceTypes(
```
conradoverta: Note that you changed the `check` to return a value vs raise an exception in a different PR. You'll end up with the wrong behavior here.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -428,6 +433,120 @@ public FutureExperimentRunDAO(
...
+                  // if we are fetching entities then we are just skip those datasets from response
+                  // and skipped PermissionDeniedException on fetch
+                  if (errorOut) {
+                    return InternalFuture.failedStage(ex);
+                  }
+                }
+              }
+
+              // If all mapped datasets are allowed for the user then we will filter datasets based
+              // on linked dataset version and allowed dataset ids.
```
conradoverta: Why do we have to filter if they are all allowed? This seems incorrect to me

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: No it is not a bug. So let say we have dataset and dataset version which is mapped with runs but after adding it as datasets  in ER, user will delete the datasets or that mapped dataset version at that time we are not deleting it from mapped ER so it will be still there with mapped ER table and while we are try to fetch, it will be not there so we should have to throw error.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Changed

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Ok, Created separate handler for this called `FilterPrivilegedDatasetsHandler` and moved code to it.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: No this is not about ablocking code. i have to decide about throwing `PermissionDeniedException` or not based on `errorOut` (errorOut : Throw error while creation (true) otherwise we will keep it silent (false) at fetch time) flag so i have to catch it here.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Ok, i will be managed it when different PR code will merge in this PR.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: they are all allowed means the all allowed dataset ids but using those ids i can not directly filter the datasets of runs so i have to applied map logic for linked_artifact_id and datasetId and then based on allowed ids i have to get those Linked datasetVersionId and want to filter with ER dataset `linked_artifact_id`

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -317,8 +321,17 @@ public FutureExperimentRunDAO(
                 .execute());
   }
 
-  private InternalFuture<Void> checkProjectPermission(
-      List<String> projId, ModelDBActionEnum.ModelDBServiceActions action) {
+  public interface CheckPermissionBasedOnResourceTypesFunction {
+    InternalFuture<Void> checkPermission(
+        List<String> projId,
```
```suggestion
        List<String> ids,
```

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -317,8 +321,17 @@ public FutureExperimentRunDAO(
                 .execute());
   }
 
-  private InternalFuture<Void> checkProjectPermission(
-      List<String> projId, ModelDBActionEnum.ModelDBServiceActions action) {
+  public interface CheckPermissionBasedOnResourceTypesFunction {
```
conradoverta: Avoid interface inside class unless it _really_ belongs there. It makes the code harder to read. In this case, this interface seems generic.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -317,8 +321,17 @@ public FutureExperimentRunDAO(
...
+    InternalFuture<Void> checkPermission(
+        List<String> projId,
+        ModelDBActionEnum.ModelDBServiceActions action,
+        ModelDBServiceResourceTypes modelDBServiceResourceTypes);
+  }
+
+  private InternalFuture<Void> checkEntityPermissionBasedOnResourceTypes(
```
conradoverta, line 326: I'm confused between `checkPermission` defined before and this. They have the same signature but somehow they're different.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -790,10 +805,41 @@ public FutureExperimentRunDAO(
...
+                                              maps -> {
+                                                Map<String, List<Artifact>> finalDatasetMap =
+                                                    new HashMap<>();
+                                                maps.forEach(finalDatasetMap::putAll);
+                                                return InternalFuture.completedInternalFuture(
+                                                    finalDatasetMap);
+                                              },
+                                              executor);
+                                    },
+                                    executor);
```
conradoverta, line 834: This has nothing to do with creation, so the title is misleading and the PR more complex than it should be. We don't need to remove from this PR, but we have to stop doing this.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/FilterPrivilegedDatasetsHandler.java
```diff
@@ -0,0 +1,151 @@
...
+              }
+
+              // If all dataset with linked dataset versions found in database then checking
+              // authorization for the current user for those datasets
+              List<InternalFuture<String>> internalFutures = new LinkedList<>();
+              for (String datasetId : datasetIds) {
+                try {
+                  internalFutures.add(
+                      InternalFuture.completedInternalFuture(datasetId)
+                          .thenCompose(
```
conradoverta, line 106: Same feedback

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/FilterPrivilegedDatasetsHandler.java
```diff
@@ -0,0 +1,151 @@
...
+                  }
+                }
+              }
+
+              // If all mapped datasets are allowed for the user then we will filter datasets based
+              // on linked dataset version and allowed dataset ids.
+              InternalFuture<Set<String>> accessibleDatasetIdsFutures =
+                  InternalFuture.sequence(internalFutures, executor)
+                      .thenCompose(
+                          strings -> InternalFuture.completedInternalFuture(new HashSet<>(strings)),
```
conradoverta: `strings` is a horrible variable name bc it has nothing to do with what they contain, just their type (which I can already know by checking the type that the IDE already gives me

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: renamed

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Created new package for interfaces and moved this interface to that package

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Above `checkPermission`  was interface method so while we are passing function as an argument in handler we have to implement interface method so that method will be `checkEntityPermissionBasedOnResourceTypes` with same argument. SO finally `checkEntityPermissionBasedOnResourceTypes` method and new introduced interface works for the function as argument in handler

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Sorry about this :( . i have take care of this in other PRs but might be forgot in this. Updated PR title

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/FilterPrivilegedDatasetsHandler.java
anandjakhaniya: Renamed it with `accessibleDatasetIds`

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Rename interface name from `checkPermission` to `checkEntityPermissionBasedOnResourceTypes` to reduce confusion

conradoverta (approved):

PR #2147, opened 2021-04-16
Defer numpy, scipy, and pandas imports to support core functionality without these dependencies

daniel-verta:


client/verta/verta/data_types/_discrete_histogram.py
```diff
@@ -37,6 +36,7 @@ class DiscreteHistogram(_VertaDataType):
 
     _TYPE_NAME = "discreteHistogram"
     _VERSION = "v1"
+    _scipy = maybe_dependency("scipy")
```
convoliution: If it's not too much trouble, for these `maybe_dependency()` calls, I think we can help the user an extra step by putting these in the classes' `__init__()`s with a
```
scipy = maybe_dependency("scipy")
if scipy is None:
    raise ImportError("scipy is not installed; try `pip install scipy`")
```

Otherwise, they'll just see a rather esoteric  `NoneType has no attribute` later on

convoliution (changes requested):

client/verta/verta/operations/monitoring/profiler.py
```diff
@@ -61,14 +60,14 @@ class ContinuousHistogramProfiler(Profiler):
     def __init__(self, columns, bins=10):
         super(ContinuousHistogramProfiler, self).__init__(columns)
         self._bins = bins
-
+        self._np = maybe_dependency("numpy")
```
convoliution, line 63: Last one. Everything else looks good.
```suggestion
        self._np = maybe_dependency("numpy")
        if self._np is None:
            raise ImportError("numpy is not installed; try `pip install numpy`")
```

convoliution (changes requested):

convoliution (approved):

PR #2159, opened 2021-04-17
[VR-10760] Take datetime as a parameter

convoliution:
## Changes
Accept `datetime` objects as parameters in:
- alerts.create()
- notification_channels.create()
- SummarySampleQuery()
- allow time_utils.epoch_millis() to take and return `None`

### DIFF
--- a/client/verta/verta/_internal_utils/time_utils/_time_utils_py2.py
+++ b/client/verta/verta/_internal_utils/time_utils/_time_utils_py2.py
@@ -35,6 +35,8 @@ def epoch_millis(dt):
         return int(round((dt - UNIX_EPOCH).total_seconds() * 1000))
     elif type(dt) == int:
         return dt
+    elif dt is None:
+        return dt
     else:
         raise ValueError("Cannot convert argument to epoch milliseconds")
 
--- a/client/verta/tests/operations/monitoring/alerts/test_entities.py
+++ b/client/verta/tests/operations/monitoring/alerts/test_entities.py
@@ -83,6 +83,7 @@ class TestIntegration:
 
 class TestAlert:
     """Tests that aren't specific to an alerter type."""
+
     def test_update_last_evaluated_at(self, monitored_entity):
         alerts = monitored_entity.alerts
         name = _utils.generate_default_name()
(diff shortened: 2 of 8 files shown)


client/verta/verta/_internal_utils/time_utils/_time_utils_py2.py
```diff
@@ -35,6 +35,8 @@ def epoch_millis(dt):
         return int(round((dt - UNIX_EPOCH).total_seconds() * 1000))
     elif type(dt) == int:
         return dt
+    elif dt is None:
+        return dt
```
convoliution, line 39: @daniel-verta Do you think this is reasonable to have in this function, or should the caller be expected to handle `None`?

daniel-verta: I think you've got a reasonable point and we can expect the caller to wrap this in `maybe` if the parameter is optional. So many of our use cases will switch to `maybe(time_utils.epoch_millis, dt)`. We might also rename this to something like `convert_to_epoch_millis` in order to make it clear that it's a function and not a variable, since it could easily look like that in those use cases.

daniel-verta: That switch isn't pressing in my mind and could be done as part of the ongoing refactoring and clean-up work instead of in this PR depending on your bandwidth.

daniel-verta (approved):

client/verta/verta/_internal_utils/time_utils/_time_utils_py2.py
convoliution: I forgot we had `maybe()`! Amazing.

conradoverta: Unlike Scala where `Maybe/Optional` is a standard thing, this would be a new thing for users to have to learn and get used to. The Pythonic way is to have `None` or a value.

convoliution: > a new thing for users to have to learn

Our end users would not be calling `maybe(). This is only about how the client is internally handling optional requirements.

PR #2166, opened 2021-04-18
Remove TODO and implemented version_input in ER creation

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+      case "ExperimentRunEntity":
+        this.entityIdReferenceColumn = "experiment_run_id";
+        break;
+      default:
+        throw new InternalErrorException("Invalid entity name: " + entityName);
+    }
+  }
+
+  public InternalFuture<ExperimentRun> validateAndInsertVersionedInputs(
+      ExperimentRun experimentRun) {
```
conradoverta: Why do you need the whole ER?

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+          .thenCompose(
+              locationBlobWithHashMap ->
+                  insetVersioningInput(
+                      experimentRun.getVersionedInputs(),
+                      locationBlobWithHashMap,
+                      experimentRun.getId()),
+              executor)
+          .thenCompose(unused -> InternalFuture.completedInternalFuture(experimentRun), executor);
+    } else {
+      return InternalFuture.completedInternalFuture(experimentRun);
```
conradoverta: Don't return the argument you just received.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+      VersioningEntry versioningEntry) {
+    String errorMessage = null;
+    if (versioningEntry.getRepositoryId() == 0L) {
+      errorMessage = "Repository Id not found in VersioningEntry";
+    } else if (versioningEntry.getCommit().isEmpty()) {
+      errorMessage = "Commit hash not found in VersioningEntry";
+    }
+
+    if (errorMessage != null) {
+      throw new ModelDBException(errorMessage, io.grpc.Status.Code.INVALID_ARGUMENT);
```
conradoverta: Don't throw exceptions. You're returning a future, so just return a failed future. Throwing exceptions is now an anti-pattern as it breaks the execution chain. It's only acceptable if you're not returning a future.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+              (session1) -> repositoryDAO.getRepositoryById(session, repositoryIdentification));
+      if (!versioningEntry.getKeyLocationMapMap().isEmpty()) {
+        Map<String, Map.Entry<BlobExpanded, String>> locationBlobWithHashMap =
+            blobDAO.getCommitBlobMapWithHash(
+                session, commitEntity.getRootSha(), new ArrayList<>(), Collections.emptyList());
+        for (Map.Entry<String, Location> locationBlobKeyMap :
+            versioningEntry.getKeyLocationMapMap().entrySet()) {
+          String locationKey = String.join("#", locationBlobKeyMap.getValue().getLocationList());
+          if (!locationBlobWithHashMap.containsKey(locationKey)) {
+            throw new ModelDBException(
```
conradoverta: Same

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+          }
+          requestedLocationBlobWithHashMap.put(
+              locationKey, locationBlobWithHashMap.get(locationKey));
+        }
+      }
+    } catch (ModelDBException ex) {
+      throw ex;
+    } catch (Exception ex) {
+      throw new ModelDBException(ex);
+    }
```
conradoverta: Remove

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+    return InternalFuture.completedInternalFuture(requestedLocationBlobWithHashMap);
```
conradoverta: This value is always known immediately, so remove the future wrapper for clarity.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+          }
+        });
+  }
+
+  /**
+   * @param versioningEntry : versioningEntry
+   * @return returns a map from location to an Entry of BlobExpanded and sha
+   * @throws ModelDBException ModelDBException
+   */
+  private InternalFuture<Map<String, Map.Entry<BlobExpanded, String>>> validateVersioningEntity(
```
conradoverta, line 333: Please say where this is copied from or if this needs review for correctness.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+      Map<String, Map.Entry<BlobExpanded, String>> locationBlobWithHashMap,
+      String entityId) {
+    return jdbi.useHandle(
+        handle -> {
+          var queryStr =
+              "INSERT INTO versioning_modeldb_entity_mapping "
+                  + " (repository_id, commit, versioning_key, versioning_location, entity_type, versioning_blob_type, blob_hash, "
+                  + entityIdReferenceColumn
+                  + ") "
+                  + " VALUES (:repository_id, :commit, :versioning_key, :versioning_location, :entity_type, :versioning_blob_type, :blob_hash, :entityId)";
```
conradoverta: I recommend doing the map pattern you did in the other PR.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  private InternalFuture<Void> insetVersioningInput(
+      VersioningEntry versioningEntry,
+      Map<String, Map.Entry<BlobExpanded, String>> locationBlobWithHashMap,
+      String entityId) {
+    return jdbi.useHandle(
+        handle -> {
+          var queryStr =
+              "INSERT INTO versioning_modeldb_entity_mapping "
```
conradoverta: Don't we have a hibernate entity for this that we can use vs doing everything manually already? The focus is to remove the ER entity first, not remove all entities at once as it leads to complex changes and more risk.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+                      locationBlobWithHashMap,
+                      experimentRun.getId()),
+              executor)
+          .thenCompose(unused -> InternalFuture.completedInternalFuture(experimentRun), executor);
+    } else {
+      return InternalFuture.completedInternalFuture(experimentRun);
+    }
+  }
+
+  private InternalFuture<Void> insetVersioningInput(
```
conradoverta:
```suggestion
  private InternalFuture<Void> insertVersioningInput(
```

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+                          handle
+                              .createUpdate(blobMappingQueryStr)
+                              .bind("repository_id", versioningEntry.getRepositoryId())
+                              .bind("commit", versioningEntry.getCommit())
+                              .bind("versioning_key", locationEntry.getKey())
+                              .bind("entityId", entityId)
+                              .bind("entity_type", entity_type)
+                              .bind("blob_hash", rs.getString("blob_hash"))
+                              .bind("config_seq_number", rs.getString("config_seq_number"))
+                              .execute();
```
conradoverta: Don't insert in the middle of mapping a select.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,245 @@
...
+                    "SELECT cb.blob_hash, cb.config_seq_number, cb.hyperparameter_type, hecbs.name, hecbs.int_value, hecbs.float_value, hecbs.string_value "
+                        + " FROM config_blob cb "
+                        + " JOIN hyperparameter_element_config_blob hecbs ON cb.hyperparameter_element_config_blob_hash = hecbs.blob_hash"
+                        + " WHERE cb.blob_hash = :blobHash";
+                handle
+                    .createQuery(configBlobQuery)
+                    .bind("blobHash", blobExpandedWithHashMap.getValue())
+                    .map(
+                        (rs, ctx) -> {
+                          String blobMappingQueryStr =
```
conradoverta: This seems to be executed multiple times, so we should use the JDBI batch functionality.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Removed

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Replaced with Internal future

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Removed

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Handled using InternalFuture because this method called UAC and database as well

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Copy logic from here https://github.com/VertaAI/modeldb/blob/2663a4857de417e94240f4495a2dee998cb09659/backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java#L272

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Yes but these entities are only related with ExperimentRun and we can not used it as hibernate entity because it is totally binding with ExperimentRunEntity which not exists in new flow.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: We are selecting mappings to insert this data by use of fetch one so this is a correct place i think so..

conradoverta: It is not. Don't do a DB operation while in the middle of another operation. That is really bad practice that kills performance. Do one thing at a time.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Implemented based on map pattern

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: separated in multiple functions instead of middle of DB operation

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Implemented batch for entire class where batch is needed.

backend/src/main/java/ai/verta/modeldb/DAOSet.java
```diff
@@ -87,6 +86,7 @@ public static DAOSet fromServices(ServiceSet services, FutureJdbi jdbi, Executor
       set.auditLogLocalDAO = new AuditLogLocalDAORdbImpl();
     }
 
+    set.futureExperimentRunDAO = new FutureExperimentRunDAO(executor, jdbi, services.uac, set);
```
conradoverta: Avoid passing the whole set. It's harder to identify dependencies this way.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,354 @@
...
+                  .thenCompose(
+                      unused ->
+                          insertHyperparameterElementMapping(
+                              experimentRun.getVersionedInputs(),
+                              locationBlobWithHashMap,
+                              experimentRun.getId()),
+                      executor),
+          executor);
+    } else {
+      return InternalFuture.runAsync(() -> {}, executor);
```
conradoverta: Just return a completed future instead of creating an extra task that does nothing.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -0,0 +1,354 @@
...
+                              + "' not found in commit blobs",
+                          Code.INVALID_ARGUMENT));
+                }
+                requestedLocationBlobWithHashMap.put(
+                    locationKey, locationBlobWithHashMap.get(locationKey));
+              }
+            }
+          } catch (Exception ex) {
+            return InternalFuture.failedStage(ex);
+          }
```
conradoverta, line 401: Because you're inside of a future, this is already handled for you.

conradoverta (commented):
This code is **way** too big and complex not to have any comments. Can we please document what it's doing?

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: That function has return type `InternalFuture<Void>` so i do not have ti return anything in response so i can not use completed future there. that's why i have added runAsync which return Void.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: I have tried to remove this catch but the code from try throwing error which is not auto handled by future.

anandjakhaniya:
> This code is **way** too big and complex not to have any comments. Can we please document what it's doing?

try to add comments on each functions.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
conradoverta: Are you absolutely sure you can't use a completed future?

conradoverta (commented):
>try to add comments on each functions.

Useful, but not what I was thinking. It's much better to have documentation in the middle of the code, on each complex code block about what it's doing instead of a 5-line description at the top.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Yes i am sure. we cant use  completed future. function return type is `public InternalFuture<Void> validateAndInsertVersionedInputs(ExperimentRun experimentRun)` so if we are trying to return `return InternalFuture.completedInternalFuture(R)`  we need one argument for the same and we can not create Void there so. `return InternalFuture.runAsync(() -> {}, executor);` is the write line there as per my view.

anandjakhaniya:
> > try to add comments on each functions.
> 
> Useful, but not what I was thinking. It's much better to have documentation in the middle of the code, on each complex code block about what it's doing instead of a 5-line description at the top.

try to add more comments in function code as well. i hope it will works.

conradoverta (approved):

PR #2177, opened 2021-04-19
Async create experiment run in transaction

anandjakhaniya:


conradoverta (approved):
Nice. But the other fields (e.g. metrics) also need to be in a transaction or we have to add the `created` flag. I recommend the flag.

anandjakhaniya:
> Nice. But the other fields (e.g. metrics) also need to be in a transaction or we have to add the `created` flag. I recommend the flag.

i'm working with `created` flag.

anandjakhaniya:
> > Nice. But the other fields (e.g. metrics) also need to be in a transaction or we have to add the `created` flag. I recommend the flag.
> 
> i'm working with `created` flag.

`created` flag PR: https://github.com/VertaAI/modeldb/pull/2178

PR #2178, opened 2021-04-19
Created flag in ER

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/entities/ExperimentRunEntity.java
```diff
@@ -466,6 +469,10 @@ public void setEnvironment(String environment) {
     this.environment = environment;
   }
 
+  public void setCreated(Boolean created) {
```
conradoverta, line 473:
```suggestion
  // TODO: update indices and code to take this flag into consideration
  public void setCreated(Boolean created) {
```

conradoverta (approved):
Added a TODO. We need to make sure we consume this flag everywhere.

PR #2179, opened 2021-04-19
Async find ER using tags

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -94,6 +94,66 @@
...
+    try {
+      final var value = predicate.getValue();
+      var operator = predicate.getOperator();
+      //  We will manage this operator at bottom of method based on `NE` and `NOT_CONTAIN` using
+      // `IN` OR `NOT IN` query
+      if (operator.equals(OperatorEnum.Operator.NOT_CONTAIN)) {
+        operator = OperatorEnum.Operator.CONTAIN;
+      } else if (operator.equals(OperatorEnum.Operator.NE)) {
+        operator = OperatorEnum.Operator.EQ;
+      }
```
conradoverta: Please don't. This kind of solution made my life hell trying to understand the original code as things that were supposed to be the same became different. Just spend the extra typing to account for these types in the checks instead of just overwriting the operator and making `operator` be different from `predicate.getOperator()`. It's just not worth the mental burden for saving a handful of characters.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -94,6 +94,66 @@
...
+      final var finalOperator = operator;
+
+      final var valueBindingName = String.format("v_t_%d", index);
+
+      var sql = "select distinct experiment_run_id from tag_mapping where entity_name=:entity_name";
+      sql += " and ";
+
+      final var colValue = "tags";
+      var queryContext =
+          new QueryFilterContext().addBind(q -> q.bind("entity_name", "ExperimentRunEntity"));
```
conradoverta: This will collide with other `entity_name`. We have to make the bindings we create here unique (so we have to use the index). It might be a bug in my current code too.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -94,6 +94,66 @@
...
+              valueList.add(item.getStringValue());
+            }
+          }
+
+          sql += applyOperator(finalOperator, colValue, "<" + valueBindingName + ">");
+          queryContext = queryContext.addBind(q -> q.bindList(valueBindingName, valueList));
+          break;
+        default:
+          return InternalFuture.failedStage(
+              new UnimplementedException("Unknown 'Value' type recognized"));
```
conradoverta: It's either unknown or recognized. It can't be both.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -94,6 +94,66 @@
...
+      }
+
+      if (predicate.getOperator().equals(OperatorEnum.Operator.NOT_CONTAIN)
+          || predicate.getOperator().equals(OperatorEnum.Operator.NE)) {
+        queryContext =
+            queryContext.addCondition(String.format("experiment_run.id NOT IN (%s)", sql));
+      } else {
+        queryContext = queryContext.addCondition(String.format("experiment_run.id IN (%s)", sql));
+      }
```
conradoverta: This is different from what I had in the key value one. Let's standardize.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: We have to do this for all the predicates for standard pattern for `NE` & `NOT_CONTAIN`

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: Removed and modify as per sync up.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: created unique using index and fixed binding

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: updated message

conradoverta (approved):

PR #2180, opened 2021-04-19
Async find ER using attributes

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -179,9 +259,13 @@ private String applyOperator(
       case OperatorEnum.Operator.NE_VALUE:
         return String.format("%s != %s", colName, valueBinding);
       case OperatorEnum.Operator.CONTAIN_VALUE:
-        return String.format("%s LIKE %s", colName, valueBinding);
+        return String.format(
+            "%s LIKE %s",
+            "lower(" + colName + ") ", Pattern.compile(valueBinding).toString().toLowerCase());
```
conradoverta: Why do we need this?

conradoverta (changes requested):
Same feedback as the other one.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: - let say if there is key like `KEY_1` and user will try to predicate with value in small case like `key` then it will not returning anything so for that we have to take that column in lower case.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -153,6 +155,84 @@
...
+    try {
+      final var value = predicate.getValue();
+      var operator = predicate.getOperator();
+      //  We will manage this operator at bottom of method based on `NE` and `NOT_CONTAIN` using
+      // `IN` OR `NOT IN` query
+      if (operator.equals(OperatorEnum.Operator.NOT_CONTAIN)) {
+        operator = OperatorEnum.Operator.CONTAIN;
+      } else if (operator.equals(OperatorEnum.Operator.NE)) {
+        operator = OperatorEnum.Operator.EQ;
+      }
```
conradoverta: Change as discussed

conradoverta (changes requested):
You have not addressed the feedback

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: i have changed it in https://github.com/VertaAI/modeldb/pull/2179/files so i will change here after merging it because both are used same operator function.

conradoverta: Sure

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: Removed this code as per suggestions

PR #2193, opened 2021-04-21
[VR-10800] Add QueryFilterContext

coryverta:
@conrado this is borrowed from your working branch but is the same class in the same package.


conradoverta (approved):
I have some other small changes for later, but not blocking

PR #2194, opened 2021-04-21
Fix review comments for PR https://github.com/VertaAI/modeldb/pull/2140

anandjakhaniya:


conradoverta (approved):
For small fixes like this, prefer to do in the same PR. Otherwise it forces me to have to carry mental context between the PRs, which only works if I have them fresh in memory already.

PR #2199, opened 2021-04-21
Find ER with Ids only flag

anandjakhaniya:


conradoverta (approved):
Is this option actually used by anything? I'm pretty sure it isn't.

anandjakhaniya:
> Is this option actually used by anything? I'm pretty sure it isn't.

i do not understand this question. Could you elaborate more please??

conradoverta:
> > Is this option actually used by anything? I'm pretty sure it isn't.
> 
> i do not understand this question. Could you elaborate more please??

Can we ask the client and webapp folks if they use this? If not, I'd rather remove this functionality. I don't think anywhere uses it

anandjakhaniya:
> > > Is this option actually used by anything? I'm pretty sure it isn't.
> > 
> > 
> > i do not understand this question. Could you elaborate more please??
> 
> Can we ask the client and webapp folks if they use this? If not, I'd rather remove this functionality. I don't think anywhere uses it

Confirm that in slack https://vertaai.slack.com/archives/CJG0NL86B/p1619759633302900, it is not used by client and webapp and safe to remove it.

conradoverta:
> > > > Is this option actually used by anything? I'm pretty sure it isn't.
> > > 
> > > 
> > > i do not understand this question. Could you elaborate more please??
> > 
> > 
> > Can we ask the client and webapp folks if they use this? If not, I'd rather remove this functionality. I don't think anywhere uses it
> 
> Confirm that in slack https://vertaai.slack.com/archives/CJG0NL86B/p1619759633302900, it is not used by client and webapp and safe to remove it.

Then we should probably close this PR and remove the field from the protos

PR #2208, opened 2021-04-22
Populate code_version in find ER response

anandjakhaniya:


conradoverta (approved):

conradoverta:
Is this everything we need? I know we have a single field _and_ we have a whole structure (see https://github.com/VertaAI/protos-all/blob/master/protos/protos/public/modeldb/ExperimentRunService.proto#L31).

anandjakhaniya:
> Is this everything we need? I know we have a single field _and_ we have a whole structure (see https://github.com/VertaAI/protos-all/blob/master/protos/protos/public/modeldb/ExperimentRunService.proto#L31).

This is for the https://github.com/VertaAI/protos-all/blob/8600f354ad023fa8c40bd381414c81e07e0db9ea/protos/protos/public/modeldb/ExperimentRunService.proto#L30 field.  so yes for that field this only required. for your pointed field (`code_version_snapshot`) i will create separate PR today.

PR #2213, opened 2021-04-22
Directly import scipy's spatial submodule

convoliution:
Some versions of scipy don't let you do
```python
import scipy
scipy.spatial
```

So this effectively changes the logic to 
```python
import scipy.spatial
scipy.spatial
```


convoliution:
I'm at a loss for how to adequate cover this in our tests, since our pipelines aren't too conducive to testing against multiple versions of external libraries. 😕

daniel-verta (approved):

PR #2215, opened 2021-04-23
[Task] Support mismatched buckets in dist for discrete histogram

daniel-verta:


client/verta/tests/test_data_types.py
```diff
@@ -71,6 +71,18 @@ def test_discrete_histogram_numpy(self):
             },
         }
 
+    def test_missing_buckets(self):
+        one = data_types.DiscreteHistogram(
```
convoliution, line 75: Instead of `import numpy as np`
```suggestion
        np = pytest.importorskip("numpy")

        one = data_types.DiscreteHistogram(
```

convoliution (changes requested):

convoliution (approved):

PR #2219, opened 2021-04-23
Populate hyperparameter from config blob in Find ER

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -44,6 +45,7 @@
   private final DatasetHandler datasetHandler;
   private final PredicatesHandler predicatesHandler;
   private final SortingHandler sortingHandler;
+  private final Config config = Config.getInstance();
```
conradoverta: Get as constructor argument. Instantiating like this is bad practice as it doesn't allow us to change the config (e.g. testing).

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -695,6 +699,42 @@ public FutureExperimentRunDAO(
...
+                                        repositoryIds ->
+                                            hyperparametersHandler
+                                                .getExperimentRunHyperparameterConfigBlobMap(
+                                                    new ArrayList<>(ids), repositoryIds),
+                                        executor);
+                            futureBuildersStream =
+                                futureBuildersStream.thenCombine(
+                                    futureHyperparamsFromConfigBlobs,
+                                    (stream, hyperparamsFromConfigBlob) ->
+                                        stream.peek(
```
conradoverta: map, don't `peek` and then do inplace operation

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
```diff
@@ -249,4 +256,66 @@ public KeyValueHandler(Executor executor, FutureJdbi jdbi, String fieldType, Str
...
+      List<String> expRunIds, List<String> selfAllowedRepositoryIds) {
+    return jdbi.withHandle(
+            handle -> {
+              String queryStr =
+                  "SELECT distinct vme.experiment_run_id, hecb.name, hecb.value_type, hecb.int_value, hecb.float_value, hecb.string_value  FROM hyperparameter_element_config_blob hecb "
+                      + "INNER JOIN config_blob cb ON cb.hyperparameter_element_config_blob_hash = hecb.blob_hash "
+                      + "INNER JOIN versioning_modeldb_entity_mapping vme ON vme.blob_hash = cb.blob_hash "
+                      + "WHERE cb.hyperparameter_type = :hyperparameterType AND vme.experiment_run_id IN (<expRunIds>) ";
+
+              if (config.populateConnectionsBasedOnPrivileges) {
```
conradoverta: Why do we care about this flag here? We already checked the flag before we call this method.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
```diff
@@ -249,4 +256,66 @@ public KeyValueHandler(Executor executor, FutureJdbi jdbi, String fieldType, Str
...
+                if (selfAllowedRepositoryIds == null || selfAllowedRepositoryIds.isEmpty()) {
+                  return new ArrayList<AbstractMap.SimpleEntry<String, KeyValue>>();
+                } else {
+                  queryStr = queryStr + " AND vme.repository_id IN (<repoIds>)";
+                }
+              }
+
+              var query = handle.createQuery(queryStr);
+              query.bind("hyperparameterType", HYPERPARAMETER);
```
conradoverta: This is a `KeyValueHandler`. Don't hardcode hyperparameter things. If you want, extract it to a hyperparameter handler that inherits from this one

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
```diff
@@ -249,4 +256,66 @@ public KeyValueHandler(Executor executor, FutureJdbi jdbi, String fieldType, Str
...
+              query.bindList("expRunIds", expRunIds);
+              if (config.populateConnectionsBasedOnPrivileges) {
+                query.bindList(
+                    "repoIds",
+                    selfAllowedRepositoryIds.stream()
+                        .map(Long::parseLong)
+                        .collect(Collectors.toList()));
+              }
+              LOGGER.debug(
```
conradoverta: trace

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed and added feature flag in constructor

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Implemented with map

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
anandjakhaniya: Remove check from here and improve code relevant code

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
anandjakhaniya: Extracted in separate hyperparameter config handler

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
anandjakhaniya: Done

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -44,16 +44,20 @@
...
+  private final boolean populateConnectionsBasedOnPrivileges;
 
   public FutureExperimentRunDAO(
       Executor executor,
       FutureJdbi jdbi,
       UAC uac,
       ArtifactStoreDAO artifactStoreDAO,
-      DatasetVersionDAO datasetVersionDAO) {
+      DatasetVersionDAO datasetVersionDAO,
+      boolean populateConnectionsBasedOnPrivileges) {
```
conradoverta: I'd recommend passing the whole config here as we'll most likely need other fields later and we don't want to add them as more constructor parameters.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -695,6 +704,53 @@ public FutureExperimentRunDAO(
...
+                                                        AbstractMap.SimpleEntry<
+                                                            String, KeyValue>>())
+                                                .thenApply(MapSubtypes::from, executor);
+                                          } else {
+                                            return hyperparametersFromConfigHandler
+                                                .getExperimentRunHyperparameterConfigBlobMap(
+                                                    new ArrayList<>(ids), selfAllowedRepositoryIds);
+                                          }
+                                        },
+                                        executor);
```
conradoverta: This is huge. We shouldn't inline this amount of logic. Extract to the handler.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -695,6 +704,53 @@ public FutureExperimentRunDAO(
...
+                                        executor)
+                                    .thenCompose(
+                                        selfAllowedRepositoryIds -> {
+                                          if (selfAllowedRepositoryIds == null
+                                              || selfAllowedRepositoryIds.isEmpty()) {
+                                            return InternalFuture.completedInternalFuture(
+                                                    new ArrayList<
+                                                        AbstractMap.SimpleEntry<
+                                                            String, KeyValue>>())
+                                                .thenApply(MapSubtypes::from, executor);
```
conradoverta: Why do we need this? You already know that the array will be empty because you just created it. Reading this code makes me confused

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -695,6 +704,53 @@ public FutureExperimentRunDAO(
                                                     hyperparams.get(builder.getId()))),
                                     executor);
 
+                            final var futureHyperparamsFromConfigBlobs =
```
conradoverta: Please comment this block of code. It's not trivial on what it's doing.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/HyperparametersFromConfigHandler.java
```diff
@@ -0,0 +1,86 @@
...
+      List<String> expRunIds, List<String> selfAllowedRepositoryIds) {
+    return jdbi.withHandle(
+            handle -> {
+              String queryStr =
+                  "SELECT distinct vme.experiment_run_id, hecb.name, hecb.value_type, hecb.int_value, hecb.float_value, hecb.string_value  FROM hyperparameter_element_config_blob hecb "
+                      + "INNER JOIN config_blob cb ON cb.hyperparameter_element_config_blob_hash = hecb.blob_hash "
+                      + "INNER JOIN versioning_modeldb_entity_mapping vme ON vme.blob_hash = cb.blob_hash "
+                      + "WHERE cb.hyperparameter_type = :hyperparameterType AND vme.experiment_run_id IN (<expRunIds>) ";
+
+              if (selfAllowedRepositoryIds != null && !selfAllowedRepositoryIds.isEmpty()) {
```
conradoverta: This check looks weird. How do you differentiate between not having access to the repo and having access to all the repos (which is what this `if` assumes to be the case).

anandjakhaniya: What is the purpose of this comment?? `selfAllowedRepositoryIds` shows it self like it will comes from UAC  right. so `How do you differentiate between not having access to the repo and having access to all the repos` why you are confuse on it. this if added because if repo ids are there then i have to add in query otherwise i do not have to add it in query. it is simple as that. Why you are confuse here?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: This is already added in separate handler 
```
return hyperparametersFromConfigHandler
                                                .getExperimentRunHyperparameterConfigBlobMap(
                                                    new ArrayList<>(ids), selfAllowedRepositoryIds);
```
and above logic is not complex it just checking repository access using `getAllowedEntitiesByResourceType`  so due to getAllowedEntitiesByResourceType function i can not move that code to handler. i can create a separate function for this inline code

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Added config instead of individual flag

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: extracted code in separate function and try to explain code with adding code comments

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Simplify logic and reduce one line

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Added comment on code for this block

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/HyperparametersFromConfigHandler.java
conradoverta: How do you differentiate between not having access to the repo and having access to all the repos?

I am confused because I don't recall the answer to the question above. An empty list representing `all` and `none` in different cases is very confusing to me. Could you answer that question to remind me? Your comment assumes I remember specifics of the API, which I don't.

If you mean to have the behavior for all ids, would you mind using the new explicit field? https://github.com/VertaAI/protos-all/blob/master/protos/protos/private/uac/RoleService.proto#L135

anandjakhaniya: We have checked repository ids access here https://github.com/VertaAI/modeldb/pull/2219/files#diff-f554e4fd8e5f43fa49db0fa2ba8f6e7a5903f2259b01a1a1e411c21189a2ce78R837 and then I have passed accessible ids from here https://github.com/VertaAI/modeldb/pull/2219/files#diff-f554e4fd8e5f43fa49db0fa2ba8f6e7a5903f2259b01a1a1e411c21189a2ce78R855 so here i do not want to check again it has access or not. it has access that why those `selfAllowedRepositoryIds` ids are here. I think you do not have checked the flow OR if naming conventions are miss leading to you then you can suggestion me the name here?

anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/2255

conradoverta (approved):
LGTM

PR #2222, opened 2021-04-23
[VR-10789] Enable artifact download from model version

### LINKED ISSUE #2183: RegisteredModelVersion should implement download_model like ExperimentRun
Right now, ExperimentRun can save a directory as a model zip file, and then `.download_model()` back into a directory, whereas a RegisteredModelVersion only implements `.get_model()`, which returns the raw pkzip file

convoliution:
Resolves #2183 

## Changes

- assign `"model"` to `_artifact_utils.REGISTRY_MODEL_KEY`
- add `download_artifact()` and `download_model()` to the `_DeployableEntity` interface
- implement `download_artifact()` and `download_model()` on `RegisteredModelVersion`
- enable `RegisteredModelVersion.log_artifact()` to handle directories
- reorganize existing model version tests
- add new tests


client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -437,6 +336,148 @@ def test_find_stage(self, client, created_entities):
...
+
+    def test_download_directory(self, model_version, strs, dir_and_files, in_tempdir):
+        key, download_path = strs[:2]
+        dirpath, _ = dir_and_files
+
+        model_version.log_artifact(key, dirpath)
+        model_version.download_artifact(key, download_path)
+
+        # contents match
+        utils.assert_dirs_match(dirpath, download_path)
```
convoliution, line 415: These are the only added tests in this class (`test_download`, `test_download_directory`)—the rest were moved from a different class.

client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -622,6 +681,56 @@ def test_arbitrary_object(self, model_version):
...
+        download_path = strs[0]
+
+        experiment_run.log_model(dirpath)
+        model_version = registered_model.create_version_from_run(
+            run_id=experiment_run.id,
+            name="From Run {}".format(experiment_run.id),
+        )
+        model_version.download_model(download_path)
+
+        utils.assert_dirs_match(dirpath, download_path)
```
convoliution, line 714: This is the only _new_ test. The rest were copied nearly verbatim from experiment run tests.

conradoverta (approved):
We should probably not copy this amount of tests, but I'm ok with this being a follow up.

PR #2230, opened 2021-04-26
[VR-10897] Added privilege logic for while populating datasets on find

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10897

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
+++ b/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
@@ -30,11 +30,13 @@ import ai.verta.modeldb.Observation;
 import ai.verta.modeldb.artifactStore.ArtifactStoreDAO;
 import ai.verta.modeldb.common.EnumerateList;
 import ai.verta.modeldb.common.connections.UAC;
+import ai.verta.modeldb.common.exceptions.ModelDBException;
 import ai.verta.modeldb.common.exceptions.NotFoundException;
 import ai.verta.modeldb.common.futures.FutureGrpc;
 import ai.verta.modeldb.common.futures.FutureJdbi;
 import ai.verta.modeldb.common.futures.InternalFuture;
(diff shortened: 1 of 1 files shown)


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -838,9 +842,36 @@ public FutureExperimentRunDAO(
...
+                                                        .thenCompose(
+                                                            artifacts -> {
+                                                              builder
+                                                                  .clearDatasets()
+                                                                  .addAllDatasets(artifacts);
+                                                              return InternalFuture
+                                                                  .completedInternalFuture(builder);
+                                                            },
+                                                            executor)
+                                                        .get();
```
conradoverta: Why are you resolving the future immediately? This is extremely bad practice. You're just turning parallel code and making it blocking, which is the whole problem we're trying to solve =)

conradoverta (changes requested):

conradoverta (commented):
Can we apply the same feedback from the config PR to this one? They're doing similar functionality

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Resolved this on the based PR so this logic has been removed from here

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -90,6 +91,7 @@
   private final SortingHandler sortingHandler;
   private final FeatureHandler featureHandler;
   private final CreateExperimentRunHandler createExperimentRunHandler;
+  private final Config config = Config.getInstance();
```
anandjakhaniya, line 94: This will be removed in other PR so while merging other this will be removed from here as well

conradoverta (approved):

PR #2231, opened 2021-04-26
[VR-10898] Check privileges for version input in find ER

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10898


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -845,6 +853,59 @@ public FutureExperimentRunDAO(
...
+    } else if (notAccessibleRepoIdIdsSet.contains(repoId)) {
+      notAccessibleRepoIdIdsSet.add(repoId);
+      experimentRunBuilder.clearVersionedInputs().build();
+    } else {
+      try {
+        Boolean isSelfAllowed =
+            IsRepositorySelfAllowed(
+                    Collections.singletonList(String.valueOf(repoId)),
+                    ModelDBActionEnum.ModelDBServiceActions.READ)
+                .get();
```
conradoverta: Same feedback here on blocking

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -845,6 +853,59 @@ public FutureExperimentRunDAO(
...
+          experimentRunBuilder.clearVersionedInputs().build();
+          notAccessibleRepoIdIdsSet.add(repoId);
+        }
+      } catch (ExecutionException | InterruptedException ex) {
+        throw new ModelDBException(ex);
+      }
+    }
+  }
+
+  private InternalFuture<Boolean> IsRepositorySelfAllowed(
```
conradoverta: We already have a method for this for any type. Let's re-use instead of coyping

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -845,6 +853,59 @@ public FutureExperimentRunDAO(
         executor);
   }
 
+  private void checkVersionInputBasedOnPrivileges(
+      ExperimentRun.Builder experimentRunBuilder,
```
conradoverta: Avoid passing an object to do side effects on it. It's very error prone. Instead, return a signal of what you want to do with the result (e.g. a boolean saying it should clean).

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -786,6 +788,8 @@ public FutureExperimentRunDAO(
                                                     observations.get(builder.getId()))),
                                     executor);
 
+                            Set<Long> accessibleRepoIdsSet = new HashSet<>();
+                            Set<Long> notAccessibleRepoIdIdsSet = new HashSet<>();
```
conradoverta: These are not used and just mutated inside `checkVersionInputBasedOnPrivileges`. That's bad practice. Don't mutate your arguments if you can avoid.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Fix code using Internal future

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed and modify existing one for re-use

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Fix code with just passing required fields and resolved code path using InternalFuture

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed and manage using InternalFuture

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -329,8 +328,10 @@ public FutureExperimentRunDAO(
                 .execute());
   }
 
-  private InternalFuture<Void> checkProjectPermission(
-      List<String> projId, ModelDBActionEnum.ModelDBServiceActions action) {
+  private InternalFuture<Boolean> checkEntityPermissionBasedOnResourceTypes(
```
```suggestion
  private InternalFuture<Boolean> getEntityPermissionBasedOnResourceTypes(
```
Now that you're returning something vs just checking.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -891,63 +933,59 @@ public FutureExperimentRunDAO(
...
+                                .thenCompose(
+                                    isSelfAllowed -> {
+                                      // Set null into map if repository is not accessible to the
+                                      // current user
+                                      VersioningEntry allowedVersioningEntry = null;
+                                      if (isSelfAllowed) {
+                                        allowedVersioningEntry = versioningEntry1;
+                                      }
+                                      return InternalFuture.completedInternalFuture(
+                                          Collections.singletonMap(runId, allowedVersioningEntry));
```
conradoverta: Is it safe for us to create a map with `null` here? I'd probably expect this to be skipped vs saving a `null` entry as it will probably cause a NPE down the line.

conradoverta (commented):
Approved conditional on you being absolutely certain that the `null` won't cause issues.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: updated

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: We can not just skip the entry because we have to return something so now i have returned empty entry instead of add the null in map.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -896,10 +933,60 @@ public FutureExperimentRunDAO(
         executor);
   }
 
+  private InternalFuture<Map<String, VersioningEntry>> filterVersionedInputsBasedOnPrivileges(
```
conradoverta: Move this to `versionInputHandler` like all the others.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -896,10 +933,60 @@ public FutureExperimentRunDAO(
...
+            // Get VersionEntry from fetched map
+            VersioningEntry versioningEntry = versionedInputsMap.get(runId);
+            if (versioningEntry == null) {
+              continue;
+            }
+            String repoId = String.valueOf(versioningEntry.getRepositoryId());
+            // Check repository is accessible or not from versionedInputs If not then we will remove
+            // it from fetched map
+            internalFutureList.add(
+                InternalFuture.completedInternalFuture(versioningEntry)
```
conradoverta: You don't have to create a future just to immediately use it. Just use the value directly.

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Moved it in separate handler called `FilterPrivilegedVersionedInputsHandler`

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Fixed code

PR #2232, opened 2021-04-26
[VR-10903] Log versioned input endpoint

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10903

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
+++ b/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
@@ -48,6 +48,7 @@ import ai.verta.modeldb.experimentRun.subtypes.ObservationHandler;
 import ai.verta.modeldb.experimentRun.subtypes.PredicatesHandler;
 import ai.verta.modeldb.experimentRun.subtypes.SortingHandler;
 import ai.verta.modeldb.experimentRun.subtypes.TagsHandler;
+import ai.verta.modeldb.experimentRun.subtypes.VersionInputHandler;
 import ai.verta.modeldb.versioning.BlobDAO;
 import ai.verta.modeldb.versioning.CommitDAO;
 import ai.verta.modeldb.versioning.RepositoryDAO;
@@ -85,6 +86,7 @@ public class FutureExperimentRunDAO {
   private final DatasetHandler datasetHandler;
   private final PredicatesHandler predicatesHandler;
   private final SortingHandler sortingHandler;
+  private final VersionInputHandler versionInputHandler;
   private final CreateExperimentRunHandler createExperimentRunHandler;
 
   public FutureExperimentRunDAO(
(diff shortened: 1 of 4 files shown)


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -831,4 +837,22 @@ public FutureExperimentRunDAO(
...
         .thenCompose(unused -> createExperimentRunHandler.createExperimentRun(request), executor);
   }
+
+  public InternalFuture<Void> logVersionedInputs(LogVersionedInput request) {
+    final var runId = request.getId();
+    final var now = Calendar.getInstance().getTimeInMillis();
+    if (!request.hasVersionedInputs()) {
+      throw new InvalidArgumentException("VersionedInput not found in request");
+    } else if (!request.getId().isEmpty()) {
+      throw new InvalidArgumentException("ExperimentRun ID not found in request");
```
conradoverta: Don't throw. Return a failed future.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -831,4 +837,22 @@ public FutureExperimentRunDAO(
             ModelDBActionEnum.ModelDBServiceActions.UPDATE)
         .thenCompose(unused -> createExperimentRunHandler.createExperimentRun(request), executor);
   }
+
+  public InternalFuture<Void> logVersionedInputs(LogVersionedInput request) {
+    final var runId = request.getId();
+    final var now = Calendar.getInstance().getTimeInMillis();
+    if (!request.hasVersionedInputs()) {
+      throw new InvalidArgumentException("VersionedInput not found in request");
```
conradoverta: This check should be inside `validateAndInsertVersionedInputs`

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
conradoverta: This check is unneeded as `checkPermission` will handle this.

conradoverta (commented):
Some changes

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Moved it as per suggestion

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -78,11 +79,15 @@ public VersionInputHandler(
    */
   public InternalFuture<Void> validateAndInsertVersionedInputs(
       String runId, VersioningEntry versioningEntry) {
-    if (versioningEntry != null) {
+    if (versioningEntry == null) {
```
conradoverta, line 82: I'm pretty sure this will never be `null` because `getXYZ` in grpc returns the zero value of the object. Could you verify?

conradoverta (approved):
One concern on input validation. LGTM otherwise

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Yes agree with you. it will never null if we will directly get it from proto but this is a PUBLIC method and if some one set null explicitly then it will throw null pointer so to protect null pointer i have added it for safe side.

conradoverta: Fair

PR #2235, opened 2021-04-27
VR-10981 change encoding for strings

evgeniilezhnin:


backend/src/main/resources/migration/utf/mdb_columns.csv
```diff
@@ -0,0 +1,214 @@
+database_name,TABLE_NAME,COLUMN_NAME,DATA_TYPE,maximum_length
```
coryverta: It looks like this file is generated.  Let's not store it in git since we can recreate it later.

coryverta (approved):
Remove the CSV file and then it looks good.

evgeniilezhnin:
> Remove the CSV file and then it looks good.

removed

PR #2237, opened 2021-04-27
[VR-10902] Async log environment in run

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10902


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -874,4 +877,30 @@ public FutureExperimentRunDAO(
...
+                    handle ->
+                        handle
+                            .createUpdate(
+                                "UPDATE experiment_run SET environment = :environment WHERE id = :runId")
+                            .bind("runId", request.getId())
+                            .bind(
+                                "environment",
+                                environmentHandler.getEnvironmentStringFromBlob(
+                                    request.getEnvironment()))
+                            .execute()),
```
conradoverta: Why is this not in the environment handler?

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
```diff
@@ -232,10 +229,7 @@ private ExperimentRun getExperimentRunFromRequest(
               runValueMap.put("parent_id", newExperimentRun.getParentId());
               runValueMap.put("owner", newExperimentRun.getOwner());
 
-              EnvironmentBlob environmentBlob =
-                  sortPythonEnvironmentBlob(newExperimentRun.getEnvironment());
-              runValueMap.put(
-                  "environment", ModelDBUtils.getStringFromProtoObject(environmentBlob));
+              runValueMap.put("environment", null);
```
conradoverta: Seems like we're just saving the wrong value now. Why this change?

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: - Before this async DAO we are not set any value in database at creation time so in database that column has NULL value for each new runs.
- but then we have introduce new DAO and i forgot that logic so i have implemented `ModelDBUtils.getStringFromProtoObject(environmentBlob)` instead of null so it will create empty JSON in database while creating each runs
- So Now while implementing `logEnvironment` endpoint i realize that that logic is not right so i have added null here. it will not affect any user or  code.

anandjakhaniya: if you would mind to take empty JSON like `{}` database instead of null then this explicitly NULL set not needed.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Moved to environment handler

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
conradoverta: I don't mind either way, as long as we keep the behavior.

conradoverta (approved):

PR #2238, opened 2021-04-27
[VR-10900] Async get experiment run by id

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10900


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunServiceImpl.java
```diff
@@ -60,7 +60,21 @@ public void getExperimentRunsInExperiment(
...
-    super.getExperimentRunById(request, responseObserver);
+    try {
+      final var response =
+          futureExperimentRunDAO
+              .findExperimentRuns(
+                  FindExperimentRuns.newBuilder().addExperimentRunIds(request.getId()).build())
+              .thenApply(
+                  findResponse ->
+                      GetExperimentRunById.Response.newBuilder()
+                          .setExperimentRun(findResponse.getExperimentRuns(0))
```
conradoverta, line 84: You have no guarantee you'll have at least one result in the response.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunServiceImpl.java
anandjakhaniya: Throw error if more than one run found for given id

conradoverta: This is exactly the opposite of what I said. I said that we might not even have one result, as the `get` assumes there is at least one element. I'm not concerned about the case of more than one result.

conradoverta: We correctly handle the case of none found in https://github.com/VertaAI/modeldb/pull/2239/files

anandjakhaniya: Handled 404 case as well

conradoverta (approved):

PR #2239, opened 2021-04-27
[VR-10901] Async get experiment run by name

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-10901

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunServiceImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunServiceImpl.java
@@ -1,8 +1,12 @@
 package ai.verta.modeldb.experimentRun;
 
+import ai.verta.common.KeyValueQuery;
+import ai.verta.common.OperatorEnum;
+import ai.verta.common.ValueTypeEnum;
 import ai.verta.modeldb.*;
 import ai.verta.modeldb.common.CommonUtils;
 import ai.verta.modeldb.common.futures.FutureGrpc;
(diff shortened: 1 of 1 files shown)


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunServiceImpl.java
```diff
@@ -67,7 +71,35 @@ public void getExperimentRunById(
...
+                              .setValue(
+                                  Value.newBuilder().setStringValue(request.getName()).build())
+                              .setOperator(OperatorEnum.Operator.EQ)
+                              .setValueType(ValueTypeEnum.ValueType.STRING)
+                              .build())
+                      .build())
+              .thenApply(
+                  findResponse -> {
+                    if (findResponse.getExperimentRunsCount() == 0) {
+                      return GetExperimentRunByName.Response.newBuilder().build();
```
conradoverta: Should we return 404 in this case?

conradoverta (changes requested):
We're just dropping the experiment id that was in the request.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunServiceImpl.java
anandjakhaniya: Added 404 while not found the run by name

anandjakhaniya:
> We're just dropping the experiment id that was in the request.

I forgot to consider it. now added it in find cause

conradoverta (approved):

PR #2242, opened 2021-04-30
Correctly handle exception on future sequence

conradoverta:


anandjakhaniya (approved):
Tested and it is works as expected

PR #2246, opened 2021-05-04
Add check to output UAC error messages

ghost:


ghost:
Handling this in UAC instead.

PR #2252, opened 2021-05-04
[VR-10872] Support alert creation from summaries

daniel-verta:
## Changes
- implements an `__add__` method to `SummaryQuery` in order to support adding queries together. The behavior is such that earlier queries are not overwritten, only added to if fields are not defined. If this doesn't seem intuitive I'm open to naming the method and not using the **add** magic method.
- adds an optional base_summary_query to the constructor for the `Alerts` object to be used as a default summary query which is extended by optional summary sample queries.
- adds an `alerts` field to summaries with the base_summary_query set to match the parent summary. In this way that `alerts` object can be used to create alerts which always select for the parent summary.

## Testing
- `TestSummaryQuery` tests the introduced `__add__` method
- `TestIntegration::test_alert_from_monitored_summary` performs an integration test for alert creation
- `TestAlert::test_alerts_base_query` tests the internal helper function which produces the summary sample query using the default query. In the absence of better unit testing abilities for this code, it felt important to validate that piece.


client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -25,6 +25,7 @@
 )
 from verta._tracking import entity
 from verta import data_types
+from verta.operations.monitoring.alert import _entities
```
convoliution: This results in a circular import in Python 2

We might have to import it in the function where it's used 😬

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -117,6 +153,23 @@ def __repr__(self):
             self._ids, self._names, self._type_names, self._monitored_entity_ids
         )
 
+    def __add__(self, other):
+        assert isinstance(other, self.__class__)
```
convoliution: Returning `NotImplemented` à la [the data model reference](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).

Using `type(self)` instead of `self.__class__` because I'd personally favor using more public functions than dunders.
```suggestion
        if not isinstance(other, type(self)):
            return NotImplemented
```

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -267,11 +337,20 @@ def __init__(self, conn, conf, msg):
         self.name = msg.name
         self.type = msg.type_name  # TODO: hide me
 
+        alerts_query = SummaryQuery(ids=[self.id])
+        self._alerts = _entities.Alerts(
+            conn, conf, self.monitored_entity_id, base_summary_query=alerts_query
+        )
```
convoliution: How would you feel about moving these lines into `def alerts` since that's the only place these would be used anyway?

client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -432,13 +458,23 @@ def create(
...
 
+    @property
+    def base_summary_query(self):
+        return self._base_summary_query
+
+    def _add_query_to_default_query(self, summary_sample_query):
+        summary_query = self._base_summary_query + summary_sample_query.summary_query
+        updated_summary_sample_query = copy.copy(summary_sample_query)
+        updated_summary_sample_query.summary_query = summary_query
+        return updated_summary_sample_query
```
convoliution: The method name makes it sound like it's an in-place operation rather than a method that returns a new object. It also feels a bit unintuitive to add a sample query to a summary query, but maybe that's a bigger design matter. 🤔

client/verta/verta/operations/monitoring/summaries.py
```diff
@@ -89,6 +93,38 @@ def __init__(
...
+
+    @type_names.setter
+    def type_names(self, names):
+        if names:
+            type_classes = data_types._VertaDataType._from_type_strings(names)
+            self._data_type_classes = type_classes
+            self._type_names = names
+        else:
+            self._data_type_classes = None
+            self._type_names = None
```
convoliution: How are these intended to be used, out of curiosity? I also think we may run into unintuitive behavior with two properties overwriting the same attributes under the hood.

convoliution (changes requested):
I've probably thought through this much less than you have, but this seems like a lot of added machinery (methods, properties) for what I kind of assumed would be a very simple and minimal change.

Is there a reason to avoid simply adding a `_summary` attribute to `Alerts` and just inject the ID into the sample query in `Alerts.create()`? I'm partially also thinking of one of my tickets where I'm changing the API of `Alerts.create()` to not take a `SummarySampleQuery` object at all 😬

client/verta/verta/operations/monitoring/alert/_entities/alert.py
daniel-verta: I'm renaming it to `_combine_query_with_default_summary`

client/verta/verta/operations/monitoring/monitored_entity.py
```diff
@@ -86,24 +87,24 @@ def workspace(self):
     def alerts(self):
         return Alerts(self._conn, self._conf, self.id)
 
+    @property
+    def summaries(self):
+        return Summaries(self._conn, self._conf)
+
```
convoliution, line 93: I'm guessing this is the starting point for that other ticket about moving summary CRUD to ME 👌

client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -348,6 +354,17 @@ class Alerts(object):
...
+    Parameters
+    ----------
+    conn
+        A connection object to the backend service.
+    conf
+        A configuration object used by conn methods.
+    monitored_entity_id : int
+        A monitored entity id to use for all alerts in this collection
+    base_summary_query : :class:`~verta.operations.monitoring.summaries.SummaryQuery`
+        A base query to use for creating and finding alerts in this collection.
```
convoliution:
```suggestion
    summary : :class:`~verta.operations.monitoring.summaries.summary.Summary`
        A summary for creating and finding alerts in this collection, and finding samples to alert on.
```

client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -439,6 +467,31 @@ def create(
             last_evaluated_at_millis=time_utils.epoch_millis(last_evaluated_at),
         )
 
+    def _build_summary_query(self):
+        if self._summary:
+            return SummaryQuery(
+                ids=[self._summary.id],
+                names=[self._summary.name],
+                monitored_entities=[self._summary.monitored_entity_id],
+            )
```
convoliution, line 476: Side note (I'm likely going to do this in my tasks), it might make sense to also have
```python
elif self._monitored_entity_id:
    return SummaryQuery(
        monitored_entities=[self._monitored_entity_id],
    )
```
since the cron alerter needs something like this, but is doing it in a rather hacky way.

client/verta/tests/operations/monitoring/alerts/test_entities.py
```diff
@@ -73,22 +76,36 @@ def test_set_status(self, monitored_entity, summary_sample, created_entities):
...
 
         alert = alerts.create(name, alerter, sample_query)
-        created_query_proto = sample_query._to_proto_request()
+        created_query_proto = alerts._combine_query_with_default_summary(
+            sample_query
+        )._to_proto_request()
         retrieved_query_proto = alert.summary_sample_query._to_proto_request()
         assert created_query_proto == retrieved_query_proto
 
+    def test_alert_from_monitored_summary(self, client, monitored_entity):
```
convoliution: Typo, I think?
```suggestion
    def test_alert_from_summary(self, client, monitored_entity):
```

convoliution (approved):
Looks good, thank you! A few minor points.

PR #2253, opened 2021-05-04
[VR-11037] Deprecate log_training_data()

convoliution:


conradoverta (approved):

conradoverta:
Probably needs to remove from docs too

convoliution:
> Probably needs to remove from docs too

Along with examples, etc, which I'm leaving for another ticket.

PR #2256, opened 2021-05-06
python sdk: support manual python version override for python Version

houqp:


houqp:
closing this since it could break the system if model is pickled with a newer version of python that's not compatible with the version being specified.

PR #2257, opened 2021-05-07
[VR-10559] Drop "+..." suffix from version pins

convoliution:
PyTorch modifies its version number in the pip environment with metadata (most notably a `+cu102` for CUDA) which makes it impossible to feed back into a `pip install`. This change removes that suffix from `torch` in environment versioning (considering it's more about the machine's environment than the Python library version).

## Changes

- Change `env._as_proto().environment` to `_as_env_proto()` (which will also help down the line when the blob –> proto API is adjusted)
- Add `Python.requirements` property to help testing
- Add logic to remove "+..." suffix in `_pip_requirements_utils.process_requirements()`
- Add tests
- Shorten `verta.environment.Python` in tests to `Python`


client/verta/verta/environment/_python.py
```diff
@@ -203,11 +219,29 @@ def _capture_constraints(self, constraints):
...
-            for req_spec
-            in req_specs
+            map(
+                self._req_spec_to_msg,
+                req_specs,
+            ),
         )
 
+    @staticmethod
+    def _remove_torch_metadata(req_specs):
```
conradoverta: We should not assume this is only an issue for pytorch, as I'd rather not have another case come up. Is `+` in the install name valid according to pip? If not, we should just remove from all.

convoliution: According to [the Python spec](https://www.python.org/dev/peps/pep-0440/), the `+` is legal in version numbers, but I haven't personally seen it in PyPI. I'm about 80% comfortable with removing from all, because I'm not sure if it's being used for packages I'm not aware of. (and `+` is terrible as an internet search term)

conradoverta: sg. let's just have torch for now then

convoliution: While looking at robustifying the test, I found this in the PEP:
> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server . . .

So it now makes more sense to remove this for all packages, here.

conradoverta (approved):

PR #2265, opened 2021-05-11
[VR-11200] Migration and app startup support for Mssql

coryverta:
Modify the app startup and DB connection building to support MSSql server.  SQL Server uses a different format for attaching parameters to the URL, and does not support the autocreation of the database at startup time via the URL.


backend/common/src/main/java/ai/verta/modeldb/common/CommonHibernateUtil.java
```diff
@@ -520,11 +513,17 @@ public void createDBIfNotExists(RdbConfig rdb) throws SQLException {
       }
     }
 
-    String quotedDBName = String.format("`%s`", rdb.RdbDatabaseName);
```
conradoverta, line 523: I believe we need this for db names with `-` in mysql

coryverta: It gets handled in the check beneath.  MSSql doesn't support ` characters around database names, instead requiring " so each database has to be explicitly handled.

PR #2267, opened 2021-05-11
VR-10965 add exception cause to constructor

evgeniilezhnin:


conradoverta (approved):
I really dislike method overloading

PR #2269, opened 2021-05-11
Add client support for creating range alerters

daniel-verta:
### Changes

- adds a `RangeAlerter` class in `_alerter.py` for the new alerter type
- removes the expectation that an `Alerter` has a `comparison`, because this does not apply to the `RangeAlerter`

### Tests

- `test_alerter.py` is updated to test that a RangeAlerter can be constructed from its proto message and vice versa
- `test_entities.py` parametrizes `TestFixed` into `TestNonReference` because the behavior tested is identical for `RangeAlerter` and `FixedAlerter` types


client/verta/verta/operations/monitoring/alert/_alerter.py
```diff
@@ -16,41 +16,36 @@ class _Alerter(object):
...
     def _from_proto(msg):
-        comparison = comparison_module._VertaComparison._from_proto(
-            msg.operator,
-            msg.threshold,
-        )
+        _protos_to_classes = {
+            _AlertService.AlertFixed: FixedAlerter,
+            _AlertService.AlertReference: ReferenceAlerter,
+            _AlertService.AlertRange: RangeAlerter,
+        }
```
convoliution: I think there's a way we could have have these proto classes returned by our client classes, so that we could do something like
```python
@classmethod
def _from_proto(cls, msg):
    ...
    _protos_to_classes = {
        subcls._get_proto_class(): subcls
        for subcls
        in cls.__subclasses__
    }
```

Maybe not required here (let me know what you think) but it might reduce later maintenance so we won't need to append to this mapping.

client/verta/verta/operations/monitoring/alert/_entities/alert.py
```diff
@@ -51,6 +51,12 @@ class Alert(entity._ModelDBEntity):
 
     """
 
+    _ONEOF_ALERTER_ATTR = {
+        _AlertService.AlerterTypeEnum.FIXED: "alerter_fixed",
+        _AlertService.AlerterTypeEnum.REFERENCE: "alerter_reference",
+        _AlertService.AlerterTypeEnum.RANGE: "alerter_range",
+    }
```
convoliution: Somewhat similar to the comment in _alerter.py: I think we could have these strings defined on our alerter classes to be used in `_create_proto_internal()` to avoiding maintain an explicit mapping.

(though maybe that'd also require catching an `AttributeError` from `getattr()`)

client/verta/verta/operations/monitoring/alert/_alerter.py
```diff
@@ -16,41 +16,36 @@ class _Alerter(object):
...
+        if type(msg) in _protos_to_classes:
+            return _protos_to_classes[type(msg)]._from_proto(msg)
```
convoliution: I don't feel too strongly about this, but it _might_ be a good idea for the subclasses' method to be called something different. Elsewhere, I had spent more time than I'd like debugging an infinite recursion error because I was implementing a subclass but didn't realize I needed to override `_from_proto()` (since the base class implemented it).

But it would probably be the same time spent either way 🤷🏼‍♂️

```suggestion
            return _protos_to_classes[type(msg)]._from_proto_inner(msg)
```

client/verta/verta/operations/monitoring/alert/_alerter.py
```diff
@@ -142,3 +152,75 @@ def _as_proto(self):
...
+    .. note::
+
+        This alerter is only intended to work with summary samples of the
+        :class:`~verta.data_types.NumericValue` type.
+
+    Parameters
+    ----------
+    comparison : :class:`~verta.common.comparison._VertaComparison`
+        Alert condition. An alert is active if a queried sample meets this
+        condition.
```
convoliution: This will need to be changed.

convoliution (changes requested):
Some ideas for code maintainability—nothing critical

client/verta/verta/operations/monitoring/alert/_alerter.py
```diff
@@ -16,41 +16,49 @@ class _Alerter(object):
...
-        return self._comparison
+    @classmethod
+    @abc.abstractmethod
+    def _get_proto_class(cls):
+        raise NotImplementedError
+
+    @classmethod
+    @abc.abstractmethod
+    def _get_alert_field(cls):
+        raise NotImplementedError
```
convoliution, line 32: Did you happen to test this in Python 2? Asking because I don't personally know if they can be chained like this, and I _vaguely_ remember hearing about issues with it before. *shakes fist angrily at Python 2*

convoliution: https://stackoverflow.com/questions/11217878/python-2-7-combine-abc-abstractmethod-and-classmethod

Ahhh yes.
I think this is also why I ended up just tacking `_TYPE` as class attributes rather than methods or properties.

client/verta/verta/operations/monitoring/alert/_entities/alert.py
convoliution: I believe this can be removed now!

convoliution (changes requested):
Question about potential Python 2 shenanigans, but maybe it works anyway?

client/verta/verta/operations/monitoring/alert/_alerter.py
daniel-verta: I've double checked and confirmed that the tests pass in Python 2.7

```
(verta-py2) daniel@Daniels-MacBook-Pro verta % pytest tests/operations/monitoring/alerts/test_alerter.py
============================================================================== test session starts ===============================================================================
platform darwin -- Python 2.7.18, pytest-4.6.11, py-1.10.0, pluggy-0.13.1
rootdir: /Users/daniel/workspace/modeldb/client/verta/tests, inifile: pytest.ini
plugins: hypothesis-4.57.1
collected 8 items

tests/operations/monitoring/alerts/test_alerter.py ........                                                                                                                [100%]

============================================================================ 8 passed in 3.17 seconds ============================================================================
(verta-py2) daniel@Daniels-MacBook-Pro verta % pytest tests/operations/monitoring/alerts/test_entities.py
============================================================================== test session starts ===============================================================================
platform darwin -- Python 2.7.18, pytest-4.6.11, py-1.10.0, pluggy-0.13.1
rootdir: /Users/daniel/workspace/modeldb/client/verta/tests, inifile: pytest.ini
plugins: hypothesis-4.57.1
collected 12 items

tests/operations/monitoring/alerts/test_entities.py ............                                                                                                           [100%]

================================================================================ warnings summary ================================================================================
operations/monitoring/alerts/test_entities.py::TestIntegration::test_add_notification_channels
operations/monitoring/alerts/test_entities.py::TestIntegration::test_set_status
operations/monitoring/alerts/test_entities.py::TestAlert::test_update_last_evaluated_at
operations/monitoring/alerts/test_entities.py::TestAlert::test_creation_query_params
operations/monitoring/alerts/test_entities.py::TestAlert::test_creation_override_datetimes
operations/monitoring/alerts/test_entities.py::TestNonReferenceAlerters::test_crud[alerter0]
operations/monitoring/alerts/test_entities.py::TestNonReferenceAlerters::test_crud[alerter1]
operations/monitoring/alerts/test_entities.py::TestNonReferenceAlerters::test_repr[alerter0]
operations/monitoring/alerts/test_entities.py::TestNonReferenceAlerters::test_repr[alerter1]
operations/monitoring/alerts/test_entities.py::TestReference::test_crud
operations/monitoring/alerts/test_entities.py::TestReference::test_repr
  /Users/daniel/workspace/modeldb/client/verta/verta/_internal_utils/_utils.py:84: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
    raise_on_status=False)  # return Response instead of raising after max retries

-- Docs: https://docs.pytest.org/en/latest/warnings.html
==================================================================== 12 passed, 11 warnings in 77.76 seconds =====================================================================
```

convoliution (approved):

PR #2271, opened 2021-05-11
[VR-11200] Fix bug in ER updateModifiedTimestamp which used mysql-specific syntax

coryverta:
mssql does not support the built-in function `greatest` so it is necessary to perform the date comparison in code.


conradoverta (approved):
Wow

PR #2274, opened 2021-05-12
[VR-11200] Use bound parameters for escaped values for mssql compatibility

coryverta:
mssql for some reason thinks these are column names otherwise.  :facepalm:


conradoverta (approved):
There's a ton more like this in the ER speed up branch. Cc @ad-47

PR #2279, opened 2021-05-13
Fix test initialization logic which is broken after App.java was restructured

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/config/Config.java
```diff
@@ -81,18 +82,31 @@ public FutureJdbi getJdbi() {
...
+      return initializeJdbi(databaseConfig);
+    }
+    return this.jdbi;
+  }
 
-      final Jdbi jdbi = Jdbi.create(hikariDataSource);
-      final Executor dbExecutor = FutureGrpc.initializeExecutor(databaseConfig.threadCount);
-      this.jdbi = new FutureJdbi(jdbi, dbExecutor);
+  public FutureJdbi getTestJdbi() {
+    if (this.jdbi == null) {
```
conradoverta, line 91: Is it safe to use the exact same object?

backend/src/main/java/ai/verta/modeldb/config/Config.java
```diff
@@ -81,18 +82,31 @@ public FutureJdbi getJdbi() {
...
+    final var dbUrl = CommonHibernateUtil.buildConnectionString(databaseConfig.RdbConfiguration);
+    hikariDataSource.setJdbcUrl(dbUrl);
+    hikariDataSource.setUsername(databaseConfig.RdbConfiguration.RdbUsername);
+    hikariDataSource.setPassword(databaseConfig.RdbConfiguration.RdbPassword);
+
+    hikariDataSource.setMaximumPoolSize(10); // TODO: configure with a valid pool size
+
+    final Jdbi jdbi = Jdbi.create(hikariDataSource);
+    final Executor dbExecutor = FutureGrpc.initializeExecutor(databaseConfig.threadCount);
+    return new FutureJdbi(jdbi, dbExecutor);
```
conradoverta, line 110: We're no longer saving `this.jdbi` anywhere. We create a new instance every single time the `get` methods are called.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/config/Config.java
ad-47: We are checking `if (this.jdbi == null) {` before calling `initializeJdbi` method so if it was already initialized then we are return existing one.

anandjakhaniya: Yes. fixed this. it is a bug.

backend/src/main/java/ai/verta/modeldb/config/Config.java
anandjakhaniya: Yes. we are used same `jdbi` but used test config so yes it is safe.

conradoverta (approved):

PR #2283, opened 2021-05-13
Fix code version handler bugs for code version snapshot

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -1057,6 +1057,9 @@ public FutureExperimentRunDAO(
                                                           builder.getId())) {
                                                         builder.setCodeVersionSnapshot(
                                                             codeVersionsMap.get(builder.getId()));
+                                                      } else {
+                                                        builder.setCodeVersionSnapshot(
+                                                            CodeVersion.getDefaultInstance());
```
anandjakhaniya, line 1069: added empty response as per old code

anandjakhaniya: it will add empty `{}`

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CodeVersionHandler.java
```diff
@@ -76,6 +76,13 @@ public CodeVersionHandler(Executor executor, FutureJdbi jdbi) {
             maybeSnapshotId -> {
               if (maybeSnapshotId.isPresent()) {
                 if (request.getOverwrite()) {
+                  jdbi.useHandle(
+                      handle ->
+                          handle
+                              .createUpdate(
+                                  "UPDATE experiment_run SET code_version_snapshot_id = null WHERE id=:run_id")
+                              .bind("run_id", request.getId())
+                              .execute());
```
anandjakhaniya: we can not delete code version snapshot directly because experimentRun have foreign key reference so first need to remove code_version_snapshot_id from experimentRun and then we can delete snapshot. so fixed it here.

conradoverta: Then we need to put this in a transaction. Otherwise we can set it to `null` but fail to add new value. I have no clue how transactions will work while mixing with hibernate.

conradoverta (changes requested):
Transaction

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CodeVersionHandler.java
conradoverta: Can we just remove the foreign key instead?

anandjakhaniya: Removing foreign key might be affect many other unwanted places so instead of that i have handled this using hibernate transition and used hibernate sql query instead of jdbi sql query so we can used hibernate transaction.

anandjakhaniya:
> Transaction

Fixed https://github.com/VertaAI/modeldb/pull/2283#discussion_r632359003

conradoverta (approved):

PR #2286, opened 2021-05-13
Fix tags handler bugs

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/TagsHandler.java
```diff
@@ -42,7 +42,7 @@ public TagsHandler(Executor executor, FutureJdbi jdbi, String entityName) {
                     "select tags from tag_mapping "
                         + "where entity_name=:entity_name and "
                         + entityIdReferenceColumn
-                        + "=:entity_id")
+                        + "=:entity_id ORDER BY tags ASC")
```
anandjakhaniya, line 45: tags not maintain orders while returning so added order by cause

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/TagsHandler.java
```diff
@@ -75,10 +75,8 @@ public TagsHandler(Executor executor, FutureJdbi jdbi, String entityName) {
     var currentFuture =
         InternalFuture.runAsync(
             () -> {
-              for (final var tag : tags) {
-                if (tag.isEmpty()) {
-                  throw new InvalidArgumentException("Empty tag");
-                }
+              if (tags.isEmpty()) {
+                throw new InvalidArgumentException("Tags not found");
```
anandjakhaniya, line 79: if tags list are empty then it was never throw exception so fixed condition

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -338,7 +339,9 @@ public FutureExperimentRunDAO(
 
     return checkPermission(
             Collections.singletonList(runId), ModelDBActionEnum.ModelDBServiceActions.UPDATE)
-        .thenCompose(unused -> tagsHandler.addTags(runId, tags), executor)
+        .thenCompose(
+            unused -> tagsHandler.addTags(runId, ModelDBUtils.checkEntityTagsLength(tags)),
+            executor)
```
anandjakhaniya, line 344: [Missing to add in new Impl from old code] Added tag length validation while logging ER tags

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/TagsHandler.java
conradoverta: It was never what?

anandjakhaniya: I mean new async code was wrongly implement. it was never throw exception if tags are empty in request. so i have checked it right way here.

conradoverta: But we should also check that the tags themselves aren't empty strings, right?

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/TagsHandler.java
anandjakhaniya: Ohh yes, i will add it.

anandjakhaniya: Added validation check for each tag

PR #2287, opened 2021-05-13
Fix find experiment run predicates

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -667,6 +667,8 @@ public FutureExperimentRunDAO(
         InternalFuture.supplyAsync(
             () -> {
               final var localQueryContext = new QueryFilterContext();
+              localQueryContext.getConditions().add("experiment_run.deleted = :deleted");
+              localQueryContext.getBinds().add(q -> q.bind("deleted", false));
```
anandjakhaniya, line 683: `deleted = false` filter is missing on find ER DAO so added it.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -207,7 +207,12 @@
             new UnimplementedException("Unknown 'Value' type recognized"));
     }
 
-    queryContext = queryContext.addCondition(String.format("experiment_run.id in (%s)", sql));
+    if (operator.equals(OperatorEnum.Operator.NOT_CONTAIN)
+        || operator.equals(OperatorEnum.Operator.NE)) {
+      queryContext = queryContext.addCondition(String.format("experiment_run.id NOT IN (%s)", sql));
+    } else {
+      queryContext = queryContext.addCondition(String.format("experiment_run.id IN (%s)", sql));
+    }
```
anandjakhaniya, line 215: missing for hyperparameters so added it

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -176,7 +176,7 @@
 
     switch (value.getKindCase()) {
       case NUMBER_VALUE:
-        sql += applyOperator(operator, columnAsNumber(colValue, true), ":" + valueBindingName);
+        sql += applyOperator(operator, colValue, ":" + valueBindingName);
```
anandjakhaniya: Removed casting in number because while doing casting it will not return result so after testing and spending some time i realize that the query is working without casting as expected.

conradoverta: Is it!? In both mysql and postgres? For numeric data types? I'm _really_ doubtful it works since the DB has no idea that it should interpret the string as a number. Let's not merge this until we have collected evidence that it works correctly without the conversion.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: I knew it that you will have question in mind for this so i have got query for this 

```
select distinct experiment_run_id from keyvalue where entity_name='ExperimentRunEntity' and field_type='hyperparameters' 
and kv_key='C'  and cast(kv_value as decimal) >= 1.0E-4
```

You can execute it in mysql with and without casting. You will get correct result while execute it without casting but not with casting. 
i was also surprise but sql is managing string column with numeric value without casting.

anandjakhaniya: Reverted casting code

conradoverta (approved):

PR #2288, opened 2021-05-13
Fix checkPermission method

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -408,7 +408,10 @@ public FutureExperimentRunDAO(
 
   private InternalFuture<Void> checkPermission(
       List<String> runIds, ModelDBActionEnum.ModelDBServiceActions action) {
-    if (runIds.isEmpty()) {
+    // If request do not have ids and we are passing request.getId() then it will create list record
+    // with `""` string
+    final var finalRunIds = runIds.stream().filter(s -> !s.isEmpty()).collect(Collectors.toList());
+    if (finalRunIds.isEmpty()) {
```
anandjakhaniya, line 414: If request do not have ids and we are passing request.getId() in `Collections.singletonList(request.getId())` then it will create list record with `""` string so here old condition broken and now i have fixed it.

conradoverta (approved):

PR #2291, opened 2021-05-13
Fix ER test class with async ER service impl

anandjakhaniya:
Some file are already from other PR but added here for compilation and proper impl


anandjakhaniya:
this is covered in other PRs so closing this.

PR #2296, opened 2021-05-14
Fix test cases for not returning entity in response

anandjakhaniya:


anandjakhaniya:
We have stopped returning entity in response. but i am using entity from response for validate correctness of inserted data so not it is not there so i have to call explicitly to check inserted data is correct or not.

conradoverta (approved):

anandjakhaniya:
We will consider this PR later when we remove entity from response

jkwatson-verta:

PR #2311, opened 2021-05-19
Fix find ER bug for the all accessible resources

anandjakhaniya:


conradoverta (commented):
I don't understand what this PR does. `QueryFilterContext` is not a gather all for every info we want to pass around. `allowedAllResources=false` means no filter, which can already be expressed

anandjakhaniya:
> I don't understand what this PR does. `QueryFilterContext` is not a gather all for every info we want to pass around. `allowedAllResources=false` means no filter, which can already be expressed

Removed `allowedAllResources` from  `QueryFilterContext` and fix logic.

conradoverta (approved):

PR #2320, opened 2021-05-20
[VR-11319] Fix findHydratedExperimentRuns bug

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-11319

**Bug Reason:** We have implemented `findExperimentRuns` using InternalFuture so while any exception like `PermissionDenied`, etc. thrown from `findExperimentRuns` from line `futureExperimentRunDAO.findExperimentRuns(request).get();` will not handled correctly and throw an internal server error instead of those proper errors.


conradoverta (approved):
Nice!

PR #2328, opened 2021-05-21
[VR-11344] Fix reconcile cron bug

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-11344


conradoverta (commented):
Why do we need to load the entity to then delete the entity?

anandjakhaniya:
> Why do we need to load the entity to then delete the entity?

Because `CommentEntity` has a list of comment message so while deleting `CommentEntity` the sub list of entity not delete due to foreign key and hibernate not allowed to delete entity without removing foreign key entity. so need to fetch entity which load all entities with all foreign key sub entities and delete it automatically by hibernate.

conradoverta (approved):
I don't like it, but ok

PR #2329, opened 2021-05-21
[VR-11339] Implemented CloneExperimentRun endpoint

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-11339


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -1446,4 +1450,127 @@ private boolean checkAllResourceAllowed(List<Resources> resources) {
...
+                  return projectId.get();
+                },
+                executor)
+            .thenCompose(
+                projectId -> {
+                  if (!projectId.equals(cloneExperimentRunBuilder.getProjectId())) {
+                    return getEntityPermissionBasedOnResourceTypes(
+                            Collections.singletonList(projectId),
+                            ModelDBActionEnum.ModelDBServiceActions.UPDATE,
+                            ModelDBServiceResourceTypes.PROJECT)
```
conradoverta, line 1727: We need to check permission even if the project IDs are equal. There's no guarantee that the user can update the project.

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -1446,4 +1450,127 @@ private boolean checkAllResourceAllowed(List<Resources> resources) {
...
+            userInfo ->
+                findExperimentRuns(
+                        FindExperimentRuns.newBuilder()
+                            .addExperimentRunIds(request.getSrcExperimentRunId())
+                            .build())
+                    .thenApply(
+                        findExperimentRuns -> {
+                          if (findExperimentRuns.getExperimentRunsList().isEmpty()) {
+                            throw new PermissionDeniedException("Permission Denied");
+                          }
```
conradoverta: This might be permission denied or not found, right?

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
```diff
@@ -321,7 +318,8 @@ private ExperimentRun getExperimentRunFromRequest(
               return InternalFuture.sequence(futureLogs, executor)
                   .thenAccept(unused2 -> {}, executor);
             },
-            executor);
+            executor)
+        .thenCompose(unused2 -> createRoleBindingsForExperimentRun(newExperimentRun), executor);
```
conradoverta: This might fail after we have created the ER, which will cause our system to be inconsistent.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
```diff
@@ -386,4 +384,8 @@ private String buildRoleBindingName(
             },
             executor);
   }
+
+  public InternalFuture<Void> createClonedExperimentRun(ExperimentRun experimentRun) {
+    return insertExperimentRun(experimentRun);
+  }
```
conradoverta: https://github.com/VertaAI/docs/blob/main/engineering-practices/api-implementation/performance.md#avoid-layers-of-abstraction

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Ok, i will update it

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: So what we should do for this?? this is have UAC service so we can not used transaction on this. Any suggestion @conradoverta ??

conradoverta: Let's put a to-do. We can handle it later.

anandjakhaniya: Ohh Sorry, i forget about `created` flag. we have handled logic using `created` flag. if that line throwing error then at MDB we do not marked ER to `created` so we have maintain inconsistent

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Removed project id equal check

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Changed to NOT_FOUND

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/CreateExperimentRunHandler.java
anandjakhaniya: Improve code to resolved abstraction for clone ER

conradoverta (changes requested):
It's very confusing to have `create` and `insert` for ERs as they sound like they do the same thing, but they don't. Pick just one of them to be public.

anandjakhaniya:
> It's very confusing to have `create` and `insert` for ERs as they sound like they do the same thing, but they don't. Pick just one of them to be public.

Fixed as per slack discussion

conradoverta (approved):

PR #2332, opened 2021-05-21
Have summary.delete() delete itself

convoliution:
Motivation here is to change `summary.delete()` to delete the `summary` itself, as other `entity.delete()`s also delete themselves.

### DIFF
--- a/client/verta/tests/monitoring/conftest.py
+++ b/client/verta/tests/monitoring/conftest.py
@@ -28,10 +28,7 @@ def summary(client, monitored_entity, created_entities):
         monitored_entity,
     )
 
-    yield summary
-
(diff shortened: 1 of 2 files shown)


client/verta/tests/monitoring/conftest.py
```diff
@@ -28,10 +28,7 @@ def summary(client, monitored_entity, created_entities):
         monitored_entity,
     )
 
-    yield summary
-
-    # TODO: use `created_entities` if/when Summary reimplements delete()
-    client.monitoring.summaries.delete([summary])
+    return summary
```
convoliution, line 31: Summary deletion in tests is actually unnecessary, since the deletion of its parent monitored entity cascades to it.

daniel-verta (approved):

PR #2338, opened 2021-05-24
Update tests to check for 404

convoliution:
Some backend entity access checks have been updated to return `404`s instead of `403s`. This PR updates client integration tests accordingly.


conradoverta (approved):
I'd recommend making the string checks case insensitive

PR #2340, opened 2021-05-24
Restore comparison property to reference alerter

daniel-verta:
This property was accidentally removed. Adding it back with a test for its existence.


convoliution (approved):
🙌

PR #2348, opened 2021-05-25
Add support for aggregation summary samples to client

daniel-verta:
### Changes

- Adds an `Aggregation` class to be used for specifying how summary samples should be aggregated
- Adds an optional `aggregation` field to `SummarySampleQuery` 

### Testing

- conversion to and from protobuf messages is tested in `test_aggregation.py`

### DIFF
--- a/client/verta/setup.py
+++ b/client/verta/setup.py
@@ -1,7 +1,6 @@
 import os
 
-from setuptools import setup, find_packages
-
+from setuptools import find_packages, setup
 
 HERE = os.path.abspath(os.path.dirname(__file__))
 
--- a/client/verta/setup.py
+++ b/client/verta/setup.py
@@ -7,22 +7,22 @@ HERE = os.path.abspath(os.path.dirname(__file__))
 
 
 about = {}
-with open(os.path.join(HERE, "verta", "__about__.py"), "r") as f:
+with open(os.path.join(HERE, 'verta', '__about__.py'), 'r') as f:
     exec(f.read(), about)
 
-with open("README.md", "r") as f:
+with open("README.md", 'r') as f:
     readme = f.read()
 
 setup(
-    name=about["__title__"],
-    version=about["__version__"],
-    maintainer=about["__maintainer__"],
-    maintainer_email=about["__maintainer_email__"],
-    description=about["__description__"],
+    name=about['__title__'],
+    version=about['__version__'],
+    maintainer=about['__maintainer__'],
+    maintainer_email=about['__maintainer_email__'],
+    description=about['__description__'],
     long_description=readme,
     long_description_content_type="text/markdown",
-    license=about["__license__"],
-    url=about["__url__"],
+    license=about['__license__'],
+    url=about['__url__'],
     packages=find_packages(),
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
     install_requires=[
@@ -33,10 +33,9 @@ setup(
         "protobuf>=3.8, <4.0",
         "pyyaml>=5.1, <5.4",
         "requests>=2.21, <3.0",
-        "pytimeparse>=1.1.8, <2.0",
     ],
     entry_points={
-        "console_scripts": [
+        'console_scripts': [
             "verta = verta._cli:cli",
         ],
     },
--- a/client/verta/setup.py
+++ b/client/verta/setup.py
@@ -7,22 +7,22 @@ HERE = os.path.abspath(os.path.dirname(__file__))
 
 
 about = {}
-with open(os.path.join(HERE, 'verta', '__about__.py'), 'r') as f:
+with open(os.path.join(HERE, "verta", "__about__.py"), "r") as f:
     exec(f.read(), about)
 
-with open("README.md", 'r') as f:
+with open("README.md", "r") as f:
     readme = f.read()
 
 setup(
-    name=about['__title__'],
-    version=about['__version__'],
-    maintainer=about['__maintainer__'],
-    maintainer_email=about['__maintainer_email__'],
-    description=about['__description__'],
+    name=about["__title__"],
+    version=about["__version__"],
+    maintainer=about["__maintainer__"],
+    maintainer_email=about["__maintainer_email__"],
+    description=about["__description__"],
     long_description=readme,
     long_description_content_type="text/markdown",
-    license=about['__license__'],
-    url=about['__url__'],
+    license=about["__license__"],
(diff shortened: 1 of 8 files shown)


client/verta/setup.py
```diff
@@ -33,9 +33,10 @@
         "protobuf>=3.8, <4.0",
         "pyyaml>=5.1, <5.4",
         "requests>=2.21, <3.0",
+        "pytimeparse>=1.1.8, <2.0",
```
convoliution: Up a couple of lines to maintain alphabetical order please!

client/verta/verta/monitoring/summaries/queries.py
```diff
@@ -162,6 +163,8 @@ class SummarySampleQuery(object):
     created_after : datetime.datetime or int, optional
         Only fetch samples created at or after this time. Either a timezone
         aware datetime object or unix epoch milliseconds.
+    aggregation : :class:`Aggregation`, optional
```
convoliution:
```suggestion
    aggregation : :class:`~verta.monitoring.summaries.aggregation.Aggregation`, optional
```

client/verta/verta/monitoring/summaries/queries.py
```diff
@@ -162,6 +163,8 @@ class SummarySampleQuery(object):
...
+        Parameters for aggregation of summary samples
```
convoliution:
```suggestion
        Parameters for aggregation of summary samples.
```

client/verta/verta/monitoring/summaries/queries.py
```diff
@@ -234,6 +239,23 @@ def _find_summaries(self, proto_summary_query):
...
+    def aggregation(self, value):
+        if value is None:
+            self._aggregation = None
+        elif isinstance(value, AggregationQuerySummary):
+            self._aggregation = Aggregation._from_proto(value)
+        elif isinstance(value, Aggregation):
+            self._aggregation = value
+        else:
+            raise ValueError(
+                "Not a valid aggregation type"
```
convoliution: Maybe something like this
```suggestion
                "value must be Aggregation object or proto, not {}".format(type(value))
```

client/verta/verta/monitoring/summaries/aggregation.py
```diff
@@ -0,0 +1,77 @@
...
+    @classmethod
+    def operations(cls):
+        return cls._OPERATIONS.keys()
+
+    @classmethod
+    def _parse_operation(cls, value):
+        if value in cls._OPERATIONS:
+            return value
+        if type(value) is str:
+            if value in AggregationQuerySummary.AggregationOperation.keys():
```
convoliution: This might be totally unnecessary, but to make it extra Python 2-proof:

```python
from verta.external import six
```
```suggestion
        if isinstance(value, six.string_types):
            value = six.ensure_str(value)
            if value in AggregationQuerySummary.AggregationOperation.keys():
```

convoliution: Or
```python
if type(value) in six.string_types
```
if preferred, but I've been favoring `isinstance`

client/verta/verta/_internal_utils/time_utils/__init__.py
```diff
@@ -33,5 +36,30 @@ def epoch_millis(dt):
         raise ValueError("Cannot convert argument to epoch milliseconds")
 
 
+def timedelta_millis(delta):
+    if isinstance(delta, timedelta):
+        return int(delta.total_seconds() * 1000)
+    elif type(delta) == int and delta > 0:
```
convoliution:
```suggestion
    elif type(delta) is int and delta > 0:
```

client/verta/verta/_internal_utils/time_utils/__init__.py
```diff
@@ -33,5 +36,30 @@ def epoch_millis(dt):
...
+        return delta
+    else:
+        raise ValueError("Cannot convert argument to duration milliseconds")
```
convoliution: Since the other functions don't use an `else` for the catch-all error:
```suggestion
    raise ValueError("Cannot convert argument to duration milliseconds")
```

convoliution (commented):
Small comments! Also:
- I've been favoring `isinstance()` over `type() is` because of PEP 8, but I could always imagine cases where `type() is` is more accurately what we want. No strong personal preference either way, but just to note the slight break in convention in the codebase.
- I've been starting exception messages with a lowercase letter rather than capital because that seems to be the convention Python builtins follow. No strong preference in terms of practicality, but again just to note!

client/verta/verta/monitoring/summaries/aggregation.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __repr__(self):
+        return "Aggregation('{}',{})".format(self.granularity, self.operation)
+
+    @property
+    def granularity(self):
+        """The granularity of aggregation.
+
+        :getter: Returns the specified time granularity of aggregation
+        :setter: Sets the time granularity from a timedelta, string, or int of epoch milliseconds.
```
convoliution, line 49: I have absolutely no clue how Sphinx's autosummary handles docstrings inside properties 😬 But it's probably fine (worst case, the class docstring is descriptive enough for this to be used from looking at the API docs).

client/verta/verta/monitoring/summaries/queries.py
```diff
@@ -173,14 +177,17 @@ class SummarySampleQuery(object):
...
+        from verta.monitoring.summaries.aggregation import Aggregation
+        from verta.monitoring.summaries.queries import SummaryQuery, SummarySampleQuery
 
         samples = Client().monitoring.summary_samples
+
         sample_query = SummarySampleQuery(
             summary_query=SummaryQuery(names=["Income Distributions"]),
             labels={"datasource": ["census2010", "census2020"]},
             created_after=datetime(year=2021, month=2, day=22, tzinfo=timezone.utc),
+            aggregation=Aggregation(timedelta(days=7), "sum")
```
convoliution, line 190: Light suggestion: I prefer maintaining trailing commas for simpler diffs if more lines are added in the future.
```suggestion
            aggregation=Aggregation(timedelta(days=7), "sum"),
```

convoliution (approved):

client/verta/verta/monitoring/summaries/aggregation.py
daniel-verta: I actually did this based on something I saw online about how to document properties so that Sphinx gets it right. I'll fix it up if it results in terrible docs.

PR #2349, opened 2021-05-26
[VR-11451] Update parent timestamp using reconcile util

anandjakhaniya:
**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-11451


backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
```diff
@@ -0,0 +1,92 @@
...
+              .append(" FROM commit cm INNER JOIN repository_commit rc ")
+              .append(" ON rc.commit_hash = cm.commit_hash ")
+              .append(" INNER JOIN commit_parent cp ")
+              .append(" ON cp.parent_hash IS NOT NULL ")
+              .append(" AND cp.child_hash = cm.commit_hash ")
+              .append(" INNER JOIN repository rp ")
+              .append(" ON rp.id = rc.repository_id AND rp.date_updated < cm.date_created ")
+              .append(" WHERE rp.id IN (:ids) ")
+              .append(" GROUP BY rc.repository_id")
+              .toString();
```
conradoverta: Running this query again can get pretty wasteful. Instead, you could select the id and desired timestamp in the `resync` step.

backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
```diff
@@ -0,0 +1,92 @@
...
+
+      Query updateDeletedChildrenQuery = session.createSQLQuery(fetchUpdatedTimeQueryString);
+      updateDeletedChildrenQuery.setParameterList("ids", ids);
+      List<Object[]> fetchUpdatedTimeQuery = updateDeletedChildrenQuery.list();
+
+      var updateDatasetTimestamp = "UPDATE repository SET date_updated = ? WHERE id = ?";
```
conradoverta: We need to set the date updated to the greatest of what is in the db and the value we have, to avoid inconsistencies.

backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
```diff
@@ -0,0 +1,92 @@
...
+      fetchUpdatedDatasetIdQuery.setMaxResults(config.maxSync);
+      fetchUpdatedDatasetIdQuery.stream().forEach(id -> this.insert(Long.parseLong(id.toString())));
+    }
+  }
+
+  @Override
+  protected ReconcileResult reconcile(Set<Long> ids) {
+    LOGGER.debug("Reconciling projects " + ids.toString());
+
+    try (Session session = modelDBHibernateUtil.getSessionFactory().openSession()) {
```
conradoverta: Can we just use jdbi? You're already not leveraging hibernate at all for this and we're trying to move away from hibernate.

conradoverta (commented):
Note that we'll have to implement this for the other types too.

backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
anandjakhaniya: Taking greatest number in `resync` and updated here

backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
anandjakhaniya: Change PR with jdbi support

backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
```diff
@@ -0,0 +1,73 @@
...
+                  })
+              .list();
+        });
+  }
+
+  @Override
+  protected ReconcileResult reconcile(Set<AbstractMap.SimpleEntry<Long, Long>> updatedMaxDateMap) {
+    LOGGER.debug("Reconciling parent time update for repository" + updatedMaxDateMap.size());
+
+    futureJdbi.useHandle(
```
conradoverta: This will create a loose future, which can add a lot of pressure to the DB. We had the same issue in another reconciler. Could you copy the solution from there?

conradoverta (commented):
Let's also delete the corresponding cronjob.

anandjakhaniya:
> Let's also delete the corresponding cronjob.

Deleted

backend/src/main/java/ai/verta/modeldb/reconcilers/UpdateParentTimestampReconcile.java
anandjakhaniya: Fixed

conradoverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
```diff
@@ -22,10 +24,18 @@
...
 
   protected final ReconcilerConfig config;
+  protected final FutureJdbi futureJdbi;
+  protected final Executor executor;
 
   protected Reconciler(ReconcilerConfig config, Logger logger) {
+    this(config, logger, null, null);
+  }
+
+  protected Reconciler(ReconcilerConfig config, Logger logger, FutureJdbi futureJdbi, Executor executor) {
```
conradoverta: Method overload is evil. Why do you need these new fields? They don't seem to be used anywhere.

anandjakhaniya: We have used `futureJdbi and executor` at new created reconciler at here https://github.com/VertaAI/modeldb/pull/2349/files#diff-b3b240b8515b4098d6415d9903ff487f282e699b892b4facb007124e37c42282R40 so we have to introduce it in main class. The old reconciler utils do not have  `futureJdbi and executor` so for those utils it will be null till we are not convert it with `futureJdbi and executor`

conradoverta: Let's fix the overload then. If it's used, it's required.

anandjakhaniya: Done

conradoverta (approved):

PR #2356, opened 2021-05-27
Convert parsed times to UTC if missing tzinfo

daniel-verta:
## Changes

- in `time_utils.epoch_millis` promote timezone naive date times to UTC date times with a warning issued

## Minor Changes

- rename `timedelta_millis` to `duration_millis`
- split `test_utils.py` into a few smaller files in the `test_utils` folder

## Testing

- tests of time utils added in `test_time_utils.py`


client/verta/verta/_internal_utils/time_utils/__init__.py
```diff
@@ -26,15 +26,24 @@ def now_in_millis():
...
         return dt
     elif dt is None:
         return dt
     else:
         raise ValueError("Cannot convert argument to epoch milliseconds")
 
 
+def _promote_naive_to_utc(dt):
+    if dt.tzinfo is None:
+        warnings.warn("Time zone naive datetime found, assuming UTC time zone")
```
convoliution, line 41: Would you know if it'd be more intuitive to use the user's system's time zone?

client/verta/tests/time_strategies.py
```diff
@@ -1,11 +1,18 @@
...
-from verta._internal_utils.time_utils import timedelta_millis
+import hypothesis.strategies as st
+from verta._internal_utils.time_utils import duration_millis
+import warnings
+
+
+def duration_millis_ignore_warn(delta):
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        return duration_millis(delta)
```
convoliution, line 13: It feels a bit odd to me to suppress warnings at the `hypothesis` strategy level, like a mixture of separate concerns? If it's to avoid polluting test output, what would you think about [using `pytest`](https://docs.pytest.org/en/stable/warnings.html#pytest-mark-filterwarnings) to filter out the timezone conversion warning, instead?

client/verta/tests/test_utils/test_time_utils.py
```diff
@@ -0,0 +1,76 @@
...
+class TestEpochMillis:
+    def test_none(self):
+        assert time_utils.epoch_millis(None) is None
+
+    def test_naive_dt(self):
+        naive_dt = datetime.now()
+        assert naive_dt.tzinfo is None
+        with pytest.warns(UserWarning):
+            millis = time_utils.epoch_millis(naive_dt)
+            _check_positive_millis(millis)
```
convoliution: Which of these two lines is expected to raise the `UserWarning`? Could we separate them?

client/verta/tests/test_utils/test_time_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        delta = timedelta(milliseconds=millis)
+        duration = time_utils.duration_millis(delta)
+        _check_positive_millis(duration)
+
+    @given(millis=millis_uint64_strategy)
+    def test_downsample_resolution(self, millis):
+        delta = timedelta(milliseconds=millis, microseconds=2)
+        with pytest.warns(UserWarning):
+            duration = time_utils.duration_millis(delta)
+            assert duration == millis
```
convoliution:
```suggestion
            duration = time_utils.duration_millis(delta)
        assert duration == millis
```

client/verta/tests/test_utils/test_time_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        delta = timedelta(milliseconds=millis)
+        duration = time_utils.parse_duration(delta)
+        assert duration == delta
+
+    @given(millis=millis_uint64_strategy)
+    def test_downsample_resolution(self, millis):
+        delta = timedelta(milliseconds=millis, microseconds=5)
+        with pytest.warns(UserWarning):
+            duration = time_utils.parse_duration(delta)
+            assert duration == timedelta(milliseconds=millis)
```
convoliution:
```suggestion
            duration = time_utils.parse_duration(delta)
        assert duration == timedelta(milliseconds=millis)
```

convoliution (changes requested):

client/verta/tests/time_strategies.py
daniel-verta: This is in here basically as a convenience function for the strategy, because the hypothesis generator of time deltas _will_ generate time deltas with microseconds, and we use that generator as the basis for generating valid utc millisecond values. Doing it here is preferable in my opinion because in several of the tests I instead want to expect that the warnings will be produced.

client/verta/verta/_internal_utils/time_utils/__init__.py
daniel-verta: I think that defaulting to local timezone would be more immediately intuitive at the cost of downstream surprise. Defaulting to UTC in the absence of a time zone has the benefit of being a predictable and consistent kind of surprise, if it is a surprise at all. Also some of our early feedback has indicated approval of defaulting to UTC.

client/verta/tests/time_strategies.py
convoliution: Ahhh, I see. Yes, that makes sense to me now as well.

client/verta/verta/_internal_utils/time_utils/__init__.py
convoliution: Sounds good!

convoliution (approved):
🎉

PR #2357, opened 2021-05-27
[VR-11492] Allow parse_duration() to handle value=0

convoliution:
```
SummarySampleQuery._from_proto_request()
-> Aggregation._from_proto()
```
fails if `granularity = 0` (which is the default for an non-set proto value)


convoliution:
Incorporated into #2356

PR #2372, opened 2021-06-04
Make test order deterministic

conradoverta:


convoliution (commented):
@conradoverta Would you be able to explain the issue this was encountering with parallelism (or provide steps to reproduce)? I haven't seen it myself and I'd like to understand it better!

Also, I _think_ we could just cast the `set` to a `list` on line 41 rather than `enumerate`ing it, and that should maintain access order a bit more simply.

conradoverta:
> @conradoverta Would you be able to explain the issue this was encountering with parallelism (or provide steps to reproduce)? I haven't seen it myself and I'd like to understand it better!
> 
> Also, I _think_ we could just cast the `set` to a `list` on line 41 rather than `enumerate`ing it, and that should maintain access order a bit more simply.

If you specify any parallelism on pytest, it will list all tests in all executors and make sure they match so that they can distribute. If their listing returns a different order or a different set, it doesn't even start testing.

Changing `set` to `list` sounds fine.

convoliution:
```sh
$ pytest test_versioning/test_code.py::TestGit::test_autocapture -n 10
===================================================== test session starts ======================================================
platform darwin -- Python 3.9.2, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
plugins: xdist-2.2.1, forked-1.3.0, hypothesis-6.8.3
gw0 [27] / gw1 [27] / gw2 [27] / gw3 [27] / gw4 [27] / gw5 [27] / gw6 [27] / gw7 [27] / gw8 [27] / gw9 [27]
...........................
====================================================== 27 passed in 3.11s ======================================================
```

PR #2373, opened 2021-06-04
Metrics are only initialized once

coryverta:
The tests do not properly isolate the initialization sequence.  This causes multiple calls to the metrics collectors, generating errors that break tests.

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/App.java
+++ b/backend/src/main/java/ai/verta/modeldb/App.java
@@ -49,6 +49,7 @@ import java.util.Optional;
 import java.util.TimerTask;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import javax.management.MalformedObjectNameException;
 import liquibase.exception.LiquibaseException;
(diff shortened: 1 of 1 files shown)


backend/src/main/java/ai/verta/modeldb/App.java
```diff
@@ -107,13 +108,16 @@ public static ApplicationContext getContext() {
 
   // Export all JMX metrics to Prometheus
   private static final String rules = "---\n" + "rules:\n" + "  - pattern: \".*\"";
+  private static final AtomicBoolean metricsInitialized = new AtomicBoolean(false);
```
conradoverta, line 111: Should we use a mutex or is this never run in parallel?

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/App.java
coryverta: It's never run in parallel, and typically is only ever run exactly once.  In the tests the setup functionality calls this once per test during the setup phase, so access is always sequential.

coryverta: FWIW: We should refactor the the tests to avoid this, since it means that setup logic that is designed to run once is getting called more than once.  Normally it just doesn't matter, but the metrics collectors must be JVM-level singletons.

PR #2374, opened 2021-06-04
Fix floating point error in computing millis for time delta

daniel-verta:
Computing milliseconds for a `timedelta` by multiplying the floating point `total_seconds` by 1000 doesn't guarantee that the round trip of "to milliseconds" and "from milliseconds" isn't broken by floating point math. I've fixed this and added a test for the bug.


convoliution (approved):
A comment in the body of `duration_millis()` would be appreciated as well!

PR #2378, opened 2021-06-08
[VR-11563, VR-11564] Provide VertaModelBase interface

convoliution:
Rendered doc page can be viewed here: https://docs.verta.ai/en/ml-model-base/_autogen/verta.registry.VertaModelBase.html


client/verta/verta/registry/_verta_model_base.py
```diff
@@ -0,0 +1,118 @@
...
+@six.add_metaclass(abc.ABCMeta)
+class VertaModelBase(object):
+    """Abstract base class for Verta Standard Models.
+
+    .. note::
+
+        ``__init__()`` and :meth:`predict` **must** be implemented by
+        subclasses. Any other methods are optional.
+
+    Parameters
```
mvartakAtVerta, line 17: Wrong spot?

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @abc.abstractmethod
+    def predict(self, input):
+        """Produce an output from `input`.
+
+        This method is called when requests are made against a Verta endpoint.
+
+        .. note::
+
+            This method should expect to recieve and return values that are
```
mvartakAtVerta: Should this expect to do the deserialization itself? what is compatible? What's not?

client/verta/verta/registry/_verta_model_base.py
convoliution: This is the right spot—Sphinx (as far as I've seen) looks for `__init__()` parameters on the class docstring.

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -0,0 +1,118 @@
...
+        model = Model(
+            artifacts=model_ver.fetch_artifacts(["np_matrix"]),
+        )
+
+        # persist to model version
+        model_ver.log_model(Model, artifacts=["np_matrix"])
+
+    """
+
+    @abc.abstractmethod
```
mvartakAtVerta, line 53: This doesn't seem to show on the docs;  expected?

client/verta/verta/registry/_verta_model_base.py
convoliution: > Should this expect to do the deserialization itself?

It should expect that the value it gets as `input` is a Python primitive, and it should return Python primitives. If it wants to handle a NumPy array, `predict()`'s body will need to manually cast it (`input = np.array(input)`). Does that make sense? Would you have wording recommendation that might make this clearer?

---
> what is compatible? What's not?

I was hoping that the
> JSON-compatible (most Python built-in types)

would explain this, but maybe it's not as clear as I expected. The answer is `tuple`, `list`, `dict`, `int`, `float`, `bool`, `str`, `None`. I can list these out explicitly.

client/verta/verta/registry/_verta_model_base.py
convoliution: Sphinx doesn't render `__init__()` by default. I can change our templates to do so, but it'll take time (maybe an hour?) if I need to change other docstrings to accommodate it.

How important would it be to show this?

daniel-verta (approved):

convoliution:
I am removing `describe()` and `example()` because they aren't part of our upcoming plans for our standard model. For posterity:

```python
def describe(self):
    """Return a rich description of the model's behavior for use by the model sandbox.

    Returns
    -------
    dict of str to str
        A mapping where values are textual descriptions, and keys are of
        the following values:

        - method
        - args
        - returns
        - description
        - input_description
        - output_description

    """
    return ""

def example(self):
    """Return example input data for :meth:`predict` for use by the model sandbox.

    Returns
    -------
    any JSON-compatible type
        Example model input

    Examples
    --------
    .. code-block:: python

        output = model.predict(
            input=model.example(),
        )

    """
    return ""
```

client/verta/verta/registry/_verta_model_base.py
convoliution: After a discussion, I have filed VR-11770 to investigate whether it would make sense to document __init__() separately; for now, it seems like the convention is to not.

PR #2394, opened 2021-06-10
Amend Python env repr and make modelversion env update more inclusive

daniel-verta:
* I've updated the `Python.__repr__` method for improved visibility of the attribute
* The modelversion update now applies to all of the environment rather than a specific subset of the environment fields

```
In [3]: client = Client()
set host from environment
set email from environment
set developer key from environment

In [4]: model = client.get_registered_model(name='Model 369011623294573428545')

In [5]: ver = model.get_version(name='version 4')

In [6]: ver.get_environment()
Out[6]:
Python Version
    Python 3.7.9
    requirements:
        cloudpickle==1.6.0
        verta==0.17.7a4
    apt packages:
        python-opencv
    command line arguments:
        /Users/daniel/.pyenv/versions/verta-py3/bin/ipython
```

### DIFF
--- a/client/verta/verta/registry/entities/_modelversion.py
+++ b/client/verta/verta/registry/entities/_modelversion.py
@@ -18,11 +18,7 @@ from ..._protos.public.common import CommonService_pb2 as _CommonCommonService
 
 from ...external import six
 
-from ..._internal_utils import (
-    _utils,
-    _artifact_utils,
-    importer, _request_utils
-)
+from ..._internal_utils import _utils, _artifact_utils, importer, _request_utils
 from ... import utils
 
 from ...tracking.entities._entity import _MODEL_ARTIFACTS_ATTR_KEY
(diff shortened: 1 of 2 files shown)


client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -499,23 +585,25 @@ def log_environment(self, env, overwrite=False):
...
+            raise ValueError(
+                "environment already exists; consider setting overwrite=True"
+            )
 
         if overwrite:
             self._fetch_with_no_cache()
             self._msg.environment.CopyFrom(env._as_env_proto())
             self._update(self._msg, method="PUT")
         else:
-            self._update(self.ModelVersionMessage(environment=env._as_env_proto()), method="PATCH",
```
daniel-verta, line 512: Apologies for the autoformat complications, but this is the one significant change here

convoliution: That certainly is simplify simpler than what it was before!

convoliution (approved):

PR #2397, opened 2021-06-10
[VR-11539] Implement convenience functions for creating Standard Verta Model versions

convoliution:
## Changes
- add `_internal_utils.model_validator` functions
- add to `RegisteredModel`:
  - `create_standard_model()`
  - `create_standard_model_from_keras()`
  - `create_standard_model_from_sklearn()`
  - `create_standard_model_from_torch()`
  - `create_standard_model_from_xgboost()`
- add so many tests

## Tests
The tests unfortunately failing in Python 2 with a segfault from TensorFlow. I am investigating how to resolve it, but it is most likely unrelated to the client implementation itself.

Here they are running in Python 3.7:
```bash
$ pytest test_model_registry/test_standard_model.py
plugins: hypothesis-6.14.0
collected 83 items                                                                                                          

test_model_registry/test_standard_model.py .......................................................................... [ 89%]
.........                                                                                                             [100%]

======================================= 83 passed, 55 warnings in 2089.98s (0:34:49) ========================================
```

## Followups
Unless proposed otherwise, I consider these to be follow-up enhancements that are not strictly necessary for this ticket. I will file follow-ups for them.
- Raise an explicit `TypeError` if the user tries to log an unsupported Keras subclassing API model. The check that I thought I could make is only available in certain versions of TF/Keras, and investigating an alternative is an indefinite rabbit hole. Keras will raise its own `NotImplementedError` during serialization anyway, so the code path is covered.
- Automatically inject known required libraries into `environment` param. For example, if the user calls `create_model_from_sklearn()`, we know it's going to need `"scikit-learn"` so we can make sure it's set as a courtesy. It's tricky to do today without something like `Python.__add__()`.


convoliution:
Moving back to draft while I debug the tests.

client/verta/tests/models/standard_models.py
```diff
@@ -0,0 +1,187 @@
...
+    # # sequential API
+    # model = keras.Sequential(
+    #     [
+    #         keras.layers.Dense(3, activation="linear", name="layer1"),
+    #         keras.layers.Dense(2, activation="relu", name="layer2"),
+    #         keras.layers.Dense(1, activation="sigmoid", name="layer3"),
+    #     ]
+    # )
+    # model(np.random.random(size=(3, 3)))  # initialize weights
+    # models.append(model)
```
convoliution, line 44: I've commented out some models because they actually do cause test failures.

But because this PR is just simple wrappers over existing functionality, I don't personally consider them a reason to block this ticket.

I've split off investigating+resolving the values to VR-11962.

daniel-verta (approved):
Approving with a note that we might actually want a single function which dispatches to the right helper function based on the model type. We can always deprecate these functions and point to a unified interface in the future though if we decide to simplify in that way.

This would look something like:

```
@singledispatch
def create_standard_model(...):
...

@create_standard_model.register(tf.keras.Sequential)
def _create_standard_model_from_keras(...)
..

@create_standard_model.register(sklearn.base.BaseEstimator)
def _create_standard_model_from_sklearn(...)
...
```
And then `create_standard_model` would be called by the user with any of the registered types as the model, with appropriate redirection to the registered functions.

(Using [PyPI Singledispatch](https://pypi.org/project/singledispatch/). A little bit of indirection might be necessary for that library to not look at what would otherwise be `self`)

PR #2425, opened 2021-06-18
Enable JDBI3 tracing

coryverta:
- Enable tracing of JDBI calls
- Demote hikari logging
- Clean up pom.xml

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/config/Config.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/config/Config.java
@@ -15,17 +15,15 @@ import io.opentracing.contrib.grpc.TracingClientInterceptor;
 import io.opentracing.contrib.grpc.TracingServerInterceptor;
 import io.opentracing.contrib.jdbc.TracingDriver;
 import io.opentracing.util.GlobalTracer;
-import io.opentracing.contrib.jdbi.OpenTracingCollector;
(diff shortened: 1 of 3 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/config/Config.java
```diff
@@ -129,7 +128,11 @@ public FutureJdbi initializeJdbi(DatabaseConfig databaseConfig, String poolName)
     hikariDataSource.setMetricsTrackerFactory(new PrometheusMetricsTrackerFactory());
     hikariDataSource.setPoolName(poolName);
 
-    final Jdbi jdbi = Jdbi.create(hikariDataSource);
+    if (enableTrace) {
+      final var tracer = Configuration.fromEnv().getTracer();
```
conradoverta, line 132: I _think_ we have to use the same tracer, vs creating a new one. I don't know if this creates a new one or not.

conradoverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/config/Config.java
coryverta: I think it must create a new one, because the spans aren't linked.

coryverta: I'll research how to use the existing tracer.

PR #2431, opened 2021-06-22
Fix SoftDeleteRepositories reconcile issue due to OptimisticLockException

anandjakhaniya:
[Kibana](https://kibana.dev.verta.ai/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2021-06-20T17:30:00.000Z',to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:f4c6aa40-32b2-11eb-b0d6-b3168b0b5e7b,key:kubernetes.namespace_name,negate:!f,params:(query:production),type:phrase),query:(match_phrase:(kubernetes.namespace_name:production))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:f4c6aa40-32b2-11eb-b0d6-b3168b0b5e7b,key:kubernetes.labels.app,negate:!f,params:(query:modeldb--backend),type:phrase),query:(match_phrase:(kubernetes.labels.app:modeldb--backend))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:f4c6aa40-32b2-11eb-b0d6-b3168b0b5e7b,key:level,negate:!f,params:(query:ERROR),type:phrase),query:(match_phrase:(level:ERROR)))),index:f4c6aa40-32b2-11eb-b0d6-b3168b0b5e7b,interval:auto,query:(language:kuery,query:''),sort:!(!('@ts',asc))))

All DeleteEntitiesCron issues like 
```
DeleteEntitiesCron : deleteRepositories : Exception: Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1;
```
in above kibana is due to OptimisticLockException which has fixed in this PR.

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
@@ -68,16 +68,30 @@ public abstract class Reconciler<T> {
               CommonUtils.registeredBackgroundUtilsCount();
               try {
                 if (deduplicate) {
-                  Set<T> processingIds =
-                      pop().stream()
-                          .filter(id -> !processingIdSet.contains(id))
-                          .collect(Collectors.toSet());
-                  if (!processingIds.isEmpty()) {
+                  Set<T> idsToProcess;
+                  // Fetch ids to process while avoiding race conditions
+                  try {
+                    lock.lock();
+                    idsToProcess = pop().stream()
+                            .filter(id -> !processingIdSet.contains(id))
+                            .collect(Collectors.toSet());
+                    if (!idsToProcess.isEmpty()) {
+                      processingIdSet.addAll(idsToProcess);
+                    }
+                  } finally {
+                    lock.unlock();
+                  }
+
+                  if (!idsToProcess.isEmpty()) {
                     try {
-                      processingIdSet.addAll(processingIds);
-                      reconcile(processingIds);
+                      reconcile(idsToProcess);
                     } finally {
-                      processingIdSet.removeAll(processingIds);
+                      lock.lock();
+                      try {
+                        processingIdSet.removeAll(idsToProcess);
+                      } finally {
+                        lock.unlock();
+                      }
                     }
                   }
                 } else {
--- a/backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
@@ -25,16 +25,18 @@ public abstract class Reconciler<T> {
   final Condition notEmpty = lock.newCondition();
   // To prevent OptimisticLockException
   private final Set<T> processingIdSet = new HashSet<>();
+  private final boolean deduplicate;
 
   protected final ReconcilerConfig config;
   protected final FutureJdbi futureJdbi;
   protected final Executor executor;
 
-  protected Reconciler(ReconcilerConfig config, Logger logger, FutureJdbi futureJdbi, Executor executor) {
+  protected Reconciler(ReconcilerConfig config, Logger logger, FutureJdbi futureJdbi, Executor executor, boolean deduplicate) {
     this.logger = logger;
     this.config = config;
     this.futureJdbi = futureJdbi;
     this.executor = executor;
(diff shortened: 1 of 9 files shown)


conradoverta (commented):
I think I'm missing something here. I don't believe `reconcile` can be called with similar ids in parallel. So `processingIdSet` should always be empty. Do you have evidence of it not being empty?

anandjakhaniya:
> I think I'm missing something here. I don't believe `reconcile` can be called with similar ids in parallel. So `processingIdSet` should always be empty. Do you have evidence of it not being empty?

Yes, i have reproduce it. lets assume ids being empty but now a days UAC will take time to process `deleteEntityResourcesWithServiceUser` so still those entities are not deleted and those are in process but at that time reconcile utils again wake up at schedule time and  it found those entity not deleted because it was on process from first wake up. So after going further, both reconcile jobs try to delete objects and get Optimistic Lock Exception in second one.

conradoverta:
> > I think I'm missing something here. I don't believe `reconcile` can be called with similar ids in parallel. So `processingIdSet` should always be empty. Do you have evidence of it not being empty?
> 
> Yes, i have reproduce it. lets assume ids being empty but now a days UAC will take time to process `deleteEntityResourcesWithServiceUser` so still those entities are not deleted and those are in process but at that time reconcile utils again wake up at schedule time and it found those entity not deleted because it was on process from first wake up. So after going further, both reconcile jobs try to delete objects and get Optimistic Lock Exception in second one.

I'd rather fix this in the reconciler tool itself not to allow duplicated ids to begin with. It has the same effect, but it fixes this issue for all reconcilers at once.

anandjakhaniya:
> I'd rather fix this in the reconciler tool itself not to allow duplicated ids to begin with. It has the same effect, but it fixes this issue for all reconcilers at once.

Ok, that would be great, i have try to implement it in `Reconciler<T>` class but i thought you will not agree with it so i do not have added it there. :)

conradoverta (commented):
As discussed in [slack](https://vertaai.slack.com/archives/CJG0NL86B/p1624515605256600?thread_ts=1624515567.256400&cid=CJG0NL86B), this has to be optional.

anandjakhaniya:
> As discussed in [slack](https://vertaai.slack.com/archives/CJG0NL86B/p1624515605256600?thread_ts=1624515567.256400&cid=CJG0NL86B), this has to be optional.

Updated PR with deduplicate logic

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
```diff
@@ -15,23 +16,28 @@
...
+import java.util.stream.Collectors;
 
 public abstract class Reconciler<T> {
   final Logger logger;
   protected final HashSet<T> elements = new HashSet<>();
   protected final LinkedList<T> order = new LinkedList<>();
   final Lock lock = new ReentrantLock();
   final Condition notEmpty = lock.newCondition();
+  // To prevent OptimisticLockException
+  private final Set<T> processingIdSet = ConcurrentHashMap.newKeySet();
```
conradoverta, line 28: As per the the conversation in slack, this is not enough to make things thread safe.

conradoverta (changes requested):

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
anandjakhaniya: `ConcurrentHashMap.newKeySet();` will create the thread safe `Set`  so it is enough as per my thought.

conradoverta: It is not. Please refer to the slack thread where I described the 2 race conditions. This fixes one (as I described in the thread) but not the other

conradoverta (changes requested):
Still has a race condition

anandjakhaniya:
> Still has a race condition

Added lock for second case.

conradoverta (commented):
Following up in slack for speed: https://vertaai.slack.com/archives/CJG0NL86B/p1625648035303300

conradoverta (approved):

PR #2435, opened 2021-06-22
Fix summary sample doc

CaptEmulation:
Fixed summary sample names


convoliution:
@CaptEmulation @daniel-verta Looking over this, I'm noticing that it's the attribute names themselves that are wrong, rather than the docs. They should be changed to match the parameters of `summary.log_sample()` (`time_window_start`, `time_window_end`, and `created_after`). I'll try to get around to this sometime soon.

daniel-verta (approved):

liuverta:
The monitoring sub-client has been removed.

PR #2438, opened 2021-06-23
Removed backgroundUtils count logic

anandjakhaniya:


backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
```diff
@@ -24,6 +26,21 @@
   public <R, S> Listener<R> interceptCall(
       ServerCall<R, S> call, Metadata requestHeaders, ServerCallHandler<R, S> next) {
     Context context = Context.current().withValue(METADATA_INFO, requestHeaders);
+
+    // To protect empty headers from request
```
conradoverta: I don't get this comment.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
```diff
@@ -24,6 +26,21 @@
...
+    Metadata.Key<String> dev_key =
+        Metadata.Key.of("developer_key", Metadata.ASCII_STRING_MARSHALLER);
+    Metadata.Key<String> source_key = Metadata.Key.of("source", Metadata.ASCII_STRING_MARSHALLER);
+
+    if (requestHeaders == null
+        || !requestHeaders.containsKey(email_key)
+        || !requestHeaders.containsKey(dev_key)
+        || !requestHeaders.containsKey(source_key)) {
+      Status status = Status.newBuilder().setMessage("Required parameter is missing in metadata").build();
+      throw StatusProto.toStatusRuntimeException(status);
```
conradoverta: I don't think we can throw exceptions from the interceptor

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthServiceChannel.java
```diff
@@ -56,10 +56,10 @@ public AuthServiceChannel(
   }
 
   private Metadata getMetadataHeaders() {
-    int backgroundUtilsCount = CommonUtils.getRegisteredBackgroundUtilsCount();
-    LOGGER.trace("Header attaching with stub : backgroundUtilsCount : {}", backgroundUtilsCount);
+    boolean metadataNotExists = metadataInfo == null || metadataInfo.get() == null;
+    LOGGER.trace("Header info exists: {}", metadataNotExists);
     Metadata requestHeaders;
-    if (backgroundUtilsCount > 0 && (metadataInfo == null || metadataInfo.get() == null)) {
+    if (metadataNotExists) {
```
conradoverta: I'd rather get rid of this altogether. It's an implicit thing whether to use a service account or not, which can easily be done by mistake and causes all sort of errors. I'd like the use of service account be a deliberate choice instead of the default to avoid these bugs.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
anandjakhaniya: If user do not set the metadata headers in request then we are protecting request in the interceptor instead of passing to the backend api.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
anandjakhaniya: Yes, agree i have fixed the code with the better solution

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthServiceChannel.java
anandjakhaniya: we have already make a service account choice deliberate from here https://github.com/VertaAI/modeldb/blob/b224fb60835c0e83440d3583eab15700a02e8269/backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java#L123 and https://github.com/VertaAI/modeldb/blob/b224fb60835c0e83440d3583eab15700a02e8269/backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java#L128 so whatever used auth channel they know the user preference  and they will initialize service based on the same.

conradoverta: You are assuming we're always implementing the code correctly when defining those methods. You should not make this assumption. If we pass `null` in the argument by mistake (which is _very_ easy because the variable itself can have value `null` so it's not even a typo in the code), then we'll use the superuser credentials. That's not ok. The default behavior must be safety: do not assume that a missing value means the call should be made with a super user; force the caller to say "yes, I want to do this as a super user".

anandjakhaniya: Updated PR with the your suggestion. Now all auth channel logic is implemented and managed by UAC.java class. The Service account status is now separated.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
```diff
@@ -24,6 +28,24 @@
   public <R, S> Listener<R> interceptCall(
       ServerCall<R, S> call, Metadata requestHeaders, ServerCallHandler<R, S> next) {
     Context context = Context.current().withValue(METADATA_INFO, requestHeaders);
+
+    // validate empty headers from user request
+    Metadata.Key<String> email_key = Metadata.Key.of("email", Metadata.ASCII_STRING_MARSHALLER);
+    Metadata.Key<String> dev_key =
+        Metadata.Key.of("developer_key", Metadata.ASCII_STRING_MARSHALLER);
```
conradoverta, line 35: We must check either `developer_key` or `developer-key` is present.

backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
```diff
@@ -75,7 +66,7 @@ public boolean IsImplemented() {
       String resourceId,
       String resourceOwnerId,
       Metadata requestHeaders) {
-    try (AuthServiceChannel authServiceChannel = new AuthServiceChannel()) {
+    try (AuthServiceChannel authServiceChannel = uac.getBlockingAuthServiceChannel()) {
```
conradoverta, line 69: I'm pretty sure that, if you do this here, we'll attempt to do `authServiceChannel.close()` at the end of the `try`, which will close the shared object, right? Could you confirm?

anandjakhaniya: Yes, it will do automatically because `AuthServiceChannel` class has implemented AutoClose interface which is give one method `close()` where i am doing `authServiceChannel.close()` already.

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
anandjakhaniya: That i will add it.

conradoverta (approved):

PR #2439, opened 2021-06-23
[VR-9109] Support Windows

mvartakAtVerta:
Various fixes for log_artifacts for windows

## Changes
- close `NamedTemporaryFile`s before passing them downstream (with `delete=False`) to make Windows happy
- use `pip list --format=freeze` instead of `pip freeze` for Windows environment compatibility

### DIFF
--- a/client/verta/tests/conftest.py
+++ b/client/verta/tests/conftest.py
@@ -514,9 +514,9 @@ def organization(client, created_entities):
 
 @pytest.fixture
 def requirements_file():
-    with tempfile.NamedTemporaryFile("w+") as tempf:
+    with tempfile.NamedTemporaryFile("w+", delete=False) as tempf:
         # create requirements file from pip freeze
-        pip_freeze = subprocess.check_output([sys.executable, "-m", "pip", "freeze"])
+        pip_freeze = subprocess.check_output([sys.executable, "-m", "pip", "list", "--format=freeze"])
         pip_freeze = six.ensure_str(pip_freeze)
         tempf.write(pip_freeze)
         tempf.flush()  # flush object buffer
--- a/client/verta/tests/conftest.py
+++ b/client/verta/tests/conftest.py
@@ -294,12 +294,12 @@ def dir_and_files(strs, tmp_path, request):
     dirpath = tmp_path / request.param
 
     filepaths = {
-        os.path.join(strs[0], strs[1], strs[2]),
-        os.path.join(strs[0], strs[1], strs[3]),
-        os.path.join(strs[0], strs[2]),
-        os.path.join(strs[0], strs[4]),
-        os.path.join(strs[2]),
-        os.path.join(strs[5]),
+        "/".join([strs[0], strs[1], strs[2]]),
+        "/".join([strs[0], strs[1], strs[3]]),
+        "/".join([strs[0], strs[2]]),
+        "/".join([strs[0], strs[4]]),
+        "/".join([strs[2]]),
+        "/".join([strs[5]]),
     }
 
     for filepath in filepaths:
--- a/client/verta/tests/test_deployment.py
+++ b/client/verta/tests/test_deployment.py
@@ -293,8 +293,8 @@ class TestFetchArtifacts:
             retrieved_filepaths = set()
             for root, _, files in os.walk(dirpath):
                 for filename in files:
-                    filepath = os.path.join(root, filename)
-                    filepath = os.path.relpath(filepath, dirpath)
+                    filepath = "/".join([root, filename])
+                    filepath = os.path.relpath(filepath, dirpath).replace('\\', '/') 
                     retrieved_filepaths.add(filepath)
 
             assert filepaths == retrieved_filepaths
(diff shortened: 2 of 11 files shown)


mvartakAtVerta:
Lots of fixes for tests on Windows. Known issues:
1) test_versioning/test_dataset.py::TestPath needs revisiting for path comparisons on windows
2) test_creation_updates_dataset_timestamp shows failures on windows and unix
3) TestPathManagedVersioning.test_mngd_ver_folder shows failure on windows

client/verta/tests/conftest.py
```diff
@@ -294,12 +294,12 @@ def dir_and_files(strs, tmp_path, request):
...
-        os.path.join(strs[0], strs[2]),
-        os.path.join(strs[0], strs[4]),
-        os.path.join(strs[2]),
-        os.path.join(strs[5]),
+        "/".join([strs[0], strs[1], strs[2]]),
+        "/".join([strs[0], strs[1], strs[3]]),
+        "/".join([strs[0], strs[2]]),
+        "/".join([strs[0], strs[4]]),
+        "/".join([strs[2]]),
+        "/".join([strs[5]]),
```
convoliution, line 302: This seems odd to me—these are filepaths on the local machine, so I'd expect `os.path.join()` to be appropriate here.

client/verta/tests/conftest.py
```diff
@@ -514,9 +514,9 @@ def organization(client, created_entities):
 
 @pytest.fixture
 def requirements_file():
-    with tempfile.NamedTemporaryFile("w+") as tempf:
+    with tempfile.NamedTemporaryFile("w+", delete=False) as tempf:
```
convoliution, line 517: Suggestion:

- `delete=False` is 👌, though a bit odd to me
- line 526: move `yield tempf` out of the `with:` block (unindent one level)'
- after the `yield`, add an `os.remove(tempf.name)`

With this, we can take out some of the `.close()`s and `os.remove(tempf.name)`s added to tests in this PR

client/verta/tests/test_versioning/test_environment.py
```diff
@@ -60,7 +60,7 @@ def requirements_file_without_versions():
 
 @pytest.fixture
 def requirements_file_with_unsupported_lines():
-    with tempfile.NamedTemporaryFile("w+") as tempf:
+    with tempfile.NamedTemporaryFile("w+", delete=False) as tempf:
```
convoliution, line 63: See my suggestion on `requirements_file()` in conftest.py

client/verta/tests/test_versioning/test_environment.py
```diff
@@ -43,9 +43,9 @@ def metadata(draw):
 
 @pytest.fixture
 def requirements_file_without_versions():
-    with tempfile.NamedTemporaryFile("w+") as tempf:
+    with tempfile.NamedTemporaryFile("w+", delete=False) as tempf:
```
convoliution, line 46: See my suggestion on `requirements_file()` in conftest.py

client/verta/tests/test_deployment.py
```diff
@@ -321,16 +321,17 @@ def test_cached_fetch_zip(self, experiment_run, strs, dir_and_files):
...
+        
+        # make archive
+        with tarfile.open(tempf.name, 'w:gz') as tarf:
+            tarf.add(dirpath, "")
+        tempf.flush()  # flush object buffer
+        os.fsync(tempf.fileno())  # flush OS buffer
+        tempf.seek(0)
+        tempf.close()
+
+        experiment_run.log_artifact(key, tempf.name)
```
convoliution, line 334: I suspect that for all of these, we could leave the `with:` block intact but simply move the `log_artifact() call out of it (unindent that one line, one level)

client/verta/tests/test_versioning/test_dataset.py
```diff
@@ -332,7 +332,7 @@ def test_list_paths(self):
     def test_base_path(self, paths, base_path):
         filepaths = _file_utils.flatten_file_trees(paths)
         expected_paths = set(
-            os.path.relpath(path, base_path)
+            os.path.relpath(path, base_path).replace("\\", "/")
```
convoliution, line 335: I'm reading more about this, and it sounds bananas.

I'm pretty sure our client uses `relpath()` in a few places, and I'm concerned that this is a wider underlying problem.

convoliution (changes requested):
I need to do some more reading on the Windows's specific changes here. In broad strokes, for this initial review:

1. I'm not personally familiar with `pip list --format freeze`! Is it functionally different from `pip freeze`? I'm going to spend some time after this initial review to look into it.
2. The `.replace("\\", "/")` in tests might not be necessary, if the `dir_and_files()`is able to stick with` os.path.join()`.
3. The use of `delete=False` may lead to leftover files on the system, if the tempfiles aren't cleaned up. This may need a broader solution.

liuverta:
I remember having difficulty picking up this PR when it was opened, and now with the numerous merge conflicts I fear it is no longer in a good state.

But make no mistake, Windows: I'm not afraid of you. One day...

PR #2440, opened 2021-06-24
Add tests and bugfixes for profilers

convoliution:
## Changes
- add Hypothesis strategies for generating `pd.Series` and `pd.DataFrame`s
- add property-based tests for `ContinuousHistogramProfiler`, `BinaryHistogramProfiler,` and MissingValuesProfiler`
- fix bugs in how `ContinuousHistogramProfiler` handles and returns data
  - use `dropna()` to prevent missing values from breaking `np.histogram()`
  - use `tolist()` to more safely convert NumPy array to a `list` of Python builtins

## Tests
```bash
$ pytest monitoring/test_profilers.py::TestProfilers
plugins: hypothesis-6.14.0
collected 3 items                                                                                                           

monitoring/test_profilers.py ...                                                                                      [100%]

============================================== 3 passed, 5 warnings in 51.87s ===============================================
```

### DIFF
--- a/client/verta/verta/monitoring/profiler.py
+++ b/client/verta/verta/monitoring/profiler.py
@@ -204,9 +204,9 @@ class ContinuousHistogramProfiler(Profiler):
             bins = self._bins[column]
         else:
             bins = self._bins
-        values, limits = self._np.histogram(df[column], bins=bins)
-        values = [v.item() for v in values]
-        limits = [lim.item() for lim in limits]
+        values, limits = self._np.histogram(df[column].dropna(), bins=bins)
+        values = values.tolist()
+        limits = limits.tolist()
         return (column + "_histogram", FloatHistogram(limits, values))
 
     # TODO: consider the case where the data is outside of the bucket list
--- a/client/verta/verta/monitoring/profiler.py
+++ b/client/verta/verta/monitoring/profiler.py
@@ -165,7 +165,7 @@ class BinaryHistogramProfiler(Profiler):
         dict
             A dictionary from summary data names to DiscreteHistogram instances.
         """
-        content = df[column].value_counts()
+        content = df[column].value_counts(dropna=True)
         keys = list(content.keys())
         values = [content[k] for k in keys]
         values = [v.item() for v in values]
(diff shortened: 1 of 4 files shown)


client/verta/verta/monitoring/profiler.py
```diff
@@ -204,9 +204,9 @@ def profile_column(self, df, column):
             bins = self._bins[column]
         else:
             bins = self._bins
-        values, limits = self._np.histogram(df[column], bins=bins)
-        values = [v.item() for v in values]
-        limits = [lim.item() for lim in limits]
+        values, limits = self._np.histogram(df[column].dropna(), bins=bins)
```
convoliution, line 207: `dropna()` omits missing values to prevent`np.histogram()` from failing due to NaNs, and matches the behavior of `BinaryHistogramProfiler` (which also omits missing values when profiling).

client/verta/verta/monitoring/profiler.py
```diff
@@ -165,7 +165,7 @@ def profile_column(self, df, column):
         dict
             A dictionary from summary data names to DiscreteHistogram instances.
         """
-        content = df[column].value_counts()
+        content = df[column].value_counts(dropna=True)
```
convoliution, line 168: `True` is [already the default](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.value_counts.html), but this makes what's happening more explicit.

client/verta/verta/monitoring/profiler.py
```diff
@@ -204,9 +204,9 @@ def profile_column(self, df, column):
             bins = self._bins[column]
         else:
             bins = self._bins
-        values, limits = self._np.histogram(df[column], bins=bins)
-        values = [v.item() for v in values]
-        limits = [lim.item() for lim in limits]
+        values, limits = self._np.histogram(df[column].dropna(), bins=bins)
+        values = values.tolist()
+        limits = limits.tolist()
```
convoliution, line 209: I was running into an issue where the `np.array` `limits` sometimes contained Python builtin `float`s rather than NumPy values (so `item()` fails).

As far as I can tell from [NumPy's docs and examples](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html), `values` and `limits` will always be 1-D arrays so it should be safe to use `array.tolist()` rather than `element.item()`.

daniel-verta (approved):

PR #2441, opened 2021-06-24
Enable adding+getting data types as RMV attrs

convoliution:
## Changes
- allow `RegisteredModelVersion`'s `add_attributes()` and `get_attributes()` to de/serialize `verta.data_type`s like `ExperimentRun`s can

## Tests
```bash
$ pytest test_model_registry/test_model_version.py::TestComplexAttributes test_model_registry/test_model_version.py::TestAutoMonitoring::test_profile_training_data
plugins: hypothesis-6.14.0
collected 4 items                                                                                                           

test_model_registry/test_model_version.py ....                                                                        [100%]

============================================== 4 passed, 9 warnings in 53.56s ===============================================
```

### DIFF
--- a/client/verta/tests/test_model_registry/test_model_version.py
+++ b/client/verta/tests/test_model_registry/test_model_version.py
@@ -23,6 +23,7 @@ import six
 import verta
 import verta.dataset
 from verta import visibility
+from verta import data_types
 from verta.data_types import _verta_data_type
 from verta.environment import Python
 from verta.tracking.entities._deployable_entity import _CACHE_DIR
@@ -33,6 +34,7 @@ from verta.monitoring import profiler
 from verta._internal_utils import (
     _artifact_utils,
     _utils,
+    importer,
 )
 from verta._protos.public.monitoring.DeploymentIntegration_pb2 import FeatureDataInModelVersion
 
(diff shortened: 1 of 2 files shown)


client/verta/tests/test_model_registry/test_model_version.py
```diff
@@ -858,6 +860,60 @@ def test_redact(self, client, client_2, organization, created_entities):
         admin_model_ver.delete()
 
 
+# TODO: combine with test_experimentrun/test_attributes.py::TestComplexAttributes
```
convoliution, line 863: This was almost exactly copied & pasted from the existing `ExperimentRun` tests. I really need to figure out how to parametrize tests that are common between `ExperimentRun` and `ModelVersion`

daniel-verta: Would it be appropriate to move `get_attribute` onto `DeployableEntity` and have this test exist in a TestDeployableEntity` class? Then you can parameterize with both types of entities to run the same test for both an experiment run and a model version. See [tests/monitoring/alerts/test_entities.py#L214](https://github.com/VertaAI/modeldb/blob/ae30afeaea972f8e56de74450bcd656159da0c7a/client/verta/tests/monitoring/alerts/test_entities.py#L214) for an example.

daniel-verta: (And if the implementations need to diverge between the two classes, we can introduce `get_attribute` as an abstract method on DeployableEntity of course...)

daniel-verta (approved):
Approved with a possible path towards test consolidation

convoliution:
> Would it be appropriate to move `get_attribute` onto `DeployableEntity` and have this test exist in a TestDeployableEntity` class?

@daniel-verta What I would actually like to do is put `add_attribute()` and `get_attribute()` abstract methods on the base class (the implementations of both already diverge, and matching them isn't an easy task).

For parametrization, it's a bit trickier because `ExperimentRun` and `ModelVersion` have setups and teardowns involved—I can't directly instantiate the classes in `pytest.mark.parametrize()`, rather they should come from a fixture. And pytest doesn't have a way to pass fixtures to `parametrize()`.

I've looked up solutions before though, and I think I've got an idea for it:
```python
@pytest.fixture(params=[ExperimentRun, ModelVersion])
def deployable_entity(cls):
    if cls == ExperimentRun:
        ...
    elif ...
```
maybe with `singledispatch` too, if it works. But this would probably be something for after this month's development push.

PR #2443, opened 2021-06-24
Use assume() instead of a while loop in composite strategy

convoliution:
I recently learned about `hypothesis.assume()` for fine-tuning desired examples, which seems more idiomatic than a `while` loop.

### DIFF
--- a/client/verta/tests/monitoring/comparison/test_comparison.py
+++ b/client/verta/tests/monitoring/comparison/test_comparison.py
@@ -21,8 +21,7 @@ from verta.monitoring.comparison import (
 @st.composite
 def lower_and_higher(draw):
     lower = draw(st.floats(allow_nan=False, allow_infinity=False))
-    higher = draw(st.floats(min_value=lower, exclude_min=True, allow_nan=False))
-    hypothesis.assume(not _VertaComparison.isclose(lower, higher))
+    higher = draw(st.floats(min_value=lower+DEFAULT_TOLERANCE, exclude_min=True, allow_nan=False))
 
     return lower, higher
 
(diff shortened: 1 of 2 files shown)


client/verta/tests/monitoring/comparison/test_comparison.py
```diff
@@ -21,10 +21,8 @@
 @st.composite
 def lower_and_higher(draw):
     lower = draw(st.floats(allow_nan=False, allow_infinity=False))
-
-    higher = lower
-    while _VertaComparison.isclose(lower, higher):
-        higher = draw(st.floats(min_value=lower, exclude_min=True, allow_nan=False))
+    higher = draw(st.floats(min_value=lower, exclude_min=True, allow_nan=False))
+    hypothesis.assume(not _VertaComparison.isclose(lower, higher))
```
daniel-verta, line 25:
```suggestion
    higher = draw(st.floats(min_value=lower+DEFAULT_TOLERANCE, exclude_min=True, allow_nan=False))
```
(With the addition of defining such a constant in `_verta_comparison`, and then importing it here. It looks to be `1e-9`.)

It's super neat to learn about hypothesis.assume, but in this case we can also just make the minimum of our draw range the lower value plus the tolerance.

daniel-verta (approved):

client/verta/tests/monitoring/comparison/test_comparison.py
convoliution: @daniel-verta I've realized that this isn't actually this simple—this tolerance value is a _relative_ tolerance, meaning `lower+DEFAULT_TOLERANCE` isn't accurate. I think I'd need to solve `abs(a-b) <= rel_tol * max(abs(a), abs(b))` for `b`, which I'm willing to do, but it's probably not going to be easy on the eyes for anyone maintaining this strategy in the future.

In light of this, what would you think about sticking with `assume()`?

PR #2444, opened 2021-06-28
[VR-7717] Propagate usable information when a model being deployed fails at runtime

convoliution:
The deployment API's `/endpoints/{}/stages/{}` pathway doesn't consistently populate its response with an error message when model deployment fails at runtime (VR-7740), so we often just get
```python
RuntimeError: endpoint update failed;
Error in model container: 
```

Until that's remedied, we'll instead directly fetch the container logs for the exception message.

## Changes
- add `endpoint.get_logs()`
- use ☝️ runtime logs when raising an exception for a failed deployment

### DIFF
--- a/client/verta/verta/endpoint/_endpoint.py
+++ b/client/verta/verta/endpoint/_endpoint.py
@@ -403,6 +403,35 @@ class Endpoint(object):
 
         return response_json
 
+    def get_logs(self):
+        """Get runtime logs of this endpoint.
+
+        Returns
+        -------
+        list of str
+            Lines of this endpoint's runtime logs.
+
+        """
+        url = "{}://{}/api/v1/deployment/workspace/{}/endpoints/{}/stages/{}/logs".format(
+            self._conn.scheme,
+            self._conn.socket,
+            self.workspace,
+            self.id,
+            self._get_or_create_stage()
+        )
+        response = _utils.make_request("GET", url, self._conn)
+        self._conn.must_response(response)
+
+        logs = response.json().get("entries", [])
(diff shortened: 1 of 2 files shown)


client/verta/verta/endpoint/_endpoint.py
```diff
@@ -403,6 +405,35 @@ def get_status(self):
 
         return response_json
 
+    def get_logs(self):
```
daniel-verta, line 408: @convoliution how big do these logs get? Is there a chance we're going to pull down a 1000 line error and put it into a `RuntimeError`? Also, dare I ask what happens when `get_logs` hits an error?

convoliution: > how big do these logs get? Is there a chance we're going to pull down a 1000 line error and put it into a RuntimeError?

In theory, there's no upper bound on their size, since it's whatever logs are coming out of the model container.

In practice, in the event of a deployment failure, it's 5 lines of Flask startup logs + any startup logs for external libraries (TensorFlow tends to have a lot) + the offending exception's stack trace. So 10–30 lines perhaps?

The previous `status_dict['components'][0].get('message', "no error message available")` relied on logic in the deployment backend to extract just the stack trace, but it's currently broken (hence this PR) and I'd prefer to not reinvent it here 😅 

> Also, dare I ask what happens when get_logs hits an error?

I suppose the user would see a rather unhelpful stack trace coming out of `Endpoint.get_logs()`. I'm not sure how many layers of robust exception handling we'd like.

daniel-verta (approved):

PR #2478, opened 2021-07-12
Only modify created_at when fetching alert's query

convoliution:
Previously, if a user doesn't specify `alerts.create(starting_from=)`, then the alert will basically only consider new samples with time windows set in the present.

This change has the alert consider _all_ newly-created samples. Even if their time windows are set in the past, as long as they were created since the alert was last evaluated then they will be counted (barring race conditions).

At a more technical level: only `created_at_after_millis` will be modified when fetching an alert's sample query.


convoliution:
Drafted while I test this change.

daniel-verta (approved):

PR #2484, opened 2021-07-13
Fix Python 2 compatibility bugs in log_training_data_profile()

convoliution:
These had gone uncaught because of an XGBoost segfault in our test suite (VR-12149 to resolve) that prevented the tests from being run in Python 2.

### DIFF
--- a/client/verta/verta/registry/entities/_modelversion.py
+++ b/client/verta/verta/registry/entities/_modelversion.py
@@ -1347,8 +1347,9 @@ class RegisteredModelVersion(_deployable_entity._DeployableEntity):
                 json.loads(feature_data.content),
             )
 
-        with ThreadPool(1) as p:
-            p.map(lambda args: log_attribute(*args), enumerate(feature_data_list))
+        p = ThreadPool(1)
+        p.map(lambda args: log_attribute(*args), enumerate(feature_data_list))
+        p.close()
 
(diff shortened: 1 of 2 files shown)


client/verta/verta/tracking/entities/_entity.py
```diff
@@ -106,7 +106,7 @@ def _normalize_attribute_key(key):
             Sanitized `key`
 
         """
-        return "".join([x for x in key if (str.isalnum(x) or x == '_')])
+        return "".join([x for x in key if (x.isalnum() or x == '_')])
```
convoliution, line 109: The old code failed if `x` is `unicode` rather than `str`

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1347,8 +1347,9 @@ def log_attribute(i, feature_data):
                 json.loads(feature_data.content),
             )
 
-        with ThreadPool(1) as p:
-            p.map(lambda args: log_attribute(*args), enumerate(feature_data_list))
+        p = ThreadPool(1)
+        p.map(lambda args: log_attribute(*args), enumerate(feature_data_list))
+        p.close()
```
convoliution, line 1352: pools don't implement the context manager interface in Python 2

daniel-verta (approved):

PR #2488, opened 2021-07-14
[VR-11633] Enable Kafka configuration on endpoints

convoliution:
Enables:
- setting Kafka on endpoint creation
- modifying Kafka on endpoint update
- removing Kafka on endpoint update

Does not support disabling Kafka while keeping the stream values (they are deleted entirely in the third bullet point ☝🏼).

Screen Shot 2021-07-14 at 4 46 09 PM

## Tests
```bash
$ pytest test_endpoint/test_kafka.py
plugins: hypothesis-6.14.0
collected 4 items                                                                                                           

test_endpoint/test_kafka.py ....                                                                                      [100%]

============================================== 4 passed, 4 warnings in 58.09s ===============================================
```

### DIFF
--- a/client/verta/verta/endpoint/_endpoint.py
+++ b/client/verta/verta/endpoint/_endpoint.py
@@ -363,15 +363,32 @@ class Endpoint(object):
         return response.json()["id"]
 
     def _set_kafka_settings(self, stage_id, kafka_settings):
+        if not isinstance(stage_id, six.integer_types):
+            raise TypeError("`stage_id` must be int, not {}".format(type(stage_id)))
+        if not isinstance(kafka_settings, KafkaSettings):
+            raise TypeError(
+                "`kafka_settings` must be of type verta.endpoint.KafkaSettings,"
+                " not {}".format(type(kafka_settings))
+            )
+
         url = "{}://{}/api/v1/deployment/stages/{}/kafka".format(
-            self._conn.scheme, self._conn.socket, stage_id,
+            self._conn.scheme,
+            self._conn.socket,
+            stage_id,
         )
-        response = _utils.make_request("PUT", url, self._conn, json=kafka_settings._as_dict())
+        response = _utils.make_request(
+            "PUT", url, self._conn, json=kafka_settings._as_dict()
+        )
         _utils.raise_for_http_error(response)
 
     def _get_kafka_settings(self, stage_id):
+        if not isinstance(stage_id, six.integer_types):
+            raise TypeError("`stage_id` must be int, not {}".format(type(stage_id)))
+
         url = "{}://{}/api/v1/deployment/stages/{}/kafka".format(
-            self._conn.scheme, self._conn.socket, stage_id,
+            self._conn.scheme,
+            self._conn.socket,
+            stage_id,
         )
         response = _utils.make_request("GET", url, self._conn)
         _utils.raise_for_http_error(response)
--- a/client/verta/verta/endpoint/_endpoint.py
+++ b/client/verta/verta/endpoint/_endpoint.py
@@ -8,22 +8,24 @@ import yaml
 
 import requests
 
-from ..external import six
+from verta.external import six
 
-from ..endpoint.autoscaling import Autoscaling
-from ..endpoint.autoscaling.metrics import _AutoscalingMetric
-from ..endpoint.resources import Resources
-from ..endpoint.update.rules import _UpdateRule
-from ..deployment import DeployedModel
-from ..endpoint.update._strategies import (
+from verta.deployment import DeployedModel
+from verta._internal_utils import _utils
+from verta.tracking.entities import ExperimentRun
+from verta.registry.entities import RegisteredModelVersion
+from verta.visibility import _visibility
+
+from . import KafkaSettings
+from .autoscaling import Autoscaling
+from .autoscaling.metrics import _AutoscalingMetric
+from .resources import Resources
+from .update.rules import _UpdateRule
+from .update._strategies import (
     _UpdateStrategy,
     DirectUpdateStrategy,
     CanaryUpdateStrategy,
 )
-from .._internal_utils import _utils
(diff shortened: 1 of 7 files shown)


client/verta/verta/endpoint/_endpoint.py
```diff
@@ -295,15 +379,63 @@ def _get_or_create_stage(self, name="production"):
...
+            self._conn.scheme,
+            self._conn.socket,
+            stage_id,
+        )
+        response = _utils.make_request(
+            "PUT", url, self._conn, json={"disabled": True},
+        )
+        _utils.raise_for_http_error(response)
+
+    # TODO: add support for KafkaSettings to _update_from_dict below or de-scope
```
convoliution: @daniel-verta Was this required as a part of the ticket?

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -295,15 +379,63 @@ def _get_or_create_stage(self, name="production"):
...
+
+        url = "{}://{}/api/v1/deployment/stages/{}/kafka".format(
+            self._conn.scheme,
+            self._conn.socket,
+            stage_id,
+        )
+        response = _utils.make_request(
+            "PUT", url, self._conn, json={"disabled": True},
+        )
+        _utils.raise_for_http_error(response)
```
convoliution, line 435: The main methods for Kafka setting manipulation

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -186,22 +253,35 @@ def update(self, model_reference, strategy=None, wait=False, resources=None,
...
-        update_body = self._create_update_body(strategy, resources, autoscaling, env_vars)
+        update_body = self._create_update_body(
+            strategy, resources, autoscaling, env_vars
+        )
+
+        # update Kafka settings
+        if kafka_settings is False:
+            self._del_kafka_settings(self._get_or_create_stage())
+        elif kafka_settings:
+            self._set_kafka_settings(self._get_or_create_stage(), kafka_settings)
```
convoliution, line 280: Here is where Kafka settings are modified during `endpoint.update()`.

This notably makes `endpoint.update()` less atomic, because updating the Kafka settings of the stage is separate from creating a build.

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -59,54 +68,99 @@ def __repr__(self):
...
+        return "\n".join(
+            (
+                "path: {}".format(data["creator_request"]["path"]),
+                "url: {}://{}/{}/endpoints/{}/summary".format(
+                    self._conn.scheme, self._conn.socket, self.workspace, self.id
+                ),
+                "id: {}".format(self.id),
+                "curl: {}".format(curl),
+                "status: {}".format(status["status"]),
+                "Kafka settings: {}".format(self.kafka_settings),
```
convoliution, line 80:
```
Kafka settings: None
```
or
```
Kafka settings: KafkaSettings('allergy', 'battery', 'cannery')
```

client/verta/verta/endpoint/_kafka_settings.py
```diff
@@ -0,0 +1,82 @@
...
+    @classmethod
+    def _from_dict(cls, d):
+        # NOTE: ignores extraneous keys in `d`
+        try:
+            input_topic = d["input_topic"]
+            output_topic = d["output_topic"]
+            error_topic = d["error_topic"]
+        except KeyError as e:
+            msg = 'expected but did not find key "{}"'.format(e.args[0])
+            six.raise_from(RuntimeError(msg), None)
```
convoliution, line 90: All three of these keys are required, right @daniel-verta?

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -59,54 +68,99 @@ def __repr__(self):
...
+        )
         if endpoint_json:
-            endpoint = cls(conn, conf, workspace, endpoint_json['id'])
-            endpoint._create_stage()  # an endpoint needs a stage to function, so might as well create it here
+            endpoint = cls(conn, conf, workspace, endpoint_json["id"])
+
+            # an endpoint needs a stage to function, so might as well create it here
+            stage_id = endpoint._create_stage()
+            if kafka_settings:
+                endpoint._set_kafka_settings(stage_id, kafka_settings)
```
convoliution, line 125: Here is where Kafka settings are set during `client.create_endpoint()`.

Notably not atomic, because our deployment API doesn't expose setting Kafka values during stage creation.

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -186,22 +252,35 @@ def update(self, model_reference, strategy=None, wait=False, resources=None,
             Autoscaling condition for the updated endpoint.
         env_vars : dict of str to str, optional
             Environment variables.
+        kafka_settings : :class:`verta.endpoint.KafkaSettings` or False, optional
+            Kafka settings. If ``False``, clears this endpoint's existing Kafka
+            settings.
```
convoliution, line 257: The acceptance criteria mentions using ``None`` to deactivate the Kafka connection, but I personally find that to run against an implicit Python convention where ``None`` should mean "no parameter provided, do nothing", and would make it difficult for a user to say "update the model, but don't change or delete the current Kafka settings".

daniel-verta: I wish there was something more elegant, but I concur and like this as a solution.

client/verta/verta/endpoint/_endpoint.py
daniel-verta: It was not. I realized that there was this pathway in the API for updating endpoints and left this as a note to explicitly not support updating the Kafka settings with this method, or to add such support. I think we should add a ticket tracking this, leave the reference here, and move on for the time being.

daniel-verta (approved):

PR #2494, opened 2021-07-16
[VR-11871] Fix filter ER by hyperparameter predicates

anandjakhaniya:
Jira Ticket: https://vertaai.atlassian.net/browse/VR-11871

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/HyperparameterPredicatesHandler.java
+++ b/backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/HyperparameterPredicatesHandler.java
@@ -5,6 +5,8 @@ import ai.verta.common.OperatorEnum;
 import ai.verta.modeldb.common.futures.InternalFuture;
 import ai.verta.modeldb.common.query.QueryFilterContext;
 import ai.verta.modeldb.exceptions.UnimplementedException;
+import ai.verta.modeldb.utils.ModelDBUtils;
+import java.util.Date;
 
 public class HyperparameterPredicatesHandler extends PredicateHandlerUtils {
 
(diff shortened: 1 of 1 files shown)


conradoverta (commented):
I don't follow what was the issue here and the ticket is empty. Could you provide a description?

anandjakhaniya:
> I don't follow what was the issue here and the ticket is empty. Could you provide a description?

there was an issue while sending empty keys while deleting hyperparameters, metrics etc. and we have implemented Optional there and implemented `Optional.isPresent()` but due to empty list it was giving a true but actual it has empty list so we should have to add empty check for that optional value which i have added now.

conradoverta:
> > I don't follow what was the issue here and the ticket is empty. Could you provide a description?
> 
> there was an issue while sending empty keys while deleting hyperparameters, metrics etc. and we have implemented Optional there and implemented `Optional.isPresent()` but due to empty list it was giving a true but actual it has empty list so we should have to add empty check for that optional value which i have added now.

I think you're taking about the wrong ticket, as this is a filtering one, not a deletion one.

anandjakhaniya:
> > > I don't follow what was the issue here and the ticket is empty. Could you provide a description?
> > 
> > 
> > there was an issue while sending empty keys while deleting hyperparameters, metrics etc. and we have implemented Optional there and implemented `Optional.isPresent()` but due to empty list it was giving a true but actual it has empty list so we should have to add empty check for that optional value which i have added now.
> 
> I think you're taking about the wrong ticket, as this is a filtering one, not a deletion one.

Ohh Ok, I am sorry. it's my bad. So here in database we have store value as a string using `ModelDBUtils.getStringFromProtoObject(value)` while inserting so in database it will keep data with `""` so without `ModelDBUtils.getStringFromProtoObject(value)` , all match are failed so fix logic to used `ModelDBUtils.getStringFromProtoObject(value)` for given predicate value

conradoverta (approved):

PR #2499, opened 2021-07-19
Introduce common pom.xml for build common module separately

anandjakhaniya:


evgeniilezhnin (approved):
works for autoformat. But to reuse it in root pom.xml dependency to that module should be included

coryverta (approved):

PR #2507, opened 2021-07-22
Mention that std lib should not be included in custom_modules

convoliution:
Python can determine when an object is coming from the standard library (assuming there was an `import` statement either in the original script or within the model itself) without needing to include it in `custom_modules`.


bts-verta (approved):
LGTM

PR #2525, opened 2021-08-02
[VR-12349] Support pagination for alerts.list()

convoliution:
To facilitate not completely overwhelming the backend during the alerter's `/findAlert` request, we're going to paginate it!

The client has a base class `LazyList` that supports paginated iteration for `/find*` requests (which it uses for `Project.expt_runs`, `RegisteredModel.version`, and so on). Here, I'm splitting `LazyList`'s paginated iteration interface into a `_PaginatedIterable` subclass (leaving the more complex and slightly outdated `find()` querying in `_LazyList`).

## Changes

- move `verta._internal_utils._utils.LazyList` to `verta._bases._LazyList`
- split `_LazyList`'s `__getitem__()` + `__iter__()` + `__len__()` interface to `_PaginatedIterable`
- return `AlertsPaginatedIterable` from `Alerts.list()`


client/verta/verta/monitoring/alert/entities/_alert.py
```diff
@@ -3,6 +3,8 @@
 import warnings
```
convoliution, line 3: This file is where you'll find the actual non-refactoring part of this PR.

daniel-verta (approved):

PR #2542, opened 2021-08-06
Switch executors to use the Fork-Join Pool instead of fixed thread pool

coryverta:


conradoverta (commented):
Documentation says https://github.com/VertaAI/modeldb/compare/cm/ForkJoinPool?expand=1 is a better approach

PR #2562, opened 2021-08-17
[VR-12836] upload_completed added in artifact entity

anandjakhaniya:
Jira Ticket: https://vertaai.atlassian.net/browse/VR-12836


conradoverta (commented):
Don't we have to set this flag at some point?

conradoverta (commented):
More specifically, only when committed. The issue is exactly that the artifact might not be complete when we just log it. Example:
1. log
2. try to push to aws and fail

conradoverta (changes requested):
We don't seem to be setting `upload_completed` during the final commit call, right?

anandjakhaniya:
> We don't seem to be setting `upload_completed` during the final commit call, right?

No, We should have to set it. In current implementation, we are setting  `upload_completed` during `commitMultipartArtifact` https://github.com/VertaAI/modeldb/blob/15017c59c907214e128a6d33f79f0285ef1a8320/backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java#L2647

and for new implementation at here https://github.com/VertaAI/modeldb/blob/15017c59c907214e128a6d33f79f0285ef1a8320/backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/ArtifactHandler.java#L297

conradoverta (approved):

conradoverta (changes requested):
Every pre-existing artifact should have the flag set to `true` as it was already uploaded before. Only new artifacts can have `upload_completed=false`.

backend/src/main/resources/liquibase/db-changelog-2.0.xml
```diff
@@ -979,4 +979,24 @@
...
+        </preConditions>
+        <update tableName="artifact">
+            <column name="upload_completed" valueBoolean="true" />
+            <where>field_type="artifacts" AND upload_completed = 0</where>
+        </update>
+        <rollback>
+            <update tableName="artifact">
+                <column name="upload_completed" valueBoolean="false" />
+                <where>field_type="artifacts" AND upload_completed = 1</where>
+            </update>
```
conradoverta: This rollback is not valid. This is an information losing change, so we can't rollback from it. Let's remove to make that clearer.

conradoverta (approved):
One change. LGTM otherwise

backend/src/main/resources/liquibase/db-changelog-2.0.xml
anandjakhaniya: Ok, got it. removed rollback tag now.

PR #2570, opened 2021-08-23
Validate VertaModelBase.__init__() param names

convoliution:
## Problem

`VertaModelBase` has the following interface [defined for `__init__()`](https://github.com/VertaAI/modeldb/blob/160dbb3/client/verta/verta/registry/_verta_model_base.py#L54):

```python
class VertaModelBase(object):
    @abc.abstractmethod
    def __init__(self, artifacts):
        raise NotImplementedError
```

Our model service [passes the `artifacts` argument by keyword](https://github.com/VertaAI/services/blob/2dc4d7403b9750642fda96efd7314eb9898c0665/deployment/python-model-service/app/wrappers/model/class_model.py#L23), meaning that a subclass _must_ have `artifacts` as a parameter on its `__init__()` method to work with our deployment system today.

However, oftentimes out of forgetfulness ~or laziness~, I personally accidentally omit the `artifacts` parameter when quickly defining a model for testing purposes:

```python
class Model(VertaModelBase):
    def __init__(self):
        pass
```

only to find out about 10 minutes later that the deployment failed, and have to start all over again.

## Changes

To facilitate catching the problem sooner (and not require waiting for artifacts to upload + the endpoint to deploy before realizing it):

- validate at logging-time that a model subclassing VertaModelBase has the correct parameters on `__init__()`

## Tests

### Python 3

```bash
$ pytest registry/test_standard_model.py                                                                   
plugins: hypothesis-6.14.0
collected 72 items                                                                                                          

registry/test_standard_model.py ..............s................s................s.................s.....              [100%]

================================== 68 passed, 4 skipped, 38 warnings in 1213.41s (0:20:13) ==================================
```

### Python 2

```bash
$ pytest registry/test_standard_model.py                                                                   
platform darwin -- Python 2.7.18, pytest-4.6.11, py-1.10.0, pluggy-0.13.1                                                    
rootdir: /Users/miliu/Documents/modeldb/client/verta/tests, inifile: pytest.ini                                              
plugins: hypothesis-4.57.1                                                                                                   
collected 72 items                                                                                                           
                                                                                                                             
registry/test_standard_model.py ..............s................s................s.................s.....              [100%] 
                                                                                                                             
=================================== 68 passed, 4 skipped, 38 warnings in 1304.61 seconds ====================================
```

## Alternatives Considered

Aside from linting on the user's end, I'm not aware of any Python mechanisms for validating that a subclass matches the interfaces of its superclass's methods at definition-time.

There is `__init_subclass__()` in Python 3.6+ which could let us bundle the code for this check into the `VertaModelBase` abstract class, but we still need to support 2.7 and 3.5, and the check wouldn't run until the class is instantiated which isn't something we can necessarily do when we log the model.

### DIFF
--- a/client/verta/tests/models/standard_models.py
+++ b/client/verta/tests/models/standard_models.py
@@ -61,11 +61,37 @@ def decorated_verta_models():
 def incomplete_verta_models():
     models = []
 
-    models.extend([
-        VertaModelNoImpl,
-        VertaModelOnlyInit,
-        VertaModelOnlyPredict,
-    ])
+    models.extend(
+        [
+            VertaModelNoImpl,
+            VertaModelOnlyInit,
+            VertaModelOnlyPredict,
+        ]
+    )
+
+    return models
+
+
+def bad_init_verta_models():
+    models = []
+
+    class InsufficientInit(VertaModelBase):
+        def __init__(self):  # no `artifacts`
+            pass
+
+        def predict(self, input):
+            pass
+
+    models.append(InsufficientInit)
+
+    class MisnamedInit(VertaModelBase):
+        def __init__(self, relics):  # not `artifacts`
+            pass
+
+        def predict(self, input):
+            pass
+
+    models.append(MisnamedInit)
 
     return models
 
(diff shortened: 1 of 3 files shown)


conradoverta (approved):
Nice catch

PR #2572, opened 2021-08-24
Allow tests to force run reconciler

ghost:


ghost:
Don't actually need this.. calling resync makes the workers start.

PR #2577, opened 2021-08-27
Fix code smells suggested by sonarqube part - 3

anandjakhaniya:
Main changes in this PR are removed `InvalidProtocolBufferException` exceptions from all places


anandjakhaniya:
it has same changes everywhere and i think not need to split this in second PR so i hope it will works.. :)

conradoverta (approved):

PR #2584, opened 2021-08-31
[VR-12511] Fall back to logging raw requirements if manual parsing fails

convoliution:
## Impact and Context

Previously, the client's `Python` utility skipped harder-to-parse configurations in requirements files. This PR changes the client to store the raw requirements/constraints if it encounters an error during parsing.

The error is presented to the user as a warning via the logger, though the object is still created (with the raw contents):

Screen Shot 2021-08-31 at 3 36 44 PM

## Risks

- This PR involves a sizeable refactor to the client's internal utilities for capturing/parsing pip items, which carries some risk of side effects. A best effort at test coverage has been made.
- Because previously-skipped lines are now being captured by the client, outdated deployment backends (i.e. images existing prior to this effort) will be unable to process them. This itself wouldn't cause errors (they would be ignored), but it may be unexpected for a user.
- `pin_verta_and_cloudpickle()` use the running environment's `verta` and `cloudpickle` versions as a proxy for the versions "when the model was serialized", but that's not necessarily the case. This issue was already existent before this PR, however.

The following **backwards incompatibilities** must be noted in the changelog:

- `ExperimentRun.log_requirements()` has been removed (deprecated since `0.18.0` in June, superseded by `log_environment()`)
- `ExperimentRun.log_model_for_deployment()` has been removed (deprecated since `0.13.13` in November 2019, superseded by `log_model()`)
- `Python.read_pip_file()` and `Python.read_pip_environment()` now include "unsupported" lines rather than skipping them (the new `skip_options` parameter can be toggled to restore the old behavior)

## Testing

  - no new service
  - this change is client-only
- [x] Added unit test(s)
  - `pytest test_utils/test_pip_requirements.py test_versioning/test_environment.py` passes in Python 2 and 3 on my machine
  - testing is enhanced in https://github.com/VertaAI/modeldb/pull/2586
- [ ] ~Added integration test(s)~
  - the remainder of the stack has not yet implemented support for this functionality

## How to Revert

1. Revert this PR


client/verta/tests/test_versioning/test_environment.py
```diff
@@ -22,25 +17,6 @@
...
-
-    return ".".join(map(str, [major, minor, patch]))
-
-
-@st.composite
-def metadata(draw):
-    """The "cu102" in "torch==1.8.1+cu102"."""
-    # https://www.python.org/dev/peps/pep-0440/#local-version-identifiers
-    alphabet = string.ascii_letters + string.digits + ".-_"
-    return draw(st.text(alphabet=alphabet, min_size=1))
```
convoliution, line 41: Moved verbatim to test_utils/test_pip_requirements.py

client/verta/tests/test_versioning/test_environment.py
```diff
@@ -84,43 +60,6 @@ def requirements_file_with_unsupported_lines():
...
-        metadata=metadata(),
-    )
-    def test_remove_torch_metadata(self, version, metadata):
-        clean_requirement = "torch=={}".format(version)
-        requirement = "+".join([clean_requirement, metadata])
-
-        requirements = [requirement]
-        _pip_requirements_utils.remove_public_version_identifier(requirements)
-        assert requirements != [requirement]
-        assert requirements == [clean_requirement]
```
convoliution, line 121: Moved verbatim to test_utils/test_pip_requirements.py

client/verta/tests/test_utils/test_pip_requirements.py
```diff
@@ -1,14 +1,77 @@
...
+        metadata=metadata(),
+    )
+    def test_remove_torch_metadata(self, version, metadata):
+        clean_requirement = "torch=={}".format(version)
+        requirement = "+".join([clean_requirement, metadata])
+
+        requirements = [requirement]
+        _pip_requirements_utils.remove_public_version_identifier(requirements)
+        assert requirements != [requirement]
+        assert requirements == [clean_requirement]
```
convoliution, line 76: Moved verbatim from test_utils/test_pip_requirements.py

client/verta/tests/test_utils/test_pip_requirements.py
```diff
@@ -1,14 +1,77 @@
...
+
+    return ".".join(map(str, [major, minor, patch]))
+
+
+@st.composite
+def metadata(draw):
+    """The "cu102" in "torch==1.8.1+cu102"."""
+    # https://www.python.org/dev/peps/pep-0440/#local-version-identifiers
+    alphabet = string.ascii_letters + string.digits + ".-_"
+    return draw(st.text(alphabet=alphabet, min_size=1))
```
convoliution, line 37: Moved verbatim from test_utils/test_pip_requirements.py

client/verta/tests/test_versioning/test_environment.py
```diff
@@ -169,14 +114,53 @@ def test_inject_verta_cloudpickle(self):
         assert "verta" in requirements
         assert "cloudpickle" in requirements
 
+    def test_raw_inject_verrta_cloudpickle(self):
```
conradoverta:
```suggestion
    def test_raw_inject_verta_cloudpickle(self):
```

client/verta/verta/_internal_utils/_pip_requirements_utils.py
```diff
@@ -267,34 +277,100 @@ def add_verta_and_cloudpickle(requirements):
...
-            their_ver = req.split('==')[-1]
-            if our_ver != their_ver:  # versions conflict, so raise exception
-                raise ValueError("Client is running with cloudpickle v{}, but the provided requirements specify v{};"
-                                 " these must match".format(our_ver, their_ver))
-            else:  # versions match, so proceed
-                break
-    else:  # if not present, add
-        requirements.append(cloudpickle_req)
+    for i, req in enumerate(requirements):
+        if req.startswith("-e git+git@github.com:VertaAI/modeldb.git"):
```
conradoverta, line 281: Do we need to check for the ssh style too?

conradoverta (approved):
LGTM. For future reference, this could probably have been a few PRs while allows faster merge (1. refactor of moving things around directly, 2. the pure deletes, 3. actually handling raw; at least).

conradoverta (commented):
One case that I don't believe I saw was when the requirement has multiple lines for a single item. Do we handle that correctly when reading the requirements file? If not, it could be a follow up. But I believe we make the strong assumption that we have 1 requirement per line when evaluating if we should inject verta/cloudpickle. Maybe it doesn't matter that we can have multiple lines, but just double checking.

client/verta/verta/_internal_utils/_pip_requirements_utils.py
convoliution: @conradoverta Hmm maybe?

My intention here was just to handle a particular case where `pip freeze` outputs something that `pip install` syntactically cannot handle. As far as I've seen, ssh style lines are syntactically valid, so I wouldn't think to check for it.

Is the concern around having the ssh keys in the model container?

convoliution:
> For future reference, this could probably have been a few PRs while allows faster merge

Yeah, I'd like to be more diligent about this! 😶 

> One case that I don't believe I saw . . .

@conradoverta Would you mind providing an example? From the wording, I'm not sure whether you're referring to "multiple lines for a single item" or "multiple requirements per line" (which I'm not sure is a thing, but wanted to confirm)

client/verta/verta/_internal_utils/_pip_requirements_utils.py
conradoverta: No, just wasn't exactly sure it supported ssh. LGTM then =)

conradoverta:
> > For future reference, this could probably have been a few PRs while allows faster merge
> 
> Yeah, I'd like to be more diligent about this! 😶
> 
> > One case that I don't believe I saw . . .
> 
> @conradoverta Would you mind providing an example? From the wording, I'm not sure whether you're referring to "multiple lines for a single item" or "multiple requirements per line" (which I'm not sure is a thing, but wanted to confirm)

```
~/workspace/cluster-setup (cm/encrypt-snapshots) [engine-dev/setup]$ pip install -r reqs.txt 
Ignoring pytest: markers 'python_version == "2.7"' don't match your environment
WARNING: You are using pip version 20.1.1; however, version 21.2.4 is available.
You should consider upgrading via the '/Users/conrado/.pyenv/versions/3.7.10/envs/client37/bin/python3.7 -m pip install --upgrade pip' command.
~/workspace/cluster-setup (cm/encrypt-snapshots) [engine-dev/setup]$ cat reqs.txt 
pytest; \
  python_version=='2.7'
```

Essentially pip interprets `\` correctly.

convoliution:
> Essentially pip interprets \ correctly.

o no I've missed the fact that requirements files support line continuations

convoliution:
> Essentially pip interprets \ correctly.

o no I've missed the fact that requirements files support line continuations

@conradoverta I'm doing some rough testing and it looks like injecting verta/cloudpickle can still work:

```python
reqs = r"""
verta; \
    python_version=='2.7'
""".strip().splitlines()
Python(requirements=reqs)
# Python Version
#     Python 3.7.10
#     raw requirements:
#         verta==0.19.2a0; \
#             python_version=='2.7'
#         cloudpickle==1.6.0
```

though not for others

```python
reqs = r"""
verta \
    == 0.18.0
""".strip().splitlines()
Python(requirements=reqs)
# Python Version
#     Python 3.7.10
#     raw requirements:
#         verta==0.19.2a0
#             == 0.18.0
#         cloudpickle==1.6.0
```

I'll file a follow-up ticket for this

PR #2602, opened 2021-09-09
Move pylint directive

convoliution:
## Impact and Context

Our code formatter `black` moved this `pylint: disable` directive to a line where it no longer has an effect. This PR moves it so it works again.

## Risks

None—this is just a localized change to linting behavior

## Testing

I watched the `no-value-for-parameter` linting error disappear from the `strategies.dataframes()` call.

## How to Revert

Revert this PR.


conradoverta (approved):
Linter fight!

PR #2604, opened 2021-09-09
Parametrize deployable entity to run with both ER and RMV

convoliution:
## Impact and Context
 #2603 consolidated tests that are conceptually applicable to all deployable entities into the `deployable_entity/` testing subdirectory. This PR actually allows those tests to run on both `ExperimentRun` and `RegisteredModelVersion` when invoked.

## Risks

Not a risk but as a side effect: Some tests fail because ER and RMV don't yet have full feature parity. 
See **Testing** for follow-ups.

## Changes

- add `deployable_entity` fixture so tests can be run using both `ExperimentRun` and `RegistreredModelVersion`
- clean up imports in `deployable_entity/test_*.py`
- find & replace `experiment_run` with `deployable_entity` in `deployable_entity/test_*.py`
- find & replace `model_version` with `deployable_entity` in `deployable_entity/test_*.py`

## Testing

No new service, functionality, or tests added—this PR is changing how tests are run.
- [ ] ~Deployed the service to dev env~
- [ ] ~Used functionality on dev env~
- [ ] ~Added unit test(s)~
- [ ] ~Added integration test(s)~

Results of `deployable_entity/test_*.py` (job/pytest-3.7/186):
- `deployable_entity/test_artifacts.py::TestArtifacts::test_upload_multipart[RegisteredModelVersion]`
  - VR-12864
- `deployable_entity/test_artifacts.py::TestModels::test_keras`
  - VR-12823
- `deployable_entity/test_artifacts.py::TestOverwrite::test_model[RegisteredModelVersion]`
  - #2611
- `deployable_entity/test_artifacts.py::TestOverwrite::test_setup_script[RegisteredModelVersion]`
  - VR-12223
- `deployable_entity/test_deployment.py::TestLogModel::test_not_class_model_artifacts_error[RegisteredModelVersion]`
  - VR-12865
- `deployable_entity/test_deployment.py::TestFetchArtifacts::test_fetch_artifacts[RegisteredModelVersion]`
  - #2610
- `deployable_entity/test_deployment.py::TestDeployability::test_log_model[ExperimentRun]`
  - #2609
- `deployable_entity/test_deployment.py::TestDeployability::test_download_docker_context[RegisteredModelVersion]`
  - #2608

## How to Revert

Revert this PR.

PR #2603, opened 2021-09-09
Consolidate deployable entity tests

convoliution:
## Impact and Context

Client tests have various duplicated/overlapping tests between `ExperimentRun` and `RegisteredModelVersion`—the subclasses of `_DeployableEntity`. This PR consolidates those tests.

Note that tests in `deployable_entity/test_*.py` are each still specific to either `ExperimentRun` or `RegisteredModelVersion` (as they were just copy & pasted); #2604 makes the change to have those tests run with _both_ `ExperimentRun` and `RegisteredModelVersion`.

## Risks

Since tests are effectively being renamed, some test executions histories on Jenkins will lose their continuity.

## Changes

This PR is pure copy + paste + delete.

- add `deployable_entity/__init__.py`
- create `test_experimentrun/test_artifacts.py` from root `test_artifacts.py`
- create `deployable_entity/test_artifacts.py` from remaining root `test_artifacts.py`
- remove redundant artifact download tests from `registry/model_version/test_artifacts.py`
- remove redundant model artifact tests from `registry/model_version/test_deployment.py`
- create `test_utils/test_histogram_utils.py` from root `test_deployment.py`
- create `test_experimentrun/test_deployment.py` from root `test_deployment.py`
- create `deployable_entity/test_deployment.py` from remaining root `test_deployment.py` plus `registry/model_version/test_deployment.py`

## Testing

No new service, functionality, or tests added—this PR is just rearranging old tests.
- [ ] ~Deployed the service to dev env~
- [ ] ~Used functionality on dev env~
- [ ] ~Added unit test(s)~
- [ ] ~Added integration test(s)~

Verifying tests that are expected to still work:
- [x] `test_experimentrun/test_artifacts.py`
  - job/pytest-3.7/176 has one failure, addressed in #2605
- [x] `test_experimentrun/test_deployment.py`
  - job/pytest-3.7/177 all passed
- [x] `registry/model_version/test_artifacts.py`
  - job/pytest-3.7/178 all passed
- [x] `registry/model_version/test_deployment.py`
  - job/pytest-3.7/179 all passed

## How to Revert

Revert this PR

PR #2616, opened 2021-09-11
Remove non-generalizable, duplicated test

convoliution:
## Impact and Context

This test from the preceding #2615 actually isn't generalizable, because `RegisteredModelVersion` today doesn't enforce a blocklist in `log_artifact()`. So this PR removes it from `deployable_entity/`.

This test does still exist (verbatim) in [`test_experimentrun/test_artifacts.py`](https://github.com/VertaAI/modeldb/blob/311dffa2e6733061a2edb2dff06e7a55f232013e/client/verta/tests/test_experimentrun/test_artifacts.py#L73-L81), and an equivalent is present—though skipped—in [`registry/model_version/test_artifacts.py`](https://github.com/VertaAI/modeldb/blob/311dffa2e6733061a2edb2dff06e7a55f232013e/client/verta/tests/registry/model_version/test_artifacts.py#L163-L169).

## Risks

None.

## Testing

See #2603 and #2604.

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/deployable_entity/test_artifacts.py
+++ b/client/verta/tests/deployable_entity/test_artifacts.py
@@ -189,16 +189,6 @@ class TestArtifacts:
             with pytest.raises(ValueError):
                 experiment_run.log_artifact(key, artifact)
 
-    def test_blocklisted_key_error(self, experiment_run, all_values):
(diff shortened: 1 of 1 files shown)


conradoverta (approved):
Interesting

PR #2617, opened 2021-09-11
Move ER-specific deployment tests

convoliution:
## Impact and Context

These tests cover functionality that only exist in `ExperimentRun` today. Specifically:

- old-style deployment
- git ops, e.g. `download_deployment_yaml()`

This PR moves them to `test_experimentrun/test_deployment.py`. This is a straight copy & paste + `import`s at the top of the new file.

## Risks

Since tests are effectively being renamed, some test executions histories on Jenkins will lose their continuity.

## Testing

See #2603 and #2604.

## How to Revert

Revert this PR.

PR #2618, opened 2021-09-11
Move remainining generalizable deployment tests

convoliution:
## Impact and Context

After the preceding #2617, all that's left in the root `test_deployment.py` are tests that conceptually apply to both `ExperimentRun` and `RegisteredModelVersion`. Since they are common to deployable entities, this PR moves those remaining tests to the new `deployable_entity/` test subdirectory.

## Risks

Since tests are effectively being renamed, some test executions histories on Jenkins will lose their continuity.

## Testing

See #2603 and #2604.

## How to Revert

Revert this PR.


convoliution:
!rebuild

conradoverta (approved):

PR #2619, opened 2021-09-11
Move more generalizable deployment tests

convoliution:
## Impact and Context

`registry/model_version/test_deployment.py` contains tests that conceptually apply to both `ExperimentRun` and `RegisteredModelVersion`. Since they are common to deployable entities, this PR moves those tests to `deployable_entity/test_deployment.py` as well.

## Risks

Since tests are effectively being renamed, some test executions histories on Jenkins will lose their continuity.

## Testing

See #2603 and #2604.

## How to Revert

Revert this PR.


conradoverta (approved):
Seems like `DeployableEntity` is mostly a win =)

PR #2628, opened 2021-09-13
Add _MODEL_KEY to _DeployableEntity interface

convoliution:
## Impact and Context

For historical reasons, `ExperimentRun` and `RegisteredModelVersion` use different keys for their artifacts. This PR sets those keys as a class attribute as a part of the `_DeployableEntity` interface.

## Risks

This is a slight refactor, which always runs the risk of introducing regressions, but that's what tests are for.

There's also a risk that I missed some lines in my find-and-replace, but I think my find (`(?:[^_]MODEL_KEY|REGISTRY_MODEL_KEY)`) got all of them.

## Testing

  - no new service added
  - no new service added
- [ ] ~Added unit test(s)~
  - covered by existing tests
- [ ] ~Added integration test(s)~
  - covered by existing tests

Test results:

- Python 2.7 `deployable_entity/test_artifacts.py` **passed with expected failures** (job/pytest-2.7/49)
- Python 3.7 `deployable_entity/test_artifacts.py` **passed with expected failures** (job/pytest-3.7/209)

For more info on "expected failures", see #2604 (tl;dr they're captured in tickets)

## How to Revert

Revert this PR.


client/verta/tests/test_experimentrun/test_deployment.py
```diff
@@ -157,13 +157,13 @@ def test_no_model_deploy_error(self, experiment_run, model_for_deployment):
...
                                                               experiment_run._conn.socket),
-            experiment_run._conn, json={'id': experiment_run.id, 'key': _artifact_utils.MODEL_KEY}
+            experiment_run._conn, json={'id': experiment_run.id, 'key': experiment_run._MODEL_KEY}
         )
         _utils.raise_for_http_error(response)
 
         with pytest.raises(RuntimeError) as excinfo:
             experiment_run.deploy()
-        assert _artifact_utils.MODEL_KEY in str(excinfo.value)
+        assert experiment_run._MODEL_KEY in str(excinfo.value)
```
convoliution, line 166: These tests aren't actually being executed anymore, but I've made the change here for completeness.

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -45,6 +45,9 @@
 
 @six.add_metaclass(abc.ABCMeta)
 class _DeployableEntity(_ModelDBEntity):
+
+    _MODEL_KEY = None
```
daniel-verta: suggestion: if `None` isn't a valid `_MODEL_KEY` we should convert this to a property which raises a `NotImplementedError` if the sub-class doesn't supply a `_MODEL_KEY`.

daniel-verta (approved):

client/verta/verta/tracking/entities/_deployable_entity.py
convoliution: I had considered this, but I had decided against it since I would like `_MODEL_KEY` to be accessible from the `class` object, not just on instances.

Although maybe that's fine—the client might not be relying on it as a class attribute at all. I'll take another look through the code.

PR #2636, opened 2021-09-16
[VR-3772] Implement Docker environment blob

convoliution:
## Impact and Context

To support Docker-based environments, the client (at last) needs to implement the Docker environment blob.

~Linting fails at the moment because the code in this PR actually depends on #2634 😬.~

## Risks

This PR is fairly self-contained, so any risks would be isolated to local bugs.

## Testing

- [ ] ~Deployed the service to dev env~
  - no new service
- [ ] ~Used functionality on dev env~
  - changes are fully clientside
- [x] Added unit test(s)
  - Python 2.7 `versioning/environment/test_docker.py` **all passed** (job/pytest-2.7/60)
  - Python 3.7 `versioning/environment/test_docker.py` **all passed** (job/pytest-3.7/228)
  - tests were run by merging this, #2634, and #2637
- [ ] ~Added integration test(s)~
  - changes are fully clientside
  - related integration tests are added in #2637

## How to Revert

Revert this PR.

PR #2637, opened 2021-09-16
Support Docker environment through log/get_environment()

convoliution:
## Impact and Context

#2636 adds a new type of environment blob, and `ExperimentRun` and `RegisteredModelVersion` need to support it so that it can actually be used in our platform.

## Risks

This PR includes a refactor to the `log`/`get_environment()` methods, so it could introduce regressions around that functionality but that's what test are for.

## Testing

- [x] ~Deployed the service to dev env~
  - no new service
- [x] Used functionality on dev env
  - integration tests POST and GET from our backends
- [x] ~Added unit test(s)~
  - clientside test coverage already exists
  - Python 2.7 `deployable_entity/test_deployment.py::TestEnvironment` **passed with 1 failure** (job/pytest-2.7/62)
  - Python 3.7 `deployable_entity/test_deployment.py::TestEnvironment` **passed with 1 failure** (job/pytest-3.7/230)
  - the test failure is being tracked in VR-12944 (`ExperimentRun` doesn't implement `del_environment()`)
  - tests were run by merging this, #2634, and #2636

## How to Revert
Revert this PR.


client/verta/verta/_blob.py
```diff
@@ -13,7 +13,7 @@ def __eq__(self, other):
         if not isinstance(other, type(self)):
             return NotImplemented
 
-        return self._as_proto() == other._as_proto()
+        return repr(self) == repr(other)
```
daniel-verta, line 16: Can you share the motivated this change?

daniel-verta (approved):

client/verta/verta/_blob.py
convoliution: Ah, yes!

What I specifically encountered was that when the client creates a new e.g. `Python()`, it doesn't necessarily sort the requirements before sending a creation request to the backend. But when the backend receives it, it sorts e.g. [`repeated PythonRequirementEnvironmentBlob requirements`](https://github.com/VertaAI/modeldb/blob/b497e4b8b82aba8557dd24145b3ef1986a5a9830/protos/protos/public/modeldb/versioning/Environment.proto#L41) before saving the blob to the database. Since the field is a list rather than a mapping, the difference in ordering affects the protos' equality.

To not be beholden to differences in internal representation that may be out of the client's control, I changed this to use the objects' `repr` for which the client has already bundled logic for sorting things (for user-friendly consistency).

PR #2638, opened 2021-09-16
Fix role service methods for migration using service account

anandjakhaniya:
## Impact and Context

- Fix role service methods to support service account for migration and doing changes all the places according to it.

## Testing

- [ ] Added unit test(s)

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
+++ b/backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
@@ -407,22 +407,19 @@ public class RoleServiceUtils extends ai.verta.modeldb.common.authservice.RoleSe
    * workspace.
    */
   @Override
-  public WorkspaceDTO getWorkspaceDTOByWorkspaceId(
+  public WorkspaceDTO getWorkspaceDTOByWorkspaceIdForServiceUser(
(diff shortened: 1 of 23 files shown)


backend/src/main/java/ai/verta/modeldb/authservice/RoleServiceUtils.java
```diff
@@ -407,22 +407,19 @@ public Workspace getWorkspaceByWorkspaceName(
...
-        Organization organization = (Organization) getOrgById(workspaceId);
+        Organization organization = (Organization) getOrgById(true, workspaceId, true);
         workspaceDTO.setWorkspaceType(WorkspaceType.ORGANIZATION);
         workspaceDTO.setWorkspaceName(organization.getName());
         return workspaceDTO;
       case WorkspaceType.USER_VALUE:
         workspaceDTO.setWorkspaceType(WorkspaceType.USER);
-        if (currentLoginUserInfo == null) {
-          currentLoginUserInfo = authService.getCurrentLoginUserInfo();
-        }
```
conradoverta, line 425: ?

anandjakhaniya: This method is no more called by any user and it will only called by service user from background so not need to fetch `getCurrentLoginUserInfo` so removed it.

conradoverta (approved):

PR #2640, opened 2021-09-16
Factor out ensure_starts_with_slash() as a separate util function

convoliution:
## Impact and Context

This small behavior of "ensure an endpoint path begins with a slash" is going to be used for our Docker image model support, so this PR factors it out to a reusable utility function.

## Risks

This could introduce bugs, but I think there's ample test coverage to rule it out.

## Testing

  - no new service
  - no new backend integration behavior
- [x] Added unit test(s)

```bash
pytest test_utils/test_arg_handler.py test_endpoint/test_endpoint.py::TestEndpoint::test_path_ensure_starts_with_slash
```
passes on my machine in both Python 2.7 and 3.7.

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/verta/_internal_utils/arg_handler.py
+++ b/client/verta/verta/_internal_utils/arg_handler.py
@@ -94,3 +94,22 @@ def maybe(fn, val):
         return fn(val)
     else:
         return None
+
+
+def prepend_slash(path):
+    """Prepends ``"/"`` to `path` if it doesn't already start with one.
+
+    Parameters
+    ----------
+    path : str
+        Request path
+
+    Returns
+    -------
+    str
+        `path` prepended with ``"/"`` if it wasn't already.
+
+    """
+    if path.startswith("/"):
+        return path
+    return "/" + path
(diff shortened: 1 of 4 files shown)


client/verta/verta/_internal_utils/arg_handler.py
```diff
@@ -94,3 +94,22 @@ def maybe(fn, val):
         return fn(val)
     else:
         return None
+
+
+def prepend_slash(path):
```
daniel-verta: Minor suggestion to name functions like this with the verbs "normalize", "standardize", or "ensure." For example: "standardize_path." This isn't always prepending a slash, only when one which ought to be present is not there.

daniel-verta (approved):

client/verta/verta/_internal_utils/arg_handler.py
convoliution: Good point! I've renamed it to `ensure_starts_with_slash()` which is a bit lengthy, but is unambiguous

PR #2642, opened 2021-09-17
Delete old deployment

conradoverta:
## Testing

- [ ] Added unit test(s)


client/verta/tests/test_experimentrun/test_deployment.py
```diff
@@ -1,318 +0,0 @@
...
-
-        filepath = experiment_run.download_deployment_yaml(download_to_path)
-        assert filepath == os.path.abspath(download_to_path)
-
-        # can be loaded as YAML
-        with open(filepath, 'rb') as f:
-            model_deployment = yaml.safe_load(f)
-
-        assert model_deployment['kind'] == "ModelDeployment"
-        assert model_deployment['metadata']['name'] == experiment_run.id
```
convoliution, line 318: ~We should keep this (unless `download_deployment_yaml()` is also defunct)! Everything else can go.~

Oops commented too soon.

I do see `download_deployment_yaml()` was removed from `ExperimentRun`. Is that not being supported anymore?

convoliution (changes requested):

client/verta/tests/test_experimentrun/test_deployment.py
convoliution: Is [`download_endpoint_manifest()`](https://verta.readthedocs.io/en/master/_autogen/verta.Client.html?highlight=download_endpoint_manifest#verta.Client.download_endpoint_manifest) its replacement??

conradoverta: Yes

convoliution (approved):

PR #2643, opened 2021-09-20
[VR-12958] Replace git-installed verta in requirements with a pinned version

convoliution:
**Recommend reviewing with "hide whitespace changes"**

## Impact and Context

This PR addresses two issues:

1. If a user's environment has `git+https://github.com/VertaAI/modeldb.git#egg=verta`, `Python(requirements)` wouldn't recognize this and would append `verta=={current_version}`, resulting in a duplicate requirement and an installation error.
2. If a user's environment has `git+ssh://git@github.com/VertaAI/modeldb.git#egg=verta`, this has the above issue, but also wouldn't work anyway because the model build machinery may not have the requisite ssh credentials.

The solution chosen is to deliberately replace _all_ git installations of the verta package with `verta=={current_version}`.

## Risks

This means that a user working locally with a non-publicly-published version of the client—when they deploy a model—will have their client version effectively "rounded down" to the nearest public version. But aside from the brief window of this past week, this has always been the case!

## Testing

  - no new service
- [x] Used functionality on dev env
   - deployed a model using `Python(Python.read_pip_environment())` and it worked
- [x] Added unit test(s)
  - `test_utils/test_pip_requirements.py::TestPinVertaAndCloudpickle::test_inject_requirement`
  - `versioning/environment/test_python.py::TestVCSInstalledVerta::test_vcs_installed_verta`

`test_utils/test_pip_requirements.py` and `versioning/environment/test_python.py` all pass in Jenkins in Python 2.7 and 3.7.

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/versioning/environment/conftest.py
+++ b/client/verta/tests/versioning/environment/conftest.py
@@ -35,7 +35,6 @@ def requirements_file_with_unsupported_lines():
             "-c some_constraints.txt",
             "-f file://dummy",
             "-i https://pypi.org/simple",
-            "-e git+ssh://git@github.com/VertaAI/modeldb.git@master#egg=verta&subdirectory=client/verta",
             "-r more_requirements.txt",
             "en-core-web-sm==2.2.5",
         ]
--- a/client/verta/verta/_internal_utils/_pip_requirements_utils.py
+++ b/client/verta/verta/_internal_utils/_pip_requirements_utils.py
@@ -277,49 +277,72 @@ def pin_verta_and_cloudpickle(requirements):
         conflicts with the version in the current environment.
 
     """
+    # replace VCS-installed verta (`pip install -e verta`) with "verta"
     for i, req in enumerate(requirements):
-        if req.startswith("-e git+git@github.com:VertaAI/modeldb.git"):
-            # `pip install -e modeldb/client/verta` can make `pip freeze` return
-            # this requirement item which is unusable by `pip install`
-            # https://github.com/pypa/pip/issues/7554
-            # https://github.com/pypa/pip/issues/9625
-            # https://github.com/pypa/pip/pull/9436
-            # https://github.com/pypa/pip/pull/9822
+        if "VertaAI/modeldb.git" in req and "#egg=verta" in req:
+            # git+git is not installable
+            #     https://github.com/pypa/pip/issues/7554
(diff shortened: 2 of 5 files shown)


client/verta/tests/versioning/environment/conftest.py
```diff
@@ -35,7 +35,6 @@ def requirements_file_with_unsupported_lines():
             "-c some_constraints.txt",
             "-f file://dummy",
             "-i https://pypi.org/simple",
-            "-e git+ssh://git@github.com/VertaAI/modeldb.git@master#egg=verta&subdirectory=client/verta",
```
convoliution, line 38: Removing this because it's no longer an "unsupported line" (instead, it gets swapped out with `"verta"` before supported-ness is even checked).

client/verta/verta/_internal_utils/_pip_requirements_utils.py
```diff
@@ -277,49 +277,72 @@ def pin_verta_and_cloudpickle(requirements):
...
+            continue
+
+    # inject verta
+    inject_requirement(requirements, __about__.__title__, __about__.__version__)
+
+    # inject cloudpickle
+    inject_requirement(requirements, cloudpickle.__name__, cloudpickle.__version__)
+
+
+def inject_requirement(requirements, library, version):
```
daniel-verta, line 309: In my own experience, functions which alter mutable arguments are harder to reason about and debug than pure functions. Unless there's a compelling performance reason, I would suggest that a method such as this should instead return the intended output instead of silently mutating inputs.

daniel-verta (changes requested):
Feel free to contest or discuss my request to favor a pure functions approach for methods such as `inject_requirements`.

Outside of the scope of this change, I'll note that we can also make our code surrounding Python requirements more robust by using the `setuptools` library and its models for Python requirements, e.g. [`Requirement` Objects](https://setuptools.pypa.io/en/latest/pkg_resources.html#requirement-objects). I'll make a ticket to capture that improvement which we can then prioritize accordingly.

client/verta/verta/_internal_utils/_pip_requirements_utils.py
convoliution: Sounds reasonable to me!

daniel-verta (approved):

PR #2662, opened 2021-09-28
[VR-12591] Display upload status of artifacts and prevent downloading incomplete artifacts

daniel-verta:
## Review Notes
I've tried to bundle formatting changes in clean formatting-only commits. Reviewing this using commit ranges may be easier than reviewing the atomic change.

## Impact and Context
Artifacts are now tracked on the backend as having completed uploads or not. This PR makes that status visible and prevents attempts to download incomplete artifacts client-side.

## Testing

- [ ] Added unit test(s)

## How to Revert
Revert this pull request.


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -814,26 +814,6 @@ def unravel_key_values(rpt_key_value_msg):
             in rpt_key_value_msg}
 
 
-def unravel_artifacts(rpt_artifact_msg):
```
daniel-verta, line 817: This was only used in the `__repr__` method of the `ExperimentRun` class.

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -118,7 +118,7 @@ def __repr__(self):
                 "registered model id: {}".format(msg.registered_model_id),
                 "experiment run id: {}".format(msg.experiment_run_id),
                 # "archived status: {}".format(msg.archived == _CommonCommonService.TernaryEnum.TRUE),
-                "artifact keys: {}".format(artifact_keys),
```
daniel-verta, line 122: This would be redundant with the prettified artifact string which includes this information in addition to the upload status.

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -627,14 +623,8 @@ def get_artifact(self, key):
 
         return artifact_stream
 
-    def download_artifact(self, key, download_to_path):
```
daniel-verta, line 679: The shared preamble code has moved to `DeployableEntity.download_artifact` with the diverging implementations following in `_internal_download_artifact`

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -448,7 +494,11 @@ def fetch_artifacts(self, keys):
             #     identical, so multiple writes would be idempotent.
             path = self._get_cached_file(filename)
             if path is None:
-                contents = self._get_artifact(key)
+                try:
```
daniel-verta, line 496: A `ValueError` is raised by `_get_artifact` if the artifact is incomplete. I have guessed that partial success is a better behavior for this method than total failure if any one of the artifacts has an incomplete upload.

client/verta/verta/tracking/entities/_experiment.py
```diff
@@ -30,11 +30,14 @@ class Experiment(_ModelDBEntity):
         Experiment Runs under this Experiment.
 
     """
+
```
daniel-verta, line 33: Changes in this file only belong to the auto-format commit

client/verta/verta/tracking/entities/_experimentruns.py
```diff
@@ -5,7 +5,9 @@
 import copy
 import warnings
 
-from verta._protos.public.modeldb import ExperimentRunService_pb2 as _ExperimentRunService
+from verta._protos.public.modeldb import (
```
daniel-verta, line 8: Changes in this file only belong to the auto-format commit

client/verta/verta/tracking/entities/_experiments.py
```diff
@@ -14,27 +14,28 @@
 class Experiments(_LazyList):
     # keys that yield predictable, sensible results
     _VALID_QUERY_KEYS = {
-        'id',
-        'name',
-        'date_created',
-        'date_updated',
-        'attributes',
-        'tags',
+        "id",
```
daniel-verta, line 17: Changes in this file only belong to the auto-format commit

client/verta/verta/tracking/entities/_projects.py
```diff
@@ -14,27 +14,28 @@
 class Projects(_LazyList):
     # keys that yield predictable, sensible results
     _VALID_QUERY_KEYS = {
-        'id',
-        'name',
-        'date_created',
-        'date_updated',
-        'attributes',
-        'tags',
+        "id",
```
daniel-verta, line 17: Auto-format only

client/verta/verta/tracking/entities/_project.py
```diff
@@ -37,17 +37,22 @@ class Project(_ModelDBEntity):
         Experiment Runs under this Project.
 
     """
+
```
daniel-verta, line 40: Auto-format only

client/verta/verta/tracking/entities/_experimentrun.py
```diff
@@ -1720,7 +1724,7 @@ def download_artifact(self, key, download_to_path):
 
         # get a stream of the file bytes, without loading into memory, and write to file
         # TODO: consolidate this with _get_artifact() and get_artifact()
-        print("downloading {} from ModelDB".format(key))
+        logger.info("downloading %s from ModelDB", key)
```
convoliution, line 1717: I'd generally be hesitant about doing piecemeal PR-by-PR transitions from `print` to `logger`, because some users 😬 _may_ expect to see print statements regarding downloads from their time using our client, and this breaks public-facing behavior (if they haven't configured their logging level).

That being said, the upcoming release will have backwards-incompatibilities anyway so this is fine to me

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -189,8 +190,21 @@ def download_artifact(self, key, download_to_path):
...
+        if artifact.upload_completed:
+            pathlib2.Path(download_to_path).parent.mkdir(parents=True, exist_ok=True)
+            # TODO: clean up empty parent dirs if something later fails
+            return self._internal_download_artifact(key, artifact, download_to_path)
+        else:
+            raise ValueError("artifact upload incomplete; download not possible")
+
+
+    @abc.abstractmethod
+    def _internal_download_artifact(self, key, artifact, download_to_path):
```
convoliution, line 211: I'd like to request this be named either `_download_artifact_internal()` (in the vein of our existing `_create_proto_internal()`s) or `_download_artifact_inner()` (in the vein of our existing `_from_dict_inner()`s).

I'd push for the former, since "internal" seems to have been the more intuitive choice ("inner" was mine 😬)

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -580,7 +580,7 @@ def log_artifact(self, key, artifact, overwrite=False, _extension=None):
         self._update(self._msg, method="PUT")
         self._upload_artifact(key, artifact_stream, artifact_type=artifact_type)
 
-    # TODO: Fix for VR-12591
+
```
convoliution, line 632: It looks like this PR introduces just a couple of extra newlines. PEP8 recommends just one blank line to separate methods in a class.

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -865,8 +856,8 @@ def _upload_artifact(
 
     def _get_artifact(self, key, artifact_type=0):
         # check to see if key exists
-        self._get_artifact_msg(key)
-
+        artifact_msg = self._get_artifact_msg(key)
+        self._validate_artifact_uploaded(artifact_msg)
```
convoliution, line 860: Since `artifact_msg` goes unused, I'd request not binding it to a variable at all, unless it seems more readable that way.

```suggestion
        self._validate_artifact_uploaded(self._get_artifact_msg(key))
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -172,6 +175,14 @@ def log_model(
         """
         raise NotImplementedError
 
+
+    @classmethod
+    def _validate_artifact_uploaded(cls, artifact_msg):
```
convoliution: Since `cls` goes unused, I'd recommend going with a `staticmethod` in the interest of being persimmonious.

```suggestion
    @staticmethod
    def _validate_artifact_uploaded(artifact_msg):
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -301,6 +323,24 @@ def _create_artifact_msg(
         )
         return artifact_msg
 
+    @classmethod
+    def _pretty_artifact_str(cls, artifact_msg):
```
convoliution, line 326: Since `cls` goes unused, I'd recommend going with a `staticmethod` in the interest of being persimmonious.

```suggestion
    @staticmethod
    def _pretty_artifact_str(artifact_msg):
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -448,7 +494,11 @@ def fetch_artifacts(self, keys):
             #     identical, so multiple writes would be idempotent.
             path = self._get_cached_file(filename)
             if path is None:
-                contents = self._get_artifact(key)
+                try:
+                    contents = self._get_artifact(key)
+                except ValueError as error:
+                    logger.exception(error)
```
convoliution, line 499: A few thoughts on this:
1. `logger.exception()` already [automatically captures exception info](https://docs.python.org/3/library/logging.html#logging.Logger.exception) (specifically the stack trace), so I think attaching the exception as the message may be redundant
2. This is a bit shakier, but to me this isn't exactly an exception when fetching the artifact per se (it's more of a prior exception on _upload_ that left it in a bad state for subsequent operations) so I'd consider a warning more appropriate than an exception (and `warning()` doesn't do the automatic thing from (1), so the exception needs to be given).
3. I'd request using `as e` which is what [most of] the rest of our codebase is using today

```suggestion
                except ValueError as e:
                    logger.warning(exc_info=e)
```

convoliution (changes requested):
Functionality looks about right—most requested changes are about coding conventions (which I'm hoping to start codifying this quarter!)

Would you also mind attaching an example of the `repr` as you go about adding tests?

client/verta/verta/tracking/entities/_deployable_entity.py
convoliution: @daniel-verta Actually, `True` can be passed to `exc_info` which will fetch the stack trace from the context, so we wouldn't have to bother with the `as e` at all!

```suggestion
                except ValueError:
                    logger.warning(exc_info=True)
```

liuverta:
This has numerous merge conflicts, and may be best handled from the beginning if the ticket is picked up again.

PR #2664, opened 2021-09-29
Document useful pytest flags

convoliution:
## Impact and Context

`pytest` has useful flags for augmenting its output when run at the terminal. These options can be a bit difficult to find from `pytest`'s documentation, so I'm listing here the ones I've found nifty.

## Risks

None.

## Testing

I ran these flags at my terminal.

## How to Revert

Revert this PR


bts-verta (approved):
awesome sauce

PR #2671, opened 2021-09-30
Skip clone tests

convoliution:
## Impact and Context

These tests are currently failing due to a backend bug regarding persisting `upload_completed` on clones (VR-13087). I've filed VR-13088 to un-skip them when the backend bug is resolved.

And while I was at it...these tests were written before our objects had a caching mechanism, so in the interest of unifying code paths I've changed the tests to use the cache instead of directly calling `_get_proto_by_id()`.

## Risks

Skipping these means we're missing coverage on our ER clone functionality until, the backend bug is resolved.

## Testing

None—they're being skipped (though I verified these pass if the equality check on `.artifacts` is skipped).

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/test_entities.py
+++ b/client/verta/tests/test_entities.py
@@ -370,28 +370,26 @@ class TestExperimentRun:
         expt_run.log_artifact("my-artifact", "README.md")
         expt_run.log_code()
 
-        # set various things in the run
-        new_run_no_art = expt_run.clone()
+        # clone run
+        new_run = expt_run.clone()
+        expt_run._fetch_with_no_cache()
+        new_run._fetch_with_no_cache()
 
-        old_run_msg = expt_run._get_proto_by_id(expt_run._conn, expt_run.id)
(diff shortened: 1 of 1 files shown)


client/verta/tests/test_entities.py
```diff
@@ -370,28 +371,26 @@ def test_clone(self, experiment_run):
...
 
-        # set various things in the run
-        new_run_no_art = expt_run.clone()
+        # clone run
+        new_run = expt_run.clone()
+        expt_run._fetch_with_no_cache()
+        new_run._fetch_with_no_cache()
 
-        old_run_msg = expt_run._get_proto_by_id(expt_run._conn, expt_run.id)
-        new_run_no_art_msg = new_run_no_art._get_proto_by_id(new_run_no_art._conn, new_run_no_art.id)
```
convoliution, line 377: The ER cloning functionality used to have a user-facing option to skip cloning artifacts. It doesn't anymore, but it appears this test still kept the `new_run_no_art_msg` variable (`no_art` for "no artifacts"). I've replaced it with simply `new_run`.

daniel-verta (approved):

PR #2674, opened 2021-10-04
[VR-13024] Event logged for experiment_run resource in local database

anandjakhaniya:
## Impact and Context

**Jira Ticket:** https://vertaai.atlassian.net/browse/VR-13024

- Event logged in local database for project service add/update endpoints

## Testing

- [ ] Added unit test(s)


evgeniilezhnin (commented):
No addEvent in deleteExperimentRun, deleteExperimentRuns in Future service and service

Also why there is so many same methods in future and regular service? They are used both? Should we delete these methods from regular service to not add events in these methods?

evgeniilezhnin (approved):
assuming deletion in separate PR

conradoverta (approved):
I won't review 1.6k LOC =)

Let's see if maybe we can provide less granularity on the events. Otherwise it will also be almost impossible to consume. For example, we can group all "ER updated" into a single event type.

conradoverta (changes requested):
Pressed the wrong button

anandjakhaniya:
> I won't review 1.6k LOC =)
> 
> Let's see if maybe we can provide less granularity on the events. Otherwise it will also be almost impossible to consume. For example, we can group all "ER updated" into a single event type.

You can see document at https://github.com/VertaAI/docs/pull/89 where there is only three event types for run CRUD which is ADD/UPDATE/DELETE .

PR #2677, opened 2021-10-06
Fix repository string examples

convoliution:
## Impact and Context

I realized that I was using "registry/repository" instead of just "repository" in these examples and tests.

## Risks

None.

## Testing

I re-ran

```
pytest registry/test_docker_image.py versioning/environment/test_docker.py
```

on my machine, and the tests still pass.

## How to Revert

Revert this PR.


daniel-verta (approved):

CaptEmulation (changes requested):
How will our backend know to pull the containers from `012345678901.dkr.ecr.apne2-az1.amazonaws.com` ?

PR #2688, opened 2021-10-22
Used entities workspace id in events instead of current login user workspace ID - 1

anandjakhaniya:
## Impact and Context

Used project workspace id in experiment and experimentRun service events instead of current login user workspace ID
Covered entities in this PR are:
- Experiment
- ExperimentRun


conradoverta (changes requested):
This will be _way_ too slow as we'll call uac on every single request on ER. We have already spent a ton of time optimizing this path and doing this now will just kill performance all over again

anandjakhaniya:
> This will be _way_ too slow as we'll call uac on every single request on ER. We have already spent a ton of time optimizing this path and doing this now will just kill performance all over again

but we do not have any other option for the same because `workspace_id` will be a required field so we need to populate that field. do you have nay better design in mind for this?

anandjakhaniya:
> > This will be _way_ too slow as we'll call uac on every single request on ER. We have already spent a ton of time optimizing this path and doing this now will just kill performance all over again
> 
> but we do not have any other option for the same because `workspace_id` will be a required field so we need to populate that field. do you have nay better design in mind for this?

We have decided to keep this in and further going with some caching mechanism to reduce UAC calls

anandjakhaniya:
Created other prs #2693, #2694, #2695 so closing this PR.

PR #2692, opened 2021-10-26
[VR-13282] startup fixes for mssql

anandjakhaniya:

### DIFF
--- a/backend/src/main/resources/liquibase/db-changelog-2.0.xml
+++ b/backend/src/main/resources/liquibase/db-changelog-2.0.xml
@@ -981,6 +981,9 @@
 
     <changeSet id="marked-existing-artifact-uploaded-true" author="anandJ">
         <preConditions onFail="MARK_RAN">
+            <not>
+                <dbms type="mssql"/>
+            </not>
             <columnExists tableName="artifact" columnName="upload_completed"/>
         </preConditions>
         <update tableName="artifact">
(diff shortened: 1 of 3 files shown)


backend/src/main/resources/liquibase/db-changelog-2.0.xml
```diff
@@ -981,6 +981,9 @@
 
     <changeSet id="marked-existing-artifact-uploaded-true" author="anandJ">
         <preConditions onFail="MARK_RAN">
+            <not>
+                <dbms type="mssql"/>
+            </not>
```
conradoverta, line 986: Does this change the hash/signature of the changeset?

conradoverta (approved):
One possible concern. Otherwise LGTM

backend/src/main/resources/liquibase/db-changelog-2.0.xml
anandjakhaniya: No, if there will be hash changed then it will break system but here whatever we are adding in `preConditions` it will not affect on the hash/signature.

evgeniilezhnin (approved):

backend/src/main/resources/liquibase/db-changelog-2.0.xml
conradoverta: cool

eheinlein-verta (approved):

PR #2693, opened 2021-10-27
project events implementation with feature flag

anandjakhaniya:

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
+++ b/backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
@@ -1444,9 +1444,6 @@ public class ProjectServiceImpl extends ProjectServiceImplBase {
 
       // Add succeeded event in local DB
       for (String projectId : deletedProjectIds) {
-        GetResourcesResponseItem projectResource =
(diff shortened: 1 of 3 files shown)


backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
```diff
@@ -1438,7 +1449,7 @@ public void deleteProjects(
                 Optional.of(projectId), Optional.empty(), ModelDBServiceResourceTypes.PROJECT);
```
conradoverta: Should we even be making this call to uac if it's disabled? Let's review to make sure we're not doing unnecessary calls.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/project/ProjectServiceImpl.java
anandjakhaniya: nice catch, i missed that place :)

anandjakhaniya: line was unused and forgot to remove it. so removed it now.

conradoverta (approved):

PR #2699, opened 2021-10-27
repository events implementation with feature flag

anandjakhaniya:


conradoverta (commented):
Why are we repeating the flag in this case vs centralizing like for project or dataset? Centralization is a much cleaner design.

anandjakhaniya:
> Why are we repeating the flag in this case vs centralizing like for project or dataset? Centralization is a much cleaner design.

Updated code with Centralization pattern

conradoverta (approved):

PR #2701, opened 2021-10-29
VR-13293 utf8 with mssql

evgeniilezhnin:
## Impact and Context

utf8 support in mssql

## How to Revert

revert the PR

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonHibernateUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonHibernateUtil.java
@@ -166,16 +166,13 @@ public abstract class CommonHibernateUtil {
     throw new ModelDBException("Unrecognized database " + rdbConfiguration.getRdbDialect());
   }
 
-  public static void changeCharsetToUtf(JdbcConnection jdbcCon,
-      RdbConfig rdbConfiguration)
+  public static void changeCharsetToUtf(JdbcConnection jdbcCon, RdbConfig rdbConfiguration)
       throws DatabaseException, SQLException {
     try (var stmt = jdbcCon.createStatement()) {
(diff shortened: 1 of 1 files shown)


evgeniilezhnin:
need first make sure that this is supported. UTF-8 is supported since SQL Server 2019

backend/common/src/main/java/ai/verta/modeldb/common/CommonHibernateUtil.java
```diff
@@ -166,13 +166,20 @@ private String getDatasourceClass(RdbConfig rdbConfiguration) {
...
-              "ALTER DATABASE `%s` CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;", dbName);
+      String sql;
+      if (rdbConfiguration.isMssql()) {
+        sql =
+            String.format(
+                "ALTER DATABASE \"%s\" COLLATE Latin1_General_100_CI_AI_SC_UTF8;", dbName);
+      } else {
+        sql =
+            String.format(
+                "ALTER DATABASE `%s` CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;", dbName);
```
eheinlein-verta, line 181: My understanding is that `utf8mb4_unicode_ci` is better to use than `utf8mb4_general_ci`.  I know this code was already using `utf8mb4_general_ci`, and making this change now might cause problems, but I wanted to make a note.

evgeniilezhnin: makes sence, but the current change was about utf-8 in Microsoft SQL, this line is about mysql

evgeniilezhnin:
will be using utf-16 instead utf-8

PR #2707, opened 2021-11-03
Fix remaining code impelementation to support mssql

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/entities/KeyValueEntity.java
```diff
@@ -29,6 +32,21 @@ public KeyValueEntity() {}
 
   public KeyValueEntity(Object entity, String fieldType, KeyValue keyValue) {
     setKey(keyValue.getKey());
+    if (App.getInstance().mdbConfig.getDatabase().getRdbConfiguration().isMssql()) {
```
eheinlein-verta: I noticed that this line is repeated in a couple of places, and some other places have shorter versions, e.g 
`config.getDatabase().getRdbConfiguration().isMssql()`

It would probably be useful to create a single method (perhaps that takes as a parameter either the general config, or perhaps the database returned from `getDatabase()`

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
```diff
@@ -243,16 +255,27 @@ private void insertKeyValue(String entityId, Handle handle, KeyValue kv) {
                     jdbi.useHandle(
                         handle -> {
                           if (exists) {
+                            String value = ModelDBUtils.getStringFromProtoObject(kv.getValue());
+                            // Logic to convert canonical number to double number
+                            if (kv.getValue().hasNumberValue()) {
```
eheinlein-verta: Can this logic be pulled into a function so it can be reused here and in `KeyValueEntity` line 37? The two places have very similar logic with only a small change, so if possible it would be nice to not repeat i t. It might also be useful in the future.

eheinlein-verta (commented):
I left a couple of questions - looks generally good but I'd like to hear your thoughts on them.

backend/src/main/java/ai/verta/modeldb/entities/KeyValueEntity.java
anandjakhaniya: Extracted duplicate code in util

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/KeyValueHandler.java
anandjakhaniya: Extracted duplicate code in util

anandjakhaniya:
> I left a couple of questions - looks generally good but I'd like to hear your thoughts on them.

Extracted duplicate code in utils as per your point of view.

eheinlein-verta (approved):
Looks great!

PR #2708, opened 2021-11-08
Fix broken test cases and Improve entities clean up code for test cases

anandjakhaniya:


backend/src/test/java/ai/verta/modeldb/TestsInit.java
```diff
@@ -165,15 +161,25 @@ public static void setServerAndService() throws Exception {
...
   }
+
+  protected static void cleanUpResources() throws InterruptedException {
+    // Remove all entities
+    // removeEntities();
+    // Delete entities by cron job
+    ReconcilerInitializer.softDeleteProjects.resync();
+    Thread.sleep(5000);
+    ReconcilerInitializer.softDeleteExperiments.resync();
+    Thread.sleep(5000);
```
conradoverta: Time-based testing is a bad, bad idea. It both makes tests slower _and_ create flaky tests. Can we do better?

conradoverta (changes requested):

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
```diff
@@ -142,4 +144,8 @@ public void insert(T element) {
   public abstract void resync();
 
   protected abstract ReconcileResult reconcile(Set<T> objs);
+
+  public boolean isElementsNotDeleted() {
+    return !processingIdSet.isEmpty();
+  }
```
conradoverta:
```suggestion
  public boolean isEmpty() {
    return processingIdSet.isEmpty();
  }
```
It's better to use nomenclature common in the libraries around. It's much clearer than trying to think about what a given name means.

backend/src/main/java/ai/verta/modeldb/reconcilers/ReconcilerInitializer.java
```diff
@@ -25,32 +26,34 @@
   public static void initialize(
       Config config, ServiceSet services, DAOSet daos, FutureJdbi futureJdbi, Executor executor) {
     LOGGER.info("Enter in ReconcilerUtils: initialize()");
+
+    ReconcilerConfig reconcilerConfig = new ReconcilerConfig();
+    reconcilerConfig.setTestReconciler(config instanceof TestConfig);
```
conradoverta: This initialization and then setting extra things is an anti-pattern. We should handle everything in the initialization.

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
```diff
@@ -40,7 +40,9 @@ protected Reconciler(
     this.executor = executor;
     this.deduplicate = deduplicate;
 
-    startResync();
+    if (!config.isTestReconciler()) {
+      startResync();
+    }
```
conradoverta, line 45: I'm not a big fan of code that behaves differently during tests. Why do we need this? Can we make things work in a way that is the same in test or out?

backend/src/test/java/ai/verta/modeldb/TestsInit.java
```diff
@@ -165,15 +166,44 @@ public static void setServerAndService() throws Exception {
...
+    }
+    softDeleteExperimentRuns.resync();
+    while (softDeleteExperimentRuns.isElementsNotDeleted()) {
+      LOGGER.trace("ExperimentRun deletion is still in progress");
+    }
 
-    // shutdown test server
-    serverBuilder.build().shutdownNow();
+    ReconcilerInitializer.softDeleteDatasets.resync();
+    ReconcilerInitializer.softDeleteRepositories.resync();
```
conradoverta: It would be fine if we had a resync cronjob here too, right? Not sure why we have to have that `if` on the testing condition.

backend/src/test/java/ai/verta/modeldb/TestsInit.java
```diff
@@ -165,15 +166,44 @@ public static void setServerAndService() throws Exception {
...
-    deleteEntitiesCron.run();
+    SoftDeleteProjects softDeleteProjects = ReconcilerInitializer.softDeleteProjects;
+    SoftDeleteExperiments softDeleteExperiments = ReconcilerInitializer.softDeleteExperiments;
+    SoftDeleteExperimentRuns softDeleteExperimentRuns =
+        ReconcilerInitializer.softDeleteExperimentRuns;
+
+    softDeleteProjects.resync();
+    while (softDeleteProjects.isElementsNotDeleted()) {
+      LOGGER.trace("Project deletion is still in progress");
+    }
```
conradoverta: Never do a kind of loop like this without a sleep. This will go crazy consuming CPU for no reason. Even sleeping for 10ms will make this better.

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
anandjakhaniya: this will be starting new periodic thread to fetched deleted entity and try to delete but in the test cases we are calling directly `resync` and not need to start the periodic thread. if we will depends on it then in between tests it will difficult to identify when it will execute and how much time we will wait to finish it. other wise all other code will used the same one.

backend/src/test/java/ai/verta/modeldb/TestsInit.java
anandjakhaniya: that if added due to https://github.com/VertaAI/modeldb/pull/2708#discussion_r756090704

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
anandjakhaniya: FIxed

backend/src/main/java/ai/verta/modeldb/reconcilers/ReconcilerInitializer.java
anandjakhaniya: Fixed and used the constructor to initialize field.

backend/src/test/java/ai/verta/modeldb/TestsInit.java
anandjakhaniya: Fixed

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
```diff
@@ -142,4 +144,8 @@ public void insert(T element) {
   public abstract void resync();
 
   protected abstract ReconcileResult reconcile(Set<T> objs);
+
+  public boolean isEmpty() {
+    return !processingIdSet.isEmpty();
+  }
```
conradoverta: This seems wrong, no? It's empty if the id set is not empty?

conradoverta (changes requested):

backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
anandjakhaniya: Yes, Renamed it with `isNotEmpty`

conradoverta: It defeats the purpose of the original comment (https://github.com/VertaAI/modeldb/pull/2708#discussion_r755682768), but ok.

conradoverta (approved):

PR #2710, opened 2021-11-08
[VR-13096] Support specifying command for user-provided container

convoliution:


liuverta:
This effort has been stale for a while, and shouldn't be difficult to pick up from the beginning.

PR #2711, opened 2021-11-09
[VR-13366] Add ModelError class to be caught in model wrapping infra

daniel-verta:
## Impact and Context

- Users have requested the ability to raise custom errors in order to have better control over the error behavior of their deployed models
- The introduced `ModelError` may be handled specially in the provided HTTP server wrapping models

## Risks and Area of Effect

- introduces public API for custom errors which will need stability after release

## Testing

```
from verta.registry.errors import CustomModelError
import traceback
import sys

msg, code, printed_tb = None, None, None

try:
    raise CustomModelError(418, "teapot does not support tensorflow")
except CustomModelError as cme:
    msg = cme.message
    code = cme.http_code
    _, _, tb = sys.exc_info()
    printed_tb = traceback.format_tb(tb)
```

## How to Revert

- revert this PR

### DIFF
--- a/client/verta/verta/registry/errors.py
+++ b/client/verta/verta/registry/errors.py
@@ -15,4 +15,4 @@ class CustomModelError(Exception):
     def __init__(self, http_code, message):
         self.http_code = http_code
         self.message = message
-        super().__init__(self.message)
+        super(CustomModelError, self).__init__(self.message)
--- a/client/verta/verta/registry/errors.py
+++ b/client/verta/verta/registry/errors.py
@@ -1,4 +1,4 @@
-class CustomModelError(Exception):
+class ModelError(Exception):
     """Exception to be raised in user-defined models.
 
     Messages of this class are caught in model containers and converted to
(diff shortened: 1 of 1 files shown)


client/verta/verta/registry/errors.py
```diff
@@ -0,0 +1,18 @@
+class CustomModelError(Exception):
```
convoliution: How would you feel about naming it `ModelError` rather than `CustomModelError`? The "Custom" feels a little bit on the nose to me.

client/verta/verta/registry/errors.py
```diff
@@ -0,0 +1,18 @@
...
+
+    Attributes
+    ----------
+    http_code : int
+        The http code to be returned by a wrapping service.
+    message : str
+        The response message.
+    """
+
+    def __init__(self, http_code, message):
```
convoliution: Would you know if it'd be more idiomatic to have the message first? I've rarely encountered exceptions that didn't have the `message` as the first parameter, but that might just be limited exposure on my part.

client/verta/verta/registry/errors.py
```diff
@@ -0,0 +1,18 @@
...
+        self.http_code = http_code
+        self.message = message
```
convoliution, line 16: I wonder if this could be a `property` for `self.args[0]`, but that might be an over-optimization.

daniel-verta: Let's keep it simple.

convoliution: I'm unsure whether having three ways to obtain the message string from this exception class (`e.message`, `str(e)`, and `e.args[0]`), with this one even being openly mutable in a way that would break internal consistency, is simpler than just having the two that `Exception` already offers.

But this is somewhat beyond my familiarity with Python conventions so I'm alright with revisiting it later if it becomes any sort of real issue 😆

convoliution (approved):

PR #2716, opened 2021-11-17
[VR-13352] Support JWT Cookies as an authn mechanism in Python client

daniel-verta:
## Impact and Context

- Adds support for authentication via `Grpc-Metadata-bearer_access_token` and `Grpc-Metadata-bearer_access_token_sig` HTTP headers
- The values for these headers may now be set with the environment variables `VERTA_JWT_TOKEN` and `VERTA_JWT_TOKEN_SIG`

## Risks and Area of Effect

- Although some of the refactoring and changes introduced touch upon the core `Client` and `Connection` classes, behavior has been checked with our extensive E2E test suite.

### Unit Testing

- Added `tests/http_requests/test_connection.py` to validate header computation behavior

```
(verta-py3) ➜  verta git:(df/vr-13352) pytest tests/http_requests/test_connection.py
============================================================================== test session starts ===============================================================================
platform darwin -- Python 3.7.9, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /Users/daniel/workspace/modeldb/client/verta/tests, configfile: pytest.ini
plugins: hypothesis-6.3.4, timeout-1.4.2
timeout: 300.0s
timeout method: signal
timeout func_only: False
collected 4 items

tests/http_requests/test_connection.py ....                                                                                                                                [100%]

================================================================================ warnings summary ================================================================================
http_requests/test_connection.py::TestConnection::test_no_headers
http_requests/test_connection.py::TestConnection::test_no_headers
http_requests/test_connection.py::TestConnection::test_no_auth
http_requests/test_connection.py::TestConnection::test_dev_key_auth
http_requests/test_connection.py::TestConnection::test_jwt_auth
  /Users/daniel/.pyenv/versions/verta-py3/lib/python3.7/site-packages/urllib3/util/retry.py:255: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
    DeprecationWarning,

-- Docs: https://docs.pytest.org/en/stable/warnings.html
========================================================================= 4 passed, 5 warnings in 2.91s ==========================================================================
```

### Manual Testing

*Checking for relevant environment variables*
```
(verta-py3) ➜  verta git:(df/vr-13352) ✗ env | grep VERTA
VERTA_HOST=df2.dev.verta.ai
VERTA_BEARER_TOKEN_SIG=XXXXX
VERTA_BEARER_TOKEN=XXXXXX
```

*Testing some client functionality*
```
Python 3.7.9 (default, May  1 2021, 12:41:31)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from verta import Client

In [2]: client = Client()
got VERTA_HOST from environment
got VERTA_BEARER_TOKEN from environment
got VERTA_BEARER_TOKEN_SIG from environment

In [3]: projects = list(client.projects)

In [4]: projects
Out[4]:
[name: Example Project 2
 url: https://df2.dev.verta.ai/Daniel_Fennelly/projects/c54728a1-bccc-49fa-a916-a82c0b2bc84a/summary,
 name: Example Project 1
 url: https://df2.dev.verta.ai/Daniel_Fennelly/projects/004a6d87-f09f-4dcb-a6a0-2dcf63456960/summary]
```

## How to Revert

- Revert this PR

### DIFF
--- a/client/verta/tests/http_requests/test_connection.py
+++ b/client/verta/tests/http_requests/test_connection.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+
+from verta._internal_utils import credentials
+from verta._internal_utils._utils import Connection, _GRPC_PREFIX
+
+
+https_scheme = "https"
+fake_socket = "test:8080"
+custom_headers = {"custom-header": "custom-header-value"}
+
+
+def assert_dictionary_is_subset(subset_dict, test_dict):
+    for key, value in subset_dict.items():
+        assert key in test_dict
+        assert test_dict.get(key) == value
+
+
+class TestConnection:
+    def test_no_headers(self):
+        conn = Connection(scheme=https_scheme, socket=fake_socket, credentials=None)
+        assert isinstance(conn.headers, dict)
+        conn.headers = None
+        assert isinstance(conn.headers, dict)
+        assert conn.headers.get(_GRPC_PREFIX + "scheme") == https_scheme
+
+    def test_no_auth(self):
+        conn = Connection(
+            scheme=https_scheme,
+            socket=fake_socket,
+            credentials=None,
+            headers=custom_headers,
+        )
+        assert conn.credentials is None
+        conn_headers = conn.headers
+        assert_dictionary_is_subset(custom_headers, conn_headers)
+
+    def test_dev_key_auth(self):
+        fake_email = "test@verta.ai"
+        fake_dev_key = "1234"
+        email_credentials = credentials.build(email=fake_email, dev_key=fake_dev_key)
+        conn = Connection(
+            scheme=https_scheme,
+            socket=fake_socket,
+            credentials=email_credentials,
+            headers=custom_headers,
+        )
+        conn_headers = conn.headers
+        expected = {
+            _GRPC_PREFIX + "scheme": https_scheme,
+            _GRPC_PREFIX + "source": "PythonClient",
+            _GRPC_PREFIX + "email": fake_email,
+            _GRPC_PREFIX + "developer_key": fake_dev_key,
+            _GRPC_PREFIX + "developer-key": fake_dev_key,
+        }
+        assert_dictionary_is_subset(custom_headers, conn_headers)
+        assert_dictionary_is_subset(expected, conn_headers)
+
+    def test_jwt_auth(self):
+        fake_token = "token"
+        fake_token_sig = "token_sig"
+        jwt_credentials = credentials.build(
+            jwt_token=fake_token, jwt_token_sig=fake_token_sig
+        )
+        conn = Connection(
+            scheme=https_scheme,
+            socket=fake_socket,
+            credentials=jwt_credentials,
+            headers=custom_headers,
+        )
+        conn_headers = conn.headers
+        expected = {
+            _GRPC_PREFIX + "scheme": https_scheme,
+            _GRPC_PREFIX + "source": "JWT",
+            _GRPC_PREFIX + "bearer_access_token": fake_token,
+            _GRPC_PREFIX + "bearer_access_token_sig": fake_token_sig,
+            _GRPC_PREFIX + "bearer-access-token": fake_token,
+            _GRPC_PREFIX + "bearer-access-token-sig": fake_token_sig,
+        }
+        assert_dictionary_is_subset(custom_headers, conn_headers)
+        assert_dictionary_is_subset(expected, conn_headers)
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+#pylint: disable=E1101
 
 import datetime
 import glob
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -112,7 +112,10 @@ class Connection:
 
     @headers.setter
     def headers(self, value):
-        self._headers = value
+        if value:
+            self._headers = value
+        else:
+            self._headers = dict()
         self._recompute_headers()
 
     def _init_headers(self):
@@ -121,7 +124,8 @@ class Connection:
         self._computed_headers = {}
 
     def _recompute_headers(self):
-        headers = self._headers.copy()
+        headers = self._headers or dict()
+        headers = headers.copy()
         headers[_GRPC_PREFIX+'scheme'] = self.scheme
         headers.update(self._auth_headers)
         self._computed_headers = headers
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -18,6 +18,7 @@ import warnings
 
 import click
 import requests
+from requests import codes as http_codes
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
@@ -43,9 +44,6 @@ _VALID_FLAT_KEY_CHARS = set(string.ascii_letters + string.digits + '_-/')
 
 THREAD_LOCALS = threading.local()
 THREAD_LOCALS.active_experiment_run = None
-
-# TODO: remove this in favor of _config_utils when #635 is merged
-HOME_VERTA_DIR = os.path.expanduser(os.path.join('~', ".verta"))
 
 
 class Connection:
@@ -82,7 +80,7 @@ class Connection:
         self.retry = Retry(total=max_retries,
                            backoff_factor=1,  # each retry waits (2**retry_num) seconds
                            method_whitelist=False,  # retry on all HTTP methods
-                           status_forcelist=(502, 503, 504),  # only retry on these status codes
+                           status_forcelist=(http_codes.bad_gateway, http_codes.unavailable, http_codes.gateway_timeout),  # only retry on these status codes
                            raise_on_redirect=False,  # return Response instead of raising after max retries
                            raise_on_status=False)  # return Response instead of raising after max retries
         self.ignore_conn_err = ignore_conn_err
@@ -161,7 +159,7 @@ class Connection:
             err.args = ("connection failed; please check `host` and `port`; error message: \n\n{}".format(err.args[0]),) + err.args[1:]
             six.raise_from(err, None)
 
-        if response.status_code == requests.codes.unauthorized:
+        if response.status_code == http_codes.unauthorized:
             try:
                 response.raise_for_status()
             except requests.HTTPError as e:
@@ -190,8 +188,8 @@ class Connection:
             response_msg = json_to_proto(body_to_json(response), response_type)
             return response_msg
         else:
-            if ((response.status_code == 403 and body_to_json(response)['code'] == 7)
-                    or (response.status_code == 404 and     body_to_json(response)['code'] == 5)):
+            if ((response.status_code == http_codes.forbidden and body_to_json(response)['code'] == 7)
+                    or (response.status_code == http_codes.not_found and body_to_json(response)['code'] == 5)):
                 return NoneProtoResponse()
             else:
                 raise_for_http_error(response)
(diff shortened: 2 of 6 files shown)


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+#pylint: disable=E1101
```
convoliution: What's this for, in particular? Considering how much this module has accumulated in it, I'm concerned about blanket-disabling a linting error.

client/verta/verta/dataset/_s3.py
```diff
@@ -196,7 +199,7 @@ def _prepare_components_to_upload(self):
             s3_loc = S3Location(component.path, component.s3_version_id)
 
             # download to file in ~/.verta/temp/
-            tempdir = os.path.join(_utils.HOME_VERTA_DIR, "temp")
+            tempdir = os.path.join(_config_utils.HOME_VERTA_DIR, "temp")
```
convoliution, line 202: 🙇🏼

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -117,52 +191,17 @@ def must_proto_response(response, response_type):
...
-        if request.body:
-            lines.append("-d '{}'".format(request.body.decode()))
-
-        curl = " \\\n    ".join(lines)
-        print(curl)
-
+    # NB: Maybe replace use of this with a "not is_json_response" method?
     @staticmethod
     def is_html_response(response):
         return response.text.strip().endswith("</html>")
```
convoliution: This is specifically used for checking whether the response is a web app page (which can happen in our OSS platform if the endpoint doesn't exist). I think if we changed code to check for non-JSON instead, it would get different results (in particular for empty contents and raw text).

client/verta/verta/client.py
```diff
@@ -119,49 +125,31 @@ class Client(object):
 
     """
     def __init__(self, host=None, port=None, email=None, dev_key=None,
-                 max_retries=5, ignore_conn_err=False, use_git=True, debug=False, extra_auth_headers={}, _connect=True):
+                 max_retries=5, ignore_conn_err=False, use_git=True, debug=False, extra_auth_headers={}, _connect=True, jwt_token=None, jwt_token_sig=None):
```
convoliution: I think new public parameters should go before our private parameters (plus I highly doubt we're passing `_connect` positionally), though my view might be incorrect!
```suggestion
                 max_retries=5, ignore_conn_err=False, use_git=True, debug=False, extra_auth_headers={}, jwt_token=None, jwt_token_sig=None, _connect=True):
```

client/verta/verta/client.py
```diff
@@ -271,6 +237,23 @@ def _set_from_config_if_none(self, var, resource_name):
                 print("setting {} from config file".format(resource_name))
         return var or None
 
+
+    def _get_with_fallback(self, parameter, env_var=None, config_var=None):
```
convoliution, line 241: 💯

client/verta/verta/client.py
```diff
@@ -119,49 +125,31 @@ class Client(object):
...
-                # without underscore, for NGINX support
-                # https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls#missing-disappearing-http-headers
-                _utils._GRPC_PREFIX+'developer-key': dev_key,
-            })
+                print("[DEBUG] using credentials: {}".format(repr(self.auth_credentials)))
             # save credentials to env for other Verta Client features
-            os.environ['VERTA_EMAIL'] = email
-            os.environ['VERTA_DEV_KEY'] = dev_key
-        else:
-            raise ValueError("`email` and `dev_key` must be provided together")
```
convoliution, line 163: Is this being kept anywhere? I think it's helpful to check for the user that both are provided together, but I'm fine if it's intentionally being removed.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -27,6 +29,11 @@
 from ..external import six
 from ..external.six.moves.urllib.parse import urljoin  # pylint: disable=import-error, no-name-in-module
 
+from .credentials import (
+    EmailCredentials,
+    JWTCredentials,
+)
```
convoliution, line 33: With full awareness that our `import` ordering is kind of a mess everywhere, I'd put this a few lines below, next to `from . import importer`. But we could clean this all up in the future™.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -68,19 +72,89 @@ def __init__(self, scheme=None, socket=None, auth=None, max_retries=0, ignore_co
             on HTTP codes {502, 503, 504} which commonly occur during back end connection lapses.
         ignore_conn_err : bool, default False
             Whether to ignore connection errors and instead return successes with empty contents.
-
+        credentials : :class:`~verta._internal_utils.credentials.EmailCredentials` or :class:`~verta._internal_utils.credentials.JWTCredentials`, optional
+            Either dev key or JWT token data to be used for authentication
```
convoliution:
```suggestion
            Either dev key or JWT token data to be used for authentication.
```

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -68,19 +72,89 @@ def __init__(self, scheme=None, socket=None, auth=None, max_retries=0, ignore_co
...
+        headers: dict, optional
+            Additional headers to attach to requests
```
convoliution: [The styleguide](https://numpydoc.readthedocs.io/en/latest/format.html) puts a newline before the closing `"""`.
```suggestion
            Additional headers to attach to requests.

```

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -68,19 +72,89 @@ def __init__(self, scheme=None, socket=None, auth=None, max_retries=0, ignore_co
...
+    def _init_headers(self):
+        self._headers = {}
+        self._auth_headers = {}
+        self._computed_headers = {}
+
+    def _recompute_headers(self):
+        headers = self._headers.copy()
+        headers[_GRPC_PREFIX+'scheme'] = self.scheme
+        headers.update(self._auth_headers)
+        self._computed_headers = headers
```
convoliution: How would you feel about recomputing headers when accessing `conn.headers` instead of when setting it? From the looks of this, I'm a bit worried about inconsistency having 3 co-dependent variables; I don't think we necessarily need `_computed_headers` to be stored as state.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -68,19 +72,89 @@ def __init__(self, scheme=None, socket=None, auth=None, max_retries=0, ignore_co
...
+
+
+    @property
+    def credentials(self):
+        return self._credentials
+
+    @credentials.setter
+    def credentials(self, value):
+        self._credentials = value
+        self._auth_headers = self._headers_for_credentials(value)
```
convoliution: I'd personally prefer having `_auth_headers` be a property that calculates the headers from `self._credentials` when accessed than tracking it as separate, duplicated-ish state.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -68,19 +72,89 @@ def __init__(self, scheme=None, socket=None, auth=None, max_retries=0, ignore_co
...
+        self._auth_headers = {}
+        self._computed_headers = {}
+
+    def _recompute_headers(self):
+        headers = self._headers.copy()
+        headers[_GRPC_PREFIX+'scheme'] = self.scheme
+        headers.update(self._auth_headers)
+        self._computed_headers = headers
+
+    def _headers_for_credentials(self, credentials):
```
convoliution: I found this name a bit confusing (I couldn't figure out what the method is for without looking at its implementation). Mild suggestion, but not too attached to this:
```suggestion
    def _auth_headers_from_credentials(self, credentials):
```

convoliution (changes requested):

client/verta/verta/_internal_utils/_utils.py
daniel-verta: Maybe there's a way we can check the content-type on the response instead of looking for an `</html>` tag?

convoliution: Oooh good idea. Way better.

client/verta/verta/_internal_utils/_utils.py
daniel-verta: This is for use of `requests.codes.not_found` and similar. I think the Requests library probably does some `__get_attr__` trickery to enable such attributes, and pylint thinks that non-existent attributes are getting referenced. I rather prefer `http_codes.not_found` over `404` for readability. A compromise may be to use `http_codes['not_found']`, although the attribute interface feels nice and clean minus the pylint issue.

convoliution: >I think the Requests library probably does some `__get_attr__` trickery to enable such attributes

Ah, yep, [looks like it](https://github.com/psf/requests/blob/main/requests/status_codes.py#L110).

I also like the attribute access, but I'd strongly oppose disabling `no-member` checks for an entire module! We might be able to add something to our [.pylintrc's ignore section](https://github.com/VertaAI/modeldb/blob/master/client/verta/.pylintrc#L274-L288).

convoliution: 😬 https://github.com/psf/requests/issues/2417#issuecomment-353740674

client/verta/verta/_internal_utils/_utils.py
daniel-verta: Renaming the helper to `_prefixed_headers_for_credentials` for clarity

client/verta/verta/client.py
daniel-verta: Adding it back into `verta._internal_utils.credentials.build`

client/verta/verta/_internal_utils/_utils.py
daniel-verta: Good point. Removed `_auth_headers`.

client/verta/verta/_internal_utils/_utils.py
daniel-verta: Resolving this per our Slack conversation about writing unit tests and moving on.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -68,19 +70,111 @@ def __init__(self, scheme=None, socket=None, auth=None, max_retries=0, ignore_co
...
+    @property
+    def auth(self):
+        return self.headers
+
+    @headers.setter
+    def headers(self, value):
+        if value:
+            self._headers = value
+        else:
+            self._headers = dict()
```
convoliution:
```suggestion
        self._headers = value or dict()
```

client/verta/tests/http_requests/test_connection.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+
+from verta._internal_utils import credentials
+from verta._internal_utils._utils import Connection, _GRPC_PREFIX
+
+
+https_scheme = "https"
+fake_socket = "test:8080"
+custom_headers = {"custom-header": "custom-header-value"}
```
convoliution, line 9: Just out of curiosity, was there a reason for making these lowercase rather than uppercase as constants?

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -68,19 +70,111 @@ def __init__(self, scheme=None, socket=None, auth=None, max_retries=0, ignore_co
...
+        self._computed_headers = {}
+
+    def _recompute_headers(self):
+        headers = self._headers or dict()
+        headers = headers.copy()
+        headers[_GRPC_PREFIX+'scheme'] = self.scheme
+        headers.update(self._prefixed_headers_for_credentials(self.credentials))
+        self._computed_headers = headers
+
+    def _prefixed_headers_for_credentials(self, credentials):
```
convoliution: No strong preference on this one.
```suggestion
    @staticmethod
    def _prefixed_headers_for_credentials(credentials):
```

convoliution (approved):
Fantastic!
May this code go on to do great things 🚀

PR #2724, opened 2021-11-23
[VR-13520] Added filter in find call to not fetch ER for deleted project and experiment

anandjakhaniya:

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
+++ b/backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
@@ -770,7 +770,13 @@ public class FutureExperimentRunDAO {
         InternalFuture.supplyAsync(
             () -> {
               final var localQueryContext = new QueryFilterContext();
-              localQueryContext.getConditions().add("deleted = :deleted");
+              localQueryContext.getConditions().add("experiment_run.deleted = :deleted");
+              localQueryContext.getBinds().add(q -> q.bind("deleted", false));
+
+              localQueryContext.getConditions().add("p.deleted = :deleted");
+              localQueryContext.getBinds().add(q -> q.bind("deleted", false));
+
+              localQueryContext.getConditions().add("e.deleted = :deleted");
               localQueryContext.getBinds().add(q -> q.bind("deleted", false));
 
               if (!request.getProjectId().isEmpty()) {
@@ -839,7 +845,12 @@ public class FutureExperimentRunDAO {
                           return jdbi.withHandle(
                                   handle -> {
                                     var sql =
-                                        "select id, date_created, date_updated, experiment_id, name, project_id, description, start_time, end_time, owner, environment, code_version, job_id, version_number from experiment_run";
+                                        "select experiment_run.id, experiment_run.date_created, experiment_run.date_updated, experiment_run.experiment_id, experiment_run.name, experiment_run.project_id, experiment_run.description, experiment_run.start_time, experiment_run.end_time, experiment_run.owner, experiment_run.environment, experiment_run.code_version, experiment_run.job_id, experiment_run.version_number from experiment_run";
+
+                                    sql +=
+                                        " left join project p ON p.id = experiment_run.project_id ";
+                                    sql +=
+                                        " left join experiment e ON e.id = experiment_run.experiment_id ";
 
                                     // Add the sorting tables
                                     for (final var item :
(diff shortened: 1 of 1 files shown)


backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -770,7 +770,13 @@ public FutureExperimentRunDAO(
...
               final var localQueryContext = new QueryFilterContext();
-              localQueryContext.getConditions().add("deleted = :deleted");
+              localQueryContext.getConditions().add("experiment_run.deleted = :deleted");
+              localQueryContext.getBinds().add(q -> q.bind("deleted", false));
+
+              localQueryContext.getConditions().add("p.deleted = :deleted");
+              localQueryContext.getBinds().add(q -> q.bind("deleted", false));
+
+              localQueryContext.getConditions().add("e.deleted = :deleted");
               localQueryContext.getBinds().add(q -> q.bind("deleted", false));
```
conradoverta: You cannot bind to the same `deleted` name.

anandjakhaniya: Yahh, we can bind because it will replace `deleted` value. but good point i will take it a single entry. updating PR

anandjakhaniya: Fixed

conradoverta (changes requested):
Is there a ticket associated with this?

anandjakhaniya:
> Is there a ticket associated with this?

No it was not but i have created it just now : https://vertaai.atlassian.net/browse/VR-13520

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -839,7 +841,12 @@ public FutureExperimentRunDAO(
                           return jdbi.withHandle(
                                   handle -> {
                                     var sql =
-                                        "select id, date_created, date_updated, experiment_id, name, project_id, description, start_time, end_time, owner, environment, code_version, job_id, version_number from experiment_run";
+                                        "select experiment_run.id, experiment_run.date_created, experiment_run.date_updated, experiment_run.experiment_id, experiment_run.name, experiment_run.project_id, experiment_run.description, experiment_run.start_time, experiment_run.end_time, experiment_run.owner, experiment_run.environment, experiment_run.code_version, experiment_run.job_id, experiment_run.version_number from experiment_run";
+
+                                    sql +=
+                                        " left join project p ON p.id = experiment_run.project_id ";
+                                    sql +=
+                                        " left join experiment e ON e.id = experiment_run.experiment_id ";
```
conradoverta: I don't believe we need left joins, just inner joins, right?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
```diff
@@ -1224,7 +1231,12 @@ public FutureExperimentRunDAO(
                         queryContext ->
                             jdbi.withHandle(
                                 handle -> {
-                                  var sql = "select count(id) from experiment_run";
+                                  var sql = "select count(experiment_run.id) from experiment_run";
+
+                                  sql +=
+                                      " left join project p ON p.id = experiment_run.project_id ";
+                                  sql +=
+                                      " left join experiment e ON e.id = experiment_run.experiment_id ";
```
conradoverta: I don't believe we need left joins, just inner joins, right?

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: We do not want any data from right table then why inner join is required? we need just left table data with right table condition that's why i used left join. is it not current?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: We do not want any data from right table then why inner join is required? we need just left table data with right table condition that's why i used left join. is it not current?

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
conradoverta: Left joins allow the match _not_ to exist on the right table. In which case the entry on the left table will still be present in the result. So left joins _do not_ reduce the number of rows on the result, if I remember correctly. If we don't use the data on the right table, then left joins are effectively not doing anything besides wasting time.

If you just want to filter things out, use inner joins.

anandjakhaniya: Fix this with inner join

backend/src/main/java/ai/verta/modeldb/experimentRun/FutureExperimentRunDAO.java
anandjakhaniya: Fix this with inner join

conradoverta (approved):

PR #2727, opened 2021-11-30
[VR-13511] Add support for credentials to DeployedModel

daniel-verta:
## Impact and Context

- Adds user-based authentication to `DeployedModel.predict` requests
- Users can supply credentials to the `DeployedModel.from_url` method with the `creds=` parameter
- If credentials are not supplied for a `DeployedModel`, but credentials exist in a user's environment variables, those credentials will be used. 

## Risks and Area of Effect

- There's a breaking API change in the initialization of a `DeployedModel`, which can no longer be constructed with a `run_id` but must be constructed with a `prediction_url` following the use-case set out in `from_url`.

### Manual Testing

Note: the below testing was done against a deployed model which had its permissions manually updated due to a gap in the product for fully enabling prediction permissions.

Testing with dev-key credentials
```
(verta-py3) ➜  verta git:(df/vr-13511) env | grep VERTA
VERTA_EMAIL=daniel@verta.ai
VERTA_DEV_KEY=0f95c63c-XXXXX
VERTA_HOST=dev.verta.ai
(verta-py3) ➜  verta git:(df/vr-13511) ipython
Python 3.7.9 (default, May  1 2021, 12:41:31)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from verta.deployment import DeployedModel

In [2]: url = "https://dev.verta.ai/api/v1/predict/conrado-echo"

In [3]: model = DeployedModel.from_url(url)

In [4]: model.predict("foo")
Out[4]: 'foo'
```

Testing with JWT credentials
```
(verta-py3) ➜  verta git:(df/vr-13511) env | grep VERTA
VERTA_HOST=dev.verta.ai
VERTA_JWT_TOKEN=eyJhbG...
VERTA_JWT_TOKEN_SIG=11cb01cc...
(verta-py3) ➜  verta git:(df/vr-13511) ipython
Python 3.7.9 (default, May  1 2021, 12:41:31)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from verta.deployment import DeployedModel

In [2]: url = "https://dev.verta.ai/api/v1/predict/conrado-echo"

In [3]: model = DeployedModel.from_url(url)

In [4]: model.predict("foo")
Out[4]: 'foo'
```

### Unit Testing
- Added `tests/test_deployment/test_unit_deployedmodel.py`

## How to Revert

- Revert this PR


client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -116,6 +124,8 @@ def from_url(cls, url, token):
         token : str or None
             Prediction token. Can be copy and pasted directly from the Verta Web App. If the deployment
             does not use a token, ``None`` should be passed in as the argument.
```
convoliution: `token` originally wasn't optional per se; rather, `None` was meant to be passed explicitly for endpoints that have token-based permissions disabled. But I'm realizing that making this optional might be functionally equivalent anyway 🤔 
```suggestion
        token : str, optional
            Prediction token. Can be copy and pasted directly from the Verta Web App.
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -116,6 +124,8 @@ def from_url(cls, url, token):
...
+        creds : :class:`~verta._internal_utils.credentials.Credentials, optional
```
convoliution:
```suggestion
        creds : :class:`~verta._internal_utils.credentials.Credentials`, optional
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -135,33 +145,39 @@ def from_url(cls, url, token):
...
+        response = self._session.get(status_url)
         _utils.raise_for_http_error(response)
         status = _utils.body_to_json(response)
         if status['status'] == 'error':
             raise RuntimeError(status['message'])
         elif status['status'] != 'deployed':
             raise RuntimeError("model is not yet ready, or has not yet been deployed")
         else:
-            self._session.headers['Access-Token'] = status.get('token')
+            access_token = AccessToken(status['token'])
```
convoliution: I _believe_ that our platform still allows endpoints without tokens, so it might not be in the status.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -54,47 +57,52 @@ class DeployedModel:
         # <DeployedModel 01234567-0123-0123-0123-012345678901>
 
     """
-    def __init__(self, _host=None, _run_id=None, _from_url=False, **kwargs):
-        # this is to temporarily maintain compatibility with anyone passing in `socket` and `model_id` as kwargs
-        # TODO: instate `host` and `run_id` params
-        # TODO: remove the following block of param checks
+    def __init__(self, host=None, run_id=None, creds=None, access_token=None, prediction_url=None, **kwargs):
```
convoliution: Since `from_url()` has its params ordered `token` then `creds`, would you mind if we put `access_token` and `creds` in the same order here?

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -54,47 +57,52 @@ class DeployedModel:
...
-                })
-            except KeyError:
-                six.raise_from(EnvironmentError("${} not found in environment".format('VERTA_DEV_KEY')), None)
+        creds = creds or credentials.load_from_os_env()
+        self._session = self._make_session(creds, access_token)
+
+        # TODO: remove the following block of param checks
+        # this is to temporarily maintain compatibility with anyone passing in `socket` and `model_id` as kwargs
+        self._host = self._validate_host(host, kwargs.get('socket'))
+        self._id = self._validate_id(run_id, kwargs.get('model_id'))
```
convoliution: (as mentioned in our earlier conversation) I'd be alright with ripping this all out and supporting the `from_url()` flow. It's unlikely that users are still instantiating this using a run/model ID.

client/verta/verta/_internal_utils/access_token.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+class AccessToken:
```
convoliution, line 3: Do you think this might be generalizable to other parts of our platform? Otherwise I'd be temped to define this class in `verta.endpoint` (or `verta.deployment`, though I'd like to merge it into `verta.endpoint` soon™).

client/verta/tests/env_fixtures.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from verta._internal_utils.credentials import EmailCredentials, JWTCredentials
+
+
+@pytest.fixture
+def mock_env_dev_key_auth(monkeypatch):
```
convoliution, line 9: TIL about the `monkeypatch` fixture! Neat.

convoliution (changes requested):
As mentioned, if it's not too out-of-the-way and/or if it helps with testing/implementation, feel free to remove the `run_id` instantiation.

Thank you for the cleanup and testing 🙇🏼

client/verta/verta/_internal_utils/access_token.py
daniel-verta: I actually put it in `verta.endpoint` originally, but some import circularity forced me to move it out of there. If we could consolidate `DeployedModel` into `verta.endpoint` that might get around the issue.

client/verta/verta/deployment/_deployedmodel.py
daniel-verta: This actually leads to a substantial simplification of this class which I'm very in favor of.

client/verta/tests/env_fixtures.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from verta._internal_utils.credentials import EmailCredentials, JWTCredentials
+from verta.credentials import EmailCredentials, JWTCredentials
```
convoliution, line 5: It looks like there are three different styles being used for this import in different files. We might want to pick one if `black` doesn't do it for us!
```python
from verta.credentials import EmailCredentials, JWTCredentials
```
```python
from verta.credentials import (EmailCredentials, JWTCredentials)
```
```python
from verta.credentials import (
    EmailCredentials,
    JWTCredentials,
)
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -57,60 +55,25 @@ class DeployedModel:
...
-        back_end_url = urlparse(host)
-        self._socket = back_end_url.netloc + back_end_url.path.rstrip('/')
-        self._scheme = back_end_url.scheme or ("https" if ".verta.ai" in self._socket else "http")
+        self._credentials = creds
+        self._access_token = token
+        self._session = self._make_session()
 
-
-    def _make_session(self, credentials, access_token):
+    def _make_session(self):
```
convoliution: Would you know if we should call `close()` on the prior session if there is one?

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -57,60 +55,25 @@ class DeployedModel:
...
         session = requests.Session()
-        if credentials:
-            creds_headers = _utils.Connection.prefixed_headers_for_credentials(credentials) # TODO: extract this?
+        if self.credentials:
+            creds_headers = _utils.Connection.prefixed_headers_for_credentials(self.credentials) # TODO: extract this?
```
convoliution: What do you have in mind for the `# TODO: extract this`?

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -54,68 +55,39 @@ class DeployedModel:
...
-        self._socket = back_end_url.netloc + back_end_url.path.rstrip('/')
-        self._scheme = back_end_url.scheme or ("https" if ".verta.ai" in self._socket else "http")
-
-        self._id = run_id
-        self._status_url = "{}://{}/api/v1/deployment/status/{}".format(self._scheme, self._socket, self._id)
-
-        self._prediction_url = None
+        self.prediction_url = prediction_url
+        creds = creds or credentials.load_from_os_env()
+        self._credentials = creds
```
convoliution: These could probably be combined.
```suggestion
        self._credentials = creds or credentials.load_from_os_env()
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -54,68 +55,39 @@ class DeployedModel:
         # <DeployedModel 01234567-0123-0123-0123-012345678901>
 
     """
-    def __init__(self, _host=None, _run_id=None, _from_url=False, **kwargs):
-        # this is to temporarily maintain compatibility with anyone passing in `socket` and `model_id` as kwargs
-        # TODO: instate `host` and `run_id` params
-        # TODO: remove the following block of param checks
+    def __init__(self, prediction_url, token=None, creds=None):
         # TODO: put automodule verta.deployment back on ReadTheDocs
```
convoliution: If we could remove this line as well, since `verta.deployment` is being rendered in our docs now 🙇🏼 
```suggestion
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -54,68 +55,39 @@ class DeployedModel:
...
-            return "<{} {}>".format(self.__class__.__name__, self._id)
-        elif self._prediction_url:
-            return "<{} at {}>".format(self.__class__.__name__, self._prediction_url)
-        else:  # if someone's messing with the object's state
-            return "<{}>".format(self.__class__.__name__)
+        return "<{} at {}>".format(self.__class__.__name__, self.prediction_url)
 
     @classmethod
-    def from_url(cls, url, token):
+    def from_url(cls, url, token=None, creds=None):
```
convoliution, line 86: Seeing this now, we could probably get rid of it and just use `__init__()` everywhere 😆 but that can wait for a future cleanup

convoliution (approved):
Fantastic 🙇🏼 A few comments about formatting / future clean ups.

I believe we'll also need to add the `credentials` submodule to [docs/python.rst](https://github.com/VertaAI/modeldb/blob/master/client/verta/docs/python.rst) to render it into the docs

client/verta/verta/deployment/_deployedmodel.py
daniel-verta: I guess it seems weird to reach into a static method of `Connection` here. I'll just remove the comment though, as I don't have a particularly strong better idea.

client/verta/verta/deployment/_deployedmodel.py
daniel-verta: I'm not sure but I've changed `_make_session` to `_init_session` to handle closing any previously existing session.

PR #2728, opened 2021-11-30
Common jdbi util for backend which used for common functionality of jdbi impl after removing hibernate util

anandjakhaniya:
## Impact and Context

All functions are extracted from CommonHibernateUtils and it is the same one, there is not have any changes


conradoverta:
>All functions are extracted from CommonHibernateUtils and it is the same one, there is not have any changes

Then let's move instead of copying >300 LOC

anandjakhaniya:
> > All functions are extracted from CommonHibernateUtils and it is the same one, there is not have any changes
> 
> Then let's move instead of copying >300 LOC

that is the new file so all moved code also show as a new lines and show removed code in CommonHibernateUtils. I did move code from `CommonHibernateUtils` to `JdbiUtil` i hope it is same like you are expecting.

conradoverta:
> > > All functions are extracted from CommonHibernateUtils and it is the same one, there is not have any changes
> > 
> > 
> > Then let's move instead of copying >300 LOC
> 
> that is the new file so all moved code also show as a new lines and show removed code in CommonHibernateUtils. I did move code from `CommonHibernateUtils` to `JdbiUtil` i hope it is same like you are expecting.

The PR had no deletions when I commented =)

backend/common/src/main/java/ai/verta/modeldb/common/CommonJdbiUtil.java
```diff
@@ -0,0 +1,324 @@
...
+import liquibase.database.jvm.JdbcConnection;
+import liquibase.exception.DatabaseException;
+import liquibase.exception.LiquibaseException;
+import liquibase.exception.LockException;
+import liquibase.lockservice.LockServiceFactory;
+import liquibase.resource.FileSystemResourceAccessor;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public abstract class CommonJdbiUtil {
```
conradoverta:
```suggestion
public abstract class CommonDBUtil {
```
This doesn't seem specific to the database method (jdbi vs hibernate), right?

conradoverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/CommonJdbiUtil.java
anandjakhaniya: Renamed

PR #2734, opened 2021-12-03
Moved tag handler to common module

anandjakhaniya:


backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
```diff
@@ -42,6 +43,23 @@ public static String appendOptionalTelepresencePath(String filePath) {
...
+        String errorMessage =
+            "Tag name can not be more than "
+                + CommonConstants.TAG_LENGTH
+                + " characters. Limit exceeded tag is: "
+                + tag;
+        throw new ModelDBException(errorMessage, Code.INVALID_ARGUMENT);
+      }
+    }
+    return tags;
+  }
```
conradoverta: Let's move this to the tags handler since it seems tag specific. It's better to consolidate logic in groups vs spread around.

conradoverta (approved):
Neat! One small ask.

backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
anandjakhaniya: Moved to common tag handler

PR #2754, opened 2021-12-08
VR-13528 transaction in FutureJDBI

evgeniilezhnin:

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureJdbi.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureJdbi.java
@@ -17,12 +17,12 @@ public class FutureJdbi {
   }
 
   @FunctionalInterface
-  public interface SupplierWithException<R, T extends Exception> {
+  private interface SupplierWithException<R, T extends Exception> {
     R get() throws T;
   }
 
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureJdbi.java
```diff
@@ -16,15 +16,36 @@ public FutureJdbi(Jdbi jdbi, Executor executor) {
...
 
+  @FunctionalInterface
+  public interface SupplierWithException<R, T extends Exception> {
+    R get() throws T;
+  }
+
+  @FunctionalInterface
+  public interface RunnableWithException<T extends Exception> {
+    void run() throws T;
+  }
```
conradoverta: These don't have to be public, right? Them seem like internal implementation details that are never actually exposed to the user.

conradoverta (approved):
One suggestion to reduce the public exposure

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureJdbi.java
evgeniilezhnin: yes, changed to private

PR #2757, opened 2021-12-09
Fix handlers to use different class as a key in mapsubtype class

anandjakhaniya:


backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -80,6 +80,7 @@ protected void releaseLiquibaseLock(DatabaseConfig config)
       if (!existsStatus) {
         LOGGER.info("Table database_change_log_lock does not exists in DB");
         LOGGER.info("Proceeding with liquibase assuming it has never been run");
+        return;
```
conradoverta, line 83: This seems unrelated, no?

conradoverta (approved):
Nice

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
anandjakhaniya: Yes but since it is just one line and related to jdbi changes so i thought it will works here. :)

PR #2766, opened 2021-12-14
Move InvalidArgumentException to common module

anandjakhaniya:


conradoverta (approved):
We should probably move all exceptions for simplicity.

anandjakhaniya:
> We should probably move all exceptions for simplicity.

Sure i will do it. Ticket: https://vertaai.atlassian.net/browse/VR-13580

PR #2799, opened 2022-01-10
test: [VRD-53] Add tests for collecting installed custom modules

liuverta:
## Impact and Context

This PR adds test coverage for custom modules' ability to collect `pip`-installed modules by name.

The last two tests validate a scenario when the module shares a name with a local directory. These tests will fail until #2805 is merged.

## Risks and Area of Effect

Low risk: Most of this PR is new test code that won't break old things and is validated through testing.
- The Client's new `CustomModules` util class is a little bit risky as a new-for-us approach to module discovery, but it's validated in testing.
- The change to the Client's `_custom_modules_as_artifact` won't break anything (it's still called and used the same way)

Low AoE: Only impacts custom modules functionality and related tests.

## Testing

See #2805.

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/custom_modules/contexts.py
+++ b/client/verta/tests/custom_modules/contexts.py
@@ -87,7 +87,8 @@ def installable_package(name, dir=None):
         shutil.rmtree(pkg_dir)
 
 
-# TODO: try to get this to work in a virtual environment
+# TODO: try to create and install into a virtual environment
+#       pytest-venv looks good, but need to be able to import from it in the test
 @contextlib.contextmanager
 def installed_local_package(pkg_dir, name):
     """pip install a locally-defined package.
--- a/client/verta/tests/custom_modules/contexts.py
+++ b/client/verta/tests/custom_modules/contexts.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 
+import ast
 import contextlib
 import os
 import shutil
 import subprocess
 import sys
 import textwrap
+
+from verta._internal_utils.custom_modules import CustomModules
 
 
 @contextlib.contextmanager
@@ -103,6 +106,7 @@ def installed_local_package(pkg_dir, name):
 
     """
     subprocess.check_call([sys.executable, "-m", "pip", "install", pkg_dir])
+    assert CustomModules.is_importable(name)  # verify installation
 
     try:
         yield
--- a/client/verta/tests/custom_modules/contexts.py
+++ b/client/verta/tests/custom_modules/contexts.py
@@ -1,0 +1,112 @@
+# -*- coding: utf-8 -*-
+
+import contextlib
+import os
+import shutil
+import subprocess
+import sys
+import textwrap
+
+
+@contextlib.contextmanager
+def installable_package(name, dir=None):
+    """Make a simple pip-installable package in the local filesystem.
+
+    Parameters
+    ----------
+    name : str
+        Name of the package.
+    dir : str, optional
+        Directory in which to create the package. If not provided, the
+        current directory will be used.
+
+    Yields
+    ------
+    str
+        Absolute path to the created package.
+
+    Warnings
+    --------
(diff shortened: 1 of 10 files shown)


client/verta/tests/deployable_entity/test_deployment.py
```diff
@@ -48,7 +48,7 @@ def test_custom_modules(self, deployable_entity, model_for_deployment):
         custom_modules_dir = "."
```
liuverta, line 48: This entire file's diff is just removing a duplicated test.

client/verta/tests/custom_modules/contexts.py
```diff
@@ -0,0 +1,117 @@
...
+        f.write(content)
+
+    try:
+        yield pkg_dir
+    finally:
+        shutil.rmtree(pkg_dir)
+
+
+# TODO: try to create and install into a virtual environment
+#       pytest-venv looks good, but need to be able to import from it in the test
```
liuverta: Unless I can get this to work, these tests are likely to conflict with each other when run in parallel.

client/verta/tests/custom_modules/test_custom_modules.py
```diff
@@ -0,0 +1,130 @@
...
+            if os.path.isfile(module):
+                assert filecmp.cmp(module, retrieved_module)
+            else:
+                utils.assert_dirs_match(module, retrieved_module)
+
+    @pytest.mark.parametrize(
+        "name",
+        sorted(module[1] for module in pkgutil.iter_modules()),
+    )
+    def test_module(self, name):
```
liuverta, line 49: This test has a number of cases that I'm skipping (see the `if`s just below); the failures are unrelated to this fix, and are outside this ticket's scope to resolve 😞

liuverta:
!rebuild

liuverta:
!rebuild

client/verta/tests/utils.py
```diff
@@ -14,6 +16,12 @@
 from hypothesis import strategies as st
 
 
+# Jenkins workers had been getting lost during artifact tests;
+# we added this env var to ensure files are written to disk
+# (and not, say, into RAM)
+TEMPDIR_ROOT = os.environ.get("TEMPDIR_ROOT")
```
ghost, line 22: Did you add this to the configuration setup?

liuverta: @hmacdonald-verta This is actually moved here from https://github.com/VertaAI/modeldb/pull/2799/files#diff-8d105224a574535507b29e892eb2171f2096c788362f47da0c4d31cf3a72b3c0L335-L337; it's already existed, so nothing needs to be added to our cluster config setup (if I'm understanding right)!

ghost: I see! Okay 👍🏼

ghost (approved):

client/verta/tests/custom_modules/test_custom_modules.py
```diff
@@ -0,0 +1,128 @@
...
+        ],
+    )
+    def test_multiple_modules(self, names):
+        """Multiple pip-installed modules can be collected at once."""
+        custom_modules = _DeployableEntity._custom_modules_as_artifact(names)
+        for name in names:
+            self.assert_in_custom_modules(custom_modules, name)
+
+    def test_module_and_local_dir_have_same_name(self, worker_id):
+        """If a pip-installed module and a local directory share a name, the module is collected."""
```
conradoverta, line 83: It would be useful to keep a note here of why we want this (vs using the local dir) as I bet we'll ask ourselves in the future.

conradoverta (commented):
Very neat composition of contexts and helpers. The tests themselves get so much clearer

client/verta/tests/custom_modules/test_custom_modules.py
liuverta: Done in #2811!

PR #2800, opened 2022-01-13
DeleteProject & DeleteProjects endpoints using jdbi

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
```diff
@@ -794,4 +796,113 @@ public FutureProjectDAO(
...
+              return allowedProjectIds;
+            },
+            executor)
+        .thenCompose(
+            allowedProjectIds -> {
+              return getResourceItemsForWorkspace(
+                  "",
+                  Optional.of(allowedProjectIds),
+                  Optional.empty(),
+                  ModelDBResourceEnum.ModelDBServiceResourceTypes.PROJECT);
```
conradoverta: Why do we need this? Don't we need just the IDs that we already have from `allowedProjectIds`?

anandjakhaniya: I need workspace ids for the `allowedProjectIds` so i have called again here and it will used while logging events.

conradoverta (approved):

PR #2810, opened 2022-01-21
test: Uncache module after deletion

liuverta:
## Impact and Context

Python 3 caches modules when they're imported, so even a `pip uninstall` doesn't clean it from the system. This PR augments the context manager to remove the module from the cache right after the `pip uninstall`.

This is specifically a problem for these tests because they use the `pytest-xdist` worker ID as the package name, so if the tests run sequentially on the same worker (e.g. when not run in parallel), the same package name gets re-used.

Also while I'm at it, I'm removing the `del worker_id`s because they were preventing pytest from introspecting their values in test failure reports.

## Risks and Area of Effect

Low risk: The `installed_local_package()` context isn't thread-safe to begin with, and removing a cached module is safer than all the pip things it does anyway.

Low AoE: Only affects the two tests that use this context manager.

## Testing

Ran `custom_modules/test_custom_modules.py` locally both with and without parallelism (our CI jobs currently can only run our tests in parallel).

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/custom_modules/test_custom_modules.py
+++ b/client/verta/tests/custom_modules/test_custom_modules.py
@@ -13,11 +13,11 @@ import six
 from verta.tracking.entities._deployable_entity import _DeployableEntity
 from verta._internal_utils.custom_modules import CustomModules
 
-from .. import strategies, utils
+from .. import utils
 from . import contexts
 
 
-class TestCollectPipInstalledModule:
+class TestPipInstalledModule:
     @staticmethod
     def assert_in_custom_modules(custom_modules, module_name):
         module = CustomModules.get_module_path(module_name)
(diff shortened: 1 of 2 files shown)


client/verta/tests/custom_modules/test_custom_modules.py
```diff
@@ -13,11 +13,11 @@
 from verta.tracking.entities._deployable_entity import _DeployableEntity
 from verta._internal_utils.custom_modules import CustomModules
 
-from .. import strategies, utils
+from .. import utils
 from . import contexts
 
 
-class TestCollectPipInstalledModule:
+class TestPipInstalledModule:
```
liuverta, line 20: Minor cleanup stuff 😶

conradoverta (approved):

PR #2815, opened 2022-01-31
Get Summary endpoint using jdbi

anandjakhaniya:


conradoverta (approved):
I think it looks correct, but I had a hard time with the indentation

PR #2818, opened 2022-02-01
set project short name using jdbi

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
```diff
@@ -1367,4 +1371,85 @@ public FutureProjectDAO(
...
+                                .findOne());
+              }
+              return countFuture.thenCompose(
+                  count -> {
+                    if (count.isPresent() && count.get() > 0) {
+                      return InternalFuture.failedStage(
+                          new AlreadyExistsException(
+                              "Project already exist with given short name"));
+                    }
+                    return InternalFuture.completedInternalFuture(null);
```
conradoverta, line 1414: I'm not sure this logic is correct. Let's say there's a workspace W and project names A and B. I only have access to A (e.g. I'm out of the org). Then if I try to rename the project to B, this check will succeed (I can't see that B exists), even though it's there.

Similarly, if I have projects named A in workspace W1 and B in workspace W2 and I have access to both, I should be able to rename B to A since they're in different workspaces.

conradoverta (changes requested):
See comment

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
anandjakhaniya: In the above both commented cases the above logic is works fine.

> I'm not sure this logic is correct. Let's say there's a workspace W and project names A and B. I only have access to A (e.g. I'm out of the org). Then if I try to rename the project to B, this check will succeed (I can't see that B exists), even though it's there.

we are checking read permission before count query like `checkProjectPermission(request.getId(), ModelDBActionEnum.ModelDBServiceActions.UPDATE),`  so if B is not accessible then it will throw Access Denined.

For second option:
We are fetching self allowed resources so we have both the workspace accessible ids and added it on above count query like: `.bindList("projectIds", selfAllowedResources)` 
```
.createQuery(
                                    "select count(p.id) from project p where p.deleted = false AND p.short_name = :projectShortName AND p.id IN (<projectIds>)")
                                .bind("projectShortName", request.getShortName())
                                .bindList("projectIds", selfAllowedResources)
                                .mapTo(Long.class)
                                .findOne());
```
so it will not allowed to rename B to A because both are in accessible ids.

conradoverta: My first example was to rename `A` to `B`. Since I don't have access to `B`, B's ID won't be inside `projectIds` and `countFuture` will be zero. Even though there _is_ a project `B`, I just don't happen to have access to it.

conradoverta (changes requested):
Could we add a test for the case I'm highlighting? I think it will be easier to talk and validate behavior if we're concrete about it.

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
anandjakhaniya: We are checking accessible workspace level existence. not for entire DB level uniqueness since `short_name` field introduce. we are always checking existence based on self allowed resources and not found that resource then we will allowed to add it. 
In Old code: https://github.com/VertaAI/modeldb/blob/daf936ab83230ea4565fd3f3e047e1d83a1b03b5/backend/src/main/java/ai/verta/modeldb/project/ProjectDAORdbImpl.java#L732

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
```diff
@@ -1357,4 +1359,85 @@ public FutureProjectDAO(
...
+
+  public InternalFuture<Void> setProjectShortName(SetProjectShortName request) {
+    // Request Parameter Validation
+    InternalFuture<Void> validateParamFuture =
+        InternalFuture.runAsync(
+            () -> {
+              if (request.getId().isEmpty() && request.getShortName().isEmpty()) {
+                throw new InvalidArgumentException(
+                    "Project ID and Project shortName not found in SetProjectShortName request");
+              } else if (request.getId().isEmpty()) {
```
liam-verta: This first check seems redundant. 

If we want to do this kind of response we should switch from immediately throwing an exception to a model where we have a validator that collects all violations and then throws a single exception listing all violations.

```suggestion
              if (request.getId().isEmpty()) {
```

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
liam-verta: So, the old code had the same bug? Is there a test covering this that shows this behavior and that it is unchanged? 

If so, I suggest we accept this PR and add a task to fix the bug.
If not, let's add a test for the existing behavior and then do the above.

liam-verta (changes requested):

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
anandjakhaniya: the old code had the same bug? : YES
Is there a test covering this that shows this behaviour and that it is unchanged? : Currently it is not but when we fix this i will add test for the same scenario.

anandjakhaniya: Created Jira ticket: https://vertaai.atlassian.net/browse/VUMM-202

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
anandjakhaniya: Removed redundant validation check

liam-verta (approved):

PR #2826, opened 2022-02-03
Implement fuzzy search using jdbi and fix remaining test cases

anandjakhaniya:

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
+++ b/backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
@@ -91,6 +91,24 @@ public class PredicatesHandler extends PredicateHandlerUtils {
             new QueryFilterContext()
                 .addCondition(String.format("%s.id = :%s", alias, bindingName))
                 .addBind(q -> q.bind(bindingName, value.getStringValue())));
+      case "date_created":
+        var date =
+            value.getKindCase().equals(Value.KindCase.STRING_VALUE)
+                ? Double.parseDouble(value.getStringValue())
+                : value.getNumberValue();
+        return InternalFuture.completedInternalFuture(
+            new QueryFilterContext()
+                .addCondition(String.format("%s.date_created = :%s", alias, bindingName))
+                .addBind(q -> q.bind(bindingName, date)));
+      case "date_updated":
+        var dateUpdated =
+            value.getKindCase().equals(Value.KindCase.STRING_VALUE)
+                ? Double.parseDouble(value.getStringValue())
+                : value.getNumberValue();
+        return InternalFuture.completedInternalFuture(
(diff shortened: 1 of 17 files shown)


backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -110,7 +123,7 @@ private String getEntityName() {
                 .addCondition(String.format("%s.date_updated = :%s", alias, bindingName))
                 .addBind(q -> q.bind(bindingName, dateUpdated)));
       case "owner":
-        // case time created/updated:
```
conradoverta, line 113: This was used because another PR had it, right?

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
```diff
@@ -624,4 +637,79 @@ private String getEntityName() {
...
+    }
+
+    return userInfoListFuture.thenCompose(
+        userInfoList -> {
+          if (userInfoList != null && !userInfoList.isEmpty()) {
+            var resourceItemsFutures = new ArrayList<InternalFuture<Set<String>>>();
+            for (var userInfo : userInfoList) {
+              resourceItemsFutures.add(
+                  getResourceItemsForLoginUserWorkspace(
+                          userInfo.getVertaInfo().getUsername(),
```
conradoverta: Not sure why we need this. `getResourceItemsForLoginUserWorkspace` will already give everything I have access to, right? I'm unsure why we need to query this for every user we find

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
```diff
@@ -68,12 +66,10 @@
 import org.apache.logging.log4j.Logger;
 import org.jdbi.v3.core.statement.Query;
 
-public class FutureProjectDAO {
+public class FutureProjectDAO extends UACApisUtil {
```
conradoverta: Could we use composition instead of inheritance? Seems like a more logical approach as we're not implementing abstract methods. It sounds like the DAO _has_ some utils to deal with UAC, vs the DAO _is_ some utils to deal with UAC.

anandjakhaniya: Updated PR with composition instead of inheritance. I hope it works now. :)

conradoverta (approved):
LGTM

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/PredicatesHandler.java
anandjakhaniya: Yes

PR #2829, opened 2022-02-04
Create project using jdbi

anandjakhaniya:


backend/common/src/main/java/ai/verta/modeldb/common/connections/UAC.java
```diff
@@ -19,6 +19,8 @@
   private final ManagedChannel authServiceChannel;
 
   private final CollaboratorServiceGrpc.CollaboratorServiceFutureStub collaboratorServiceFutureStub;
+  private final CollaboratorServiceGrpc.CollaboratorServiceFutureStub
+      ServiceAccountCollaboratorServiceFutureStub;
```
conradoverta:
```suggestion
      serviceAccountCollaboratorServiceFutureStub;
```

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
```diff
@@ -1558,4 +1578,219 @@ public FutureProjectDAO(
...
+                    .thenCompose(
+                        deletedProjectIds ->
+                            deleteEntityResourcesWithServiceUser(
+                                deletedProjectIds,
+                                ModelDBResourceEnum.ModelDBServiceResourceTypes.PROJECT),
+                        executor)
+                    .thenApply(status -> project, executor),
+            executor)
+        .thenCompose(createProjectHandler::insertProject, executor)
+        .thenCompose(
```
conradoverta, line 1592: This lambda is way too nested. I can't follow what it's doing. Let's split this method into helper methods for clarity. I _think_ it's correct, but I'm not confident since I can't be sure I follow.

backend/src/main/java/ai/verta/modeldb/project/FutureProjectServiceImpl.java
```diff
@@ -110,7 +110,30 @@ public FutureProjectServiceImpl(ServiceSet serviceSet, DAOSet daoSet, Executor e
   @Override
   public void createProject(
       CreateProject request, StreamObserver<CreateProject.Response> responseObserver) {
-    super.createProject(request, responseObserver);
+    // super.createProject(request, responseObserver);
+    try {
```
conradoverta:
```suggestion
    try {
```

backend/src/main/java/ai/verta/modeldb/project/subtypes/CreateProjectHandler.java
```diff
@@ -0,0 +1,192 @@
...
+                      try {
+                        int count = query.execute();
+                        LOGGER.trace("Project Inserted : " + (count > 0));
+                      } catch (UnableToExecuteStatementException exception) {
+                        // take a brief pause before resubmitting its query/transaction
+                        Thread.sleep(config.getJdbi_retry_time()); // Time in ms
+                        LOGGER.trace("Retry to insert Project");
+                        int count = query.execute();
+                        LOGGER.trace("Project Inserted after retry : " + (count > 0));
+                      }
```
conradoverta: Don't we already have a retry mechanism we can re-use?

backend/src/main/java/ai/verta/modeldb/project/subtypes/CreateProjectHandler.java
```diff
@@ -0,0 +1,192 @@
...
+                            newProject.getArtifactsList(),
+                            false);
+                      }
+                      if (newProject.getCodeVersionSnapshot().hasCodeArchive()
+                          || newProject.getCodeVersionSnapshot().hasGitSnapshot()) {
+                        codeVersionHandler.logCodeVersion(
+                            handleForTransaction,
+                            newProject.getId(),
+                            false,
+                            newProject.getCodeVersionSnapshot());
```
conradoverta: Shouldn't we capture the futures that are being created by the handlers? right now, if we have an error, it's just ignored

backend/src/main/java/ai/verta/modeldb/project/subtypes/CreateProjectHandler.java
anandjakhaniya: No we can not use that mechanism here because here exception is deferent and use case is also deferent.

backend/src/main/java/ai/verta/modeldb/project/subtypes/CreateProjectHandler.java
anandjakhaniya: Here used method are not retuning future because we have passed handler as an argument so if we used future then we lost the handler connection before executing DB operation so above all methods are used single transaction without using future.

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/HandlerUtil.java
```diff
@@ -0,0 +1,59 @@
...
+    // Created comma separated query bind arguments for the values
+    // based on the
+    // keys of
+    // above the map
+    // Ex: VALUES (:project_id, :experiment_id, :name) etc.
+    var bindArguments =
+        String.join(",", Arrays.stream(fieldsArr).map(s -> ":" + s).toArray(String[]::new));
+
+    return String.format(
+        "insert into %s (%s) values (%s) ", entityName, commaFields, bindArguments);
```
jared-verta, line 25: At first glance, this appears to be a SQL Injection vector in all three arguments passed here. Is there an input validation stage somewhere before this call? Should we consider moving validation to directly before constructing dynamic queries like this? Any method calling this method has the potential to introduce a vulnerability.

backend/common/src/main/java/ai/verta/modeldb/common/connections/UAC.java
anandjakhaniya: Resolved

backend/src/main/java/ai/verta/modeldb/experimentRun/subtypes/HandlerUtil.java
anandjakhaniya: Yes we have many other argument and logical validation before calling this. this is an internal parameters with validation so there is no any possibilities for SQL injection.

backend/src/main/java/ai/verta/modeldb/project/subtypes/CreateProjectHandler.java
conradoverta: Please expand. Also, I just caught the `sleep` in another PR, which is something we cannot do. Under which conditions does this exception happen? We need a better strategy than blocking the thread.

backend/src/main/java/ai/verta/modeldb/project/subtypes/CreateProjectHandler.java
conradoverta: Oh yeah, sorry.

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
anandjakhaniya: Fix and try to split method in helper functions

backend/src/main/java/ai/verta/modeldb/project/subtypes/CreateProjectHandler.java
anandjakhaniya: it might be thrown when some DB concurrency accord while project insertion then it will throw `UnableToExecuteStatementException` so at that time if we can take some milli sec pause then it will goes succeed. so i have added pause there instead of throwing error to the user.

anandjakhaniya: So what do you thing here? should we return error message to the user to do a retry?

anandjakhaniya: Removed pause and implemented retry logic using InternalFuture

liam-verta (approved):

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
liam-verta: This is still somewhat heavily nested, but It is better than before.  I think it is good enough to merge.

PR #2831, opened 2022-02-06
fix: [VUMM-105] Bump protobuf to v3.19.2

liam-verta:


liam-verta:
Unclear why #2830 would pass and this would fail.  Issues with agents not being cleaned up?

liam-verta:
!rebuild

conradoverta (approved):

PR #2836, opened 2022-02-08
create experiment handler using jdbi

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
```diff
@@ -0,0 +1,295 @@
...
+
+                      String queryString = buildInsertQuery(valueMap, "experiment");
+
+                      LOGGER.trace("insert experiment query string: " + queryString);
+                      var query = handleForTransaction.createUpdate(queryString);
+
+                      // Inserting fields arguments based on the keys and value of map
+                      for (Map.Entry<String, Object> objectEntry : valueMap.entrySet()) {
+                        query.bind(objectEntry.getKey(), objectEntry.getValue());
+                      }
```
conradoverta: I really like this pattern, btw. Makes things so much more modular =)

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
```diff
@@ -0,0 +1,295 @@
...
+
+                      try {
+                        int count = query.execute();
+                        LOGGER.trace("Experiment Inserted : " + (count > 0));
+                      } catch (UnableToExecuteStatementException exception) {
+                        // take a brief pause before resubmitting its query/transaction
+                        Thread.sleep(config.getJdbi_retry_time()); // Time in ms
```
conradoverta: This is really, really bad. It adds latency in the whole system, which is exactly what we're trying to optimize for here. When would this exception happen? And can we do a better job than block the whole system?

backend/src/main/java/ai/verta/modeldb/utils/HandlerUtil.java
```diff
@@ -0,0 +1,59 @@
...
+
+  protected void buildUpdateQueryAndExecute(
+      Logger LOGGER,
+      Handle handleForTransaction,
+      long entityId,
+      Map<String, Object> rmValueMap,
+      String entityName) {
+    StringBuilder queryStrBuilder = new StringBuilder(String.format("UPDATE %s SET ", entityName));
+
+    AtomicInteger count = new AtomicInteger();
```
conradoverta, line 36: Nit: we don't need an atomic counter as we won't be calling it in parallel.

conradoverta (changes requested):

backend/src/main/java/ai/verta/modeldb/utils/HandlerUtil.java
anandjakhaniya: AtomicInteger is there because we used for loop as a lambda and in lambda if we used integer from out side then we have to make it AtomicInteger OR integer array with single value. but i have modified code with simple for loop and remove atomic. :)

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
anandjakhaniya: https://github.com/VertaAI/modeldb/pull/2829#discussion_r807760593

anandjakhaniya: Removed pause and implemented retry logic using InternalFuture

conradoverta (approved):

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
```diff
@@ -0,0 +1,291 @@
...
+    String errorMessage = null;
+    if (request.getProjectId().isEmpty()) {
+      errorMessage = "Project ID not found in CreateExperiment request";
+    } else if (request.getName().isEmpty()) {
+      request = request.toBuilder().setName(MetadataServiceImpl.createRandomName()).build();
+    }
+
+    if (errorMessage != null) {
+      throw new InvalidArgumentException(errorMessage);
+    }
```
liam-verta: This code as written works but I'd rather us do all the fatal checks in one block and non-fatal checks in a separate block.

```suggestion
    if (request.getProjectId().isEmpty()) {
      errorMessage = "Project ID not found in CreateExperiment request";
    }

    if (errorMessage != null) {
      throw new InvalidArgumentException(errorMessage);
    }
    
    if (request.getName().isEmpty()) {
      request = request.toBuilder().setName(MetadataServiceImpl.createRandomName()).build();
    }
```

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
```diff
@@ -0,0 +1,291 @@
...
+            .setVersionNumber(1L);
+
+    if (request.getDateCreated() != 0L) {
+      experimentBuilder
+          .setDateCreated(request.getDateCreated())
+          .setDateUpdated(request.getDateCreated());
+    } else {
+      experimentBuilder
+          .setDateCreated(Calendar.getInstance().getTimeInMillis())
+          .setDateUpdated(Calendar.getInstance().getTimeInMillis());
```
liam-verta: Calendar isn't deprecated, but it is clearer to us `Instant`.  Also, 
```suggestion
          .setDateCreated(Instant.now().toEpochMilli())
          .setDateUpdated(Instant.now().toEpochMilli());
```

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
```diff
@@ -0,0 +1,291 @@
...
+            .addAllArtifacts(
+                request.getArtifactsList().stream()
+                    .sorted(Comparator.comparing(Artifact::getKey))
+                    .collect(Collectors.toList()))
+            .setVersionNumber(1L);
+
+    if (request.getDateCreated() != 0L) {
+      experimentBuilder
+          .setDateCreated(request.getDateCreated())
+          .setDateUpdated(request.getDateCreated());
```
liam-verta, line 128: Do we want to allow these to be set by the request?  Shouldn't this be controlled on the server side?

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
anandjakhaniya: No it is not deprecated. but we can take `Instant.now().toEpochMilli()` instead of Calendar

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
anandjakhaniya: We have given both the functionality so if it user is not setting those fields then we are populate at server side. so it is already there.

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/experiment/subtypes/CreateExperimentHandler.java
anandjakhaniya: Fixed

PR #2839, opened 2022-02-09
VUMM-181 add session id for headers validation

evgeniilezhnin:
## Impact and Context

it is needed for dev key hashing effort: https://vertaai.atlassian.net/browse/VUMM-140 

UAC PR is waiting for other changes to be merged: https://github.com/VertaAI/verta-uacservice/pull/1143

## Risks and Area of Effect

This change is just to ease header validation, shouldn't break anything

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
@@ -56,7 +56,7 @@ public class AuthInterceptor implements ServerInterceptor {
                   || requestHeaders.containsKey(devKey))) {
             parameterMissing = true;
           }
-        } else if (sourceValue.equals("InternalToken")) {
+        } else if (sourceValue.equals("SessionId")) {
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
```diff
@@ -54,6 +56,10 @@
                   || requestHeaders.containsKey(devKey))) {
             parameterMissing = true;
           }
+        } else if (sourceValue.equals("InternalToken")) {
```
conradoverta: Wouldn't be a better name here "session" since that's what we're checking? 🤔  `InternalToken` makes me think of a secret dev key like we had

conradoverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/authservice/AuthInterceptor.java
evgeniilezhnin: renamed

PR #2859, opened 2022-02-15
Fix attribute handler for registry backend

anandjakhaniya:
## Impact and Context

Here in MDB backend we are not used `updateAttributes` while inserting any main entities like project, er etc. we have used just for individual endpoint but in registry we have just one endpoint for update attributes via `UpdateModelVersion` and for that we have to used transaction while updating `model_version` with attribute so we have added handler as a parameter in this PR for registry backend.


conradoverta (changes requested):
Please add a clearer description of the change. What was wrong? How did you fix?

anandjakhaniya:
> Please add a clearer description of the change. What was wrong? How did you fix?

Updated description

conradoverta (approved):
_Much_ clearer, thanks!!

PR #2861, opened 2022-02-15
Fix test init logic with reconciler to clean up test data

anandjakhaniya:

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/reconcilers/Reconciler.java
@@ -42,8 +42,8 @@ public abstract class Reconciler<T> {
 
     if (!config.isTestReconciler()) {
       startResync();
-      startWorkers();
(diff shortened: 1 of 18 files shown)


conradoverta (changes requested):
Making private/protected methods public is usually the wrong way to solve a problem, especially as it opens ourselves to bugs. If you soft delete the items, the reconciler should automatically cleanup right away. You can check if it's done with `isNotEmpty`, right? I'm unclear why we should expose `reconcile` as passing the wrong IDs there can be catastrophic.

anandjakhaniya:
> Making private/protected methods public is usually the wrong way to solve a problem, especially as it opens ourselves to bugs. If you soft delete the items, the reconciler should automatically cleanup right away. You can check if it's done with `isNotEmpty`, right? I'm unclear why we should expose `reconcile` as passing the wrong IDs there can be catastrophic.

Improve code based on discussion  and tested on test suite job it works fine now.

conradoverta (approved):

PR #2862, opened 2022-02-15
[VUMM-214] Modifed logic for untrusted artifacts

anandjakhaniya:


backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
```diff
@@ -122,15 +119,22 @@ public void logArtifacts(Handle handle, T entityId, List<Artifact> artifacts, bo
...
       if (artifact.getUploadCompleted()) {
         uploadCompleted = true;
+      } else {
+        path =
+            artifactStoreConfig.storeTypePathPrefix()
+                + this.entityName
+                + "/"
+                + entityId
+                + "/"
+                + artifact.getKey();
```
conradoverta: Let's make sure to do some input hygiene since this value is provided directly by the user. That's _always_ a dangerous thing to just pass along. Maybe base64 it so it's always safe?

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
```diff
@@ -122,15 +119,22 @@ public void logArtifacts(Handle handle, T entityId, List<Artifact> artifacts, bo
       validateAndThrowErrorAlreadyExistsArtifacts(entityId, artifacts, handle);
     }
 
-    for (final var artifact : artifacts) {
+    for (var artifact : artifacts) {
+      var path = artifact.getPath();
       var uploadCompleted = !artifactStoreConfig.getArtifactStoreType().equals(CommonConstants.S3);
       if (artifact.getUploadCompleted()) {
         uploadCompleted = true;
+      } else {
```
conradoverta: I'm confused why we don't overwrite the path in this case. Shouldn't we overwrite the path _always_ since it's untrusted?

backend/src/main/java/ai/verta/modeldb/entities/ArtifactEntity.java
```diff
@@ -51,9 +53,21 @@ public ArtifactEntity(Object entity, String fieldType, Artifact artifact) {
...
     var uploadCompleted = !artifactStoreConfig.getArtifactStoreType().equals(CommonConstants.S3);
     if (artifact.getUploadCompleted()) {
       uploadCompleted = true;
+    } else {
+      path =
+          artifactStoreConfig.storeTypePathPrefix()
+              + entityName
+              + (entityId.map(id -> "/" + id + "/").orElse("/"))
+              + artifact.getKey();
+    }
```
conradoverta: Similar concerns

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
```diff
@@ -720,7 +720,9 @@ public Experiment logExperimentCodeVersion(String experimentId, CodeVersion upda
               RdbmsUtils.generateArtifactEntity(
                   experimentEntity,
                   ModelDBConstants.CODE_ARCHIVE,
-                  updatedCodeVersion.getCodeArchive()));
+                  updatedCodeVersion.getCodeArchive(),
+                  CodeVersionEntity.class.getSimpleName(),
+                  Optional.empty()));
```
conradoverta: This seems wrong to me as it's prone to collisions. Why would we ever not have the entity id?

conradoverta (changes requested):

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
conradoverta: @liuverta opinions is the s3 final location name matters when downloading on the client?

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
anandjakhaniya: We have code_version in `gitsnapshot` and we have used same `artifact` entity for them and that code_version entity have auto generated id we can not set it in our entity before insertion that's the issue so i have make it optional

conradoverta: It doesn't solve the collision problem. It just confirms that collision _will_ be a problem. We need to have an id before saving to S3. No two artifacts in different places in the system should point to the same s3 location.

conradoverta (changes requested):
As previously discussed, the collision concern is a real issue.

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
liuverta: @conradoverta For regular `get_`/`log_artifact()`, no the S3 final location name doesn't matter—the client just cares about the artifact key and the pre-signed URL.

But as I took a deeper dive into the client codebase, and there are some considerations I hadn't realized before 😰 :

1. In `fetch_artifacts()`, client [uses](https://github.com/VertaAI/modeldb/blob/299d4525bb579cf58b257cbbd47f475750770d77/client/verta/verta/tracking/entities/_deployable_entity.py#L438-L440) the artifact proto's `path` field as a filepath to locally cache the downloaded artifact (b/c it's a checksum). This shouldn't be an issue though; client doesn't care what it is as long as it's unique.
2. The ZFS/local artifacts functionality we built a couple years ago involves [prepending `$VERTA_ARTIFACT_DIR` to the artifact proto's `path`](https://github.com/VertaAI/modeldb/blob/36a9000607a7e9635e8cf7d3684f3a7a57e9cc7d/client/verta/verta/tracking/entities/_experimentrun.py#L235-L245). Also, there's a `run.log_artifact_path()` that explicitly relies on [using the artifact proto's `path`.](https://github.com/VertaAI/modeldb/blob/36a9000607a7e9635e8cf7d3684f3a7a57e9cc7d/client/verta/verta/tracking/entities/_experimentrun.py#L393). I wish I knew whether these are still being actively used 🤔

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
anandjakhaniya: Simplify logic and implemented it for nfs as well as per our discussion

backend/src/main/java/ai/verta/modeldb/entities/ArtifactEntity.java
anandjakhaniya: Simplify logic and implemented it for nfs as well as per our discussion

backend/src/main/java/ai/verta/modeldb/experiment/ExperimentDAORdbImpl.java
anandjakhaniya: Used UUID for uniqueness as per discussion and removed optional from codebase

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
conradoverta: 1. yep, it should still be unique
2. I'm not particularly concerned about these

backend/src/test/java/ai/verta/modeldb/ProjectTest.java
```diff
@@ -2632,10 +2641,19 @@ public void getProjectCodeVersionTest() {
...
+      assertNotEquals(
+          "Project codeVersion not match with expected project codeVersion",
+          logProjectCodeVersionRequest.getCodeVersion(),
+          project.getCodeVersionSnapshot());
+    } else {
+      assertEquals(
+          "Project codeVersion not match with expected project codeVersion",
+          logProjectCodeVersionRequest.getCodeVersion(),
+          project.getCodeVersionSnapshot());
+    }
```
conradoverta: I'm not sure I follow why we have different behaviors for S3 🤔

conradoverta (commented):
Nit: maybe prefix "fake-" before the UUIDs so that we don't get confused with real UUIDs? Like project IDs

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
liuverta: 👌

anandjakhaniya:
> Nit: maybe prefix "fake-" before the UUIDs so that we don't get confused with real UUIDs? Like project IDs

Added prefix `fake-`  before UUIDs as per review suggestion

conradoverta (approved):

PR #2866, opened 2022-02-17
Fix find ER based on experiment name

anandjakhaniya:
## Impact and Context

Resolved issue: https://vertaai.slack.com/archives/C01RBD7LQ67/p1645092558656029
it was written on wrong place and we do not have test case for that so till now it is not catch but added test case and fix the logic in this PR

### DIFF
--- a/backend/src/test/java/ai/verta/modeldb/FindProjectEntitiesTest.java
+++ b/backend/src/test/java/ai/verta/modeldb/FindProjectEntitiesTest.java
@@ -2580,4 +2580,71 @@ public class FindProjectEntitiesTest extends TestsInit {
 
     LOGGER.info("FindExperimentRuns by experiment test stop.........");
   }
+
+  @Test
+  public void findExperimentRunsByExperimentNameTest() {
+    LOGGER.info("FindExperimentRuns by experiment name test start.........");
+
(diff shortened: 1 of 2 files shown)


backend/src/test/java/ai/verta/modeldb/FindProjectEntitiesTest.java
```diff
@@ -2580,4 +2580,71 @@ public void findExperimentRunsByExperimentTest() {
...
+        experimentRunServiceStub.findExperimentRuns(findExperimentRuns);
+    LOGGER.info("FindExperimentRuns Response : " + response.getExperimentRunsCount());
+    assertEquals(
+        "ExperimentRun count not match with expected experimentRun count",
+        2,
+        response.getExperimentRunsCount());
+    assertEquals(
+        "ExperimentRun not match with expected experimentRun",
+        experiment1.getId(),
+        response.getExperimentRunsList().get(0).getExperimentId());
```
liam-verta, line 2608: Is the output sorted such that we can depend on ordering?  
Perhaps better to assert the list contains experiment1 id.

liam-verta (approved):

conradoverta:
Merging to unblock release. The question should have a follow up!

PR #2874, opened 2022-02-19
fix: test_endpoint.test_endpoint.TestEndpoint.test_get_deployed_model

conradoverta:
## Impact and Context
Header keys are case insensitive


liuverta (approved):
Oh, interesting 😬

PR #2896, opened 2022-02-28
Removed old ExperimentRunService implementation

anandjakhaniya:


conradoverta (approved):
Do we really need to return ER that many times?

backend/src/main/java/ai/verta/modeldb/utils/UACApisUtil.java
```diff
@@ -232,4 +234,39 @@ public boolean checkAllResourceAllowed(List<Resources> resources) {
...
+              }
+              Optional<GetResourcesResponseItem> responseItem = responseItems.stream().findFirst();
+              if (responseItem.isPresent()) {
+                return responseItem.get();
+              } else {
+                StringBuilder errorMessage =
+                    new StringBuilder("Failed to locate ")
+                        .append(modelDBServiceResourceTypes.name())
+                        .append(" resources in UAC for ")
+                        .append(modelDBServiceResourceTypes.name())
```
conradoverta, line 264: These two fields are the same. I believe they probably should be different, right?

anandjakhaniya: No, Both are the same one. actual the message would be looks like `Failed to locate PROJECT resources in UAC for PROJECT ID 'ID of resource'` . So there is a used builder pattern to populate that resource name .

anandjakhaniya:
> Do we really need to return ER that many times?

Yes. those are individual endpoint so we needs to populate it in response along with use for logging events

PR #2961, opened 2022-03-22
[VRD-193] When creating RMV from ER, we should reuse the ER's artifact's paths

anandjakhaniya:

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
@@ -130,18 +130,20 @@ public abstract class CommonArtifactHandler<T> {
         uploadCompleted = true;
       }
 
-      var validPrefix = artifactStoreConfig.getPathPrefixWithSeparator() + this.entityName;
-      var path = validPrefix + "/" + entityId + "/" + artifact.getKey();
+      var path = artifact.getPath();
+      var validEntityPrefix = path.split("/")[0];
+      if (!startPrefixWithValidEntity(validEntityPrefix)) {
+        validEntityPrefix = artifactStoreConfig.getPathPrefixWithSeparator() + this.entityName;
+        path = validEntityPrefix + "/" + entityId + "/" + artifact.getKey();
+      }
+
+      artifact = artifact.toBuilder().setPath(path).build();
       var storeTypePath =
           !artifact.getPathOnly() ? artifactStoreConfig.storeTypePathPrefix() + path : "";
 
       if (overwrite && isExists(entityId, artifact.getKey(), handle)) {
-        if (!artifact.getPath().startsWith(validPrefix)) {
-          artifact = artifact.toBuilder().setPath(path).build();
-        }
         updateArtifactWithHandle(entityId, handle, artifact, uploadCompleted, storeTypePath);
       } else {
-        artifact = artifact.toBuilder().setPath(path).build();
         insertArtifactInDB(entityId, handle, artifact, uploadCompleted, storeTypePath);
       }
       pathUpdatedArtifacts.add(artifact);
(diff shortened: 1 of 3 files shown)


conradoverta (changes requested):
I'm very confused why we're adding this whole layer of parsing and then potentially re-creating the path. This makes the code very hard to read by embedding logic from a top level command (create from ER) all the way down to the artifact handler. From reading the code, I'd have no clue what this condition is trying to solve and I'm even having a hard time telling if the code is correct by reading it.

conradoverta (commented):
Also, could we at least have a test for this given how critical this is?

anandjakhaniya:
> I'm very confused why we're adding this whole layer of parsing and then potentially re-creating the path. This makes the code very hard to read by embedding logic from a top level command (create from ER) all the way down to the artifact handler. From reading the code, I'd have no clue what this condition is trying to solve and I'm even having a hard time telling if the code is correct by reading it.

We have multiple place where we are creating artifact and for all possibilities the last point is CommonArtifactHandler so i have handler path logic there.

- Logic will check about the path started with entity name as prefix 
- If it is there then we are considered path has new logic whatever entity is created that path. let say if path created in ER then it was started with ER entity name. we don't care about it will populated in MV or not because that common logic check for all entities at single place.
- If path does not started with entity name as prefix then we are creating new path with new logic in that entity.

I don't think there will be good place to add this logic and if we will try to simplify that logic for readability then we have to add that same logic multiple times at many places so i think it will create more errors instead of handling this logic at single place.

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
```diff
@@ -130,18 +134,17 @@ protected abstract Query buildGetArtifactsQuery(
...
-      var path = validPrefix + "/" + entityId + "/" + artifact.getKey();
-      var storeTypePath =
-          !artifact.getPathOnly() ? artifactStoreConfig.storeTypePathPrefix() + path : "";
+      var path = artifact.getPath();
+      if (!isPopulateFromOtherEntity) {
+        var validPrefix = artifactStoreConfig.getPathPrefixWithSeparator() + this.entityName;
+        path = validPrefix + "/" + entityId + "/" + artifact.getKey();
+        artifact = artifact.toBuilder().setPath(path).build();
+      }
+      var storeTypePath = !artifact.getPathOnly() ? path : "";
```
conradoverta, line 149: `storeTypePath` here has a different value from before, so it has to be checked with care (we can't rely on previously working tests)

anandjakhaniya: no, it has same value. Before changes we are setting it if path_only is false other wise we will leave it empty. so here after changes we keep it as it is. it is not populate in artifact proto. it is populate in DB only.

conradoverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
```diff
@@ -130,18 +134,23 @@ protected abstract Query buildGetArtifactsQuery(
...
+      var path = artifact.getPath();
+      if (!isPopulateFromOtherEntity) {
+        var validPrefix = artifactStoreConfig.getPathPrefixWithSeparator() + this.entityName;
+        path = validPrefix + "/" + entityId + "/" + artifact.getKey();
 
-      if (overwrite && isExists(entityId, artifact.getKey(), handle)) {
-        if (!artifact.getPath().startsWith(validPrefix)) {
-          artifact = artifact.toBuilder().setPath(path).build();
+        var filenameExtension = artifact.getFilenameExtension();
+        if (!filenameExtension.isEmpty() && !filenameExtension.endsWith("." + filenameExtension)) {
```
liuverta:
```suggestion
        if (!filenameExtension.isEmpty() && !path.endsWith("." + filenameExtension)) {
```

backend/src/main/java/ai/verta/modeldb/entities/ArtifactEntity.java
```diff
@@ -56,6 +56,11 @@ public ArtifactEntity(
     var validPrefix = artifactStoreConfig.getPathPrefixWithSeparator() + this.entityName;
     var path = validPrefix + "/" + entityId + "/" + artifact.getKey();
 
+    var filenameExtension = artifact.getFilenameExtension();
+    if (!filenameExtension.isEmpty() && !filenameExtension.endsWith("." + filenameExtension)) {
```
liuverta:
```suggestion
    if (!filenameExtension.isEmpty() && !path.endsWith("." + filenameExtension)) {
```

liuverta (changes requested):

backend/common/src/main/java/ai/verta/modeldb/common/subtypes/CommonArtifactHandler.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/entities/ArtifactEntity.java
anandjakhaniya: Fixed

liuverta (approved):

PR #2962, opened 2022-03-23
feat: [VMM-400] Add monitoring client patch extensions

coryverta:


client/verta/verta/registry/entities/__init__.py
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 "Entities for registering ML models to the Verta backend."
 
+from tempfile import NamedTemporaryFile
+
```
liuverta:
```suggestion

import pandas as pd

```

liuverta (approved):
I think `predict_with_id()`, `log_ground_truth()`, and the changes to `update()` will need tests and cleanup down the line.

PR #2986, opened 2022-04-10
fix: Include requirements.txt in source distribution

liuverta:
## Impact and Context

We upload our client to PyPI as both a wheel as well as a source `.tar.gz` distribution. And since `0.19.3`, we [split out](https://github.com/VertaAI/modeldb/pull/2747) the list of our dependencies from within `setup.py` to a standalone `requirements.txt`.

This results in an issue: That `requirements.txt` is not included when `setuptools` [creates a source distribution](https://github.com/VertaAI/modeldb/blob/356cfc0e1050c3f261741ce7067a022638a04bb2/client/verta/upload.sh#L1), so anyone downloading and installing our package from source via PyPI won't be able to run `setup.py` since it tries to open a missing file. Specifically, this breaks our package's [conda-forge build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=489118&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d) which does precisely that.

This PR introduces a `MANIFEST.in` file, which is [the way](https://packaging.python.org/en/latest/guides/using-manifest-in/) to specify extra files to be included in a source distribution.

## Risks and Area of Effect

Low risk: An extra couple of files won't break anything.

Low AoE: This [only] benefits users installing our client from source.

## Testing

I built our library locally without then with this change, and installation from the source `tar.gz` failed then succeeded respectively.

This could probably be added to our CI 🤔 

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/MANIFEST.in
+++ b/client/verta/MANIFEST.in
@@ -1,0 +1,1 @@
+include requirements.txt requirements-unit-tests.txt
--- a/client/verta/Makefile
+++ b/client/verta/Makefile
@@ -1,0 +1,11 @@
+# build wheel and source distributions
+build: clean
+	python setup.py sdist bdist_wheel --universal
+
+# clean previously-created builds
+clean:
+	rm -rf build dist verta.egg-info
+
+# upload distributions to PyPI
+upload: build
+	python -m twine upload dist/* -u \${PYPI_USERNAME} -p \${PYPI_PASSWORD}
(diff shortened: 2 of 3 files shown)


conradoverta (approved):
magic

PR #2988, opened 2022-04-12
feat: [VRD-233] Add cluster kafka config ID to client

ghost:
## Impact and Context

The cluster config ID is required when deploying an endpoint with kafka; it is what associates the endpoint with the kafka cluster set up by the system admin.

## Testing

I used the client to successfully deploy an endpoint to `hm`.


client/verta/tests/test_endpoint/test_kafka.py
```diff
@@ -12,55 +12,61 @@
 class TestKafkaSettings:
 
     @hypothesis.given(
+        cluster_config_id=st.text(min_size=1),
```
ghost: Note that the tests currently fail with `500 Server Error: Kafka config ID does not correlate with valid kafka configuration for url: https://hm.dev.verta.ai/api/v1/deployment/stages/2/kafka`, which checks out considering we don't have actually kafka configuration IDs to provide to dep-api.

client/verta/verta/environment/_environment.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, env_vars, autocapture):
         self._msg = _EnvironmentService.EnvironmentBlob()
 
         if env_vars:
-            self.add_env_vars(env_vars)
+            self.add_env_vars(env_vars)/app/app/api/predict_resource.py
```
liuverta: This looks unintentional 😬

client/verta/verta/endpoint/_kafka_settings.py
```diff
@@ -38,12 +41,13 @@ class KafkaSettings(object):
 
     """
 
-    def __init__(self, input_topic, output_topic, error_topic):
+    def __init__(self, cluster_config_id, input_topic, output_topic, error_topic):
```
liuverta: As a random side note, adding this parameter to the beginning instead of the end means that any existing code that instantiates `KafkaSettings("input", "output", "error")` will start throwing

```python
TypeError: __init__() missing 1 required positional argument: 'error_topic'
```

which looks kinda confusing ("What do you mean, my error topic is right there!"), as opposed to adding the parameter to the end which will throw a marginally more directly helpful

```python
TypeError: __init__() missing 1 required positional argument: 'cluster_config_id'
```

But I don't think anyone's using this today so it's not a worry.

liuverta (changes requested):
minor initial feedback

client/verta/tests/test_endpoint/test_kafka.py
liuverta: `cluster_config_id=st.text(min_size=1)` might not end up working out for this test 🤔 

We were able to use `st.text(min_size=1)` for the topic names because there was no need for them to be real values in the old Kafka functionality; this test could get away with sending garbage values.

Now that this needs to be tied to a real, existing, backend-saved Kafka config, we might need to overhaul this test to create one in the backend

client/verta/verta/environment/_environment.py
ghost: how did this even happen 😂

liuverta (approved):
😬 I had meant for my "moving `cluster_config_id` to the end" comment to be more of a "fyi for the future, but here we're good".

_**If**_ you happen to find it more intuitive/logical as an API for `cluster_config_id` to be the first parameter, I think that would be alright to change it back for this feature. Otherwise, 🤷 🚢

ghost:
> 😬 I had meant for my "moving `cluster_config_id` to the end" comment to be more of a "fyi for the future, but here we're good".
> 
> _**If**_ you happen to find it more intuitive/logical as an API for `cluster_config_id` to be the first parameter, I think that would be alright to change it back for this feature. Otherwise, 🤷 🚢

:salute:

PR #2999, opened 2022-04-18
[VUMM-293] Fix initialization using spring boot

anandjakhaniya:


backend/common/src/main/java/ai/verta/modeldb/common/CommonApp.java
```diff
@@ -0,0 +1,113 @@
...
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.GenericBeanDefinition;
+import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory;
+import org.springframework.boot.web.servlet.ServletRegistrationBean;
+import org.springframework.boot.web.servlet.server.ServletWebServerFactory;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.annotation.Bean;
+
+public abstract class CommonApp implements ApplicationContextAware {
```
jkwatson, line 29: Is it possible to avoid using a class hierarchy for this? Can we "favor composition over inheritance" rather than requiring subclassing?

anandjakhaniya: @jkwatson No we can not use `"favor composition over inheritance"` pattern here because the use cases is different. Here the matter is not about variables but it is about common methods implementation which is common for all other backend as well. and some methods are different but signature are same for all backend so the abstraction is the perfect design for this as per my point of view.

jkwatson-verta: If your goal is just to expose some spring beans, that doesn't need to be done via a class hierarchy at all. It can just be a class that exposes them, included in the packages that the application scans for startup.

backend/common/src/main/java/ai/verta/modeldb/common/CommonConstants.java
```diff
@@ -13,6 +13,10 @@ public interface CommonConstants {
   String EMPTY_STRING = "";
   String S3 = "S3";
   Integer TAG_LENGTH = 40;
+  Integer THREAD_COUNT = 256;
```
jkwatson-verta: recommend naming this with its purpose. Is this the thread count for ... grpc clients? grpc servers? internal futures?

backend/common/src/main/java/ai/verta/modeldb/common/CommonConstants.java
```diff
@@ -13,6 +13,10 @@ public interface CommonConstants {
...
+  String LIQUIBASE_MIGRATION = "LIQUIBASE_MIGRATION";
```
jkwatson-verta: recommend naming these env vars with something that indicates their usage as such. like `ENABLE_LIQUIBASE_MIGRATION_ENV_VAR` or something along those lines.

backend/src/main/java/ai/verta/modeldb/health/HealthServiceImpl.java
```diff
@@ -77,13 +78,27 @@ private ServingStatus getServingStatus(HealthCheckRequest request) {
       globalStatus = getStatus("");
       if (request.getService().equals("ready")) {
         if (globalStatus == ServingStatus.SERVING) {
-          setStatus("ready", modelDBHibernateUtil.checkReady());
+          ServingStatus hibernateConnectionStatus = modelDBHibernateUtil.checkReady();
```
jkwatson-verta, line 81: do we still have hibernate stuff around???

backend/common/src/main/java/ai/verta/modeldb/common/CommonApp.java
anandjakhaniya: @jkwatson-verta Please consider follow up PR as well while reviewing because this PR is just the initial setup which might be expanded in follow up PRs like: https://github.com/VertaAI/modeldb/pull/3005/files#diff-49c276c2dafbfa6661d2997338e0594e9bbf58374e75622216d9a30b102dee87 so there will be abstract methods where it implementation is different for all other backends.

backend/src/main/java/ai/verta/modeldb/health/HealthServiceImpl.java
anandjakhaniya: Yes. we have it. Still we have the Repository and Dataset service impl with the hibernate implementation.

backend/common/src/main/java/ai/verta/modeldb/common/CommonApp.java
jkwatson-verta: I do not think this is a good design. Please reconsider using a base class for this.

backend/common/src/main/java/ai/verta/modeldb/common/CommonConstants.java
anandjakhaniya: Fixed

backend/common/src/main/java/ai/verta/modeldb/common/CommonConstants.java
anandjakhaniya: currently not in used so removed it

ad-47:
This is no more valid after merging https://github.com/VertaAI/modeldb/pull/3036

anandjakhaniya:
This is no more valid after merging https://github.com/VertaAI/modeldb/pull/3036

PR #3005, opened 2022-04-20
[VUMM-301] Improve App.java using CommonApp.java class for use same template at multiple backend

anandjakhaniya:


anandjakhaniya:
We have taken another approach so this is not required now.

PR #3013, opened 2022-04-26
VUMM-306 check dbchangelog exists

evgeniilezhnin:
## Impact and Context

It is needed because UAC had before another table name for changelog lock, but now it will change to using common name

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
@@ -90,13 +90,18 @@ public abstract class CommonDBUtil {
 
       try (var stmt = jdbcCon.createStatement()) {
 
-        var sql = "SELECT * FROM database_change_log_lock WHERE ID = 1";
-        ResultSet rs = stmt.executeQuery(sql);
+        ResultSet rs;
+        try {
+          var sql = "SELECT * FROM database_change_log_lock WHERE ID = 1";
+          rs = stmt.executeQuery(sql);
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -90,13 +90,18 @@ protected void releaseLiquibaseLock(DatabaseConfig config)
...
       try (var stmt = jdbcCon.createStatement()) {
 
-        var sql = "SELECT * FROM database_change_log_lock WHERE ID = 1";
-        ResultSet rs = stmt.executeQuery(sql);
+        ResultSet rs;
+        try {
+          var sql = "SELECT * FROM database_change_log_lock WHERE ID = 1";
+          rs = stmt.executeQuery(sql);
+        } catch (Exception e) {
+          rs = null;
```
liam-verta, line 98: Add a log message about this table being missing. 

Maybe instead of setting to `null`, we should `return` out of this method? 

The rest of this method doesn't make sense without this ResultSet.

liam-verta: Or maybe UAC should perform this check? This code is not needed to for any other projects.

evgeniilezhnin:
Actually I revalidated the results of application run without this change for different database types that we use and it works ok. I can't find an original config with which I reproduced this error, but it might be java database driver which we don't use (mysql, but we use mariadb for connection because of its license). So I think it will be better to close this change for now. Sorry for the inconvenience

PR #3016, opened 2022-05-03
Fix test cases init

anandjakhaniya:

### DIFF
--- a/backend/src/test/java/ai/verta/modeldb/TestsInit.java
+++ b/backend/src/test/java/ai/verta/modeldb/TestsInit.java
@@ -24,6 +24,7 @@ import io.grpc.ManagedChannelBuilder;
 import io.grpc.inprocess.InProcessChannelBuilder;
 import io.grpc.inprocess.InProcessServerBuilder;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.AfterClass;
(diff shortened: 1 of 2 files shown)


backend/src/test/java/ai/verta/modeldb/TestsInit.java
```diff
@@ -73,8 +73,11 @@ public class TestsInit {
   protected static HydratedServiceGrpc.HydratedServiceBlockingStub
       hydratedServiceBlockingStubClient2;
   protected static ArtifactStoreGrpc.ArtifactStoreBlockingStub artifactStoreBlockingStub;
+  private static Boolean setServerAndServiceRan = false;
 
   protected static void init() throws Exception {
+    if (setServerAndServiceRan) return;
```
jkwatson: style nit: I would greatly prefer we have brackets around all blocks

anandjakhaniya: Changed it.

backend/src/test/java/ai/verta/modeldb/TestsInit.java
```diff
@@ -73,8 +73,13 @@ public class TestsInit {
   protected static HydratedServiceGrpc.HydratedServiceBlockingStub
       hydratedServiceBlockingStubClient2;
   protected static ArtifactStoreGrpc.ArtifactStoreBlockingStub artifactStoreBlockingStub;
+  private static Boolean setServerAndServiceRan = false;
```
jkwatson: I think this should be an AtomicBoolean with proper CAS access to make sure there aren't race conditions with the initialization.

anandjakhaniya: We are not executing test classes parallelly so race condition will never accrued. but lets make it for safety.

anandjakhaniya: Fixed it.

jkwatson-verta (approved):

backend/src/test/java/ai/verta/modeldb/TestsInit.java
jkwatson-verta: The other reason would be cross-thread visibility. I don't think there's any guarantee that junit will run all the tests on the same thread, even if they aren't running concurrently. Thanks for the change! 👍

PR #3019, opened 2022-05-05
chore: Remove upper constraints on dependencies

liuverta:
## Impact and Context

I've held a philosophy that the purpose of dependency constraints is to allow versions that our library is known to work with.

It seems the ecosystem largely takes the stance to _dis_allow versions that the library is known to _not_ work with. Now that our platform is using stricter dependency resolution, this seems to be the less strict approach.

Pros:

- If one of these dependencies—in the future—releases a compatible version, we no longer need to explicitly increase its upper bound in a new release to unblock users.

## Risks and Area of Effect

Cons:

- If one of these dependencies—in the future—releases an incompatible version, the client may misbehave/error at runtime because there was no installation-time guard against it.
  - But the user does have the freedom to pin a compatible version in their environment.

## How to Revert

Revert this PR.


liuverta:
A bit unnecessary for the time being, but a solid proof of concept perhaps.

PR #3026, opened 2022-05-17
handle illegal argument exception

evgeniilezhnin:

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
@@ -2,6 +2,7 @@ package ai.verta.modeldb.common;
 
 import ai.verta.common.Pagination;
 import ai.verta.modeldb.common.exceptions.InternalErrorException;
+import ai.verta.modeldb.common.exceptions.InvalidArgumentException;
 import ai.verta.modeldb.common.exceptions.ModelDBException;
 import ai.verta.modeldb.common.exceptions.UnavailableException;
 import ai.verta.modeldb.common.futures.Handle;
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
```diff
@@ -130,6 +130,15 @@ public static <T extends GeneratedMessageV3> StatusRuntimeException logError(
...
                 .build();
+      } else if (e instanceof IllegalArgumentException) {
+        var illegalArgumentException = (IllegalArgumentException) e;
+        logBasedOnTheErrorCode(
+            isClientError(Code.INVALID_ARGUMENT_VALUE), illegalArgumentException);
+        status =
+            Status.newBuilder()
+                .setCode(Code.INVALID_ARGUMENT_VALUE)
+                .setMessage(illegalArgumentException.getMessage())
+                .build();
```
anandjakhaniya, line 141: illegalArgumentException will thrown due to some internal mistakes not due to client's invalid argument i think. NO?

evgeniilezhnin: Looks like it is used for client argument validation in library code and also in generated grpc code. Also we use it in our code for agrument validation. As an alternative we can remove all usages of this class from our code. Is it really needed?

anandjakhaniya: In the modeldb and registry backend i am sure i do not used it anywhere for client parameter validation.

evgeniilezhnin: For example: https://github.com/VertaAI/modeldb/blob/553e74086a34ce65254b8f0e89dea1c16b1a902d/backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java#L1735

evgeniilezhnin: That doesn't look like internal error code, more like invalid argument

anandjakhaniya: Yeah got it. Ok, then we can add it.

anandjakhaniya (approved):

PR #3027, opened 2022-05-17
VUMM-329 use all allowed flag

evgeniilezhnin:


evgeniilezhnin:
Tested on el4.dev.verta.ai

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
```diff
@@ -863,51 +863,38 @@ public InternalFuture<List<GetResourcesResponseItem>> deleteProjects(List<String
...
-              var resourcesIds = new ArrayList<String>();
-              if (!getAllowedResourcesResponse.getResourcesList().isEmpty()) {
-                for (Resources resources : getAllowedResourcesResponse.getResourcesList()) {
-                  resourcesIds.addAll(resources.getResourceIdsList());
-                }
+              boolean allowedAllResources =
+                  RoleServiceUtils.checkAllResourceAllowed(getAllowedResourcesResponse);
+              if (allowedAllResources) {
+                return new ArrayList<>(requestedResourcesIds);
+              } else {
```
jkwatson-verta: this else isn't needed, since you return in the if block

backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
```diff
@@ -30,6 +32,16 @@ public RoleServiceUtils(AuthService authService, Integer timeout, UAC uac) {
     this.uac = uac;
   }
 
+  public static boolean checkAllResourceAllowed(Collection<Resources> resources) {
+    return resources.stream().allMatch(Resources::getAllResourceIds) && !resources.isEmpty();
```
jkwatson-verta: might as well put the `!resources.isEmpty()` first, since that's most likely cheaper than creating a stream (at the very least, it'll allocate one less object).

jkwatson-verta:
No tests for this change in functionality?

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
evgeniilezhnin: removed

backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
evgeniilezhnin: changed the order

evgeniilezhnin:
> No tests for this change in functionality?

Yes, you are right, I can add test case for that with the scenario that is described in the jira ticket

evgeniilezhnin:
> No tests for this change in functionality?

added project and dataset integration test

backend/common/src/main/java/ai/verta/modeldb/common/authservice/RoleServiceUtils.java
```diff
@@ -713,19 +716,29 @@ private void isSelfAllowed(
...
-    if (requestedIdList.size() == 1) {
-      isSelfAllowed(modelDBServiceResourceTypes, modelDBServiceActions, requestedIdList.get(0));
-      return requestedIdList;
+      ModelDBServiceActions modelDBServiceActions,
+      List<String> requestedResourceIds) {
+    if (requestedResourceIds.size() == 1) {
+      isSelfAllowed(
+          modelDBServiceResourceTypes, modelDBServiceActions, requestedResourceIds.get(0));
+      return requestedResourceIds;
     } else {
```
jkwatson-verta: redundant `else` here could be removed.

evgeniilezhnin: removed

evgeniilezhnin:
!rebuild

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -1216,7 +1218,14 @@ public List<String> getAccessibleExperimentRunIDs(
...
     Set<String> projectIdSet = new HashSet<>(projectIdExperimentRunIdMap.values());
-    allowedProjectIds.retainAll(projectIdSet);
+    boolean allowedAllResources = RoleServiceUtils.checkAllResourceAllowed(allowedProjects);
+    Set<String> allowedProjectIds;
+    if (allowedAllResources) {
+      allowedProjectIds = projectIdSet;
+    } else {
+      allowedProjectIds = RoleServiceUtils.getResourceIds(allowedProjects);
+      allowedProjectIds.retainAll(projectIdSet);
+    }
```
jkwatson-verta: this same block of code seems to be repeated. Worth making it a method in RoleServiceUtils so it's not repeated everywhere?

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
jkwatson-verta: looks like these redundant `else`s are still here....

jkwatson-verta (commented):
Seems reasonable. Just a few minor comments.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: yes, extracted

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
evgeniilezhnin: I open the file from the last commit and else on line FutureProjectDAO.java:880 is removed. But github shows for this conversation that FutureProjectDAO.java:880 is unchanged. Please try to open the changelog in github and select all commits. Or try to open the file or changelog not from github. We are discussing FutureProjectDAO.java:880, right?

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -1535,26 +1547,30 @@ private ExperimentRun filteredDatasetsBasedOnPrivileges(
...
         "Select vme.experimentRunEntity.id, cb From ConfigBlobEntity cb INNER JOIN VersioningModeldbEntityMapping vme ON vme.blob_hash = cb.blob_hash WHERE cb.hyperparameter_type = :hyperparameterType AND vme.experimentRunEntity.id IN (:expRunIds) ";
 
+    Set<String> accessibleResourceIds;
     if (mdbConfig.isPopulateConnectionsBasedOnPrivileges()) {
-      if (selfAllowedRepositoryIds == null || selfAllowedRepositoryIds.isEmpty()) {
+      boolean allowedAllResources =
+          RoleServiceUtils.checkAllResourceAllowed(selfAllowedRepositories);
+      if (allowedAllResources) {
         return new HashMap<>();
       } else {
```
jkwatson-verta: redundant else here.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -1535,26 +1547,30 @@ private ExperimentRun filteredDatasetsBasedOnPrivileges(
...
+        accessibleResourceIds = RoleServiceUtils.getResourceIds(selfAllowedRepositories);
         queryBuilder = queryBuilder + " AND vme.repository_id IN (:repoIds)";
       }
+    } else {
+      accessibleResourceIds = null;
```
jkwatson-verta: null? not an empty set? also, if you initialized the variable to this value, you wouldn't need to do it in an else block.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
```diff
@@ -1621,21 +1637,25 @@ private Map<String, Map<String, CodeVersion>> getExperimentRunCodeVersionMap(
...
+      boolean allowedAllResources =
+          RoleServiceUtils.checkAllResourceAllowed(selfAllowedRepositories);
+      if (allowedAllResources) {
         return new HashMap<>();
       } else {
+        accessibleRepositoryIds = RoleServiceUtils.getResourceIds(selfAllowedRepositories);
         queryBuilder = queryBuilder + " AND vme.repository_id IN (:repoIds)";
       }
+    } else {
+      accessibleRepositoryIds = null;
```
jkwatson-verta: same comment here. null seems bad vs. an empty Set, and you can assign this value in the declaration, rather than in an `else` block.

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
jkwatson-verta: this comment is about lines 887 and 889.

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: removed

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: null should be alright here too, but ok, replaced it with empty set and moved to initializer

backend/src/main/java/ai/verta/modeldb/experimentRun/ExperimentRunDAORdbImpl.java
evgeniilezhnin: replaced it with empty set and moved to initializer

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
```diff
@@ -453,6 +453,11 @@ public DeleteRepositoryRequest.Response deleteRepository(
   @Override
   public void deleteRepositories(
       Session session, ExperimentRunDAO experimentRunDAO, List<String> allowedRepositoryIds) {
+    deleteRepositoriesCollection(session, experimentRunDAO, allowedRepositoryIds);
+  }
+
+  public void deleteRepositoriesCollection(
```
jkwatson-verta: do we really need both these methods? Can't we just have one of them public?

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
jkwatson-verta: weird that github is showing this comment pointing at two different places, depending on the view.

evgeniilezhnin: removed

backend/src/main/java/ai/verta/modeldb/versioning/RepositoryDAORdbImpl.java
evgeniilezhnin: removed one of them

backend/src/main/java/ai/verta/modeldb/project/FutureProjectDAO.java
```diff
@@ -863,51 +863,36 @@ public InternalFuture<List<GetResourcesResponseItem>> deleteProjects(List<String
...
+              }
+              Set<String> allowedProjectIds =
+                  RoleServiceUtils.getResourceIds(getAllowedResourcesResponse);
+              // Validate if current user has access to the entity or not
+              // resourcesIds.retainAll(requestedResourcesIds);
+              if (requestedResourcesIds.isEmpty()) {
+                return allowedProjectIds;
+              } else if (allowedProjectIds.containsAll(requestedResourcesIds)) {
+                return requestedResourcesIds;
+              }
```
jkwatson-verta: one more redundant else here. ;)

evgeniilezhnin: I don't understand how it is reduntant. It has `if` in it

jkwatson-verta: line `886` returns, so if the return happens, the `else` will always execute. Hence, you don't need the `else`...just the next `if`.

evgeniilezhnin: removed

jkwatson-verta (approved):

conradoverta (commented):
This change is massive. I'm unsure how it solves the problem since I see a lot of unrelated refactoring (which makes it hard to spot the actual change). I'm also concerned about having to do such massive changes to handle particular cases of system admin permissions vs one consolidated approach (why does it need special treatment? it's unclear)

evgeniilezhnin:
> This change is massive. I'm unsure how it solves the problem since I see a lot of unrelated refactoring (which makes it hard to spot the actual change). 

We change utility functions to accept list of resources instead of resource ids, that is related to the change. Also I did some refactoring because there was a comment asking for a refactoring: https://github.com/VertaAI/modeldb/pull/3027#discussion_r878553111 . Please point to an unrelated refactoring line if there is another example of unrelated refactoring.

> I'm also concerned about having to do such massive changes to handle particular cases of system admin permissions vs one consolidated approach (why does it need special treatment? it's unclear)

We need all allowed flag in UAC because for system admin we don't know which resource ids are allowed

evgeniilezhnin:
> > This change is massive. I'm unsure how it solves the problem since I see a lot of unrelated refactoring (which makes it hard to spot the actual change).
> 
> We change utility functions to accept list of resources instead of resource ids, that is related to the change. Please point to an unrelated refactoring line.
> 
> > I'm also concerned about having to do such massive changes to handle particular cases of system admin permissions vs one consolidated approach (why does it need special treatment? it's unclear)
> 
> We need all allowed flag in UAC because for system admin we don't know which resource ids are allowed

I did some refactoring because there was a comment asking for a refactoring: https://github.com/VertaAI/modeldb/pull/3027#discussion_r878553111

conradoverta (approved):
LGTM

PR #3029, opened 2022-05-17
chore: [VR-13888] turn on OpenTelemetry http server tracing

jkwatson-verta:


backend/pom.xml
```diff
@@ -18,7 +18,8 @@
     <properties>
         <start-class>ai.verta.modeldb.App</start-class>
         <sonar.exclusions>src/main/java/ai/verta/modeldb/versioning/autogenerated/**/*</sonar.exclusions>
-        <otel.core.version>1.14.0</otel.core.version>
+        <otel.core.version>1.13.0</otel.core.version>
+        <otel.instrumentation.version>1.13.1-alpha</otel.instrumentation.version>
```
coryverta: Out of curiosity, what is the reason for the downgrade?

coryverta (approved):

backend/pom.xml
jkwatson-verta: at the time of writing, the instrumentation hadn't been updated to match 1.13 of the core. Looks like they've released in the meantime. I will update to the latest!

jkwatson-verta: done!

PR #3033, opened 2022-05-23
[VUMM-344] Exception handling for spring async endpoint

anandjakhaniya:


evgeniilezhnin (approved):

evgeniilezhnin:
LGTM, but please check build issues

evgeniilezhnin (approved):

evgeniilezhnin (approved):

anandjakhaniya:
!rebuild

backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
```diff
@@ -172,6 +175,64 @@ public static <T extends GeneratedMessageV3> void observeError(
...
+          message = e.getMessage();
+        }
+        logger.debug(logMessage, e.getMessage());
+      }
+    } else {
+      code = HttpStatus.INTERNAL_SERVER_ERROR;
+      // If getting 500 then we will not expose about error to the user.
+      message = "Internal Server Error";
+      logger.error(e.getMessage());
+      printStackTrace(logger, e);
```
jkwatson-verta, line 234: why not log with the exception and let the logger handle it rendering the stack trace?

anandjakhaniya: `printStackTrace` is our own function where we are logging stack trace as error using LOGGER and some extra stack trace as warn. 
```
public static void printStackTrace(Logger logger, Throwable t) {
    StackTraceElement[] stack = t.getStackTrace();
    logger.error("Stacktrace with {} elements for {}", stack.length, t);
    int n = 0;
    boolean isLongStack = stack.length > STACKTRACE_LENGTH;
    if (isLongStack) {
      for (; n < STACKTRACE_LENGTH + 1; ++n) {
        logger.warn("{}: {}", n, stack[n].toString());
      }
    }
    for (; n < stack.length; ++n) {
      if (stack[n].getClassName().startsWith("ai.verta") || !isLongStack) {
        logger.warn("{}: {}", n, stack[n].toString());
      }
    }
  }
```

conradoverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
jkwatson-verta: if it's a common pattern to log the message, then call this, we should make a function that does that, rather than have to do this everywhere.

backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
```diff
@@ -174,6 +178,64 @@ public static <T extends GeneratedMessageV3> void observeError(
     responseObserver.onError(logError(e));
   }
 
+  public static <T> void observeError(CompletableFuture<T> completableFuture, Throwable e) {
+    completableFuture.completeExceptionally(observeErrorWithResponse(e, e.getMessage(), LOGGER));
+  }
+
+  public static ResponseStatusException observeErrorWithResponse(
```
jkwatson-verta, line 185: There's a lot of code in here. We should have unit tests for this.

anandjakhaniya: We do not have a unit test for a modeldb backend so I do not have it for this but we have many Integration tests which used this code and validate exceptions.

jkwatson-verta: but, you could definitely add unit tests for this code, correct?

anandjakhaniya: Added unit test for this code.

jkwatson-verta (approved):

PR #3036, opened 2022-06-03
Improve app class without abstraction

anandjakhaniya:

PR #3042, opened 2022-06-10
fix: Fix up PR #3036

jkwatson-verta:
This is the same as #3036 with the addition of one additional conditional annotation on the ReconcilerInitializer to make sure it doesn't try to get created when doing migrations.


backend/src/main/java/ai/verta/modeldb/configuration/ReconcilerInitializer.java
```diff
@@ -23,10 +33,13 @@ public class ReconcilerInitializer {
   public static UpdateProjectTimestampReconcile updateProjectTimestampReconcile;
   public static SendEventsWithCleanUp sendEventsWithCleanUp;
 
-  public static void initialize(
-      Config config, ServiceSet services, DAOSet daos, FutureJdbi futureJdbi, Executor executor) {
+  @Bean
+  @ConditionalOnBean(DAOSet.class) // the DAOSet is explicitly not valid when doing migrations
```
jkwatson-verta, line 37: this is the important bit to make sure we don't try to create this bean while doing migrations. It's a bit of a hack. I recommend that we create a bean that is specifically crafted to be the thing that we use for this, rather than hacking it via the `null` return in the creation of the `DAOSet` bean at the moment.

coryverta (approved):

backend/src/test/java/ai/verta/modeldb/TestsInit.java
```diff
@@ -96,22 +101,25 @@ protected static void init() throws Exception {
 
     testConfig = TestConfig.getInstance();
     handleExecutor = FutureGrpc.initializeExecutor(testConfig.getGrpcServer().getThreadCount());
-    // Initialize services that we depend on
-    services = ServiceSet.fromConfig(testConfig, testConfig.artifactStoreConfig);
+
+    // TODO: FIXME: fix init flow as per spring bean initialization
```
liam-verta, line 105: Is this an upcoming task?  JIRA instead of to do in code?

liam-verta (approved):
Any way to write automated tests for this?

backend/src/test/java/ai/verta/modeldb/TestsInit.java
jkwatson-verta: this is just a re-do of Anand's original PR; I only modified the one line that I called out.

jkwatson-verta:
> Any way to write automated tests for this?

This is just a re-work of the spring initialization, so...either the app starts up or it doesn't. In this case, it was failing to start up during migrations...it would be good if we had an automated-test way to verify that migrations worked, and hopefully that becomes part of what we're going to accomplish with our testing rework in the upcoming quarter.

backend/src/test/java/ai/verta/modeldb/TestsInit.java
jkwatson-verta: @ad-47 do we need a Jira for this TODO?

PR #3044, opened 2022-06-13
chore: Improve the java HttpClient wrapper

jkwatson-verta:
* make it extend the HttpClient class
* make it propagate the tracing context


backend/common/src/main/java/ai/verta/modeldb/common/httpclient/TracingHttpClient.java
```diff
@@ -5,30 +5,44 @@
...
+import java.time.Duration;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
 
-public class TracingHttpClient {
-  private final HttpClient httpClient;
+public class TracingHttpClient extends HttpClient {
```
jkwatson-verta, line 28: I really wish the JDK had chosen to create an interface for `HttpClient` rather than forcing this inheritance on us. :(

coryverta (approved):

PR #3045, opened 2022-06-14
chore: allow H2 as a database type, and disable db creation with it.

jkwatson-verta:
This is required to allow integration tests to run against an in-memory H2 instance.

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
@@ -316,8 +316,13 @@ public abstract class CommonDBUtil {
     final var dbUrl = RdbConfig.buildDatabaseServerConnectionString(rdbConfiguration);
     LOGGER.debug("Connecting to DB URL " + dbUrl);
     Connection connection =
-        DriverManager.getConnection(
-            dbUrl, rdbConfiguration.getRdbUsername(), rdbConfiguration.getRdbPassword());
+            DriverManager.getConnection(
+                    dbUrl, rdbConfiguration.getRdbUsername(), rdbConfiguration.getRdbPassword());
+
+    if (rdbConfiguration.isH2()) {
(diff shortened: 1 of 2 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/config/RdbConfig.java
```diff
@@ -61,7 +61,14 @@ public boolean isMssql() {
     return RdbDialect.equals("org.hibernate.dialect.SQLServer2008Dialect");
   }
 
+  public boolean isH2() {
+    return RdbDialect.equals("org.hibernate.dialect.H2Dialect");
+  }
+
   public static String buildDatabaseConnectionString(RdbConfig rdb) {
+    if (rdb.isH2()) {
+      return buildDatabaseServerConnectionString(rdb);
```
jkwatson-verta, line 70: why do we have these 2 ways of building connection strings?

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -318,6 +318,11 @@ protected static void createDBIfNotExists(RdbConfig rdbConfiguration) throws SQL
     Connection connection =
         DriverManager.getConnection(
             dbUrl, rdbConfiguration.getRdbUsername(), rdbConfiguration.getRdbPassword());
+
+    if (rdbConfiguration.isH2()) {
+      LOGGER.debug("database driver is H2...skipping creation");
+      return;
+    }
```
liam-verta, line 325: Do we not need to create the DB when using H2?

liam-verta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
jkwatson-verta: H2 auto-creates the DB when it's accessed.

PR #3046, opened 2022-06-21
[VUMM-411] Fix bean initialization for when migration is enabled

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/configuration/Migration.java
```diff
@@ -13,10 +14,7 @@ public class Migration {
   private static final Logger LOGGER = LogManager.getLogger(Migration.class);
 
   public static boolean migrate(MDBConfig mdbConfig) throws Exception {
-    var liquibaseMigration =
-        Boolean.parseBoolean(
-            Optional.ofNullable(System.getenv(CommonConstants.ENABLE_LIQUIBASE_MIGRATION_ENV_VAR))
-                .orElse("false"));
+    var liquibaseMigration = new MigrationSetupConfig().isMigration();
```
jkwatson-verta: rather than constructing this here, can you pass in a `MigrationSetupConfig` instance that wraps the MDBConfig and provides everything that is needed for doing migrations?

anandjakhaniya: Improve code with better Bean creation

backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
```diff
@@ -156,6 +155,36 @@ public ServerBuilder<?> serverBuilder(
...
+    return args -> {
+      migration.migrate();
+      LOGGER.info("Migrations have completed");
+
+      var runLiquibaseSeparate = migrationSetupConfig.isRunSeparateMigration();
+      LOGGER.trace("run Liquibase separate: {}", runLiquibaseSeparate);
+      if (runLiquibaseSeparate) {
+        LOGGER.info("System exiting");
+        this.appContext.initiateShutdown(0);
+        System.exit(0);
```
jkwatson-verta, line 181: do we need this system.exit here? Won't this stop any graceful shutdown that might be in progress?

jkwatson-verta (approved):
I like it!

backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
anandjakhaniya: It will shutdown gracefully first and then we will exit the system

jkwatson-verta: I'm surprised something called "initiateShutdown" blocks until the shutdown is complete. /shrug

jkwatson-verta:
I tried deploying this and the app didn't finish starting up, with problems building the tracing instrumentation:

```
{"level":"ERROR","@timestamp":"2022-06-23T17:54:55.684Z","location":"ai.verta.modeldb.common.CommonDBUtil","method":"checkDBConnection","line":74,"message":"JdbiUtil checkDBConnection() got error ","exception":{"exception_class":"java.sql.SQLException","exception_message"
:"Unable to find a driver that accepts url: jdbc:mariadb://dev-env.rds.dev.verta.ai:3308/jw_modeldb?createDatabaseIfNotExist=true&useUnicode=yes&characterEncoding=UTF-8&sslEnabled=false&sslMode=DISABLED","stacktrace":"java.sql.SQLException: Unable to find a driver that ac
cepts url: jdbc:mariadb://dev-env.rds.dev.verta.ai:3308/jw_modeldb?createDatabaseIfNotExist=true&useUnicode=yes&characterEncoding=UTF-8&sslEnabled=false&sslMode=DISABLED\n\tat io.opentracing.contrib.jdbc.TracingDriver.findDriver(TracingDriver.java:237)\n\tat io.opentracin
g.contrib.jdbc.TracingDriver.connect(TracingDriver.java:164)\n\tat java.sql/java.sql.DriverManager.getConnection(DriverManager.java:677)\n\tat java.sql/java.sql.DriverManager.getConnection(DriverManager.java:228)\n\tat ai.verta.modeldb.common.CommonDBUtil.getDBConnection(
CommonDBUtil.java:66)\n\tat ai.verta.modeldb.common.CommonDBUtil.checkDBConnection(CommonDBUtil.java:71)\n\tat ai.verta.modeldb.common.CommonDBUtil.checkDBConnectionInLoop(CommonDBUtil.java:49)\n\tat ai.verta.modeldb.utils.CommonHibernateUtil.createOrGetSessionFactory(Com
monHibernateUtil.java:114)\n\tat ai.verta.modeldb.utils.CommonHibernateUtil.getSessionFactory(CommonHibernateUtil.java:165)\n\tat ai.verta.modeldb.reconcilers.SoftDeleteRepositories.resync(SoftDeleteRepositories.java:64)\n\tat ai.verta.modeldb.common.reconcilers.Reconcile
r.lambda$startResync$0(Reconciler.java:53)\n\tat java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)\n\tat java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)\n\tat java.base/java.util.concurrent.ScheduledThreadPoolExec
utor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)\n\tat java.ba
se/java.lang.Thread.run(Thread.java:829)\n"},"hostName":"modeldb--backend-64d78c899f-fbg4m"}
```

jkwatson-verta:
I think the simplest solution to the Reconciler initialization issue is to define a spring condition something like this:

```java
  public static class EnableReconcilers implements Condition {
    @Override
    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
      return liquibaseMigrationDisabled() || !liquibaseRunSeparately();
    }

    boolean liquibaseMigrationDisabled() {
      return "false".equalsIgnoreCase(System.getenv(CommonConstants.ENABLE_LIQUIBASE_MIGRATION_ENV_VAR));
    }

    boolean liquibaseRunSeparately() {
      return "true".equalsIgnoreCase(System.getenv(CommonConstants.RUN_LIQUIBASE_SEPARATE));
    }
  }
```

And then put this annotation on the ReconcilerInitializer:

```
@Conditional(EnableReconcilers.class)
```

It doesn't do anything to clean the rest of the bean initialization up, but this seems like a good first step to get things back running correctly.

backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
anandjakhaniya: `initiateShutdown` is our own method which called `SpringApplication.exit(this.applicationContext, () -> returnCode)` so it is also run on main thread so it will block and then at last `System.exit(0);` will call.

anandjakhaniya: If we are not adding `System.exit(0);` then `SpringApplication.exit(this.applicationContext, () -> returnCode)` will only triggered graceful shutdown but it will not exist the system and stay hold on manually exist system so we need to add `System.exit(0);`

anandjakhaniya:
> I think the simplest solution to the Reconciler initialization issue is to define a spring condition something like this:
> 
> ```java
>   public static class EnableReconcilers implements Condition {
>     @Override
>     public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
>       return liquibaseMigrationDisabled() || !liquibaseRunSeparately();
>     }
> 
>     boolean liquibaseMigrationDisabled() {
>       return "false".equalsIgnoreCase(System.getenv(CommonConstants.ENABLE_LIQUIBASE_MIGRATION_ENV_VAR));
>     }
> 
>     boolean liquibaseRunSeparately() {
>       return "true".equalsIgnoreCase(System.getenv(CommonConstants.RUN_LIQUIBASE_SEPARATE));
>     }
>   }
> ```
> 
> And then put this annotation on the ReconcilerInitializer:
> 
> ```
> @Conditional(EnableReconcilers.class)
> ```
> 
> It doesn't do anything to clean the rest of the bean initialization up, but this seems like a good first step to get things back running correctly.

Fix code as per this suggestion

backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
jkwatson-verta: ah, ok. we should probably change the name of that so it doesn't sound like it just triggers the shutdown, rather than also completing the shutdown process.

jkwatson-verta:
> > I think the simplest solution to the Reconciler initialization issue is to define a spring condition something like this:
> > ```java
> >   public static class EnableReconcilers implements Condition {
> >     @Override
> >     public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
> >       return liquibaseMigrationDisabled() || !liquibaseRunSeparately();
> >     }
> > 
> >     boolean liquibaseMigrationDisabled() {
> >       return "false".equalsIgnoreCase(System.getenv(CommonConstants.ENABLE_LIQUIBASE_MIGRATION_ENV_VAR));
> >     }
> > 
> >     boolean liquibaseRunSeparately() {
> >       return "true".equalsIgnoreCase(System.getenv(CommonConstants.RUN_LIQUIBASE_SEPARATE));
> >     }
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And then put this annotation on the ReconcilerInitializer:
> > ```
> > @Conditional(EnableReconcilers.class)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It doesn't do anything to clean the rest of the bean initialization up, but this seems like a good first step to get things back running correctly.
> 
> Fix code as per this suggestion

I'll try deploying it and see if it works!

backend/src/main/java/ai/verta/modeldb/configuration/EnableReconcilers.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+public class EnableReconcilers implements Condition {
+  @Override
+  public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+    return !liquibaseRunSeparately();
+  }
+
+  boolean liquibaseRunSeparately() {
+    return "true".equalsIgnoreCase("true");
```
jkwatson-verta: ??? this is not right, obviously. Also, I think we also need to make sure that we don't have liquibase enabled at all, correct?

jkwatson-verta (changes requested):

backend/src/main/java/ai/verta/modeldb/configuration/EnableReconcilers.java
jkwatson-verta: see my example I posted for what I think the correct logic is.

anandjakhaniya: I saw your logic and tried it but it did not work. Let say, if liquibase migration is enabled but the separate migration is not  then it will not work.. it will fail to initialise reconcilers.

anandjakhaniya: So I have Removed liquibase migration flag logic and just made a reliable in a separate migration flag.

jkwatson-verta: seems like this covers that case: ` return liquibaseMigrationDisabled() || !liquibaseRunSeparately();`

anandjakhaniya: Ok.. I have tried and it was not working in some cases. so I have removed it.  Let me try to check it again.

anandjakhaniya: @jkwatson-verta it was my problem we have used flag at multiple places so while doing testing i was forget to change flag at other place so i got errors. So your logic is right. i am updating PR.

anandjakhaniya: Updated PR.

backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
anandjakhaniya: @jkwatson-verta So what would you suggest for this? Which name you suggest for it?

jkwatson-verta: just `shutdown()` I guess? Let's not change it for this PR, though. Just something to think about.

backend/src/main/java/ai/verta/modeldb/configuration/EnableReconcilers.java
jkwatson-verta: ok, I'll give it a try and see if it works on deployment

jkwatson-verta: This still has the same issue with the database connectivity not working that I mentioned yesterday. :(

anandjakhaniya: the database connectivity not working is a separate issue. No?

jkwatson-verta: sure, but this PR leaves the db connectivity broken, so we can't merge it.

anandjakhaniya: I have checked at local and it is working fine. i do not face any issue with DB connectivity.

jkwatson-verta: that may be so, but this branch does not work when deployed to kubernetes, so we can't merge it.

anandjakhaniya: Fix DB connection issue at: https://github.com/VertaAI/modeldb/pull/3046/commits/6a735369a89d1f2e280a84dd04ed0e2d77969669 it was due session factory initialization. before this code we will initialized it from migration code but now it is conditional so when service up it was skip and further health check not getting that. so now it is fixed.

anandjakhaniya: After deploying on my dev environment `aj` i have noticed it is works fine as expected.

jkwatson-verta: cool. will verify, approve and merge! Thanks for tracking it down.

backend/src/main/java/ai/verta/modeldb/configuration/MigrationSetupConfig.java
```diff
@@ -0,0 +1,65 @@
...
+import ai.verta.modeldb.config.MigrationConfig;
+import java.util.List;
+import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+public class MigrationSetupConfig implements Condition {
```
jkwatson-verta, line 16: My only complaint with this is that "MigrationSetupConfig" doesn't sound at all like a condition, and hence we end up with a comment where it's used as a Condition. A cleaner, easier-to-understand approach would be to have a separate bean that is just the condition we need, i.e. something called (maybe) `MigrationsIncludedInAppStartup` so it's super clear what condition we're relying on. 

Could you make that change in a followup PR so we can get this one merged?

jkwatson-verta (approved):

backend/src/main/java/ai/verta/modeldb/configuration/MigrationSetupConfig.java
anandjakhaniya: Yes i will do it in follow up PR instead of blocking this.

anandjakhaniya: ticket for this: https://vertaai.atlassian.net/browse/VUMM-432

anandjakhaniya: PR: https://github.com/VertaAI/modeldb/pull/3051

PR #3049, opened 2022-06-29
chore: uppercase the liquibase table names

jkwatson-verta:
This is to account for a bug in liquibase when working with an H2 database.

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
@@ -288,14 +288,15 @@ public abstract class CommonDBUtil {
       throws InterruptedException, LiquibaseException, SQLException {
     // Change liquibase default table names
     if (config.getRdbConfiguration().isH2()) {
-      //H2 upper cases all table names, and liquibase has issues if you don't make this also upper case.
+      // H2 upper cases all table names, and liquibase has issues if you don't make this also upper
+      // case.
       System.getProperties().put("liquibase.databaseChangeLogTableName", "DATABASE_CHANGE_LOG");
       System.getProperties()
           .put("liquibase.databaseChangeLogLockTableName", "DATABASE_CHANGE_LOG_LOCK");
     } else {
       System.getProperties().put("liquibase.databaseChangeLogTableName", "database_change_log");
       System.getProperties()
-              .put("liquibase.databaseChangeLogLockTableName", "database_change_log_lock");
+          .put("liquibase.databaseChangeLogLockTableName", "database_change_log_lock");
     }
 
     // Lock to RDB for now
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
@@ -292,8 +292,8 @@ public abstract class CommonDBUtil {
     if (config.getRdbConfiguration().isH2()) {
       // H2 upper cases all table names, and liquibase has issues if you don't make this also upper
       // case.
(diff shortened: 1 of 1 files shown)


eheinlein-verta (commented):
There is code in UAC that expects the table name to be lowercased (https://github.com/VertaAI/verta-uacservice/blob/013d8424b06144d7968b766a9957b4b5c847e612/src/main/java/ai/verta/uac/App.java#L644) specifically because we switched from the old name to the new one that was being used in modeldb. Does this take that into account? (By which I mean has it been tested in UAC, and if need be is there a separate PR for UAC?)

jkwatson-verta:
> There is code in UAC that expects the table name to be lowercased (https://github.com/VertaAI/verta-uacservice/blob/013d8424b06144d7968b766a9957b4b5c847e612/src/main/java/ai/verta/uac/App.java#L644) specifically because we switched from the old name to the new one that was being used in modeldb. Does this take that into account? (By which I mean has it been tested in UAC, and if need be is there a separate PR for UAC?)

Tables names are not supposed to be case-sensitive (you can always query against either case in every RDBMS I've ever come across), which is why I believe there is a bug in liquibase about this. If we're dependent on table name casing, then we also have a bug.

It looks like UAC already deals with this, though: https://github.com/VertaAI/verta-uacservice/blob/main/src/main/java/ai/verta/uac/App.java#L599-L601

eheinlein-verta (approved):

jkwatson-verta:
!rebuild

jkwatson-verta:
!rebuild

jkwatson-verta:
> > There is code in UAC that expects the table name to be lowercased (https://github.com/VertaAI/verta-uacservice/blob/013d8424b06144d7968b766a9957b4b5c847e612/src/main/java/ai/verta/uac/App.java#L644) specifically because we switched from the old name to the new one that was being used in modeldb. Does this take that into account? (By which I mean has it been tested in UAC, and if need be is there a separate PR for UAC?)
> 
> Tables names are not supposed to be case-sensitive (you can always query against either case in every RDBMS I've ever come across), which is why I believe there is a bug in liquibase about this. If we're dependent on table name casing, then we also have a bug.
> 
> It looks like UAC already deals with this, though: https://github.com/VertaAI/verta-uacservice/blob/main/src/main/java/ai/verta/uac/App.java#L599-L601

I decided to verify this worked before I merged, and good thing I did. It looks like Amazon Aurora actually has case-sensitive table names, so this totally broke things. I updated it to only uppercase when H2 is the database. Thanks for spotting that, @eheinlein-verta , as that would have been a really bad thing to merge!

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -288,9 +287,16 @@ protected void runLiquibaseMigration(
...
-    System.getProperties().put("liquibase.databaseChangeLogTableName", "database_change_log");
-    System.getProperties()
-        .put("liquibase.databaseChangeLogLockTableName", "database_change_log_lock");
+    String changeLogTableName = "database_change_log";
+    String changeLogLockTableName = "database_change_log_lock";
+    if (config.getRdbConfiguration().isH2()) {
+      // H2 upper cases all table names, and liquibase has issues if you don't make this also upper
+      // case.
+      changeLogLockTableName = changeLogLockTableName.toUpperCase();
+      changeLogTableName = changeLogTableName.toUpperCase();
```
jkwatson-verta:
```suggestion
      changeLogTableName = changeLogTableName.toUpperCase();
      changeLogLockTableName = changeLogLockTableName.toUpperCase();
```

eheinlein-verta (approved):

PR #3051, opened 2022-07-05
[VUMM-411] improve code as per suggestion

anandjakhaniya:


backend/src/main/java/ai/verta/modeldb/configuration/MigrationsIncludedInAppStartup.java
```diff
@@ -13,14 +13,14 @@
 import org.springframework.context.annotation.ConditionContext;
 import org.springframework.core.type.AnnotatedTypeMetadata;
 
-public class MigrationSetupConfig implements Condition {
-  private static final Logger LOGGER = LogManager.getLogger(MigrationSetupConfig.class);
+public class MigrationsIncludedInAppStartup implements Condition {
```
jkwatson-verta: Heh. I think my intention was to separate the condition from the config, and have 2 different beans, rather than just renaming the config bean.

anandjakhaniya: Fix trying to implement more clearly using separate beans

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
```diff
@@ -0,0 +1,36 @@
...
+
+import ai.verta.modeldb.common.CommonConstants;
+import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+public class RunLiquibaseSeparate implements Condition {
```
jkwatson-verta: how about `RunLiquibaseSeparately`?

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
```diff
@@ -0,0 +1,36 @@
...
+  private static final Logger LOGGER = LogManager.getLogger(RunLiquibaseSeparate.class);
+
+  public RunLiquibaseSeparate() {}
```
jkwatson-verta: you don't need an explicit no-arg constructor, You get one for free from the compiler.

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public boolean liquibaseRunSeparately() {
```
jkwatson-verta: does this method need to be public?

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
```diff
@@ -0,0 +1,36 @@
...
+    LOGGER.info("RUN_LIQUIBASE_SEPARATE: {}", liquibaseRunSeparately);
+    return false;
+  }
+
+  @Override
+  public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+    return liquibaseRunSeparately();
+  }
+
+  public static class InversedRunLiquibaseSeparate extends RunLiquibaseSeparate {
```
jkwatson-verta: how about `RunMigrationsWithApp`?

anandjakhaniya: it is not about migration. If `liquibaseRunSeparately=true` then it will decide to run only Liquibase and migration separately from the actual service and it will not run the registry backend but if `liquibaseRunSeparately=false` then it will execute registry backend and not run the Liquibase.

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
anandjakhaniya: Removed

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
anandjakhaniya: make it private

backend/src/main/java/ai/verta/modeldb/configuration/RunLiquibaseSeparate.java
jkwatson-verta: how is liquibase not about migration? liquibase *is* our migration system, no? 

the "inverse" of running migrations separately is that we run them *with* the app, hence my suggestion. Calling something "InverseX" is definitely going to be confusing to read.

anandjakhaniya: Ok, All confusion is due to the name of the flags. 
So we have two flags with two different conditions.

1. ENABLE_LIQUIBASE_MIGRATION_ENV_VAR
- This flag will decide whether to run the Liquibase migration and the code migrations or not

2. RUN_LIQUIBASE_SEPARATE
- This flag will decide whether to execute the above process (first point) separately from the actual registry backend or execute it with the actual registry backend.

**Usecases:**  
1. If I want to run the Liquibase migration separately from the actual registry backend then I will set the above two flags like 
```
ENABLE_LIQUIBASE_MIGRATION_ENV_VAR = true
RUN_LIQUIBASE_SEPARATE = true
```
Which is currently we are using in our cluster setup I think. where in the prehook of Kubernetes pod we are setting flags like 
```
ENABLE_LIQUIBASE_MIGRATION_ENV_VAR = true
RUN_LIQUIBASE_SEPARATE = true
```

and then when the actual backend service pod will start after webhook we will set those false like 
```
ENABLE_LIQUIBASE_MIGRATION_ENV_VAR = false
RUN_LIQUIBASE_SEPARATE = false
```

2. Now in the second scenario, let's say I do not want to configure prehook but still I have to run the migration at that time we will not execute migration separately so instead of that we will execute it with our main service. at that time those flags should be like 
```
ENABLE_LIQUIBASE_MIGRATION_ENV_VAR = true
RUN_LIQUIBASE_SEPARATE = false
```
In this scenario, I am using this for my local setup as well because I do not have any prehook steps like Kubernetes.

jkwatson-verta: Sure. All of that is correct. I just want to make sure we name our classes/conditions something that makes it clear what they are for, rather than starting with the word "Inverse", which is very confusing and hard to understand the purpose of.

anandjakhaniya: Renamed it

backend/src/main/java/ai/verta/modeldb/configuration/EnabledMigration.java
```diff
@@ -27,7 +27,7 @@ public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata)
     return isMigration();
   }
 
-  public static class InversedEnabledMigration extends EnabledMigration {
+  public static class DisabledMigration extends EnabledMigration {
```
jkwatson-verta: *please* don't introduce a class hierarchy for this. The logic is so simple, just repeat it if you need to.

jkwatson-verta: Also, would personally prefer this were named "MigrationDisabled"

anandjakhaniya: Fixed

backend/src/main/java/ai/verta/modeldb/configuration/EnabledMigration.java
```diff
@@ -0,0 +1,36 @@
...
+    LOGGER.info("ENABLE_LIQUIBASE_MIGRATION_ENV_VAR: {}", enableLiquibaseMigrationEnvVar);
+    return enableLiquibaseMigrationEnvVar;
+  }
+
+  @Override
+  public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+    return isMigration();
+  }
+
+  public static class MigrationDisabled implements Condition {
```
jkwatson-verta, line 30: does this need `@Component` on it as well, or does spring assume that because the outer class is?

jkwatson-verta (approved):
Thanks! I'm not sure if we need `@Component` on the conditions or not. But, this is looking good.

backend/src/main/java/ai/verta/modeldb/configuration/EnabledMigration.java
anandjakhaniya: it is not compulsory. spring will assume based on the outer class or even based on the package name added in the component scan list.

PR #3052, opened 2022-07-06
chore: Add the capability of processing futures synchronously

jkwatson-verta:
Also: fixed a broken http client tracing test
Also: removed un-needed no-op tracing method in the InternalFuture class

 

## Impact and Context
In many places in our code, we use asynchronous processing where it is not actually needed, including some very simple input validation. By forcing asynchronous execution, we can introduce a considerable amount of unneeded thread jumping, which is a relatively expensive operation. 

This PR will enable processing future responses synchronously, but only if explicitly requested via the new `*Sync` methods. 

## Risks and Area of Effect
Since this is only additive to the InternalFuture API, it has very low immediate risk.


jkwatson-verta:
closing. will reopen with just the cleanup and bugfix in it for now.

PR #3071, opened 2022-07-29
feat: Split out url property

liuverta:
## Impact and Context

Occasionally-requested and low-cost!

Exposes `experiment_run.url` and such.

## Risks and Area of Effect

Low risk: Straightforward copy-and-paste.

The URL was a hack I did years ago for the novelty, and this cements it in our public API, but technically having it in the `__repr__()` already made it public anyway.

## Testing

I ran this snippet and verified that the URLs are correct:

```python
from verta import Client

client = Client()

reg_model = client.create_registered_model()
for entity in [
    client.create_dataset(),
    client.get_or_create_endpoint("banana"),
    reg_model,
    reg_model.create_version(),
    client.create_experiment_run(),
    client.proj,
]:
    print(entity.url)
```

I also ran all of our `test_repr` tests in all supported Python versions, and there are no related failures.

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/test_entities.py
+++ b/client/verta/tests/test_entities.py
@@ -196,6 +196,14 @@ class TestProject:
         with pytest.raises(ValueError):
             client.set_project(id="nonexistent_id")
 
+    def test_repr(self, client):
+        proj = client.create_project()
+        str_repr = str(proj)
+
+        assert proj.name in str_repr
+        assert proj.url in str_repr
+        assert proj.id in str_repr
+
     @pytest.mark.parametrize("tags", [TAG, [TAG]])
     def test_tags_is_list_of_str(self, client, tags):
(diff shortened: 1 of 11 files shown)


client/verta/tests/test_entities.py
```diff
@@ -339,9 +347,12 @@ def test_get_nonexistent_id_error(self, client):
         with pytest.raises(ValueError):
             client.set_experiment_run(id="nonexistent_id")
 
-    def test_no_experiment_error(self, client):
-        with pytest.raises(AttributeError):
-            client.set_experimennt_run()
```
liuverta, line 344: This test provides literally no value who let this happen wow

conradoverta (approved):
Nice

PR #3076, opened 2022-08-08
feat: [CAT-48] add new task type attributes to proto

mgonc-verta:
## Impact and Context
Add new props to be served in registry service


protos/scripts/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
```
conradoverta, line 1: ?

conradoverta (commented):
Jenkins should be able to handle this for you

PR #3077, opened 2022-08-09
[VUMM-469] ModelDB fails to create database in fresh SQL Server environment

anandjakhaniya:
## Impact and Context

Before migration, we have code about creating hibernate session factory which causes the issue. If we do not have a database then hibernate tries to get a connection for DB and it failed so now I have fixed logic and initialize hibernate after the DB creation.

### DIFF
--- a/backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
+++ b/backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
@@ -126,9 +126,6 @@ public class AppConfigBeans {
   @Bean
   @Conditional(RunLiquibaseWithMainService.class)
   public DAOSet daoSet(MDBConfig config, ServiceSet services, Executor grpcExecutor) {
-    var modelDBHibernateUtil = ModelDBHibernateUtil.getInstance();
-    modelDBHibernateUtil.initializedConfigAndDatabase(config, config.getDatabase());
-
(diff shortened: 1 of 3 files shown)


backend/src/main/java/ai/verta/modeldb/configuration/AppConfigBeans.java
```diff
@@ -170,6 +167,15 @@ public Migration migration(MDBConfig mdbConfig) throws Exception {
     return migration;
   }
 
+  @Bean
+  public ModelDBHibernateUtil hibernateUtil(MDBConfig mdbConfig) {
```
jkwatson-verta, line 171: what is this `@Bean` definition used for?

anandjakhaniya: Nothing it just gives a surety about session factory is created. that bean is not used anywhere.

jkwatson-verta: shouldn't it be injected into the `Migration` bean, then?

jkwatson-verta: this can be addressed as a separate change, but we should be sure to use the beans when we can

jkwatson-verta (approved):
approving for the bugfix; needs a followup on the bean usage

PR #3080, opened 2022-08-12
chore: [VUMM-392] Add InternalFuture with default Executor

liam-verta:
## Impact and Context
This change adds an implementation of `InternalFuture` that keeps the same `executor` across method calls rather than passing it to every call.


backend/common/src/main/java/ai/verta/modeldb/common/futures/InternalFuture.java
```diff
@@ -86,24 +90,37 @@ public static <T> InternalFuture<List<T>> sequence(
...
+    return from(CompletableFuture.completedFuture(value));
+  }
+
+  public static InternalFutureWithDefaultExecutor.FactoryWithExecutor withExecutor(
+      Executor executor) {
+    return new InternalFutureWithDefaultExecutor.FactoryWithExecutor(executor);
+  }
+
+  public InternalFutureWithDefaultExecutor<T> thenWithExecutor(Executor executor) {
+    return new InternalFutureWithDefaultExecutor<>(this.stage).thenWithExecutor(executor);
```
jkwatson-verta: we could just construct the IFWDE with the executor, yeah?

backend/common/src/main/java/ai/verta/modeldb/common/futures/InternalFutureWithDefaultExecutor.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  @Override
+  @Deprecated
+  public <U> InternalFutureWithDefaultExecutor<U> thenSupply(
+      Supplier<InternalFuture<U>> supplier, Executor executor) {
+    return super.thenSupply(supplier, executor).thenWithExecutor(defaultExecutor);
+  }
+
+  public static class FactoryWithExecutor {
```
jkwatson-verta: i think I'd either call this simply `Factory` if we want to leave it as an inner class, or pull it up and call it `InternalFutureFactory`.

backend/common/src/test/java/ai/verta/modeldb/common/futures/InternalFutureTest.java
```diff
@@ -14,35 +14,35 @@ class InternalFutureTest {
...
   }
 
   @Test
   void thenSupply() {
     AtomicBoolean firstWasCalled = new AtomicBoolean();
-    Executor executor = MoreExecutors.directExecutor();
-    InternalFuture<Void> testFuture =
-        InternalFuture.supplyAsync(
+    final Executor executor = MoreExecutors.directExecutor();
+    final InternalFutureWithDefaultExecutor.FactoryWithExecutor factory =
```
jkwatson-verta: nit: these `final` modifiers don't add much value, and clutter up the code.
```suggestion
   Executor executor = MoreExecutors.directExecutor();
   InternalFutureWithDefaultExecutor.FactoryWithExecutor factory =
```

backend/common/src/test/java/ai/verta/modeldb/common/futures/InternalFutureTest.java
```diff
@@ -51,22 +51,21 @@ void thenSupply() {
   void thenSupply_exception() {
     AtomicBoolean secondWasCalled = new AtomicBoolean();
     Executor executor = MoreExecutors.directExecutor();
-    InternalFuture<Void> testFuture =
-        InternalFuture.supplyAsync(
+    final InternalFutureWithDefaultExecutor.FactoryWithExecutor factory =
```
jkwatson-verta:
```suggestion
   InternalFutureWithDefaultExecutor.FactoryWithExecutor factory =
```

jkwatson-verta:
This is looking pretty good. I'd add some unit tests for the new functionality. I'm not 100% sure how to verify that the proper executor is being used... maybe with some custom executor implementation?  Would definitely at the very least add tests to verify that things work (we have needed these tests for a long time now), and that a null executor makes things fail fast.

PR #3088, opened 2022-08-24
[VUMM-528] Fixes to make registry test cases to spring boot compatible

anandjakhaniya:

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/configuration/AppContext.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/configuration/AppContext.java
@@ -12,8 +12,12 @@ import org.springframework.lang.NonNull;
 @Configuration
 public class AppContext implements ApplicationContextAware {
 
-  private static final AppContext appContext = new AppContext();
+  private static AppContext appContext;
   private ApplicationContext applicationContext;
+
+  public AppContext () {
+    appContext = this;
+  }
 
   public static AppContext getInstance() {
     return appContext;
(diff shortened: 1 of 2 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/configuration/AppContext.java
```diff
@@ -12,9 +12,13 @@
 @Configuration
 public class AppContext implements ApplicationContextAware {
 
-  private static final AppContext appContext = new AppContext();
+  private static AppContext appContext;
   private ApplicationContext applicationContext;
 
+  public AppContext() {
+    appContext = this;
```
jkwatson-verta: Why are you overriding a static method in this constructor? This doesn't seem like a good pattern to me. If two callers construct one of these, you've now remove a context altogether.

anandjakhaniya: I want to access the app context class at the init test setup. If we create an App context like code before changes then I can't access application context in the Test Setup class. So I have introduced this context pattern so that it will take the same reference of class and I can access it there.

jkwatson-verta: I understand, but you need to find a better way. This also leaks the class reference out of the constructor which is a dangerous antipattern, since it could be accessed by another thread before it has finished being initialized.

anandjakhaniya: Ok I will check and will find a new good pattern..

anandjakhaniya: Reverted changes and added context from App.java class at the registry backend.

jkwatson-verta (approved):

PR #3089, opened 2022-08-24
[VUMM-612] Make a backend to H2 Compatible and resolve Liquibase path issue after version upgrade

anandjakhaniya:

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/config/RdbConfig.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/config/RdbConfig.java
@@ -94,6 +94,10 @@ public class RdbConfig {
       return rdb.DBConnectionURL;
     }
 
+    if (rdb.isH2()) {
+      return rdb.RdbUrl + ";DB_CLOSE_DELAY=-1";
+    }
+
     if (rdb.isMssql()) {
       return rdb.RdbUrl;
     }
--- a/backend/common/pom.xml
+++ b/backend/common/pom.xml
@@ -244,6 +244,12 @@
             <version>3.22.0</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/com.h2database/h2 -->
+        <dependency>
+            <groupId>com.h2database</groupId>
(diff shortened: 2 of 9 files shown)


anandjakhaniya:
@liam-verta  I have taken h2 dependencies from the UAC service. I don't know what I am doing wrong here. Could check it, please?

backend/common/src/main/java/ai/verta/modeldb/common/config/RdbConfig.java
```diff
@@ -94,6 +94,10 @@ public static String buildDatabaseServerConnectionString(RdbConfig rdb) {
       return rdb.DBConnectionURL;
     }
 
+    if (rdb.isH2()) {
+      return rdb.RdbUrl + ";DB_CLOSE_DELAY=-1";
```
jkwatson-verta, line 98: any idea why we didn't need this in UAC?

anandjakhaniya: > any idea why we didn't need this in UAC?

I have tried to run uac without this and I am getting the error. To reproduce the issue. You have to start uac using the H2 database configuration.

jkwatson-verta: but... we run UAC without this on every PR, and we don't get an error?

anandjakhaniya: > but... we run UAC without this on every PR, and we don't get an error?

It surprised me too. If we are running uac service not test cases then I get the error..

jkwatson-verta: I just started it and it worked fine. 🤔

jkwatson-verta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/config/RdbConfig.java
jkwatson-verta: I do see that this helps things in the registry though. 🤔

PR #3132, opened 2022-08-30
test: Update expected error message

liuverta:
## Impact and Context

We've switched from Flask to CherryPy so this assertion has been failing. I'm changing it to just check that the error (probably) contains a stack trace.

## Risks and Area of Effect

—

## Testing

—

## How to Revert

Revert this PR.


liuverta:
The failed check is Jenkins being mischievous. This PR is highly unlikely to cause a linting issue.

ghost (approved):

PR #3153, opened 2022-09-02
[VUMM-297] Move artifact store to common module

anandjakhaniya:
## Impact and Context

We have an artifact store with two different options which are S3 and NFS. so for that, we have created two services using spring boot to serve artifact store functionality. In the current situation, This functionality is used by two backends which are Registry and ModelDB and both the backends have their own code so whenever we have updated something in one place we have to manage it on the other backends as well. so to protect duplication and uniqueness of the code we need these changes in a common place so that both backends are used same code and we can manage and maintain the code from one place.


anandjakhaniya:
@jkwatson-verta

backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
```diff
@@ -65,6 +66,18 @@ public static String getStringFromProtoObject(MessageOrBuilder object) {
     }
   }
 
+  public static void logAmazonServiceExceptionErrorCodes(Logger LOGGER, AmazonServiceException e) {
+    LOGGER.info("Amazon Service Status Code: " + e.getStatusCode());
+    LOGGER.info("Amazon Service Error Code: " + e.getErrorCode());
+    LOGGER.info("Amazon Service Error Type: " + e.getErrorType());
+    LOGGER.info("Amazon Service Error Message: " + e.getErrorMessage());
```
jkwatson-verta: Is info the right level for this? If this is truly for errors, should it at least be at WARN? Or if it's not actually errors, maybe DEBUG would be better.

anandjakhaniya: Added logs with DEBUG

jkwatson-verta (approved):
Looks good. Thanks!

PR #3157, opened 2022-09-02
test: Sidestep parallelism issue with config file tests

liuverta:
## Impact and Context

The tests in `test_config.py` involve the creation and deletion of a Verta config file in one's home directory. This results in conflicts when the tests are run concurrently.

`pytest-xdist` offers a way to group tests to indicate that they should be run on the same worker, which has the effect of preventing tests in the same group from running simultaneously (since the worker can only run one at a time).

To make use of the group, `pytest` does need to be run with the `--dist loadgroup` option; I've opened https://github.com/VertaAI/cluster-setup/pull/5185 to do just that in our CI. There doesn't seem to be an easy way to just set this as the default in the pytest config 😞 

## Risks and Area of Effect

—

### Before

A failure and an error on most test runs:

```
% pytest -n 10 test_config.py -ra
. . .
================================================== short test summary info ==================================================
ERROR test_config.py::TestRead::test_discovery - FileNotFoundError: [Errno 2] No such file or directory: '/Users/miliu/.ve...
FAILED test_config.py::TestRead::test_discovery - AssertionError: assert ['/private/va..._config.yaml'] == ['/private/va.....
==================================== 1 failed, 6 passed, 13 warnings, 1 error in 10.69s =====================================
```

### After

Success every time!

```
% pytest -n 10 --dist loadgroup test_config.py -ra
. . .
============================================== 7 passed, 13 warnings in 9.00s ===============================================
```

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/test_config.py
+++ b/client/verta/tests/test_config.py
@@ -11,6 +11,9 @@ from verta._internal_utils import _config_utils
 from . import utils
 
 
+pytestmark = [pytest.mark.xdist_group(name="verta_config")]
+
+
 @pytest.fixture
 def expected_config(in_tempdir):
     """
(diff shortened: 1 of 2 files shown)


liuverta:
The 3.10 pylint job is failing because we don't have the CI fully set up for that yet 😕

jared-verta (approved):

PR #2927, opened 2022-03-08
[VUMM-214] Modified logic for untrusted artifacts

anandjakhaniya:
## Impact and Context

Reopen with fixes after reverting PR: https://github.com/VertaAI/modeldb/pull/2925

PR #3159, opened 2022-09-02
chore!: Remove artifact path-only logging

liuverta:
## Impact and Context

Since our implementation of untrusted artifact paths in March (https://github.com/VertaAI/modeldb/pull/2927), the backend rejects the client's artifact path and substitutes its own. This breaks `ExperimentRun`'s `log_artifact_path()` and `log_image_path()`, and it would be a significant undertaking to re-enable that functionality. This PR removes the methods from our API.

## Risks and Area of Effect

These APIs will no longer be available through the client, though they also have not been operational at all since early this year.

## Testing

A search of the client source and tests for `\.log[_a-z]+_path` confirms that it has been removed completely.

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/tests/test_experimentrun/test_artifacts.py
+++ b/client/verta/tests/test_experimentrun/test_artifacts.py
@@ -9,18 +9,6 @@ from verta._internal_utils import _artifact_utils
 
 
 class TestArtifacts:
-    def test_log_path(self, experiment_run, strs):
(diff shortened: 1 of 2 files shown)


liuverta:
!rebuild

liuverta:
The 3.10 pylint job is failing because we don't have the CI fully set up for that yet 😕

conradoverta (approved):

PR #3162, opened 2022-09-04
chore!: Remove get_artifact_parts() from public API

liuverta:
## Impact and Context

- removes `ModelVersion.get_artifact_parts()` because its endpoint doesn't exist in the backend (https://github.com/VertaAI/registry/pull/1024)
- makes `ExperimentRun.get_artifact_parts()` private since it's not meant for public use

## Risks and Area of Effect

Negligible: `ModelVersion.get_artifact_parts()` never worked, and `ExperimentRun.get_artifact_parts()` doesn't have much real-world use and was never visible in our API documentation.

## Testing

I ran the test locally and it succeeds now that it's unburdened by `ModelVersion.get_artifact_parts()`.

And honestly, this test should involve a mocked server or artifact uploader, but the client test suite doesn't have the pieces in place to make that happen 🤷 

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/verta/tracking/entities/_experimentrun.py
+++ b/client/verta/verta/tracking/entities/_experimentrun.py
@@ -408,6 +408,24 @@ class ExperimentRun(_DeployableEntity):
 
             return response.content, artifact.path_only
 
+    def _get_artifact_parts(self, key):
+        endpoint = "{}://{}/api/v1/modeldb/experiment-run/getCommittedArtifactParts".format(
+            self._conn.scheme,
+            self._conn.socket,
+        )
+        data = {'id': self.id, 'key': key}
+        response = _utils.make_request(
+            "GET", endpoint, self._conn, params=data)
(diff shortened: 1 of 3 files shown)


client/verta/verta/tracking/entities/_experimentrun.py
```diff
@@ -408,6 +408,24 @@ def _get_artifact(self, key):
...
+        _utils.raise_for_http_error(response)
+
+        committed_parts = _utils.body_to_json(
+            response).get('artifact_parts', [])
+        committed_parts = list(sorted(
+            committed_parts,
+            key=lambda part: int(part['part_number']),
+        ))
+        return committed_parts
+
```
liuverta, line 428: Moved from below; only change is prepending `_` to the function name.

coryverta (approved):

conradoverta (approved):

PR #3175, opened 2022-09-07
Fix liquibase path after liquibase version upgrade

anandjakhaniya:
## Impact and Context

After Liquibase version migration, modeldb was unable to up and threw error like 
```
Specifying files by absolute path was removed in Liquibase 4.0. Please use a relative path or add '/' to the classpath parameter.
```
Solution: An added solution based on the Liquibase official suggestion at [How the Liquibase searchPath works in 4.0 and later versions](https://docs.liquibase.com/concepts/changelogs/how-liquibase-finds-files.html)


jkwatson-verta (approved):
good find! thanks!

PR #3185, opened 2022-09-07
chore: don't start up the server if we shouldn't

jkwatson-verta:

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/configuration/ServerEnabled.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/configuration/ServerEnabled.java
@@ -14,6 +14,10 @@ import org.springframework.stereotype.Component;
 public class ServerEnabled implements Condition {
   @Override
   public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+    return serverIsEnabled();
+  }
+
+  public static boolean serverIsEnabled() {
     var enableLiquibase =
         Boolean.parseBoolean(
             Optional.ofNullable(System.getenv(CommonConstants.ENABLE_LIQUIBASE_MIGRATION_ENV_VAR))
--- a/backend/src/main/java/ai/verta/modeldb/App.java
+++ b/backend/src/main/java/ai/verta/modeldb/App.java
@@ -45,5 +45,4 @@ public class App {
     application.addListeners(new ApplicationPidFileWriter(pathToPidFile));
     application.run(args);
   }
-
 }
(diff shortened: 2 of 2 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -173,7 +173,8 @@ protected void createTablesLiquibaseMigration(
       DatabaseConfig config,
       String changeSetToRevertUntilTag,
       String liquibaseRootPath,
-      ResourceAccessor resourceAccessor)
+      ResourceAccessor resourceAccessor,
+      boolean newLiquibaseFilenames)
```
liam-verta: 😭

liam-verta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
jkwatson-verta: yeah, this is not awesome. Especially as I can't figure out how to know which filename structure a given app is going to use!

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -248,6 +256,22 @@ protected void createTablesLiquibaseMigration(
...
   }
 
+  private static boolean changeLogTableExists(JdbcConnection jdbcCon, String changeLogTableName)
+      throws DatabaseException, SQLException {
+    boolean changeLogTableExists = false;
+    DatabaseMetaData metaData = jdbcCon.getMetaData();
+    ResultSet tables = metaData.getTables(null, null, "%", null);
+    while (tables.next()) {
+      String tableName = tables.getString(3);
+      if (changeLogTableName.equalsIgnoreCase(tableName)) {
```
eheinlein-verta: I think it might be a good idea to consider camelCase vs snake_case here, too.  I was looking back on the names that we had to deal with before, and [one was `DATABASECHANGELOG` and the other is `database_change_log`](https://github.com/VertaAI/verta-uacservice/blob/3324568498b17231ddeb0542e17ec5d184bb3745/src/main/java/ai/verta/uac/configuration/Migration.java#L26-L29). Not certain if that is applicable here, but wanted to mention it.

jkwatson-verta: hmm. the main purpose here was not to try to update a table that doesn't exist if we had an empty database. Do you have another idea about how to do that?

jkwatson-verta: if it turns out that the liquibase table (as configured) doesn't exist, then we shouldn't try to update it. So, I *think* this is a safe way to check.

eheinlein-verta: Cool 👍

jkwatson-verta:
I'm going to clean this up and get rid of the liquibase changes in it... and just make it so we don't start up the server, like it says on the tin. :)

eheinlein-verta (approved):

PR #3209, opened 2022-09-12
chore: Publish client v0.20.4

liuverta:
## Impact and Context

This is a branch off of tag `client-v0.20.3`. During the publication process, the commit resulting from this PR will be tagged `client-v0.20.4`—as per the usual process.

## Risks and Area of Effect

Since this is a branch—because `main` contains breaking client changes queued for the `0.21.0` release—further published patches to `0.20.x` **must** be based on the commit resulting from this PR, **not** `main`. This no longer becomes an issue at the end of September, as we effectively drop support for patches on minor versions when the next minor version is released.

## Testing

See https://github.com/VertaAI/modeldb/pull/3208.

## How to Revert

Yank `0.20.4` from PyPI.

PR #3208, opened 2022-09-12
fix: [VUMM-656] Add safeguards for client config discovery

liuverta:
## Impact and Context

The Verta client may be instantiated in environments where the user doesn't have full read access to the filesystem. This raises exceptions on instantiation due to the client config file functionality, which checks parent directories all the way to the root.

This PR adds two solutions for this:

1. Skip directories that can't be read.
2. Add an environment variable (`VERTA_DISABLE_CLIENT_CONFIG`) to disable config file functionality altogether.

## Risks and Area of Effect

AoE: This could affect any use of the client config file, which today isn't all that much—it basically lets a user set a default workspace, project name, registered model name, etc. for their filesystem.

Risk: Low for breakage, with test coverage!

## Testing

I've added a test for the environment variable, and the config tests continue to pass

```
% pytest test_config.py -vvra
==================================================== test session starts ====================================================
platform darwin -- Python 3.9.2, pytest-7.1.2, pluggy-1.0.0 -- /Users/miliu/.pyenv/versions/3.9.2/envs/dev3/bin/python3.9
cachedir: .pytest_cache
hypothesis profile 'default' -> suppress_health_check=[HealthCheck.too_slow], database=DirectoryBasedExampleDatabase('/Users/
miliu/Documents/modeldb/client/verta/tests/.hypothesis/examples')
rootdir: /Users/miliu/Documents/modeldb/client/verta/tests, configfile: pytest.ini
plugins: xdist-2.5.0, forked-1.4.0, hypothesis-6.53.0
collected 8 items                                                                                                           

test_config.py::TestRead::test_discovery PASSED                                                                       [ 12%]
test_config.py::TestRead::test_merge PASSED                                                                           [ 25%]
test_config.py::TestRead::test_merge_and_overwrite PASSED                                                             [ 37%]
test_config.py::TestRead::test_skip_client_load_config PASSED                                                         [ 50%]
test_config.py::TestWrite::test_create_empty[~] PASSED                                                                [ 62%]
test_config.py::TestWrite::test_create_empty[.] PASSED                                                                [ 75%]
test_config.py::TestWrite::test_update_closest PASSED                                                                 [ 87%]
test_config.py::TestWrite::test_does_not_update_home PASSED                                                           [100%]

=============================================== 8 passed, 5 warnings in 6.66s ===============================================
```

I've also manually tested the directory skip behavior

Screen Shot 2022-09-12 at 10 46 32 AM

## How to Revert

Revert this PR.

PR #3210, opened 2022-09-12
fix: [VUMM-656] Add safeguards for client config discovery (#3208)

liuverta:
Cherry-pick of #3209; see that PR for details.


liuverta:
Checks are going wild because Jenkins is going wild

coryverta (approved):

PR #3246, opened 2022-09-27
fix: Catch exceptions from the future retrier, to prevent dead future…

jkwatson-verta:
… executions.

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/InternalFuture.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/InternalFuture.java
@@ -76,7 +76,7 @@ public class InternalFuture<T> {
                   values.add(future.get());
                 }
                 promise.complete(values);
-              } catch (RuntimeException | InterruptedException | ExecutionException t) {
+              } catch (Throwable t) {
                 if (t instanceof InterruptedException) {
                   // Restore interrupted state...
                   Thread.currentThread().interrupt();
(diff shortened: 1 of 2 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/futures/InternalFuture.java
```diff
@@ -275,7 +275,13 @@ public static <U> InternalFuture<U> retriableStage(
         .get()
         .whenComplete(
             (value, throwable) -> {
-              boolean retryCheckerFlag = retryChecker.apply(throwable);
+              boolean retryCheckerFlag = false;
+              try {
+                retryCheckerFlag = retryChecker.apply(throwable);
+              } catch (Throwable e) {
+                promise.completeExceptionally(
+                    new ExecutionException("retryChecker threw an exception. Not retrying.", e));
```
conradoverta, line 283: Wouldn't this make the promise complete twice (another time on line 305)? Maybe we have to return here?

conradoverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/futures/InternalFuture.java
jkwatson-verta: It can't actually complete twice (once it's been completed, the other calls are no-ops), but it definitely would be clearer to return here. 👍

jkwatson-verta: fixed.

jkwatson-verta:
!rebuild

PR #3247, opened 2022-09-27
fix: Catch exceptions from the future retrier, to prevent dead futures

jkwatson-verta:
cherry pick from #3246

* fix: Catch exceptions from the future retrier, to prevent dead future executions.

* little cleanup from code review


jkwatson-verta:
this didn't help with the UAC lockups, so closing as a patch on the release.

PR #3248, opened 2022-09-27
ci: Finish makefile recipes for client publication

liuverta:
## Impact and Context

- Change `make upload` to read PyPI credentials from the environment instead of passing them as arguments so that `make` doesn't print their values to the console.
- Add `make tag`, which performs this step in [our client publication instructions](https://github.com/VertaAI/modeldb/blob/222d2b1/client/CONTRIBUTING.md#package-publication):
  > Once the pull request is approved and merged, a Verta core team member shall tag the merge commit using e.g. `git tag -a client-v0.0.0 -m '' && git push --follow-tags`, with the appropriate version number.

CI will start using these in https://github.com/VertaAI/cluster-setup/pull/3469.

## Risks and Area of Effect

—

## Testing

I checked that `build` matches [what we currently do in CI](https://github.com/VertaAI/cluster-setup/blob/0c3bd1f/ci/python_client/publish-pypi.Jenkinsfile#L56) to publish the client.

I tested `upload` against [PyPI's testing server](https://packaging.python.org/en/latest/guides/using-testpypi/).

I tested `tag` (cleaning up the dummy tag afterwards).

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/Makefile
+++ b/client/verta/Makefile
@@ -1,6 +1,6 @@
 # build wheel and source distributions
 build: clean
-	python setup.py sdist bdist_wheel --universal
+	python3 setup.py sdist bdist_wheel --universal
 
 # clean previously-created builds
 clean:
@@ -9,10 +9,10 @@ clean:
 # upload distributions to PyPI
 upload:
 	# checking TWINE_USERNAME and TWINE_PASSWORD env vars
-	python -m twine upload --non-interactive dist/*
+	python3 -m twine upload --non-interactive dist/*
 
 # tag the current commit using the client's version number
-tag: VERSION=$$(python -c 'import verta; print(verta.__version__)')
+tag: VERSION=$$(python3 -c 'import verta; print(verta.__version__)')
 tag: TAG="client-v$(VERSION)"
 tag:
 	git tag -a $(TAG) -m ''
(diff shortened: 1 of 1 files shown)


eheinlein-verta (approved):
LGTM 👍

client/verta/Makefile
```diff
@@ -1,11 +1,19 @@
...
-upload: build
-	python -m twine upload dist/* -u \${PYPI_USERNAME} -p \${PYPI_PASSWORD}
+upload:
+	# checking TWINE_USERNAME and TWINE_PASSWORD env vars
+	python3 -m twine upload --non-interactive dist/*
+
+# tag the current commit using the client's version number
+tag: VERSION=$$(python3 -c 'import verta; print(verta.__version__)')
+tag: TAG="client-v$(VERSION)"
+tag:
```
conradoverta: I'm confused by having the target defined 3 times

liuverta: Oh no I was worried that I was accidentally using something obscure haha

The first two are [target-specific variable](https://www.gnu.org/software/make/manual/make.html#Target_002dspecific) assignments. It doesn't re-define `tag`—rather, it defines variables `VERSION` and `TAG` solely for use in the `tag` target.

PR #3251, opened 2022-09-28
chore: Publish client v0.21.0

liuverta:
## Impact and Context

🚀 

## Risks and Area of Effect

—

## Testing

Linting passes in all Python versions.

The complete client test suite passes in Python 3.9.

3.10 has two failures that I would not consider blockers for release:

- `custom_modules/test_custom_modules.py::TestPipInstalledModule::test_module[distutils]` fails with some quirk in how `setuptools`'s `distutils` in Python 3.10 is being captured, and I wouldn't expect a customer to suddenly start encountering this.
- `registry/model_version/test_crud.py::TestCRUD::test_attributes_overwrite` fails because `pytest.warns(None)` is deprecated; the test code is to be fixed in #3253.

## How to Revert

Yank `0.21.0` from PyPI.


liuverta:
!rebuilt

liuverta:
!rebuild

jared-verta (approved):

liuverta:
Looks like the modeldb backend check failed but we don't need it where we're going!

PR #3267, opened 2022-10-04
feat: Use backend-generated entity names

liuverta:
## Impact and Context

Use backends' automatically-generated human-readable three-word-phrase names.

## Risks and Area of Effect

I wonder if experiment run names would ever collide 🤔 

## Testing

I ran our entire suite in Python 3.9 and it passed!

Here's every entity that implements `_create_proto_internal()` (besides `DatasetVersion`, which doesn't use names):

Screen Shot 2022-10-04 at 3 24 00 PM

Deployment-api still requires a client-provided path for endpoints (`422 Client Error: path in body is required`), so it's not included in this change.

## How to Revert

Revert this PR.

### DIFF
--- a/client/verta/verta/tracking/entities/_entity.py
+++ b/client/verta/verta/tracking/entities/_entity.py
@@ -126,7 +126,7 @@ class _ModelDBEntity(object):
     @classmethod
     def _get_or_create_by_name(cls, conn, name, getter, creator, checker):
         if name is None:
-            name = cls._generate_default_name()
+            return creator(name)
 
         obj = getter(name)
         if obj is None:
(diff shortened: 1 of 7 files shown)


client/verta/verta/tracking/entities/_entity.py
```diff
@@ -126,7 +126,7 @@ def _get_proto_by_id(cls, conn, id):
     @classmethod
```
liuverta, line 126: This file's five-line diff is the entirety of the real code change—everything else is pruning unnecessary uses of `_generate_default_name()` out of the test suite.

PR #3277, opened 2022-10-11
feat: Add Try creation in InternalFuture

conradoverta:
## Impact and Context
When trying to refactor some parts of the codebase, it's hard to keep track of which pieces can throw an exception or not, so that we can have a safe path of execution. Some of the refactoring revolve around making APIs clearer on intent instead of using exceptions that we have to interpret manually. This method provides an easy way to capture the exception so that it becomes clearer. It could be done outside of the class, but that would make it harder to know of the existence or to use. Since we're using a 3rd party `Try`, we can't modify that either.

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -263,6 +263,11 @@
             <version>1.18.24</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>io.vavr</groupId>
+            <artifactId>vavr</artifactId>
+            <version>0.10.4</version>
+        </dependency>
 
     </dependencies>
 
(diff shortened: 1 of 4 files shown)


liam-verta:
Updated the pom.xml to get it to build.

backend/pom.xml
```diff
@@ -263,6 +263,11 @@
             <version>1.18.24</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>io.vavr</groupId>
+            <artifactId>vavr</artifactId>
+            <version>0.10.4</version>
+        </dependency>
```
liam-verta:
```suggestion
```

liam-verta (changes requested):
Add a test to show the functionality and make sure it continues to work?

PR #3290, opened 2022-10-19
feature: [VRD-535] Have client use endpoint url from backend when available.

ewagner-verta:
## Impact and Context

The full prediction url was added to the API response when fetching the details of an deployed model via an existing endpoint.  However, this value may not always be present. (Just what I was told) This change expands the get_deployed_model method with a call to _get_json_by_id (where the newly added "full_url" value resides) so it can be passed along to the deployed_model object.  If it does not exist, the default behavior remains the same and the url is assembled.

`os` and `copy` are the only net new modules added, both of which are standard Python library, hence no changes to any requirements files.
## Risks and Area of Effect

This change impacts the Endpoint class, specifically the way the `get_deployed_model` method returns data.  Given that a failure to fetch the value of the new `full_url` field simply results in the same expected behavior as before the change, this seems fairly low risk.
## Testing

Unit tests to cover this change are included and all are passing.
The change has also been tested successfully against a dev environment (Special thanks to Heather for making that super easy.)
## How to Revert

These changes do not create any artifacts or contain anything irreversible. Simply revert this PR and there should be no fallout.

### DIFF
--- a/client/verta/tests/unit_tests/test_endpoint_class.py
+++ b/client/verta/tests/unit_tests/test_endpoint_class.py
@@ -1,0 +1,145 @@
+# -*- coding: utf-8 -*-
+# Unit tests for the Endpoint class.
+#
+#  Mocked API responses are declared as static test fixtures for now.  This is a brittle approach, as any schema changes
+#  in those calls will not be reflected here unless manually updated.  Long term, we'll want to generate these fixtures
+#  dynamically from a shared schema, like swagger.  Additionally, some of the patching of methods is a tad cumbersome.
+#  We'll want to consider a less verbose approach for patching in the future (decorators).
+
+import os
+import pytest
+
+from copy import deepcopy
+from typing import Dict, Any
+from unittest.mock import patch
+
+from client.verta.verta.endpoint import Endpoint
+from client.verta.verta._internal_utils._utils import Connection, Configuration
+from client.verta.verta.credentials import EmailCredentials
+
+#---------------------------------------------------------------------------------------------------------------------
+# TEST FIXTURES
+#---------------------------------------------------------------------------------------------------------------------
+# This section contains the test fixtures required for testing the Endpoint class,
+
+@pytest.fixture
+@patch.dict(os.environ, {'VERTA_EMAIL': 'test_email@verta.ai', 'VERTA_DEV_KEY':'123test1232dev1232key123'})
+def mock_conn() -> Connection:
+    """ Return a mocked object of the _internal_utils._utils.Connection class for use in tests """
+    return Connection(
+        scheme='https',
+        socket='test_socket',
+        credentials=EmailCredentials.load_from_os_env()
+    )
+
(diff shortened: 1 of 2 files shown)


ghost (approved):

client/verta/tests/unit_tests/test_endpoint_class.py
```diff
@@ -0,0 +1,145 @@
...
+import os
+import pytest
+
+from copy import deepcopy
+from typing import Dict, Any
+from unittest.mock import patch
+
+from client.verta.verta.endpoint import Endpoint
+from client.verta.verta._internal_utils._utils import Connection, Configuration
+from client.verta.verta.credentials import EmailCredentials
```
liuverta, line 18: Would you be opposed to importing from `verta` instead?

```suggestion
from verta.endpoint import Endpoint
from verta._internal_utils._utils import Connection, Configuration
from verta.credentials import EmailCredentials
```

Our tests in ci are executed [from the `client/verta/tests/` directory](https://github.com/VertaAI/cluster-setup/blob/aecb4b8ea191b18b2fbedc4fb6938d1ae754db01/ci/client/python/pytest-oss-3.10.Jenkinsfile#L67) so this `import` wouldn't work (and I'm not sure if there's a reason to require that `pytest` be run from the root of the repository).

client/verta/tests/unit_tests/test_endpoint_class.py
```diff
@@ -0,0 +1,145 @@
...
+#  in those calls will not be reflected here unless manually updated.  Long term, we'll want to generate these fixtures
+#  dynamically from a shared schema, like swagger.  Additionally, some of the patching of methods is a tad cumbersome.
+#  We'll want to consider a less verbose approach for patching in the future (decorators).
+
+import os
+import pytest
+
+from copy import deepcopy
+from typing import Dict, Any
+from unittest.mock import patch
```
liuverta, line 14: This is very cool!

client/verta/tests/unit_tests/test_endpoint_class.py
```diff
@@ -0,0 +1,145 @@
+# -*- coding: utf-8 -*-
+# Unit tests for the Endpoint class.
+#
```
liuverta, line 3:
```suggestion

""" Unit tests for the Endpoint class. """

**Review by liuverta — COMMENTED** [review]
This is a personal preference rather than an informed recommendation, but I find the use of

```
#---------------------------------------------------------------------------------------------------------------------
```

section headers to be a bit unnecessary and clutter-y. The contents of each section seem self-explanatory to me. And I think it would a lot of overhead for us to standardize what the sections of tests files should be.

This comes in part from complaints _I've_ received in the past about the client's `.py` files being too long 😅 

I'd _would_ recommend marking "in the future do this" notes as

```python
# TODO: generate response bodies from Swagger schemas
```

since most IDEs will color-highlight them, so they won't get visually lost as just another comment.

PR #3298, opened 2022-10-21
fix: Unwrap exception when getting a future value

conradoverta:
## Impact and Context
Without this change, completing a future with a MDBException type can result in a wrapped error and incorrect code used on the API return.


evgeniilezhnin:
it makes UAC test fail because they expect wrapped exception

PR #3307, opened 2022-10-24
feat: Future executor

conradoverta:
## Impact and Context
Using a explicit type of executor facilitates making sure that we're using compatible executors. Otherwise, we might end up with incorrect behavior by not passing the right thread local variables around.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -0,0 +1,53 @@
...
+public class FutureExecutor implements Executor {
+  final Executor other;
+
+  FutureExecutor(Executor other) {
+    this.other = other;
+  }
+
+  // Wraps an Executor and make it compatible with grpc's context
+  public static FutureExecutor makeCompatibleExecutor(Executor ex) {
+    return new FutureExecutor(ex);
```
jkwatson-verta, line 20: I think we should make sure to wrap the executor with otel's `Context.taskWrapping(` executor so that we get tracing-context wrapping automatically, without having to use the open-tracing APIs like we have below.

jkwatson-verta (approved):
other than making sure to wrap the executor with the otel context helper, 👍🏼

PR #3309, opened 2022-10-24
chore: Introduce a new Future that has an API that we prefer over the existing InternalFuture

jkwatson-verta:
This option is to have a global FutureExecutor that is used by the Future implementation.

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
@@ -1,6 +1,8 @@
 package ai.verta.modeldb.common.futures;
 
 import ai.verta.modeldb.common.exceptions.ModelDBException;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.trace.Span;
@@ -13,8 +15,16 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.function.*;
 import java.util.stream.Collectors;
+import lombok.extern.log4j.Log4j2;
 
-public class InternalFuture<T> {
+@Log4j2
+public class Future<T> {
+  /**
+   * This instance is the global executor for all Futures. Everything will fail if it is not set
+   * properly.
+   */
+  private static FutureExecutor futureExecutor;
+
   private static final boolean DEEP_TRACING_ENABLED;
   public static final AttributeKey<String> STACK_ATTRIBUTE_KEY = AttributeKey.stringKey("stack");
 
@@ -25,13 +35,13 @@ public class InternalFuture<T> {
 
   private final Tracer futureTracer = GlobalOpenTelemetry.getTracer("futureTracer");
   private final String formattedStack;
-  private CompletionStage<T> stage;
+  private final CompletionStage<T> stage;
 
   // keep the calling OpenTelemetry context, so we can use it to wrap the actual invocation of the
   // future's implementation
   private final Context callingContext = Context.current();
 
-  private InternalFuture() {
+  private Future(CompletionStage<T> stage) {
     if (!DEEP_TRACING_ENABLED) {
       formattedStack = null;
     } else {
@@ -44,14 +54,14 @@ public class InternalFuture<T> {
               .map(StackTraceElement::toString)
               .collect(Collectors.joining("\nat "));
     }
+    this.stage = stage;
   }
 
   // Convert a list of futures to a future of a list
   @SuppressWarnings("unchecked")
-  public static <T> InternalFuture<List<T>> sequence(
-      final List<InternalFuture<T>> futures, FutureExecutor executor) {
+  public static <T> Future<List<T>> sequence(final List<Future<T>> futures) {
     if (futures.isEmpty()) {
-      return InternalFuture.completedInternalFuture(new LinkedList<>());
+      return Future.of(List.of());
     }
 
     final var promise = new CompletableFuture<List<T>>();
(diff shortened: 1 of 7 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
```diff
@@ -0,0 +1,387 @@
...
+import java.util.function.*;
+import java.util.stream.Collectors;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class Future<T> {
+  /**
+   * This instance is the global executor for all Futures. Everything will fail if it is not set
+   * properly.
+   */
```
conradoverta, line 24: Should we set `@NonNull` or something like that? Not sure how that annotation works in java, but maybe it helps?

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
```diff
@@ -0,0 +1,387 @@
...
+    return new Future<R>(other);
+  }
+
+  /** @deprecated Use {@link #of(Object)} instead. */
+  @Deprecated
+  public static <R> Future<R> completedInternalFuture(R value) {
+    Preconditions.checkNotNull(
+        futureExecutor, "A FutureExecutor is required to create a new Future.");
+    return of(value);
+  }
```
conradoverta, line 108: should we just remove given that these are now two classes?

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
```diff
@@ -0,0 +1,387 @@
...
+      futureExecutor = executor;
+    } else {
+      log.warn(
+          "A FutureExecutor has already been set. Ignoring this call.",
+          new RuntimeException("call-site capturer"));
+    }
+  }
+
+  /** @deprecated Only use this as a part of the conversion process between versions of Futures. */
+  @Deprecated
```
conradoverta, line 374: What does this annotation do? will it make compilation fail?

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -1,23 +1,21 @@
...
 import org.springframework.lang.NonNull;
 
 public class FutureExecutor implements Executor {
-  final Executor other;
+  private final Executor delegate;
 
-  FutureExecutor(Executor other) {
-    this.other = io.opentelemetry.context.Context.taskWrapping(other);
+  private FutureExecutor(Executor delegate) {
+    this.delegate = Context.taskWrapping(delegate);
```
conradoverta: nit: I personally find very confusing to just have `Context` for otel given that +80% of the times we're using it for the grpc context. So my brain just assumes this is talking about the grpc context

conradoverta (approved):
Couple of notes, but looks pretty clean and easy.

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
jkwatson-verta: it's a fair question. While doing a test migration from InternalFuture in UAC, I found this handy to have to make the conversion a little easier. I don't feel really strongly about it, though.

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
jkwatson-verta: Setting that unfortunately won't help, since it won't be detected at compile time as an issue (which is the only place that annotation is ever used).

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
jkwatson-verta: It will generate warnings at compile time by default. And, in IDEA, usages will show up as a warning and ~crossedOut~ in the editor.

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
jkwatson-verta: heh. Of course, I have the opposite confusion, having lived OTel for so long. I actually liked your solution I saw in another PR to just fully qualify both of them. I'll update!

jkwatson-verta: or...I'll wait for your changes to this to land, and just use them as-is.

jkwatson-verta: done!

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureJdbi.java
```diff
@@ -84,4 +84,61 @@ private <T extends Exception> InternalFuture<Void> useHandleOrTransaction(
 
     return InternalFuture.from(promise);
   }
+
+  public <R, T extends Exception> Future<R> handleCompose(HandleCallback<Future<R>, T> callback) {
```
liam-verta, line 90: Should we mark the `InternalFuture` returning methods as deprecated?

liam-verta (approved):
Maybe marking `InternalFuture` and all methods of it as deprecated would make downstream conversion easier?  Otherwise, yay!

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureJdbi.java
jkwatson-verta: I think we want to wait and let the new one bake a little bit before we commit all the way on this. Should be able to do this as soon as we're happy with how the new one works and feels.

PR #3311, opened 2022-10-26
fix: Make context propagation deterministic

conradoverta:
## Impact and Context
By wrapping the executor, we end up passing context in a non-deterministic way, based on which thread it runs, which depends on whether the future had already been completed. This changes the futures implementation to be deterministic.

## Risks and Area of Effect
Some parts of the code might have made assumptions about when the context is propagated and this might make them fail.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
@@ -1,12 +1,10 @@
 package ai.verta.modeldb.common.futures;
 
-import io.grpc.Context;
 import io.opentracing.Scope;
 import io.opentracing.util.GlobalTracer;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
-
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.With;
@@ -19,7 +17,8 @@ public class FutureExecutor implements Executor {
   @With private final io.grpc.Context grpcContext;
 
   public FutureExecutor captureContext() {
-    return withOtelContext(io.opentelemetry.context.Context.current()).withGrpcContext(io.grpc.Context.current());
+    return withOtelContext(io.opentelemetry.context.Context.current())
+        .withGrpcContext(io.grpc.Context.current());
   }
 
   // Wraps an Executor and make it compatible with grpc's context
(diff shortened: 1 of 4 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -35,19 +41,25 @@ public static FutureExecutor newSingleThreadExecutor() {
...
-                  }));
+          () -> {
+            try (Scope s = tracer.scopeManager().activate(span)) {
+              finalR.run();
+            }
+          });
     } else {
-      other.execute(Context.current().wrap(r));
+      other.execute(r);
     }
```
jkwatson-verta: with OTel in place, I think we can replace all of this with just `other.execute(r)`. OpenTelemetry implements this in the tracer shim implementation itself.

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -35,19 +41,25 @@ public static FutureExecutor newSingleThreadExecutor() {
 
   @Override
   public void execute(@NonNull Runnable r) {
+    if (otelContext != null) {
+      r = otelContext.wrap(r);
```
jkwatson-verta, line 43: this is nice. I think this will let us clean up a ton of cruft in the InternalFuture itself that was doing this the hard way.

jkwatson-verta:
I think you had a failing test case that you had posted in slack. Can you add that test case here?

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
conradoverta: I follow none of these words 😅 
Could you just suggest how the code should look like?

jkwatson-verta: github won't let me suggest on this whole block. so, replace lines 51-63 with `other.execute(r)`

jkwatson-verta (approved):
Thanks!

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
jkwatson-verta: I cleaned it up.

PR #3332, opened 2022-11-01
refactor!: [VRD-553] Refactor DeployedModel class and handle custom prediction IDs.

ewagner-verta:
## Impact and Context
This PR refactors the DeployedModel class to achieve the following:

### Segregate deployed model code from underlying HTTP logic.
Prior to this change, the predict method was performing a lot of custom http logic, including retry, sleep, and custom responses to specific status codes related to challenges with model warm-up and auto-scaling.  This complexity and scale made it challenging to unit test effectively.  Additionally the design pattern used created a true _endless_ loop of request retries.

This solution:
* Utilizes well-maintained standard libraries for handling of complex http logic (_urllib3_, _requests_, etc.) allowing us to fall back on sensible defaults and more robust error handling on edge cases.
* Removes all the HTTP handling from the DeployedModel class itself and puts it into a reusable utility package covered by unit tests, thus greatly simplifying the code.
* Allows users of the client to customize handling of ANY status codes they may encounter while sending predictions.  Since we can neither predict all the odd status codes customer systems might throw, nor respond to every case with a new custom parameter, this allows users the flexibility to respond to their unique network challenges without any changes to the client code.  Total retries and backoff factor uses sensible defaults but can be fully customized.
* All parameters for the prediction method, outside of the prediction value itself, are optional, making the method easy to use
* Handles 404 and 429 status codes gracefully.  The default retry logic will result in automatic retries of the prediction requests for any 404 and 429 errors by default.  Rather than an infinite loop of retries, default behavior will be to retry 13 times spread over a total of ~ 10 minutes.  (Max retry sleep = 120 seconds).  These defaults can be overridden by the user for any given call to predict/predict_with_id.
```
Default Retry Pattern on 404, 429:

Retry #     -    Seconds to wait before retrying
1.                        0.15
2.                        0.3
3.                        0.6
4.                        1.2
5.                        2.4
6.                        4.8
7.                        9.6
8.                        19.2
9.                        38.4
10.                        76.8
11.                       120 (max)
12.                       120 (max)
13.                       120 (max)
```

### Enable use of custom prediction IDs
After this change, users will be able to pass in a custom ID value to be used as the `verta-request-id` header in the prediction request.  While both the `predict` and `predict_with_id` methods enable users to send a custom request ID, only the `predict_with_id` method will actually return that value to the user in the form tuple(id, result).

```python
# predict method with no custom id provided by user.
model = verta.Client().get_or_create_endpoint('endpoint_name').get_deployed_model()
model.predict(['sample_prediction'])
>>> [-0.123]

# predict_with_id method with no custom id provided by user.
model = verta.Client().get_or_create_endpoint('endpoint_name').get_deployed_model()
model.predict_with_id(['sample_prediction'])
>>> ('backend-generated-id-value', [-0.123])

# predict method with custom id provided by user. (ID only sent to back-end, not requested)
model = verta.Client().get_or_create_endpoint('endpoint_name').get_deployed_model()
model.predict_with_id(['sample_prediction'], prediction_id='your-sweet-custom-id-here')
>>> [-0.123]

# predict_with_id method with custom id provided by user.  Custom ID sent to back-end, and returned.
model = verta.Client().get_or_create_endpoint('endpoint_name').get_deployed_model()
model.predict_with_id(['sample_prediction'], prediction_id='your-sweet-custom-id-here')
>>> ('your-sweet-custom-id-here', [-0.123])
```

## Risks and Area of Effect
This is a significant refactor of the DeployedModel class.  However, it should be fully backwards compatible.  Warnings are provided for all planned future deprecations

## Testing
- [ X] Unit test
- [X ] Deployed to dev env (Tested against personal dev environment with successful results)
- [ X] Other - Original test suite being run now (will update later) but individual existing file for testing deployed model specifically passed already.

## Reverting


client/verta/tests/test_utils/test_http_session_utils.py
```diff
@@ -0,0 +1,110 @@
...
+                                            backoff_factor=1.23
+                                            )
+
+# A different retry object for testing on-the-fly alteration
+ALTERED_MOCK_RETRY_OBJECT: Retry = http_session_utils.retry_config(max_retries=22,
+                                                           status_forcelist={789},
+                                                           backoff_factor=2.45
+                                                           )
+# HTTPAdapter object for patching
+MOCK_HTTP_ADAPTER = HTTPAdapter(max_retries=MOCK_RETRY_OBJECT)
```
conradoverta: Nit: should we use pytest fixtures instead of global variables? Global variables assumes they don't have state that will make the tests coupled.

client/verta/verta/_internal_utils/http_session_utils.py
```diff
@@ -0,0 +1,69 @@
...
+from urllib3.util import Retry
+
+DEFAULT_MAX_RETRIES: int = 13
+DEFAULT_BACKOFF_FACTOR: float = 0.3
+DEFAULT_STATUS_FORCELIST: Set = {404, 429}
+
+
+def retry_config(max_retries: int,
+                 status_forcelist: Set[int],
+                 backoff_factor: float) -> Retry:
```
liuverta: I think we need a consistent style for breaking function signatures into multiple lines.

I'm seeing that sometimes the `) -> RetType` is on the same line as the last parameter (like here), sometimes it's on the next line at the same indentation as the parameters (like [here for `DeployedModel.predict()`](https://github.com/VertaAI/modeldb/pull/3332/files#diff-43335cbae366a467266b824f5dc001bffc653b4f7fa813981b756cfcbc7bcb2bR164)), and sometimes it's on the next line at the same indentation as `def` (like [here for `DeployedModel.predict_with_id()`](https://github.com/VertaAI/modeldb/pull/3332/files#diff-43335cbae366a467266b824f5dc001bffc653b4f7fa813981b756cfcbc7bcb2bR240)).

client/verta/verta/_internal_utils/http_session_utils.py
```diff
@@ -0,0 +1,69 @@
...
+                 connect=0,  # Connection errors before the request is sent to the remote server should raise an error
+                 status=max_retries,
+                 other=0,  # recommended by urllib3 docs to guard against infinite retry loops
+                 status_forcelist=status_forcelist,
+                 backoff_factor=backoff_factor,
+                 allowed_methods=False
+                 )
+
+
+def alter_retry_config(session: Session,
```
conradoverta: Nit: It's unclear if the existing `Retry` will be modified or if we consider `Retry` to be immutable. A clarification here would be useful.

client/verta/verta/_internal_utils/http_session_utils.py
```diff
@@ -0,0 +1,69 @@
...
+                       max_retries: Optional[int],
+                       status_forcelist: Optional[Set[int]],
+                       backoff_factor: Optional[float]) -> Session:
```
conradoverta: nit: should we provide default `None` for these?

client/verta/tests/unit_tests/test_deployed_model_class.py
```diff
@@ -0,0 +1,250 @@
...
+from verta._internal_utils import http_session_utils
+
+PREDICTION_URL: str = 'https://test.dev.verta.ai/api/v1/predict/test_path'
+TOKEN: str = '12345678-xxxx-1a2b-3c4d-e5f6g7h8'
+MOCK_RETRY: Retry = http_session_utils.retry_config(max_retries=http_session_utils.DEFAULT_MAX_RETRIES,
+                                            status_forcelist=http_session_utils.DEFAULT_STATUS_FORCELIST,
+                                            backoff_factor=http_session_utils.DEFAULT_BACKOFF_FACTOR
+                                            )
+MOCK_SESSION: Session = http_session_utils.init_session(retry=MOCK_RETRY)
+VERTA_CLASS = 'verta.deployment._deployedmodel'
```
liuverta: fyi, this is inconsistent with how we defined this variable [in the endpoint unit tests](https://github.com/VertaAI/modeldb/blob/121254d9de29ebfba673ed6d233b7581a1da8d38/client/verta/tests/unit_tests/test_endpoint_class.py#L36), where `VERTA_CLASS` includes the trailing period:

```python
VERTA_CLASS: str= 'verta.endpoint.Endpoint.'
```

(also `_deployedmodel` is a module rather than a class 😆)

client/verta/verta/_internal_utils/http_session_utils.py
liuverta: Would you be opposed to calling this `update_retry_config`? "alter" isn't a verb we're currently using in the codebase, and introducing it feels...inconsistent? unclear? unless it really is the most accurate one to use.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +207,136 @@ def predict(
...
+                                                              status_forcelist=retry_status,
+                                                              backoff_factor=backoff_factor)
+
+        if prediction_id:
+            http_session_utils.update_session_headers(self._session, {'verta-request-id': prediction_id})
+            response = self._predict(x, compress)
+            http_session_utils.update_session_headers(self._session, {'verta-request-id': prediction_id}, remove=True)
+        else:
+            response = self._predict(x, compress)
+        return _utils.body_to_json(response)
```
conradoverta: Should we just call `predict` and drop the first returned value? The implementation is about hte same

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +207,136 @@ def predict(
...
+        # Update retry config and re-mount it to the existing session if it differs from the current one.
+        self._session = http_session_utils.alter_retry_config(self._session,
+                                                              max_retries=max_retries,
+                                                              status_forcelist=retry_status,
+                                                              backoff_factor=backoff_factor)
+
+        if prediction_id:
+            http_session_utils.update_session_headers(self._session, {'verta-request-id': prediction_id})
             response = self._predict(x, compress)
+            http_session_utils.update_session_headers(self._session, {'verta-request-id': prediction_id}, remove=True)
```
conradoverta: This update seems not to be thread safe (mutating a shared object). That's usually a code smell as we're adding and removing things. Imagine that the user is using a threadpool to send us a bunch of stuff with different IDs. This will likely cause collision

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +207,136 @@ def predict(
...
-            ):  # clientside error
-                break
-
-            sleep = 0.3 * (2 ** (num_retries + 1))
-            print(
-                "received status {}; retrying in {:.1f}s".format(
-                    response.status_code, sleep
-                )
+
+    # TODO: Remove this method after release of 0.22.0
```
conradoverta, line 338: I personally really like well documented docstrings like these ones. Makes the API spec easier to consume and less burdensome for folks not dealing with it on the day to day. Do you think we could keep them somewhere?

client/verta/verta/_internal_utils/http_session_utils.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
```
liuverta: If you're up for it, would you mind renaming this file `http_session.py`? The `*_utils.py` scheme was kind of a mistake 😬

liuverta (commented):
Some initial comments, will take another look later.

client/verta/tests/test_utils/test_http_session_utils.py
ewagner-verta: I would have strongly preferred to use fixtures here, but python throws an error if you try to use a pytest.fixture as the return value of a patch.  A global variable allowed me to inject a singular mocked object wherever I wanted without jumping through extra hoops to work around that limitation.  Perhaps we can explore this further

client/verta/verta/_internal_utils/http_session_utils.py
```diff
@@ -0,0 +1,69 @@
...
+""" Utilities for working with request.Session objects, and handling of retries """
+
+# TODO: Unit test for prediction_input_unpack method
+# TODO: Unit test for prediction_io_cleanup method
+import requests
+
+from requests import Session
+from requests.adapters import HTTPAdapter
+from typing import Dict, Optional, Set
+from urllib3.util import Retry
```
liuverta:
```suggestion
from typing import Dict, Optional, Set

import requests
from requests import Session
from requests.adapters import HTTPAdapter
from urllib3.util import Retry
```

[PEP 8](https://peps.python.org/pep-0008/#imports):

>  Imports should be grouped in the following order:
> 
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.
> 
> You should put a blank line between each group of imports.

client/verta/tests/unit_tests/test_deployed_model_class.py
```diff
@@ -0,0 +1,250 @@
...
+from requests import Session
+from requests.exceptions import RetryError, HTTPError
+from typing import Any, Dict, List
+from unittest import TestCase
+from unittest.mock import Mock, patch
+from urllib3 import Retry
+
+from verta.credentials import EmailCredentials
+from verta.deployment import DeployedModel
+from verta._internal_utils import http_session_utils
```
liuverta:
```suggestion
import os
from typing import Any, Dict, List

import pytest
from requests import Session
from requests.exceptions import RetryError, HTTPError
import responses
from responses import matchers
from unittest import TestCase
from unittest.mock import Mock, patch
from urllib3 import Retry

from verta.credentials import EmailCredentials
from verta.deployment import DeployedModel
from verta._internal_utils import http_session_utils
```

[PEP 8](https://peps.python.org/pep-0008/#imports):

>  Imports should be grouped in the following order:
> 
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.
> 
> You should put a blank line between each group of imports.

client/verta/tests/unit_tests/test_deployed_model_class.py
```diff
@@ -0,0 +1,250 @@
+# -*- coding: utf-8 -*-
+""" Unit tests for the verta.deployment.DeployedModel class. """
+
+import os
+import pytest
+import responses
```
liuverta: Is `responses` new? It'll need to be added to [the unit test requirements](https://github.com/VertaAI/modeldb/blob/main/client/verta/requirements-unit-tests.txt).

client/verta/verta/_internal_utils/http_session_utils.py
ewagner-verta: The intention is for any object of the DeployedModel class to have an attribute that is a valid Retry configuration.  As predict is called with changes to the retry pattern, we create a new retry object that replaces the existing one, which is then re-mounted to the existing session.  This allows us to change up the retry logic without impacting the open session and any connection-pools it is currently maintaining.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -1,19 +1,20 @@
...
-from urllib.parse import urlparse
+import json
 import warnings
 
-import requests
+from requests import Session
+from typing import Any, Dict, List, Optional, Set, Tuple
+from urllib3 import Retry
+from urllib.parse import urlparse
+from verta import credentials
```
liuverta: See other comments about import order or `typing`, `urllib`, and `verta` (which maybe could be changed to `..`).

client/verta/verta/_internal_utils/http_session_utils.py
ewagner-verta: Upon closer look, I think these shouldn't actually be `Optional`.  The predict methods are always going to use defaults for those values when calling `update_retry_config`.

client/verta/verta/deployment/_deployedmodel.py
liuverta: The info in this function's docsting is covered in `DeployedModel`'s class docstring!

https://github.com/VertaAI/modeldb/blob/main/client/verta/verta/deployment/_deployedmodel.py#L26-L63

Although, as an aside, perhaps we should add `endpoint.get_deployed_model()` to the `Examples` section there

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +207,136 @@ def predict(
...
-                )
+
+    # TODO: Remove this method after release of 0.22.0
+    @classmethod
+    def from_url(cls, url, token=None, creds=None):
+        """
+        Returns a :class:`DeployedModel` based on a custom URL and token.
+
+        .. deprecated:: 0.22.0
+           Simply call the ``DeployedModel`` class directly with the same parameters instead.
```
liuverta: Since this is a public, user-facing API, I'd suggest:

```suggestion
           For most use cases, call :meth:`Endpoint.get_deployed_model() <verta.endpoint.Endpoint.get_deployed_model>`. For lowel-level needs, call the ``DeployedModel`` class directly with the same parameters.
```

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Ah, so are you suggesting that all this logic should be in `predict_with_id` and `predict` should just call that and drop the id to return only the results?  I'm not sure why I didn't think of that!  I think I just had the inverse relationship in my head.  I agree that would be much simpler

client/verta/verta/_internal_utils/http_session_utils.py
liuverta: In light of

> we create a new retry object that replaces the existing one, which is then re-mounted to the existing session.

Maybe `set_retry_config` might be more appropriate 🤷 "alter" and "update" imply to me that the `Retry` object is being mutated.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -183,43 +145,55 @@ def get_curl(self):
...
-        .. versionadded:: 0.13.14
-           The `always_retry_404` parameter.
-        .. versionadded:: 0.13.12
-           The `always_retry_429` parameter.
+        .. versionadded:: 0.22.0
+           The `prediction_id` parameter.
+        .. versionadded:: 0.22.0
+           The `retry_status` parameter.
+        .. versionadded:: 0.22.0
+           The `backoff_factor` parameter.
```
liuverta: I'd recommend adding `.. deprecated` directives for the `always_retry_*` parameters—with a half-sentence about favoring `max_retries` and `retry_status`—unless there's a particular reason not to.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -183,43 +145,55 @@ def get_curl(self):
...
-            Whether to retry on 429s indefinitely. This is to accommodate third-party cluster
+            Deprecated: Whether to retry on 429s indefinitely. This is to accommodate third-party cluster
             autoscalers, which may take minutes to launch new pods for servicing requests.
-
+        retry_status : set, default {404, 429}
+            Set of status codes, as integers, for which retry attempts should be made.  Overwrites default value of
+            {404, 429}.  Expand the set to include more: retry_status={404, 429, <NEW_STATUS_CODE>}
+        backoff_parameter : float, default 0.3
+            A backoff factor to apply between attempts after the second try.  Uses standard urllib3 sleep pattern:
+            `{backoff factor} * (2 ** ({number of total retries} - 1))` with a maximum sleep time between requests of
```
liuverta: As you may or may not have noticed, Sphinx renders single backticks as italics—double backticks are needed for `monospace` (which I forget all the time myself).

```suggestion
            ``{backoff factor} * (2 ** ({number of total retries} - 1))`` with a maximum sleep time between requests of
```

client/verta/verta/_internal_utils/http_session_utils.py
ewagner-verta: Updating to the convention mentioned in Slack.  Fixup commits coming soon with those changes.

client/verta/tests/unit_tests/test_deployed_model_class.py
ewagner-verta: Good points.  I'll address both

client/verta/verta/_internal_utils/http_session_utils.py
ewagner-verta: No objections here.  Will update

ewagner-verta: Pending commit will change to `set_retry_config` with a more informative doc string

client/verta/verta/_internal_utils/http_session_utils.py
ewagner-verta: Pending commit will drop `Optional` from type declaration.

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Refactored to be thread safe.  We will now discard any existing session header for `verta-request-id` before sending the next request.  If a `prediction_id` is provided by the user, we add the header back in with the new value.  If no `prediction_id` is provided, we simply leave it off and the back-end assigns a random value.

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Good catch.  Thank you

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Doc strings updated

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Fixed

client/verta/tests/test_utils/test_http_session_utils.py
conradoverta: Interesting. Maybe a comment with such note here might help bc I'm sure I'll forget it in the future

client/verta/verta/deployment/_deployedmodel.py
conradoverta: yeah. the logic is already in `predict_with_id` from what I could tell

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +239,148 @@ def predict(
...
 
-        _utils.raise_for_http_error(response)
 
     def predict_with_id(
-        self,
-        x,
-        compress=False,
-        max_retries=5,
-        always_retry_404=True,
-        always_retry_429=True,
```
conradoverta, line 283: API break? maybe we should keep just like we did for `predict`?

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -53,7 +53,13 @@ class DeployedModel(object):
     Examples
     --------
     .. code-block:: python
+       :emphasize-lines: 3,4
+
+        # Preferred method for instantiating an object.
+        deployed_model_x = verta.Client().get_or_create_endpoint('endpoint_name').get_deployed_model()
+        deployed_model_x.predict(['here is a prediction'])
```
liuverta: For some consistency with the rest of our docs:

```suggestion
        endpoint = client.get_or_create_endpoint('endpoint_name')
        deployed_model = endpoint.get_deployed_model()
        deployed_model.predict(['here is a prediction'])
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -53,7 +53,13 @@ class DeployedModel(object):
     Examples
     --------
     .. code-block:: python
+       :emphasize-lines: 3,4
```
liuverta: fyi line `4` is empty (line `1` would be your `# Preferred method . . .` comment)

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Updated in last push.

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Import order fixed in all places.  Apologies for the confusion

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: These have been updated to `False` by default for backwards compatibility

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -183,43 +169,63 @@ def get_curl(self):
...
     def predict(
         self,
-        x,
+        x: List[Any],
         compress=False,
-        max_retries=5,
-        always_retry_404=True,
-        always_retry_429=True,
-    ):
+        prediction_id: Optional[str] = None,
```
liuverta: For API compatibility, new parameters should be added after all the existing ones. Otherwise, this breaks user code that calls

```python
deployed_model.predict(X, False, 20)
```

I'm not wholly opposed to shipping API breakages, but I think we'd like to avoid it here for at least a release or two.

Admittedly, this is frustrating because it becomes difficult to group related parameters together.

client/verta/verta/deployment/_deployedmodel.py
ewagner-verta: Main doc string for DeployedModel class has been updated with examples.  After looking at the resulting [docs page](https://verta--3332.org.readthedocs.build/en/3332/_autogen/verta.deployment.DeployedModel.html#verta.deployment.DeployedModel) the `from_url` method to be deprecated is fairly prominent still.  Given that this info is already in the main doc string for the class, should we strip out all the info and just leave the deprecation message?

client/verta/verta/deployment/_deployedmodel.py
liuverta: For extra fun, might I float the idea of implementing the new parameters as [keyword-only arguments](https://peps.python.org/pep-3102/), which frees them in the future from this positional compatibility concern and I think is how they will generally be passed anyway (e.g. in the tests you've written).

It's something I've always wanted to use after we dropped Python 2, but I'm not sure how common they are in the Python ecosystem

liuverta (changes requested):

client/verta/verta/deployment/_deployedmodel.py
liuverta: Ugh I wish our docs theme made deprecation notices more eye-catching.

> should we strip out all the info and just leave the deprecation message?

I like this idea.

ewagner-verta:
!rebuild

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -183,43 +170,63 @@ def get_curl(self):
...
+        x: List[Any],
         compress=False,
-        max_retries=5,
-        always_retry_404=True,
-        always_retry_429=True,
-    ):
+        max_retries: Optional[int] = http_session.DEFAULT_MAX_RETRIES,
+        prediction_id: Optional[str] = None,
+        always_retry_404=False,
+        always_retry_429=False,
```
liuverta: Expanding on my previous comment, we should not insert new parameters between _any_ of the existing ones, which include `always_retry_404` and `always_retry_429`. Everything new, like `prediction_id`, has to come _after_.

Honestly I wonder if we should just put `predict_id` at the very end of the argument list so that the retry parameters are grouped together 🤔

client/verta/verta/_internal_utils/http_session.py
```diff
@@ -0,0 +1,88 @@
...
+    current_retry = session.get_adapter('https://').max_retries
+    if max_retries != current_retry.status or \
+            status_forcelist != current_retry.status_forcelist or \
+            backoff_factor != current_retry.backoff_factor:
+        new_retry_config = retry_config(
+            max_retries=max_retries if max_retries else DEFAULT_MAX_RETRIES,
+            status_forcelist=status_forcelist if status_forcelist \
+                else DEFAULT_STATUS_FORCELIST,
+            backoff_factor=backoff_factor if backoff_factor \
+                else DEFAULT_BACKOFF_FACTOR
```
liuverta: fyi line breaks aren't required here since we're inside parentheses (the `retry_config()` function call), though feel free to keep them for clarity.

client/verta/verta/_internal_utils/http_session.py
```diff
@@ -0,0 +1,88 @@
...
+        backoff_factor: float
+        ) -> Session:
+    """ Creates a new Retry object with the given params and mounts it to
+        the provided session in place of the existing retry config. This
+        allows the DeployedModel.predict method to change the behavior of
+        the Session on the fly without dropping open connections. """
+    current_retry = session.get_adapter('https://').max_retries
+    if max_retries != current_retry.status or \
+            status_forcelist != current_retry.status_forcelist or \
+            backoff_factor != current_retry.backoff_factor:
```
liuverta: fyi in general, PEP 8 discourages line breaks in favor of parentheses.

> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

Here I would recommend (as I've personally been using in the codebase):

```suggestion
    if (max_retries != current_retry.status
            or status_forcelist != current_retry.status_forcelist
            or backoff_factor != current_retry.backoff_factor):
```

based on one of the acceptable options listed in the PEP:

> ```
> # Add some extra indentation on the conditional continuation line.
> if (this_is_one_thing
>         and that_is_another_thing):
>     do_something()
> ```

Though personally I don't mind either way as it doesn't hurt clarity much.

client/verta/verta/_internal_utils/http_session.py
```diff
@@ -0,0 +1,88 @@
...
+        new_retry_config = retry_config(
+            max_retries=max_retries if max_retries else DEFAULT_MAX_RETRIES,
```
liuverta: fyi for brevity, each of these can be written as

```suggestion
            max_retries=max_retries or DEFAULT_MAX_RETRIES,
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -160,18 +132,33 @@ def _predict(self, x, compress=False):
...
+            response = self._session.post(
                 self._prediction_url,
-                headers={"Content-Encoding": "gzip"},
                 data=gzstream.read(),
             )
+            http_session.update_session_headers(
+                self._session,
+                {"Content-Encoding": "gzip"},
+                remove=True,
+            )
```
liuverta: I'm a bit unclear about the benefits of this change. To me, mutating the state of the session and then undoing it after one function call is less parallel-safe and more code-complex than just passing `headers` to `Session.post()`, which I'd believe is the intended usecase for that parameter.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -62,61 +69,41 @@ class DeployedModel(object):
...
+            creds=None
+            ):
+        self.prediction_url: str = prediction_url
+        self._credentials: credentials.Credentials = creds or credentials.load_from_os_env()
+        self._access_token: str = token
+        self._retry_config: Retry = http_session.retry_config(
+            max_retries=http_session.DEFAULT_MAX_RETRIES,
+            status_forcelist=http_session.DEFAULT_STATUS_FORCELIST,
+            backoff_factor=http_session.DEFAULT_BACKOFF_FACTOR
+            )
```
liuverta: I'd suggest actually not storing `self._retry_config` at all. From my experience, it's generally not good to have an object ("this `DeployedModel`'s retry config") accessible through multiple pathways (`self._retry_config` and via `self._session.adapters`) because it's confusing for developers which one to refer to as a source of truth, and can lead to inconsistency issues.

In fact, we already have that issue in `predict_with_id()` below, where we call `set_retry_config()` on `self._session`, but don't update `self._retry_config`, so already it's unclear what should be considered the `DeployedModel`'s true retry configuration.

I'd recommend moving this into `_init_session()`'s body.

client/verta/verta/_internal_utils/http_session.py
```diff
@@ -0,0 +1,88 @@
...
+        x: Dict,
+        remove: Optional[bool] = False,
+        ) -> Session:
+    """ Update the headers of the provided session object with new values """
+    if remove:
+        for k in x.keys():
+            session.headers.pop(k)
+    else:
+        session.headers.update(x)
+    return session
```
liuverta: I'm not a huge fan of having one function that does two completely different things (removal vs addition) depending on a boolean parameter value. 

And the fact that we _have_ to pass dummy data (an empty string as a dictionary value) to delete an item

```python
            http_session.update_session_headers(
                self._session,
                {'verta-request-id': ''},
                remove=True
                )
```

makes me think this might not be the right level of abstraction for a helper method.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -183,43 +170,63 @@ def get_curl(self):
...
-        max_retries : int, default 5
-            Maximum number of times to retry a request on a connection failure.
+        max_retries : int, default 13
+            Maximum number of times to retry a request.
+        prediction_id: str, default None
+            A custom str to use as the ID for the prediction request.  Defaults to a randomly generated numeric id.
         always_retry_404 : bool, default True
-            Whether to retry on 404s indefinitely. This is to accommodate model deployment warm-up.
+            Deprecated: Whether to retry on 404s indefinitely. This is to accommodate model deployment warm-up.
         always_retry_429 : bool, default True
```
liuverta: Might want to change these to `default False` to be technically correct, though I'm also fine with just not touching them.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +240,122 @@ def predict(
...
+            backoff_factor=backoff_factor,
+            )
+
+        # Remove request id header from previous requests if present.
+        if 'verta-request-id' in self._session.headers.keys():
+            http_session.update_session_headers(
+                self._session,
+                {'verta-request-id': ''},
+                remove=True
+                )
```
liuverta: Along the lines of my comment in `http_session.py` about `update_session_headers()`, I feel that this would be much more simply with a direct `del self._session.headers("verta-request-id")`. Or if we're feeling daring since we have a comment about the intent anyway:

```suggestion
        self._session.headers.pop("verta-request-id", None)
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -62,61 +69,41 @@ class DeployedModel(object):
 
     """
```
liuverta, line 70: Do you have any thoughts on having a

```suggestion
_PREDICTION_ID_HEADER_NAME = "verta-request-id"

```

? I've had people recommend to me using constants for commonly-used strings to help with clarity and avoiding typos, but I personally go either way.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +240,122 @@ def predict(
...
+                remove=True
+                )
+
+        if prediction_id:
+            http_session.update_session_headers(
+                self._session,
+                {'verta-request-id': prediction_id},
                 )
+        response = self._predict(x, compress)
+        id = response.headers['verta-request-id']
```
liuverta: My instincts tell me that we should gracefully guard against potential backend misbehavior by not always assuming `"verta-request-id"` will be returned.

```suggestion
        id = response.headers.get("verta-request-id")
```

(This was alright before, because a user had to explicitly call `predict_with_id()`, but now we're checking the header on all `predict()` calls)

client/verta/tests/test_utils/test_http_session.py
```diff
@@ -0,0 +1,133 @@
...
+    """  The set_retry_config method correctly updates the retry config of the
+     session object """
+    sesh = http_session.init_session(MOCK_RETRY_OBJECT)
+    http_session.set_retry_config(
+        sesh,
+        max_retries=22,
+        status_forcelist={789},
+        backoff_factor=2.45,
+    )
+    retry_object = sesh.get_adapter('https://').max_retries
```
liuverta: Since we mount two adapters, we might actually want to check both of them by looping through `sesh.adapters.values()` here.

client/verta/tests/test_utils/test_http_session.py
```diff
@@ -0,0 +1,133 @@
...
+        'trust_env': True,
+        'cookies': RequestsCookieJar(),
+        'adapters': OrderedDict([
+            ('https://', mock_adapter.return_value),
+            ('http://', mock_adapter.return_value)
+        ])
+    }
+    created_session = http_session.init_session(MOCK_RETRY_OBJECT)
+    assert created_session.__dict__ == expected_session_config
+    assert created_session.get_adapter('https://').max_retries == MOCK_RETRY_OBJECT
```
liuverta, line 97: Ditto on my above comment about checking both adapters.

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -0,0 +1,346 @@
...
+        status=999,
+        )
+    creds = EmailCredentials.load_from_os_env()
+    dm = DeployedModel(
+        prediction_url=PREDICTION_URL,
+        creds=creds,
+        token=TOKEN,
+        )
+    dm.predict(x=['test_prediction'])
+    mocked_responses.assert_call_count(PREDICTION_URL, 1)
```
liuverta, line 135: This is really, really cool

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -0,0 +1,346 @@
...
+        )
+    creds = EmailCredentials.load_from_os_env()
+    dm = DeployedModel(
+        prediction_url=PREDICTION_URL,
+        creds=creds,
+        token=TOKEN,
+        )
+    try:
+        dm.predict(x=['test_prediction'])
+    except(RetryError):
```
liuverta: Could we use `with pytest.raises(RetryError):` here, too? or does `mocked_responses.assert_call_count()` preclude that?

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -0,0 +1,346 @@
...
+        json={'test2': 'test2'},
+        status=200,
+        )
+    creds = EmailCredentials.load_from_os_env()
+    dm = DeployedModel(
+        prediction_url=PREDICTION_URL,
+        creds=creds,
+        token=TOKEN,
+        )
+    with mocked_responses as mr:
```
liuverta: What is the difference between using `with mocked_responses` in some tests but not others?

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +240,122 @@ def predict(
...
+                {'verta-request-id': ''},
+                remove=True
+                )
+
+        if prediction_id:
+            http_session.update_session_headers(
+                self._session,
+                {'verta-request-id': prediction_id},
                 )
+        response = self._predict(x, compress)
```
liuverta: We should preserve this behavior where we raise a `RuntimeError` for `400`s: https://github.com/VertaAI/modeldb/blob/04b902a79b6d532f5ad7ac4ce4ec4e802223adbd/client/verta/verta/deployment/_deployedmodel.py#L251-L255

and raise other `4XX`/`5XX`s: https://github.com/VertaAI/modeldb/blob/04b902a79b6d532f5ad7ac4ce4ec4e802223adbd/client/verta/verta/deployment/_deployedmodel.py#L275

liuverta (changes requested):
Finally found time for an in-depth review 😅 Thank you for helping streamline my old code 🙇‍♂️

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -132,46 +117,52 @@ def prediction_url(self, value):
...
-            return self._session.post(self.prediction_url, json=x)
+            response = self._session.post(
+                self.prediction_url,
+                headers=request_headers,
+                json=x,
+                )
+        if not response.ok:
+            _utils.raise_for_http_error(
+                response=response,
+                component_msg='Deployed model encountered an error: ')
```
liuverta: This is not quite what I was trying to indicate with my suggestion https://github.com/VertaAI/modeldb/pull/3332#discussion_r1014483505

- This behavior should only be for `400`s. This would apply to all error codes, but would not be accurate to prepend other error codes with "Deployed model encountered an error".
- I'd prefer if we preserved the old behavior of raising a `RuntimeError` for `400`s, rather than changing it to an `HTTPError`, for compatibility with existing user code that might be trying to catch it, but I'm not wholly opposed to this.
- This approach renders a massive stack trace that exposes all of our middleware, and users tend to have a hard time interpreting large stack traces (not to mention the logging verbosity if this is thrown in a microservice):
  ```python
  ---------------------------------------------------------------------------
  HTTPError                                 Traceback (most recent call last)
  Cell In [3], line 2
        1 data = [1, 2, 3]
  ----> 2 assert endpoint.get_deployed_model().predict(data) == data
  
  File ~/Documents/modeldb/client/verta/verta/deployment/_deployedmodel.py:255, in DeployedModel.predict(self, x, compress, max_retries, always_retry_404, always_retry_429, retry_status, backoff_factor, prediction_id)
      247 if always_retry_404 or always_retry_429:
      248     warnings.warn(
      249         "Deprecation warning: The `always_retry_404` and `always_retry_429`"
      250         "parameters are deprecated and will removed in an upcoming version; "
     (...)
      253         category=FutureWarning,
      254     )
  --> 255 prediction_with_id: Tuple[str, Any] = self.predict_with_id(
      256     x=x,
      257     compress=compress,
      258     prediction_id=prediction_id,
      259     max_retries=max_retries,
      260     retry_status=retry_status,
      261     backoff_factor=backoff_factor,
      262     )
      263 return prediction_with_id[1]
  
  File ~/Documents/modeldb/client/verta/verta/deployment/_deployedmodel.py:325, in DeployedModel.predict_with_id(self, x, compress, max_retries, retry_status, backoff_factor, prediction_id)
      317 # Set the retry config if it differs from current config.
      318 self._session = http_session.set_retry_config(
      319     self._session,
      320     max_retries=max_retries,
      321     status_forcelist=retry_status,
      322     backoff_factor=backoff_factor,
      323     )
  --> 325 response = self._predict(x, compress, prediction_id)
      326 id = response.headers['verta-request-id']
      327 return (id, _utils.body_to_json(response))
  
  File ~/Documents/modeldb/client/verta/verta/deployment/_deployedmodel.py:155, in DeployedModel._predict(self, x, compress, prediction_id)
      149     response = self._session.post(
      150         self.prediction_url,
      151         headers=request_headers,
      152         json=x,
      153         )
      154 if not response.ok:
  --> 155     _utils.raise_for_http_error(
      156         response=response,
      157         component_msg='Deployed model encountered an error: ')
      158 return response
  
  File ~/Documents/modeldb/client/verta/verta/_internal_utils/_utils.py:596, in raise_for_http_error(response, component_msg)
      593 message = f"{component_msg or ''}{response.status_code} {cause} Error:" \
      594           f" {reason} for url: {response.url}"
      595 message += time_str  # attach time to error message
  --> 596 six.raise_from(requests.HTTPError(message, response=response), None)
  
  File <string>:3, in raise_from(value, from_value)
  
  HTTPError: Deployed model encountered an error: 400 Client Error: Traceback (most recent call last):
    File "/Users/miliu/Documents/modeldb/client/verta/verta/registry/_verify_io.py", line 76, in wrapper
    File "/var/folders/_b/kwcscwf12kl4213t2kwdctq00000gn/T/ipykernel_54729/3205215150.py", line 10, in predict
  KeyError: 'non existent key'
   for url: https://app.verta.ai/api/v1/predict/simple-deploy at 2022-11-16 18:59:01.442000 UTC
  ```
  I'd recommend either finding a way to suppress the inner frames, or using the old approach I linked to in my original suggestion and we can revisit this later. I think grabbing `raise_for_http_error()` is a good idea, but it'd need some more work.

liuverta (changes requested):

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +240,109 @@ def predict(
...
+            A backoff factor to apply between retry attempts.  Uses standard urllib3 sleep pattern:
+            ``{backoff factor} * (2 ** ({number of total retries} - 1))`` with a maximum sleep time between requests of
+            120 seconds.
+        prediction_id: str, default None
+            A custom str to use as the ID for the prediction request.  Defaults to a randomly generated numeric id.
+        Returns
+        -------
+        id, prediction: Tuple(str, list)
+            The first element of the tuple is the prediction ID.
+            The second element of the tuple is the output returned by the deployed model for `x`.
```
liuverta: fyi, to be faithful to numpydoc conventions, this would be [more like](https://numpydoc.readthedocs.io/en/stable/format.html#returns)

```suggestion
        id : str
            The prediction ID.
        prediction: List[Any]
            The output returned by the deployed model for `x`.
```

(since it's arguably self-explanatory that a function with "multiple return values" returns a `tuple`)

though I'm not personally sure how that interacts with static type analysis tooling.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +240,109 @@ def predict(
...
-                    response.status_code, sleep
-                )
+            self,
+            x: List[Any],
+            compress=False,
+            max_retries: int = http_session.DEFAULT_MAX_RETRIES,
+            retry_status: Set[int] = http_session.DEFAULT_STATUS_FORCELIST,
+            backoff_factor: float = http_session.DEFAULT_BACKOFF_FACTOR,
+            prediction_id: str = None,
+            ) -> Tuple[str, Dict[str, Any]]:
```
liuverta: `Dict[str, Any]` might not be correct; I think it'd be `List[Any]`?

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +240,109 @@ def predict(
...
+        max_retries : int, default 13
+            Maximum number of retries on status codes listed in ``retry_status`` parameter only.
+        retry_status : set, default {404, 429}
+            Set of status codes, as integers, for which retry attempts should be made.  Overwrites default value of
+            {404, 429}.  Expand the set to include more: ``retry_status={404, 429, <NEW_STATUS_CODE>}``
+        backoff_factor : float, default 0.3
+            A backoff factor to apply between retry attempts.  Uses standard urllib3 sleep pattern:
+            ``{backoff factor} * (2 ** ({number of total retries} - 1))`` with a maximum sleep time between requests of
+            120 seconds.
+        prediction_id: str, default None
```
liuverta: fyi, numpydoc seems to [recommend](https://numpydoc.readthedocs.io/en/stable/format.html#parameters)

```suggestion
        prediction_id: str, optional
```

rather than `default None`

client/verta/tests/unit_tests/test_deployed_model.py
liuverta: @ewagner-verta This didn't quite address my question!

I'm asking if we could use `with pytest.raises(RetryError)` (not `except RetryError`) because we should be explicitly checking that a `RetryError` is raised.

If we simply use `except` rather than `with pytest.raises()`, then if there is a bug in `dm.predict()` that ends up _not_ raising a `RetryError`, this test would happily pass none the wiser (and similarly for the other tests using `except RetryError`).

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -584,11 +589,11 @@ def raise_for_http_error(response):
...
                 cause = "Unexpected"
-            message = "{} {} Error: {} for url: {}".format(
-                response.status_code, cause, reason, response.url
-            )
-            message += time_str  # attach time to error message
-            six.raise_from(requests.HTTPError(message, response=response), None)
+            message = f"{response.status_code} {cause} Error: {reason} " \
+                      f"for url: {response.url}{time_str}"
+            e.args = (message,)
+        traceback.print_exc(file=None, limit=traceback_limit)
```
liuverta: Unfortunately, this doesn't seem to do what we'd like 😕

I was hoping that it would cut down on the stack trace of the exception itself. Instead, what it seems to do is print the traceback to `stderr` and leave the exception unchanged.

liuverta: This means that `HTTPError`s throughout the client now get effectively duplicated output

Screen Shot 2022-12-02 at 3 27 52 PM

liuverta: And model prediction errors still output their massive stack trace (and it's not visually clear to a user what part of this—if any—came from their model, vs being some other error with the client/platform)
Screen Shot 2022-12-02 at 3 18 42 PM

liuverta: Perhaps at this point—if `traceback` doesn't provide another way to do this—we ought to keep the original logic

```python
            if response.status_code in (
                400,
                502,
            ):  # possibly error from the model back end
                try:
                    data = _utils.body_to_json(response)
                except ValueError:  # not JSON response; 502 not from model back end
                    pass
                else:  # from model back end; contains message (maybe)
                    # try to directly print message, otherwise line breaks appear as '\n'
                    msg = data.get("message") or json.dumps(data)
                    raise RuntimeError(
                        "deployed model encountered an error: {}".format(msg)
                    )
```

maybe in a standalone function so that it's not entangled with other code, and would be easier to tear out in the near future.

liuverta (changes requested):

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -156,7 +156,7 @@ def _predict(
             else:  # from model back end; contains message (maybe)
                 # try to directly print message, otherwise line breaks appear as '\n'
                 msg = data.get("message") or json.dumps(data)
-                raise RuntimeError(f"deployed model encountered an error: {msg}")
+                raise RuntimeError(f"Deployed model encountered an error: {msg}")
```
liuverta: Minor note: I had deliberately made this lowercase, since the convention in Python _seems_ to be to have error and warning messages begin with lowercase (e.g. builtins, std lib), though admittedly the ecosystem (and even our client itself) is very inconsistent about it 🤷 so either way is fine.

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -0,0 +1,377 @@
...
+    creds = EmailCredentials.load_from_os_env()
+    dm = DeployedModel(
+        prediction_url=PREDICTION_URL,
+        creds=creds,
+        token=TOKEN,
+        )
+    with pytest.raises(RetryError):
+        dm.predict(x=['test_prediction'])
+        mocked_responses.assert_call_count(PREDICTION_URL, http_session.DEFAULT_MAX_RETRIES + 1)
+        # max_retries + 1 original attempt = total call count
```
liuverta: Minor note: I'd recommend unindenting the `mocked_responses.assert...`s out of the `with pytest.raises(RetryError)` (for this test, and a few others below) since strictly speaking, we're only expecting a `RetryError` to come out of `dm.predict()`. 

But of course I don't expect it to cause any issues in these cases.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -584,13 +580,10 @@ def raise_for_http_error(response):
...
                 cause = "Unexpected"
-            message = "{} {} Error: {} for url: {}".format(
-                response.status_code, cause, reason, response.url
-            )
-            message += time_str  # attach time to error message
+            message = f"{response.status_code} {cause} Error: {reason} " \
+                      f"for url: {response.url}{time_str}"
             six.raise_from(requests.HTTPError(message, response=response), None)
 
-
```
liuverta, line 593: Minor note:

> [Surround top-level function and class definitions with two blank lines.](https://peps.python.org/pep-0008/#blank-lines)

We ought to set up auto-formatting in CI or as a pre-commit hook 🤦

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -0,0 +1,377 @@
...
+        mocked_responses.assert_call_count(PREDICTION_URL, 10)
+
+
+def test_predict_with_prediction_id_provided(mocked_responses) -> None:
+    """ Calling predict while providing a value for `prediction_id` updates and includes the headers in the request. """
+    mocked_responses.post(
+        PREDICTION_URL,
+        json={'test1': 'test1'},
+        status=200,
+        headers={'verta-request-id': 'hereISaTESTidFROMtheUSER'},
```
liuverta, line 187: This test's description says

> includes the headers in the request.

but it doesn't look like we're asserting that `dm.predict()` sends the header; I think this argument means to return it in the response.

Based on https://github.com/getsentry/responses/blob/master/README.rst#request-headers-validation, I think we'd need

```suggestion
        match=[matchers.header_matcher({"verta-request-id": "hereISaTESTidFROMtheUSER"})],
```

with

```python
from responses import matchers
```

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -0,0 +1,377 @@
...
+        prediction_id='hereISaTESTidFROMtheUSER',
+        )
+    mocked_responses.assert_call_count(PREDICTION_URL, 1)
+
+
+def test_predict_with_id_response_includes_id(mocked_responses) -> None:
+    """ Calling predict_with_id returns both the ID from teh request response, and the prediction results """
+    mocked_responses.post(
+        PREDICTION_URL,
+        headers={'verta-request-id': 'hereISanotherTESTidFROMtheUSER'},
```
liuverta: I think we should change this header value to something like

```suggestion
        headers={'verta-request-id': 'AutogeneratedTestId'},
```

since this one isn't provided by the caller.

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -0,0 +1,377 @@
...
+    )
+    with pytest.raises(RuntimeError) as err:
+        dm.predict(x=['test_prediction'])
+    assert str(err.value) == (
+        'Deployed model encountered an error: Here be a message in the response'
+        )
+
+
+def test_predict_400_error_message_missing(mocked_responses) -> None:
+    """ Getting a 400 will render the attached message form the backend if present """
```
liuverta: The name and description of this test don't match what it's doing.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -132,46 +111,63 @@ def prediction_url(self, value):
...
+            ):  # possibly error from the model back end
+            try:
+                data = _utils.body_to_json(response)
+            except ValueError:  # not JSON response; 502 not from model back end
+                pass
+            else:  # from model back end; contains message (maybe)
+                # try to directly print message, otherwise line breaks appear as '\n'
+                msg = data.get("message") or json.dumps(data)
+                raise RuntimeError(f"Deployed model encountered an error: {msg}")
+
```
liuverta, line 160: This is an extremely late realization, but I've just noticed that the original behavior

https://github.com/VertaAI/modeldb/blob/0e69da8d10dd3ba969a5c0d9e73b716f0e524dfc/client/verta/verta/deployment/_deployedmodel.py#L256-L259

was to

1. retry on `5XX`s up to 5 times
2. retry on `404` and `429` indefinitely, optionally

Looks like this PR removes (1), which is...maybe alright? But if it's straightforward to do so, I think we should consider adding `500`, `503`, and `504` to our `DEFAULT_STATUS_FORCELIST` unless we've discussed otherwise and I've forgotten 😬—potentially in a follow-up PR.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -132,46 +111,63 @@ def prediction_url(self, value):
...
+        elif not response.ok:
```
liuverta: I think this should be

```suggestion
        if not response.ok:
```

to catch the

```python
except ValueError
    pass
```

from above (and I think it's safe to be able to assume that `_predict()` will always return a successful `response`, and catch + raise errors otherwise)

liuverta (changes requested):

liuverta:
!rebuild

client/verta/tests/unit_tests/test_deployed_model.py
liuverta: @ewagner-verta This still doesn't seem correct to me. The test name and docstring say `400` but the test is returning and handling a `401` (which is not the same as `400`)

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -183,43 +179,63 @@ def get_curl(self):
...
+            Maximum number of retries on status codes listed in ``retry_status``.
+        always_retry_404 : bool, default False
+            Deprecated: Whether to retry on 404s indefinitely. This is to accommodate model deployment warm-up.
+        always_retry_429 : bool, default False
+            Deprecated: Whether to retry on 429s indefinitely. This is to accommodate third-party cluster
             autoscalers, which may take minutes to launch new pods for servicing requests.
-
+        retry_status : set, default {404, 429, 500, 503, 504}
+            Set of status codes, as integers, for which retry attempts should be made.  Overwrites default value.
+            Expand the set to include more: ```retry_status={404, 429, 500, 503, 504, <NEW_STATUS_CODE>}```
```
liuverta: Should we use just two backticks instead?

```suggestion
            Expand the set to include more: ``retry_status={404, 429, 500, 503, 504, <NEW_STATUS_CODE>}``
```

Three gets rendered as

Screen Shot 2022-12-21 at 1 52 00 PM

which I _think_ might not be the intention, but I'm not sure.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -233,99 +249,110 @@ def predict(
...
+        ----------
+        x : list
+            A batch of inputs for the model.
+        compress : bool, default False
+            Whether to compress the request body.
+        max_retries : int, default 13
+            Maximum number of retries on status codes listed in ``retry_status`` parameter only.
+        retry_status : set, default {404, 429, 500, 503, 504}
+            Set of status codes, as integers, for which retry attempts should be made.  Overwrites default value.
+            Expand the set to include more: ```retry_status={404, 429, 500, 503, 504, <NEW_STATUS_CODE>}```
```
liuverta: See above comment

```suggestion
            Expand the set to include more: ``retry_status={404, 429, 500, 503, 504, <NEW_STATUS_CODE>}``
```

liuverta (approved):
🚀 `HTTPError` messages throughout the client look good, `400` model `RuntimeError`s look good (stack trace is a little longer than it used to be, but close enough), and retries appear to be happening.

One note about documentation.

One thing we should follow up on is that retries no longer emit anything for the user to know that they're happening. The client used to `print`

```
received status 500; retrying in 0.6s
received status 500; retrying in 1.2s
received status 500; retrying in 2.4s
received status 500; retrying in 4.8s
received status 503; retrying in 9.6s
```

but currently we get nothing, I think. I don't see any logs emitted by `requests`/`urllib3` on retries. So to the user, it just looks like the client is hanging on predictions. I'd strongly recommend **not** trying to add it to this PR because it's already gotten unwieldy to review, but we should see if we can add back some visibility there.

PR #3334, opened 2022-11-03
[VUMM-745] "java.lang.NullPointerException: extraHeaders" when loading projects list

anandjakhaniya:
## Impact and Context
The problem was happen because we have to call UAC from `FutureJdbi.withHandle` so our GRPC context is not shared with `FutureJdbi.withHandle` so when we are trying to fetch headers that context was null and we got the error.

Solution: I have passed the GRPC context to `FutureJdbi`'s methods while executing.
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [x] Other (explain) 
- I can reproduce it locally so I have tested the same using postman.
- Also tested deploying on aj.dev and it works fine now.

## Reverting


conradoverta (commented):
We should probably wrap the executor like we do in futures to automatically capture all the right thread local things

jkwatson-verta (approved):

PR #3335, opened 2022-11-03
feat: [CAT-155] - Log dataset into RMV using Client

marcelo-verta-ai:
# Summary

Log dataset into RMV using Client

# References

https://vertaai.atlassian.net/browse/CAT-155?atlOrigin=eyJpIjoiNmU5YjM3YTg2N2M1NDUxMmIyMGU1NWE2OWNkMDUwYTUiLCJwIjoiaiJ9

# Test Result (2022-11-07 10:00 GMT -3)
```
export VERTA_HOST="https://modelcatalog.dev.verta.ai/"
export VERTA_EMAIL="marcelo@verta.ai"
export VERTA_DEV_KEY="7ae98d72-d0dd-4034-9fb1-e2f79e894fd9"
python3.9 -m pytest -k "TestLogDataset" -rfE
============================================================================= test session starts ==============================================================================
platform linux -- Python 3.9.15, pytest-7.2.0, pluggy-1.0.0
rootdir: /opt/repositorio-git/verta/modeldb/client/verta/tests, configfile: pytest.ini
plugins: xdist-3.0.2, hypothesis-6.56.4
collected 1258 items / 1257 deselected / 3 skipped / 1 selected                                                                                                                

registry/model_version/test_log_dataset.py .                                                                                                                             [100%]

=============================================================================== warnings summary ===============================================================================
test_utils/test_pip_requirements.py:170
  /opt/repositorio-git/verta/modeldb/client/verta/tests/test_utils/test_pip_requirements.py:170: NonInteractiveExampleWarning: The `.example()` method is good for exploring strategies, but should only be used interactively.  We recommend using `@given` for tests - it performs better, saves and replays failures to avoid flakiness, and reports minimal examples. (strategy: versions())
    version=versions().example(),  # pylint: disable=no-member

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================== 1 passed, 3 skipped, 1257 deselected, 1 warning in 28.00s ===========================================================


```

### DIFF
--- a/client/verta/tests/registry/model_version/test_log_dataset.py
+++ b/client/verta/tests/registry/model_version/test_log_dataset.py
@@ -24,10 +24,20 @@ class TestLogDataset:
             model_version.get_dataset_version("fake")
 
         dataset = model_version.get_dataset_version(key1)
-        assert dataset_version1.id == dataset.linked_artifact_id
+        assert dataset_version1.id == dataset.id
 
         dataset = model_version.get_dataset_version(key2)
-        assert dataset_version2.id in dataset.linked_artifact_id
+        assert dataset_version2.id == dataset.id
 
         dataset = model_version.get_dataset_version(key3)
-        assert dataset_version3.id in dataset.linked_artifact_id
+        assert dataset_version3.id == dataset.id
+
+        with pytest.raises(KeyError, match="no dataset found with key"):
+            model_version.del_dataset_version("fake")
+
+        model_version.del_dataset_version(key3)
+
+        model_version = client.get_registered_model_version(id=model_version.id)
+
+        with pytest.raises(KeyError, match="no dataset found with key"):
+            model_version.get_dataset_version(key3)
--- a/client/verta/tests/registry/model_version/test_log_dataset.py
+++ b/client/verta/tests/registry/model_version/test_log_dataset.py
@@ -20,10 +20,8 @@ class TestLogDataset:
 
         model_version = client.get_registered_model_version(id=model_version.id)
 
-        with pytest.raises(KeyError) as excinfo:
+        with pytest.raises(KeyError, match="no dataset found with key"):
             model_version.get_dataset_version("fake")
-
-        assert "no dataset found with key" in str(excinfo.value)
 
         dataset = model_version.get_dataset_version(key1)
         assert dataset_version1.id == dataset.linked_artifact_id
--- a/client/verta/tests/registry/model_version/test_log_dataset.py
+++ b/client/verta/tests/registry/model_version/test_log_dataset.py
@@ -2,61 +2,36 @@
 
 import pytest
 
-from verta._protos.public.common import CommonService_pb2 as _CommonCommonService
+from verta.dataset import Path
 
 pytestmark = pytest.mark.not_oss  # skip if run in oss setup. Applied to entire module
 
 class TestLogDataset:
-    def test_log_dataset(self, model_version):
+    def test_log_dataset(self, client, model_version, dataset):
         key1, key2, key3 = "version1", "version2", "version3"
 
-        model_version.log_dataset(
-            key=key1,
-            path="ModelVersionEntity/191/train",
-            path_only=True,
-            artifact_type=_CommonCommonService.ArtifactTypeEnum.DATA,
-            linked_artifact_id="fbcbd389db61d78c479634b0824cf063ffc63a6af7ba5d763388a42001ed6191",
-        )
+        dataset_version1 = dataset.create_version(Path("conftest.py"))
+        dataset_version2 = dataset.create_version(Path(["modelapi_hypothesis/"]))
+        dataset_version3 = dataset.create_version(Path(["modelapi_hypothesis/"]))
 
-        model_version.log_dataset(
-            key=key2,
-            path="ModelVersionEntity/192/train",
-            path_only=True,
-            artifact_type=_CommonCommonService.ArtifactTypeEnum.BLOB,
-            linked_artifact_id="fbcbd389db61d78c479634b0824cf063ffc63a6af7ba5d763388a42001ed6192",
-        )
+        model_version.log_dataset_version(key=key1, dataset_version=dataset_version1)
+        model_version.log_dataset_version(key=key2, dataset_version=dataset_version2)
+        model_version.log_dataset_version(key=key3, dataset_version=dataset_version3)
 
-        model_version.log_dataset(
-            key=key3,
-            path="ModelVersionEntity/193/train",
-            path_only=True,
-            artifact_type=_CommonCommonService.ArtifactTypeEnum.MODEL,
-            linked_artifact_id="fbcbd389db61d78c479634b0824cf063ffc63a6af7ba5d763388a42001ed6193",
-        )
-
-        keys = model_version.get_dataset_keys()
-
-        assert len(keys) == 3
-        assert key1 in keys
-        assert key2 in keys
-        assert key3 in keys
-
-        datasets = model_version.get_datasets()
-        assert len(datasets) == 3
+        model_version = client.get_registered_model_version(id=model_version.id)
 
         with pytest.raises(KeyError) as excinfo:
-            model_version.get_dataset("fake")
+            model_version.get_dataset_version("fake")
 
         assert "no dataset found with key" in str(excinfo.value)
 
-        dataset = model_version.get_dataset(key1)
-        assert dataset.artifact_type == _CommonCommonService.ArtifactTypeEnum.DATA
-        assert "fbcbd389db61d78c479634b0824cf063ffc63a6af7ba5d763388a42001ed6191" in dataset.linked_artifact_id
+        dataset = model_version.get_dataset_version(key1)
+        assert dataset_version1.id in dataset.linked_artifact_id
 
-        dataset = model_version.get_dataset(key2)
-        assert dataset.artifact_type == _CommonCommonService.ArtifactTypeEnum.BLOB
-        assert "fbcbd389db61d78c479634b0824cf063ffc63a6af7ba5d763388a42001ed6192" in dataset.linked_artifact_id
+        dataset = model_version.get_dataset_version(key2)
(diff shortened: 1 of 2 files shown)


client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1516,98 @@ def set_hide_output_label(self, hide):
...
+        Get a dataset.
+
+        Parameters
+        ----------
+        key : str
+            Artifact key.
+
+        Returns
+        -------
+        `CommonService_pb2.Artifact`.
```
liuverta: We should never return an object from `*_pb2` to the user because they are not user-friendly. Please see Technical Implementation Plan item 3 in [the Jira ticket](https://vertaai.atlassian.net/browse/CAT-155) for how to return a client `DatasetVersion` object, if possible.

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1516,98 @@ def set_hide_output_label(self, hide):
...
+        if not key:
+            raise ValueError("key is not specified")
+        if not path:
+            raise ValueError("artifact_path is not specified")
+        if not path_only:
+            raise ValueError("path_only is not specified")
+        if not artifact_type:
+            raise ValueError("artifact_type is not specified")
+        if not linked_artifact_id:
+            raise ValueError("linked_artifact_id is not specified")
```
liuverta: I do not think these checks are necessary, because Python will require all of these arguments when a user calls `log_dataset()`.

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1516,98 @@ def set_hide_output_label(self, hide):
     def get_hide_output_label(self):
         self._refresh_cache()
         return self._msg.hide_output_label
+
+    def log_dataset(self, key, path, path_only, artifact_type, linked_artifact_id):
```
liuverta: This is likely a confusing API for the user. `path_only`, `artifact_type`, and `linked_artifact_id` are internal implementation details of our platform that the user shouldn't need to worry about. Please see the description of [the Jira ticket](https://vertaai.atlassian.net/browse/CAT-155):

> log_dataset_version() should be similar to ExperimentRun.log_dataset_version()

and it should accept a client `DatasetVersion` object.

liuverta: And also, unless there is a reason not to, would you mind renaming these functions to `log_dataset_version()`, `get_dataset_version()`, `del_dataset_version()`, etc., as per the Jira ticket description?

liuverta (changes requested):
I recommend following my API recommendations in the Jira ticket description (I was worried they would be easy to miss 😞) to match the API for `ExperimentRun`.

But this looks good as a first step! I know it was a lot of hard work between you and Matheus to test this

client/verta/tests/registry/model_version/test_log_dataset.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from verta.dataset import Path
+
+pytestmark = pytest.mark.not_oss  # skip if run in oss setup. Applied to entire module
+
+class TestLogDataset:
+    def test_log_dataset(self, client, model_version, dataset):
```
ewagner-verta, line 10: Perhaps we should start moving away from unit tests that make actual API calls to the backend and start mocking them as a general practice?  It seems like our full test suite is very heavy on calls over the network and that makes it painfully slow to run.  I'll let @liuverta weigh in on that, as I'm not sure it needs to be addressed immediately in this PR.

liuverta: This test is alright to me, for now.

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1519,59 @@ def set_hide_output_label(self, hide):
...
+        Gets the DatasetVersion with name `key` from this Model Version.
+
+        Parameters
+        ----------
+        key : str
+            Name of the dataset version.
+
+        Returns
+        -------
+        `DatasetVersion <dataset.html>`_
```
liuverta: We should also try returning

```
:class:`~verta.dataset.entities.DatasetVersion`
```

here, instead of the protobuf `self._msg.datasets`

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -42,6 +42,9 @@
 from .. import lock, DockerImage
 from ..stage_change import _StageChange
 
+from verta.dataset.entities import (
+    _dataset
```
liuverta: I recommend importing the `_dataset_version` module, because that is the more "accurate" one for `DatasetVersion`

```suggestion
    _dataset_version
```

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1519,59 @@ def set_hide_output_label(self, hide):
...
+
+        Returns
+        -------
+        `DatasetVersion <dataset.html>`_
+            DatasetVersion associated with the given key.
+
+        """
+        for dataset in self._msg.datasets:
+            if dataset.key == key:
+                return dataset
```
liuverta: I'm not 100% sure this works exactly, but based on [`ExperimentRun.get_dataset()`](https://github.com/VertaAI/modeldb/blob/b1c01068ca1abae05c8126f7fd78e84660555ad9/client/verta/verta/tracking/entities/_experimentrun.py#L1222-L1227):

```suggestion
                return _dataset_version.DatasetVersion(
                    self._conn,
                    self._conf,
                    _dataset_version.DatasetVersion._get_proto_by_id(
                        self._conn, dataset.linked_artifact_id
                    ),
                )
```

and then _I think_ for the test we can check `dataset.id` instead of `dataset.linked_artifact_id`

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1519,59 @@ def set_hide_output_label(self, hide):
     def get_hide_output_label(self):
         self._refresh_cache()
         return self._msg.hide_output_label
+
```
liuverta, line 1522: In [the Jira ticket](https://vertaai.atlassian.net/browse/CAT-155), Andy also requested being able to delete a dataset version (`del_dataset_version()`). Do you know if this is possible through the backend? 🤔

client/verta/tests/registry/model_version/test_log_dataset.py
```diff
@@ -0,0 +1,35 @@
...
+
+        model_version = client.get_registered_model_version(id=model_version.id)
+
+        with pytest.raises(KeyError) as excinfo:
+            model_version.get_dataset_version("fake")
+
+        assert "no dataset found with key" in str(excinfo.value)
+
+        dataset = model_version.get_dataset_version(key1)
+        assert dataset_version1.id in dataset.linked_artifact_id
```
liuverta: These three `asserts` should be `==` instead of `in`.

```suggestion
        assert dataset_version1.id == dataset.linked_artifact_id
```

client/verta/tests/registry/model_version/test_log_dataset.py
```diff
@@ -0,0 +1,35 @@
...
+        model_version.log_dataset_version(key=key1, dataset_version=dataset_version1)
+        model_version.log_dataset_version(key=key2, dataset_version=dataset_version2)
+        model_version.log_dataset_version(key=key3, dataset_version=dataset_version3)
+
+        model_version = client.get_registered_model_version(id=model_version.id)
+
+        with pytest.raises(KeyError) as excinfo:
+            model_version.get_dataset_version("fake")
+
+        assert "no dataset found with key" in str(excinfo.value)
```
liuverta: fyi, we can do this if it's clearer!

```suggestion
        with pytest.raises(KeyError, match="no dataset found with key"):
            model_version.get_dataset_version("fake")
```

liuverta (changes requested):
Just a few more comments.

client/verta/verta/registry/entities/_modelversion.py
liuverta: Actually, maybe it is possible to do through the client, similar to what we do in `del_artifact()`: https://github.com/VertaAI/modeldb/blob/c8e78ed50153469a171dbfe0091f0725c7a3bd19/client/verta/verta/registry/entities/_modelversion.py#L742-L753

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1519,95 @@ def set_hide_output_label(self, hide):
...
+        ----------
+        key : str
+            Name of the dataset version.
+
+        Returns
+        -------
+        :class:`~verta.dataset.entities.DatasetVersion`
+            DatasetVersion associated with the given key.
+
+        """
```
liuverta, line 1572: I forgot to mention! We should refresh the cache like this on any `get_` to get `self._msg` from registry backend

```suggestion
        """
        self._refresh_cache()
```

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -1514,3 +1519,95 @@ def set_hide_output_label(self, hide):
...
+        ----------
+        key : str
+            Name of dataset version.
+
+        """
+        if key == self._MODEL_KEY:
+            raise ValueError(
+                "model can't be deleted through del_dataset_version(); consider using del_model() instead"
+            )
+
```
liuverta: We actually do not need this here—it was only relevant for `del_artifacts()` because it modifies `self._msg.artifacts`. Sorry for the confusion!

```suggestion
```

client/verta/verta/registry/entities/_modelversion.py
liuverta: Also, this change means we will not _need_ the multiple

```python
model_version = client.get_registered_model_version(id=model_version.id)
```

in the test.

liuverta (changes requested):
A few more small changes, but this looks good overall.

liuverta (approved):
Looks great, thank you!

PR #3338, opened 2022-11-07
[VUMM-748] mysql: extract ddl modeldb schema

anandjakhaniya:
## Impact and Context
- [x] SQL
- [ ] MSSQL
- [ ] H2: Not supported in this release 

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


backend/src/main/resources/migration/releases_ddl/mysql/mysql_release_2022_08.up.sql
```diff
@@ -0,0 +1,1338 @@
+-- MySQL dump 10.13  Distrib 8.0.12, for Win64 (x86_64)
```
jkwatson-verta: this file will need to be named by the naming scheme that is being used by the migrator. I think it should be
`1_release_2022_08.up.sql` , so that it can be properly ordered by the migration tool to be the first one run.

anandjakhaniya: Fixed it

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migration.java
```diff
@@ -3,7 +3,7 @@
 import lombok.Value;
 
 @Value
-class Migration implements Comparable<Migration> {
+public class Migration implements Comparable<Migration> {
```
jkwatson-verta, line 6: not too thrilled with making this stuff public, but I see why you needed it. Once things are stable, we should remove the single-migration tests from the server module, and just rely on the standard integration tests to validate. Then, we can tighten this back up again.

jkwatson-verta (approved):

PR #3340, opened 2022-11-07
chore: Display keys, not protos, in RMV repr

liuverta:
## Impact and Context

#3335 added raw protobuf objects to `RegisteredModelVersion.__repr__()`, which can appear rather cluttered and difficult to read. This changes the `repr()` to only show their keys instead.

### Before

```
...
artifact keys: []
datasets: [key: "train"
path: "liu2/ModelVersionEntity/699/train"
path_only: true
artifact_type: DATA
linked_artifact_id: "4d0e855b6531e37b52856a345f95b9e822e70a0ae38f4d54b3b014bf3668a485"
]
code_blob_map: {'train': git {
  repo: "git@github.com:VertaAI/models.git"
  hash: "52f3d22"
}
}
```

### After

```
...
artifact keys: []
dataset version keys: ['train']
code version keys: ['train']
```

## Risks and Area of Effect

None—the user-facing code change (in `__repr__()`) is both currently-unpublished and self-isolated, and existing code coverage is maintained (albeit now in unit test form).

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

```zsh
 % pytest unit_tests
platform darwin -- Python 3.10.6, pytest-7.1.3, pluggy-1.0.0
plugins: xdist-2.5.0, forked-1.4.0, hypothesis-6.56.1
collected 4 items                                                                                                           

unit_tests/test_endpoint_class.py ...                                                                                 [ 75%]
unit_tests/registry/test_registered_model_version.py .                                                                [100%]

===================================================== warnings summary ======================================================
unit_tests/test_endpoint_class.py::test_get_deployed_model_call_get_status
unit_tests/test_endpoint_class.py::test_get_deployed_model_call_get_status
  /Users/miliu/Documents/modeldb/client/verta/verta/_internal_utils/_utils.py:89: DeprecationWarning: Using 'method_whitelist
' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
    self.retry = Retry(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================== 4 passed, 2 warnings in 5.34s ===============================================

```

## Reverting

Revert this PR.

### DIFF
--- a/client/verta/tests/unit_tests/conftest.py
+++ b/client/verta/tests/unit_tests/conftest.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+
+"""Pytest fixtures for use in client unit tests."""
+
+import os
+from unittest.mock import patch
+
+import pytest
+
+from verta._internal_utils._utils import Connection, Configuration
+from verta.credentials import EmailCredentials
+
+
+@pytest.fixture(scope="session")
+def mock_conn() -> Connection:
+    """ Return a mocked object of the _internal_utils._utils.Connection class for use in tests """
+    with patch.dict(os.environ, {'VERTA_EMAIL': 'test_email@verta.ai', 'VERTA_DEV_KEY':'123test1232dev1232key123'}):
+        credentials = EmailCredentials.load_from_os_env()
+
+    return Connection(
+        scheme='https',
+        socket='test_socket',
+        credentials=credentials,
+    )
+
+@pytest.fixture(scope="session")
+def mock_config() -> Configuration:
+    """ Return a mocked object of the _internal_utils._utils.Configuration class for use in tests """
+    return Configuration(use_git=False, debug=False)
--- a/client/verta/tests/unit_tests/registry/test_registered_model_version.py
+++ b/client/verta/tests/unit_tests/registry/test_registered_model_version.py
@@ -62,14 +62,13 @@ def test_repr(mock_conn, mock_config, model_ver_proto, workspace):
     msg: RegistryService_pb2.ModelVersion = model_ver._msg
 
     assert f"version: {msg.version}" in repr_lines
-    expected_url = "{}://{}/{}/registry/{}/versions/{}".format(
+    assert "url: {}://{}/{}/registry/{}/versions/{}".format(
         mock_conn.scheme,
         mock_conn.socket,
         workspace,
         msg.registered_model_id,
         msg.id,
-    )
-    assert f"url: {expected_url}" in repr_lines
+    ) in repr_lines
     assert f"time created: {timestamp_to_str(msg.time_created)}" in repr_lines
     assert f"time updated: {timestamp_to_str(msg.time_updated)}" in repr_lines
     assert f"labels: {msg.labels}" in repr_lines
--- a/client/verta/tests/unit_tests/registry/test_registered_model_version.py
+++ b/client/verta/tests/unit_tests/registry/test_registered_model_version.py
@@ -14,7 +14,7 @@ from verta._protos.public.registry import RegistryService_pb2
 from verta.registry.entities import RegisteredModelVersion
 
 from ..strategies import (
-    artifact_protos,
+    artifact_proto,
     code_blob_proto,
     int64,
     model_artifact_proto,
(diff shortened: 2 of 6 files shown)


liuverta:
I need to add test coverage.

client/verta/tests/unit_tests/registry/test_registered_model_version.py
```diff
@@ -0,0 +1,90 @@
...
+@patch.object(RegisteredModelVersion, "_refresh_cache", return_value=None)
+@hypothesis.given(model_ver_proto=model_ver_proto(), workspace=st.text(ascii_letters, min_size=1))
+def test_repr(mock_conn, mock_config, model_ver_proto, workspace):
+    """
+    Verify that RegisteredModelVersion.__repr__() renders expected fields and correct values.
+
+    This test does not yet cover all available fields, but exists in its current form to cover newly-added model catalog fields.
+
+    """
+    model_ver = RegisteredModelVersion(conn=mock_conn, conf=mock_config, msg=model_ver_proto)
```
liuverta, line 64: This ought to pave the way for fully-mocked `RegisteredModelVersion` tests.

client/verta/verta/registry/entities/_modelversion.py
```diff
@@ -125,9 +125,9 @@ def __repr__(self):
                 "registered model id: {}".format(msg.registered_model_id),
                 "experiment run id: {}".format(msg.experiment_run_id),
                 # "archived status: {}".format(msg.archived == _CommonCommonService.TernaryEnum.TRUE),
-                "artifact keys: {}".format(artifact_keys),
-                "datasets: {}".format(msg.datasets),
-                "code_blob_map: {}".format(msg.code_blob_map),
+                "artifact keys: {}".format(sorted(artifact_keys)),
+                "dataset version keys: {}".format(sorted(dataset.key for dataset in msg.datasets)),
+                "code version keys: {}".format(sorted(msg.code_blob_map.keys())),
```
liuverta, line 130: @marcelo-verta-ai This is the main code change that is relevant for you. Everything else is a testing change that is more relevant for @ewagner-verta (but feel free to review the whole PR!).

client/verta/tests/unit_tests/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+from unittest.mock import patch
+
+import pytest
+
+from verta._internal_utils._utils import Connection, Configuration
+from verta.credentials import EmailCredentials
+
+
+@pytest.fixture(scope="session")
```
liuverta, line 14: I scoped these to the entire test session (instead of the default per-function) because Hypothesis asked for it, plus it didn't seem necessary to recreate these objects for every single test.

client/verta/tests/unit_tests/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+def mock_conn() -> Connection:
+    """ Return a mocked object of the _internal_utils._utils.Connection class for use in tests """
+    with patch.dict(os.environ, {'VERTA_EMAIL': 'test_email@verta.ai', 'VERTA_DEV_KEY':'123test1232dev1232key123'}):
+        credentials = EmailCredentials.load_from_os_env()
```
liuverta, line 18: I changed this from a decorator on the function to a context here, because I'm not sure whether patching `os.environ` (which as far as I know is global) on a session-scoped fixture would affect other tests.

marcelo-verta-ai (approved):

client/verta/tests/unit_tests/conftest.py
ewagner-verta: Makes sense to me.  We should be able to recycle a single fixture here.  Good call

client/verta/tests/unit_tests/conftest.py
ewagner-verta: I see no downside to this approach.  We only need those env vars in place to create the fixture once anyway.

client/verta/tests/unit_tests/registry/test_registered_model_version.py
ewagner-verta: This is really cool.  I've been digging into `hypothesis` documentation and it seems like a rather brilliant testing solution.  I've written extra fixtures to mock edge cases and used them to parameterize queries but this approach is like that only a thousand times more comprehensive.

ewagner-verta (approved):
This is great work.  Thanks for pulling me into the PR.  I am a full blown convert to the ways of property-testing now.

PR #3341, opened 2022-11-07
chore: [VR-14170] Implement database locking during migrations.

jkwatson-verta:
## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -16,6 +19,18 @@ public Migrator(Connection connection, String resourcesDirectory) {
     this.resourcesDirectory = resourcesDirectory;
   }
 
+  /** Assumes that the underlying database has already been created. */
+  public void performMigration(RdbConfig config) throws SQLException, MigrationException {
```
liam-verta, line 23: `performMigration` vs `executeMigration`? 

Are these named differently just to avoid overloading?

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationDatastore.java
```diff
@@ -0,0 +1,29 @@
+package ai.verta.modeldb.common.db.migration;
+
+import ai.verta.modeldb.common.config.RdbConfig;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public interface MigrationDatastore {
+  /** Lock the database in order to perform migrations. */
```
liam-verta, line 8:
```suggestion
/**
 * A datastore that can perform migrations. 
 */
public interface MigrationDatastore {
  /**
   * Lock the database in order to perform migrations.
   */
```

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
```diff
@@ -0,0 +1,35 @@
+package ai.verta.modeldb.common.db.migration;
+
+import java.nio.charset.StandardCharsets;
+import java.sql.SQLException;
+import java.util.zip.CRC32;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class MigrationTools {
```
liam-verta, line 9: Perhaps `MigrationDatastore` should be a base class instead of an interface and these methods should be in that class?

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
```diff
@@ -0,0 +1,35 @@
...
+  public static String generateLockId(String database, String otherLockInfo) {
+    CRC32 crc32 = new CRC32();
+    crc32.update((database + ":" + otherLockInfo).getBytes(StandardCharsets.UTF_8));
+    return String.valueOf(crc32.getValue());
+  }
+
+  public static void lockDatabase(MigrationDatastore datastore) throws MigrationException {
+    int tries = 0;
+    while (tries++ < 15) {
```
liam-verta, line 18: Maybe make this count a constant instead of a literal?

liam-verta (approved):
I'm not going to block on my question about `MigrationDatastore` and `MigrationTools` please consider it before moving on.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
jkwatson-verta: yeah, that's a good idea. honestly, this whole block of code feels super hacky to me, and I don't like it much.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
jkwatson-verta: I would like to avoid building class hierarchies. They always end up growing in ways that are confusing and hard to understand.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: the execute migration method is just for a single migration (and is non-public), rather than "perform" which is the whole orchestration of migrations.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
jkwatson-verta: But, this should be an internal implementation detail, and I should make this utility class non-public, for sure.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
jkwatson-verta: I'll pick this up in a later PR, since I've got two follow-on branches to be managed right now. If that's ok.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
jkwatson-verta: I'll make sure to do that in a follow-on PR, since I've got two branches I'd need to update with this change, otherwise.

jkwatson-verta: Just as a note in favor of not having a class hierarchy... the H2 implementation doesn't need to actually do any real database locking...just a local lock, since the db is in-memory. That would mean that the H2 implementation had a method attached to it that wasn't relevant to its job at all.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: I'm going to rename the "execute" one to "executeSingleMigration" downstream, to make it a bit more clear what's up.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/MigrationTools.java
liam-verta: I agree that class hierarchies can get out of hand really fast.  But avoiding them completely (while still using interfaces) results situations like this where you have a utility class whose methods are really only used in one scenario.  The `MigrationDatastore` interface might as well be the internal too since we provide implementations for all the database types that we support.  But older Java only supports public interfaces...

jkwatson-verta: Can definitely make the interface non-public and am happy to do so.

PR #3346, opened 2022-11-08
chore: [VR-14175] Implement migration orchestration

jkwatson-verta:
Does not include dirty-handling functionality.

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -22,19 +31,147 @@ public Migrator(Connection connection, String resourcesDirectory) {
...
+          .collect(Collectors.toCollection(TreeSet::new));
+    } catch (URISyntaxException e) {
+      throw new MigrationException("Failed to read migration files.", e);
+    }
+  }
+
+  private MigrationState findCurrentState() throws SQLException {
+    try (PreparedStatement ps =
+        connection.prepareStatement(
+            "select version, dirty from " + MigrationDatastore.SCHEMA_MIGRATIONS_TABLE)) {
```
coryverta, line 185: Is this table guaranteed to contain exactly one row?

jkwatson-verta: that is the intent, yes. If it contains more than one row, things have gone wrong. The only insert is on initialization; everything after that should only be an update.

jkwatson-verta: (see the `setupDatastore` method for the implementation)

backend/common/src/test/java/ai/verta/modeldb/common/db/migration/MigrationTest.java
```diff
@@ -11,14 +10,16 @@ class MigrationTest {
   void ordering() {
```
coryverta, line 10: This is a great candidate for a property test.  
1. Build a generator that produces the set of migrations. 
2. Write a clean-room brute force method that processes the set and returns the known correct result.
3. Execute the clean-room method and the method-under-test and compare the results

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migration.java
```diff
@@ -3,25 +3,35 @@
...
   }
 
-  public int getNumber() {
+  int getNumber() {
     String[] pieces = filename.split("_");
     return Integer.parseInt(pieces[0]);
   }
 
-  public boolean isUp() {
+  boolean isUp() {
```
eheinlein-verta, line 33: I'm not clear on what `up` and `down` mean here. I can see that it is part of the filename, but I don't follow the meaning. Is that documented elsewhere?

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migration.java
```diff
@@ -3,25 +3,35 @@
...
 
 @Value
-public class Migration implements Comparable<Migration> {
+class Migration implements Comparable<Migration> {
   String filename;
 
   @Override
   public int compareTo(Migration o) {
-    return this.getNumber() - o.getNumber();
+    int versionDiff = this.getNumber() - o.getNumber();
```
coryverta, line 11: Are null checks on `o` required here for safety?

coryverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
coryverta: Cool

backend/common/src/test/java/ai/verta/modeldb/common/db/migration/MigrationTest.java
jkwatson-verta: Heh. I _just_ commented in slack that I wanted to write one.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -22,19 +31,147 @@ public Migrator(Connection connection, String resourcesDirectory) {
...
-    MigrationDatastore migrationDatastore = MigrationDatastore.create(config, connection);
+    performMigration(config, null);
+  }
 
+  void performMigration(RdbConfig config, Integer desiredVersion)
+      throws SQLException, MigrationException {
+    MigrationDatastore migrationDatastore = setupDatastore(config);
+    MigrationState currentState = findCurrentState();
+
+    if (currentState.isDirty()) {
```
eheinlein-verta, line 46: it looks like `findCurrentState()` can return null, so should there be a null check here?

backend/common/src/test/java/ai/verta/modeldb/common/db/migration/MigrationTest.java
jkwatson-verta: oh, not for this, but for the migrator itself. a good idea to add one for this as well, though.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migration.java
jkwatson-verta: I *think* I'm ok with things blowing up if somehow a null record got into the set that uses this.  I think if we put explicit null handling into this method, it would just blow up in a more explicit way? Also, I don't think TreeSet allows null values to be put into it, so I _think_ we're safe with assuming no nulls?

backend/common/src/test/java/ai/verta/modeldb/common/db/migration/MigrationTest.java
jkwatson-verta: I do worry a little bit about the "clean room" sort also being tested by a property test like this. What if they were both wrong?

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -22,19 +31,147 @@ public Migrator(Connection connection, String resourcesDirectory) {
...
+                .map(Migration::getNumber)
+                .max(Integer::compareTo)
+                .orElse(0);
+    log.info("Starting database migration process to version: " + versionToTarget);
+
+    if (currentState.getVersion() == versionToTarget) {
+      log.info("No migrations to perform. Database is already at version " + versionToTarget);
+      return;
+    }
+    NavigableSet<Migration> migrationsToPerform =
```
eheinlein-verta: TIL about `NavigableSet` :)

backend/common/src/test/java/ai/verta/modeldb/common/db/migration/MigrationTest.java
jkwatson-verta: Better might be: generate a random set of valid Migration instances, put them in a set, then validate the order desired is strictly enforced (that is... monitonically increasing numbers with up always before down).

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: this is a good candidate for a property test invariant!

eheinlein-verta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migration.java
jkwatson-verta: In the design doc: https://github.com/VertaAI/docs/blob/main/design-docs/2022-09-migrations.md

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: If we have a null current state, then something has failed during initialization. The null check would just throw an IllegalStateException, so not sure that's preferable to the NPE that we'd get now?  I'm not against it, though, if you think that would be clearer.

eheinlein-verta: I think that's fine. 👍

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migration.java
eheinlein-verta: Got it, that's kind of what I was expecting. At first I had it in my head that we were trying to avoid any rollbacks, which is why I wasn't sure. But then I realized it was that we don't want to have any **changes** to existing migrations, not that we don't provide rollback options.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: I re-learned about it during this work, myself. I found it when I went to reverse the order of the TreeSet and found that it returned a NavigableSet!

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: I'd rather see an explicit IllegalStateException with a good message, rather than having to decrypt an NPE from the logs. Will add!

jkwatson-verta: done! good callout

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migration.java
liam-verta: Maybe add javadoc explanation? Folks with access do this repo (open source) don't all have access to the `docs` repo.

jkwatson-verta: good idea. will add!

jkwatson-verta: done!

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
liam-verta: Are you using any of the methods of `NavigableSet`?  If not, let's use  `SortedSet` or even `Set`.  Unless it being a `NavigableSet` is meaningful (or is likely to be meaningful), we should ignore that detail.

jkwatson-verta: We are using the `descendingSet` method from NavigableSet when applying the `down` migrations.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -22,19 +31,150 @@ public Migrator(Connection connection, String resourcesDirectory) {
...
+  private NavigableSet<Migration> gatherMigrationsToPerform(
+      MigrationState currentState, int versionToTarget) throws MigrationException {
+    NavigableSet<Migration> migrationsToPerform;
+    if (currentState.getVersion() < versionToTarget) {
+      migrationsToPerform =
+          gatherMigrations(
+              Migration::isUp,
+              migration ->
+                  (migration.getNumber() > currentState.getVersion())
+                      && (migration.getNumber() <= versionToTarget));
```
liam-verta: These need comments and/or methods with meaningful names.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: I wouldn't want to return a bare `Set` or declare this variable as one, since the ordering is critical to the in-order application of the migrations. I could see returning SortedSet, but that's also a pretty arcane interface in the JDK, so I don't know that it would be any clearer in this case. I don't feel strongly about that, though.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -22,19 +31,150 @@ public Migrator(Connection connection, String resourcesDirectory) {
...
+      int rowsUpdated = ps.executeUpdate();
+      if (rowsUpdated != 1) {
+        throw new IllegalStateException(
+            "Failed to update schema_migrations table. rowsUpdated: " + rowsUpdated);
+      }
+    }
+  }
+
+  private NavigableSet<Migration> gatherMigrations(
+      Predicate<Migration> upOrDown, Predicate<Migration> versionPredicate)
```
liam-verta: Suggest different names.
```suggestion
      Predicate<Migration> migrationDirectionFilter, Predicate<Migration> migrationVersionRangeFilter)
```

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: apropos of nothing, I recently saw a JEP that proposes introducing [IIRC] `Traversable` as an interface that would encompass anything that allowed backward and forward in-order navigation of data in a collection.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: name suggestions welcome!

jkwatson-verta: how about `migrationGreaterThanTargetedVersion` and `migrationLessThanTargetedVersion` ?

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: filter rather than predicate?

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -22,19 +31,150 @@ public Migrator(Connection connection, String resourcesDirectory) {
...
+      throw new MigrationException(
+          "Database is in a dirty state. Bailing out until dirty handling is implemented.");
+    }
+    int versionToTarget =
+        desiredVersion != null
+            ? desiredVersion
+            : gatherMigrations(migration -> true, migration -> true).stream()
+                .map(Migration::getNumber)
+                .max(Integer::compareTo)
+                .orElse(0);
```
liam-verta: Move to a method called `findVersionToTarget()`?

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -22,19 +31,150 @@ public Migrator(Connection connection, String resourcesDirectory) {
...
+                  (migration.getNumber() > currentState.getVersion())
+                      && (migration.getNumber() <= versionToTarget));
+    } else {
+      migrationsToPerform =
+          gatherMigrations(
+                  Migration::isDown,
+                  migration ->
+                      migration.getNumber() > versionToTarget
+                          && migration.getNumber() <= currentState.getVersion())
+              .descendingSet();
```
liam-verta: This is the only place I see using a method from `NavigableSet`.

jkwatson-verta: yes, but the orderedness is critical to the processing, so keeping that semantic information in the return type seems very important to me.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
liam-verta: `SortedSet` is more immediately understandable to me that `NavigableSet`.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: changed return type to SortedSet

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: done

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
liam-verta: My thinking was that the names should indicate intended use/meaning.  It's less about `Filter` vs `Predicate` and more that the rest of the names didn't feel like they helped me much. 

Though now that I think about it probably makes just as much sense to have a `Predicate<Migration>... migrationFilters`.  That would simplify the call in `findVersionToTarget()`. :shrug:

liam-verta (approved):

PR #3351, opened 2022-11-09
chore: [VR-14177] Provide a utility in the Migrator to facilitate moving from liquibase.

jkwatson-verta:
## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


coryverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -218,6 +221,44 @@ void executeSingleMigration(Migration migration) throws IOException, SQLExceptio
     }
   }
 
+  /**
+   * Pre-initializes the schema migrations table, in cases where the database is being converted
+   * from a legacy migration tool (i.e. liquibase).
+   *
+   * <p>_If and only if_ there are legacy migrations in place, there is no existing
+   * schema_migrations table, AND an assumed current version is provided will the schema_versions
+   * table be initialized to the provided value.
```
liam-verta:
```suggestion
   * <p>_If and only if_ ALL the following conditions are met, then the <code>schema_migrations</code>
   * table will be initialized with the value of <code>assumedCurrentVersion</code>: 
   * <ul>
   * <li> there are legacy migrations present
   * <li> there is no existing <code>schema_migrations</code> table
   * <li> an <code>assumedCurrentVersion</code> value is provided
   * </ul>
```

liam-verta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -218,6 +221,44 @@ void executeSingleMigration(Migration migration) throws IOException, SQLExceptio
...
+      MigrationState currentState = findCurrentState();
+      log.info(
+          "schema_versions table already exists. Skipping pre-initialization step. current state: "
+              + currentState);
+    } catch (SQLException e) {
+      log.info(
+          "No schema_versions table found. Initializing to version " + assumedCurrentVersion.get());
+      setupDatastore();
+      updateVersion(false, assumedCurrentVersion.get());
+    }
```
anandjakhaniya, line 264: Handling existence like this. could we use `Optional<MigrationState> currentState = findCurrentState();` and instead of retuning null from `findCurrentState();` method we can return Option?

anandjakhaniya (approved):
small comment otherwise looks good.

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: sure, that can be done. It won't impact this code path, though, which will throw an exception if the table doesn't exist.

jkwatson-verta: we need to be able to distinguish between "the table doesn't exist" and "the table exists, but is empty"

jkwatson-verta: I'm planning on doing some refactoring of this code once this is merged, so I'll think about how to improve this when I do that bit of work.

PR #3360, opened 2022-11-14
chore: add the capability to turn on NPE stack-traces for Futures

jkwatson-verta:
## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
@@ -29,11 +29,15 @@ public class Future<T> {
   private static FutureExecutor futureExecutor;
 
   private static final boolean DEEP_TRACING_ENABLED;
+  private static boolean captureStacksAtCreation;
+
   public static final AttributeKey<String> STACK_ATTRIBUTE_KEY = AttributeKey.stringKey("stack");
 
   static {
     String internalFutureTracingEnabled = System.getenv("IFUTURE_TRACING_ENABLED");
     DEEP_TRACING_ENABLED = Boolean.parseBoolean(internalFutureTracingEnabled);
+    captureStacksAtCreation =
+        Boolean.parseBoolean(System.getProperty(FUTURE_TESTING_STACKS_ENABLED));
   }
 
   private final Tracer futureTracer = GlobalOpenTelemetry.getTracer("futureTracer");
@@ -41,8 +45,7 @@ public class Future<T> {
   private final CompletionStage<T> stage;
 
   private Future(CompletionStage<T> stage) {
-    if (DEEP_TRACING_ENABLED
-        || Boolean.parseBoolean(System.getProperty(FUTURE_TESTING_STACKS_ENABLED))) {
+    if (DEEP_TRACING_ENABLED || captureStacksAtCreation) {
       StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
       // get rid of the top of the stack, which is not useful
       stackTrace = Arrays.copyOfRange(stackTrace, 1, stackTrace.length);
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
```diff
@@ -37,17 +41,20 @@
   private final CompletionStage<T> stage;
 
   private Future(CompletionStage<T> stage) {
-    if (!DEEP_TRACING_ENABLED) {
-      formattedStack = null;
-    } else {
+    if (DEEP_TRACING_ENABLED
+        || Boolean.parseBoolean(System.getProperty(FUTURE_TESTING_STACKS_ENABLED))) {
```
liam-verta: Let's not do `Boolean.parseBoolean(System.getProperty(FUTURE_TESTING_STACKS_ENABLED))` for every call to `Future()`. Make this a `static` field?

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
```diff
@@ -37,17 +41,20 @@
   private final CompletionStage<T> stage;
 
   private Future(CompletionStage<T> stage) {
-    if (!DEEP_TRACING_ENABLED) {
-      formattedStack = null;
-    } else {
+    if (DEEP_TRACING_ENABLED
```
liam-verta: Why not use `DEEP_TRACING_ENABLED` instead of adding another property?  I assume it is because `DEEP_TRACING_ENABLED` has broader effects?

coryverta (approved):

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
```diff
@@ -380,4 +400,12 @@ public InternalFuture<T> toInternalFuture() {
...
     futureExecutor = null;
   }
+
+  public static void enableCallSiteStackCapture() {
+    System.setProperty(FUTURE_TESTING_STACKS_ENABLED, "true");
+  }
+
+  public static void disableCallSiteStackCapture() {
+    System.setProperty(FUTURE_TESTING_STACKS_ENABLED, "false");
+  }
```
liam-verta: Yeah, definitely a static field.  I think it should default to the value of `DEEP_TRACING_ENABLED`, but can be overridden.  Maybe have one method `public static void forceCallSiteStackCapture(boolean value)`?  If `DEEP_TRACING_ENABLED` is `true`, this method has no effect.

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
jkwatson-verta: Yes, deep tracing also includes creating spans for every future execution.

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
jkwatson-verta: so, are you suggesting that we don't allow enabling this via system property at JVM startup, but only programmatically?

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
jkwatson-verta: also, setting env vars is quite tricky to do programmatically for unit tests, where this is going to be most useful.

jkwatson-verta: Are you worried about performance or something else?

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
jkwatson-verta: ok, best of both worlds is probably to initialize a static field based on the system property, but then allow overriding via these methods. I don't think we should do any special handling around the DEEP_TRACING_ENABLED value, since they only coincidentally happen to share the stack trace capture.

liam-verta (approved):

PR #3386, opened 2022-11-22
Fix connection commit issue when we use migrator tool

anandjakhaniya:
## Impact and Context
https://jenkins.dev.verta.ai/job/build/job/registry-backend-test/873/consoleFull
Problem: We have two test suites in Registry Backend. 
1. MonitoringTestSuite
2. RegistryTestSuite
So when we are trying to run `PullRequestSuite` it will trigger migration two times from each test suite. but when the second suite tries to execute the migration, it will get the last migration (ex: 4 number) as dirty due to the connection not being closed. In the actual scenario, it should not be dirty. so 

Solution: when I am closing that connection after successful migration and try to run it again then it works fine.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
@@ -387,15 +387,17 @@ public abstract class CommonDBUtil {
       throws SQLException, MigrationException {
     RdbConfig rdbConfiguration = config.getRdbConfiguration();
     createDBIfNotExists(rdbConfiguration);
-    Connection connection = acquireDatabaseConnection(config, rdbConfiguration);
-    Migrator migrator =
-        new Migrator(
-            connection,
-            findResourcesDirectory(migrationResourcesRootDirectory, rdbConfiguration),
-            rdbConfiguration);
-    boolean liquibaseTableExists = tableExists(connection, config, "database_change_log");
-    migrator.preInitializeIfRequired(liquibaseTableExists, currentVersion);
-    migrator.performMigration();
+    try (Connection connection = acquireDatabaseConnection(config, rdbConfiguration)) {
+      Migrator migrator =
+          new Migrator(
+              connection,
+              findResourcesDirectory(migrationResourcesRootDirectory, rdbConfiguration),
+              rdbConfiguration);
+      boolean liquibaseTableExists = tableExists(connection, config, "database_change_log");
+      migrator.preInitializeIfRequired(liquibaseTableExists, currentVersion);
+      migrator.performMigration();
+      connection.commit();
+    }
   }
 
   private String findResourcesDirectory(String rootDirectory, RdbConfig rdbConfiguration) {


evgeniilezhnin (approved):

jkwatson-verta:
migrations are intended to include any commits needed in their scripts, so this should not be necessary.

jkwatson-verta (changes requested):
this should not be necessary if migrations are properly constructed to include any transactions that are necessary.

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -387,15 +387,17 @@ public void runMigrations(
...
+    try (Connection connection = acquireDatabaseConnection(config, rdbConfiguration)) {
+      Migrator migrator =
+          new Migrator(
+              connection,
+              findResourcesDirectory(migrationResourcesRootDirectory, rdbConfiguration),
+              rdbConfiguration);
+      boolean liquibaseTableExists = tableExists(connection, config, "database_change_log");
+      migrator.preInitializeIfRequired(liquibaseTableExists, currentVersion);
+      migrator.performMigration();
+      connection.commit();
```
jkwatson-verta, line 399:
```suggestion
```

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -387,15 +387,17 @@ public void runMigrations(
...
-    Connection connection = acquireDatabaseConnection(config, rdbConfiguration);
-    Migrator migrator =
-        new Migrator(
-            connection,
-            findResourcesDirectory(migrationResourcesRootDirectory, rdbConfiguration),
-            rdbConfiguration);
-    boolean liquibaseTableExists = tableExists(connection, config, "database_change_log");
-    migrator.preInitializeIfRequired(liquibaseTableExists, currentVersion);
-    migrator.performMigration();
+    try (Connection connection = acquireDatabaseConnection(config, rdbConfiguration)) {
```
jkwatson-verta, line 390: closing the connection seems good. I don't think we should need to be doing a commit, though.

evgeniilezhnin:
> this should not be necessary if migrations are properly constructed to include any transactions that are necessary.

I think Anand meant that the issue happening not during script execution, but during updateVersion java call which calls sql command for H2 database.

So here the dirty flag becomes true: https://github.com/VertaAI/modeldb/blob/407f5c3d7ad0c3dfc3f0231ff5d43df0ba8db1f8/backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java#L146
We execute script, commit. Then we try to update dirty to false here: https://github.com/VertaAI/modeldb/blob/407f5c3d7ad0c3dfc3f0231ff5d43df0ba8db1f8/backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java#L148
But commit doesn't happen automatically for some reason

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
jkwatson-verta: hmm. shouldn't we be doing the commit when we update the schema_migration table? Not just after the entire thing?

jkwatson-verta (approved):
after more pondering and thinking, I think this should be ok. If any single migration fails, though, we'll end up going all the way back to the beginning of the migration block, which might end up being confusing for the dev who is debugging a migration, but shouldn't be an issue in the real world when all the migrations are rock solid, post development.

jkwatson-verta:
> > this should not be necessary if migrations are properly constructed to include any transactions that are necessary.
> 
> I think Anand meant that the issue happening not during script execution, but during updateVersion java call which calls sql command for H2 database.
> 
> So here the dirty flag becomes true:
> 
> https://github.com/VertaAI/modeldb/blob/407f5c3d7ad0c3dfc3f0231ff5d43df0ba8db1f8/backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java#L146
> 
> 
> We execute script, commit. Then we try to update dirty to false here:
> https://github.com/VertaAI/modeldb/blob/407f5c3d7ad0c3dfc3f0231ff5d43df0ba8db1f8/backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java#L148
> 
> 
> But commit doesn't happen automatically for some reason

yeah, this is why I was wondering if we shouldn't just put the commit there, rather than around the whole thing.

jkwatson-verta:
It is interesting/confusing that we don't see this issue on mysql or sqlserver, and just on H2. 🤔

PR #3387, opened 2022-11-22
Fix dirty migration issue

anandjakhaniya:
## Impact and Context
- Issue: When we try to execute downgrade migration and at some point, some migration is failing then we are marking that migration as dirty. but we are not running a downgrade for that dirty migration and whenever we trying to execute it again it is giving error for already exists.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


backend/server/src/main/resources/migration/h2/2_release_2022_09.down.sql
```diff
@@ -0,0 +1,2 @@
+SET DB_CLOSE_DELAY -1;
+DROP TABLE "TEST_2";
```
evgeniilezhnin, line 2: why is this needed? I think test fails because of that

jkwatson-verta:
This is not the intended behavior. Since each migration should be self-contained and not commit if it fails, this isn't necessary.

PR #1915, opened 2021-01-28
VR-9379: Get defaultWorkspace

convoliution:
On top of https://github.com/VertaAI/modeldb/pull/1914

In progress:

- [x] make compatible with OSS
- [x] test against visibility-enabled backend

PR #3404, opened 2022-11-28
chore: [VUMM-758] Do not use personal workspace for permission v2

evgeniilezhnin:
## Impact and Context

The concept of a "personal workspace" is being removed in permission v2. This PR removes `Connection.get_personal_workspace()`.

## Risks and Area of Effect

This should have no impact for users. New clients can continue to be used with pre-permission v2 platforms.

`Connection.get_personal_workspace()` was only being used as a fallback for `Connection.get_default_workspace()` for backends older than Jan 2021 (https://github.com/VertaAI/modeldb/pull/1915), and all users are assigned a `default_workspace_id` as soon as they are created ([Slack thread](https://vertaai.slack.com/archives/C0306V04EJC/p1674757850181529)).

There is an issue where internal accounts bootstrapped with UAC aren't assigned a `default_workspace_id`, but that's a separate matter for the backend, and can be worked around clientside.

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

Test suite passes against a v1 environment ([Slack thread](https://vertaai.slack.com/archives/C03PWNZU5CM/p1674775415283039?thread_ts=1674666909.564999&cid=C03PWNZU5CM)).

## Reverting

Revert this PR.

### DIFF
--- a/client/verta/verta/_internal_utils/_utils.py
+++ b/client/verta/verta/_internal_utils/_utils.py
@@ -329,24 +329,6 @@ class Connection(object):
         workspace = self.must_proto_response(response, Workspace_pb2.Workspace)
         return workspace.username or workspace.org_name
 
-    def get_personal_workspace(self):
-        email = self.auth.get("Grpc-Metadata-email")
-        if email is not None:
-            msg = UACService_pb2.GetUser(email=email)
-            response = self.make_proto_request(
-                "GET", "/api/v1/uac-proxy/uac/getUser", params=msg
-            )
-
-            if (
-                response.ok and self.is_html_response(response)
-            ) or response.status_code == requests.codes.not_found:  # fetched webapp  # UAC not found
-                pass  # fall through to OSS default workspace
-            else:
-                return self.must_proto_response(
-                    response, UACService_pb2.UserInfo
-                ).verta_info.username
-        return self._OSS_DEFAULT_WORKSPACE
-
     def get_default_workspace(self):
         response = self.make_proto_request(
             "GET", "/api/v1/uac-proxy/uac/getCurrentUser"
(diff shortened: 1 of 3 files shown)


liuverta:
~fyi all: Please don't merge this until Wednesday-ish! I need to publish a new version first before we merge a bunch of these behavior-breaking changes 😭~

☝️ This is done.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -361,7 +362,10 @@ def get_default_workspace(self):
         if workspace_id:
             return self.get_workspace_name_from_id(workspace_id)
         else:  # old backend
-            return self.get_personal_workspace()
+            if self._PERMISSION_V2:
```
liuverta: Honestly, I think we could just remove this entire `if-else`. And instead, we can simply `return self.get_workspace_name_from_id(workspace_id)`.

The `else:  # old backend` is for before January 2020, so I'd say it's reasonable to remove. And then we can also remove `def get_personal_workspace()`

client/verta/verta/client.py
```diff
@@ -376,7 +376,7 @@ def get_project(self, name=None, workspace=None, id=None):
             Name of the Project.
         workspace : str, optional
             Workspace under which the Project with name `name` exists. If not provided, the current
-            user's personal workspace will be used.
+            user's default workspace will be used.
```
liuverta, line 379: Good catch! 👍

liuverta (changes requested):

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: removed

liuverta (approved):

liuverta (approved):
Looks good to merge!

evgeniilezhnin:
will wait some time until permission v2 flag enabled on test environments before merging this PR

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -361,7 +344,7 @@ def get_default_workspace(self):
         if workspace_id:
             return self.get_workspace_name_from_id(workspace_id)
         else:  # old backend
```
liuverta:
```suggestion
        else:
```

Would you mind also removing this comment, for clarity?

liuverta (approved):

PR #3407, opened 2022-11-28
chore: [VUMM-760] Remove organization for permission v2

evgeniilezhnin:
## Impact and Context

The concept of "organizations"—as they existed pre-permission v2—is being replaced with "workspaces". This PR refactors tests to use v2 workspaces instead of organizations.

## Risks and Area of Effect

This should have no impact for users. New clients can continue to be used with pre-permission v2 platforms.

`Connection.get_workspace_name_from_legacy_id()` is not needed because every entity has been using the newer workspace service for at least two years (https://github.com/VertaAI/modeldb/pull/1914/files#r566375578).

Organization CRUD in the client was restricted to tests, so it won't impact publicly-supported user code.

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain)

Test suite passes against a v1 environment ([Slack thread](https://vertaai.slack.com/archives/C03PWNZU5CM/p1674775415283039?thread_ts=1674666909.564999&cid=C03PWNZU5CM)).

Evgenii also stated

> verified it by running test_permission, test_endpoint and registry test packages

## Reverting


liuverta (commented):
Let me know if you encounter issues with specific tests. Some of them may be worth removing because they are very old.

client/verta/tests/conftest.py
```diff
@@ -75,50 +75,6 @@ def mark_time():
...
-            if not path.is_file():
-                _create_dummy_workspaces()
-                path.touch(exist_ok=False)
-
-    yield
-
-    for org in dummy_orgs:
-        org.delete()
-
-
```
liuverta, line 121: 👍 I am very glad to get rid of this 🔥

client/verta/verta/client.py
```diff
@@ -1736,13 +1735,3 @@ def find_datasets(
             datasets = datasets.find(predicates)
 
         return datasets
-
-    def _create_organization(
-        self, name, desc=None, collaborator_type=None, global_can_deploy=None
-    ):
-        return Organization._create(
-            self._conn, name, desc, collaborator_type, global_can_deploy
-        )
```
liuverta, line 1745: Unfortunately, I realize that a lot of our tests right now depend on creating custom workspace during tests 🤔 

I will need to talk to Elizabeth about whether we can remove these specific Python tests in the future, but for now I think maybe we need a `_create_workspace()`, `Workspace.add_member()` and `Workspace.delete()` just for the tests

liuverta (commented):
Just to let you know: When these tests run in Jenkins, `client` is a sys admin account, so maybe we can assume it has permissions to create and delete workspaces.

client/verta/verta/client.py
evgeniilezhnin: Is it necessary to create a workspace during test? Why is it a problem to use existing workspace?

It is possible to create a workspace during test, but the problem is that we don't have workspace.add_member. Members are added to workspaces using groups. So the problem is that if we create workspaces in test we also need to call endpoints to access or create groups, roles, users and organizations

evgeniilezhnin: Also please note that in permission v2 we can't set individual permission by resource id. The system is simplified comparing to v1 so now admin can set permission to use resource types by groups of users in specified workspaces.
So as I see some of permission tests in Python Client can't be migrated to v2.

liuverta:
The client tests should be able to run against any new, fresh dev
environment. If the workspace should not be created when the tests run,
then I am not sure when it will be created at all.

I would prefer implementing the minimum necessary client functions to
create a workspace and group for the tests to use. If that isn't possible,
I will ask Elizabeth if these tests can just be removed.

evgeniilezhnin:
> The client tests should be able to run against any new, fresh dev environment. If the workspace should not be created when the tests run, then I am not sure when it will be created at all. I would prefer implementing the minimum necessary client functions to create a workspace and group for the tests to use. If that isn't possible, I will ask Elizabeth if these tests can just be removed.

actually there is always a default workspace Default that is created during a migration. We can use it, but anyway it is not a problem to create a new besides that. So it is possible.
There is a second separate problem that there is no more visibility for individual resources. That one is not possible I think.

liuverta:
Hmmmm I see. Maybe I am misremembering the tests, and the default workspace
is okay to use for most of them 🤔 In that case I will leave it to your
judgment, sorry for any confusion!

Yes, I remember there were some per-entity `visibility` and maybe even
`public_within_org` tests. I agree those may simply need to be removed for
v2.

evgeniilezhnin:
I'm planning to add workspace creation because I think it will be useful at least in a couple of tests where we can use private workspace instead of private resource type, also it can be useful for cleanup

evgeniilezhnin:
I'm currently trying to test workspaces creation: https://github.com/VertaAI/modeldb/pull/3439
But at the moment there is a problem with permission v2 proxy: https://vertaai.slack.com/archives/C03PWNZU5CM/p1670581917882709?thread_ts=1669730220.761939&cid=C03PWNZU5CM

In the current PR tests use default workspace where possible. I verified it by running test_permission and registry test packages for now

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -265,9 +265,19 @@ def _get_visible_orgs(self):
         )
         response = self.must_proto_response(response, Workspace_pb2.Workspaces)
 
-        org_names = map(lambda workspace: workspace.org_name, response.workspace)
-        org_names = filter(None, org_names)
-        return list(org_names)
+        workspace_names = map(lambda workspace: "{}:{}".format(workspace.org_id, workspace.org_name), response.workspace)
+        workspace_names = filter(None, workspace_names)
+        return list(workspace_names)
```
liuverta: Could you please help rename this function to `_get_visible_workspaces()`? 🙇‍♂️

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -265,9 +265,19 @@ def _get_visible_orgs(self):
...
+
+    def _get_visible_org(self):
```
liuverta:
```suggestion
    def _get_organization_id(self):
```

client/verta/tests/test_permissions/test_workspace.py
```diff
@@ -61,47 +61,7 @@ def test_model_ver(self, client, organization, created_entities):
...
 
         # organization workspace
-        reg_model = client.create_registered_model(workspace=organization.name)
+        reg_model = client.create_registered_model(workspace=workspace.name)
         created_entities.append(reg_model)
-        assert reg_model.create_version().workspace == organization.name
+        assert reg_model.create_version().workspace == workspace.name
 
-
-class TestClientGetWorkspace:
```
liuverta, line 69: fyi, this test is very important. We should add it again when workspace creation is implemented.

client/verta/tests/registry/model_version/test_crud.py
```diff
@@ -193,10 +193,9 @@ def test_creation_from_run(self, registered_model, experiment_run, lock_level):
         assert model_ver._msg.lock_level == lock_level._as_proto()
```
liuverta, line 193: If possible, I do not think we should be removing any of the `pytest.raises(requests.HTTPError, match="^403")` checks in this file. They seem important to make sure non-admins cannot change lock level. Is it possible to keep them, or bring them back after we implement workspace creation?

(by the way: when tests run in Jenkins, `client` is admin and `client_2` is non-admin)

client/verta/tests/test_permissions/test_sharing_old.py
```diff
@@ -10,163 +10,67 @@
```
liuverta: Does the [`ORG_SCOPED_PUBLIC` concept](https://github.com/VertaAI/protos-all/blob/4f8323557c43ad324456bb8ee044a6169a87933b/protos/protos/public/common/CommonService.proto#L162-L173) still exist in our platform? 🤔 if not, we can probably just remove this whole file.

client/verta/tests/test_permissions/test_visibility_e2e.py
```diff
@@ -1,247 +0,0 @@
...
-from verta.visibility import (
-    OrgCustom,
-    Private,
-)
-from verta.dataset import Path
-from verta.environment import Python
-
-
-pytestmark = pytest.mark.not_oss
-
```
liuverta, line 19: @evgeniilezhnin Instead of deleting this file completely, could we add

```python
pytest.skip("visibility deprecated; tests disabled for permissions v2", allow_module_level=True)
```

? I'd like to check later to see if we can adjust these tests for v2.

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -747,7 +747,7 @@ def predict(self, _):
     def test_update_from_run_diff_workspace(
         self,
         client,
-        organization,
+            workspace,
```
liuverta:
```suggestion
        workspace,
```

client/verta/tests/test_dataset_versioning/test_dataset_version.py
```diff
@@ -18,6 +19,10 @@ def test_description(self, client, dataset, strs):
         assert version.get_description() == second_desc
 
         assert dataset.get_version(id=version.id).get_description() == second_desc
+        client.set_workspace("42379216512358509179263")
+        dataset = client.create_dataset("dataset_name3")
+        version = dataset.create_version(Path("conftest.py"), desc=first_desc)
```
liuverta: Should this be removed? I'm not sure if it is a part of the test, or if it was for debugging.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -297,27 +307,6 @@ def is_workspace(self, workspace_name):
 
         return response.ok
 
-    def get_workspace_name_from_legacy_id(self, workspace_id):
```
liuverta, line 302: Just to confirm @evgeniilezhnin: Are you sure we no longer need this? For context, this was because projects and datasets used an older ID system from before WorkspaceService existed.

Are we migrating project and dataset workspace IDs to the new permissions v2 workspace IDs?

liuverta (changes requested):
I have some feedback and questions.

Could you also add more details about how you tested this in the PR description under **Testing**?

Also, do you know if we should wait until permissions v2 is deployed to dev.verta.ai before merging this?

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: when changing from permissions v1 to permissions v2 there is no workspace IDs format change. We changed ids format a long time ago. After that we refactored services to store resource-workspace information in UAC only in the new format. So now it should not be stored in old format anywhere and migration from old workspace ids format to new format should be already executed everywhere

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: renamed. This change was in personal workspace removal PR, but can add it here too

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: renamed

client/verta/tests/test_permissions/test_workspace.py
evgeniilezhnin: reverted removal. I think it can be added here

client/verta/tests/registry/model_version/test_crud.py
evgeniilezhnin: I think workspace creation is needed for this test, I couldn't make it work with a default workspace. So I can add these checks again after workspace creation implemented in client

client/verta/tests/test_permissions/test_sharing_old.py
evgeniilezhnin: no, in v2 we can't set individual resources permissions. Removed the test

client/verta/tests/test_dataset_versioning/test_dataset_version.py
evgeniilezhnin: yes, removed it now

client/verta/tests/test_permissions/test_visibility_e2e.py
evgeniilezhnin: added

evgeniilezhnin:
> I have some feedback and questions.
> 
> Could you also add more details about how you tested this in the PR description under **Testing**?
> 
> Also, do you know if we should wait until permissions v2 is deployed to dev.verta.ai before merging this?

Added

yes, I think permission v2 flag should be enabled before merging client change on dev.verta.ai, testsqlserver, testregular

evgeniilezhnin:
tried running the test from this branch for testregular (after deploy), but there are authorization problems which were also there in previous runs: https://jenkins.dev.verta.ai/job/client/job/python/job/pytest-3.7/388/. Now I'm trying to run test for testsqlserver

evgeniilezhnin:
when testing on testsqlserver there is a problem that default workspace is not set. I think it can be fixed either by implementing migration on UAC or by using custom workspaces in Client which is the thing I'm working on now

evgeniilezhnin:
> when testing on testsqlserver there is a problem that default workspace is not set. I think it can be fixed either by implementing migration on UAC or by using custom workspaces in Client which is the thing I'm working on now

migration helped, but there are still test issues for testsqlserver, that I will fix

client/verta/tests/registry/model_version/test_crud.py
evgeniilezhnin: reverted here: https://github.com/VertaAI/modeldb/pull/3439/files

evgeniilezhnin:
> when testing on testsqlserver there is a problem that default workspace is not set. I think it can be fixed either by implementing migration on UAC or by using custom workspaces in Client which is the thing I'm working on now

I think it is better to merge the current change when https://github.com/VertaAI/modeldb/pull/3439/files is approved

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -260,15 +260,25 @@ def email(self):
...
+
+    def _get_organization_id(self):
+        response = self.make_proto_request(
+            "GET", "/api/v1/uac-proxy/workspace/getVisibleWorkspaces"
+        )
+        response = self.must_proto_response(response, Workspace_pb2.Workspaces)
+
+        workspace_names = map(lambda workspace: workspace.org_id, response.workspace)
+        workspace_names = filter(None, workspace_names)
+        return list(workspace_names)[0]
```
liuverta, line 281: Is it possible for `getVisibileWorkspaces` to return workspaces in other organizations? `list(workspace_names)[0]` seems like it will basically return the `org_id` of a random workspace, so I'm worried that if we have a multi-organization setup, we might get the wrong ID 🤔 but I'm also not sure how permissions v2 works 😅

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -260,15 +260,25 @@ def email(self):
...
+    def _get_visible_workspaces(self):
         response = self.make_proto_request(
             "GET", "/api/v1/uac-proxy/workspace/getVisibleWorkspaces"
         )
         response = self.must_proto_response(response, Workspace_pb2.Workspaces)
 
-        org_names = map(lambda workspace: workspace.org_name, response.workspace)
-        org_names = filter(None, org_names)
-        return list(org_names)
+        workspace_names = map(lambda workspace: "{}:{}".format(workspace.org_id, workspace.org_name), response.workspace)
```
liuverta: fyi @evgeniilezhnin, I had been talking with Elizabeth about changing UAC so that the client (and client users) would not need to worry about `org_id`, and instead only use workspace name.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -327,7 +316,7 @@ def get_workspace_name_from_id(self, workspace_id):
         )
 
         workspace = self.must_proto_response(response, Workspace_pb2.Workspace)
-        return workspace.username or workspace.org_name
+        return "{}:{}".format(workspace.org_id, workspace.org_name)
```
liuverta: See my other comment about maybe not needing `org_id`: https://github.com/VertaAI/modeldb/pull/3407/files#r1067587726

client/verta/verta/tracking/entities/_experimentrun.py
```diff
@@ -132,8 +132,8 @@ def workspace(self):
             # workspace is OSS default
             return self._conn._OSS_DEFAULT_WORKSPACE
         else:
-            return self._conn.get_workspace_name_from_legacy_id(
-                project_json["workspace_id"]
+            return self._conn.get_workspace_name_from_id(
+                project_json["workspace_service_id"]
```
liuverta: Sorry this is an extra request, but could we do this so it is more similar to the other `def workspace()`s?

```python
    @property
    def workspace(self):
        self._refresh_cache()

        msg = ProjectService_pb2.GetProjectById(id=self._msg.project_id)
        url = "/api/v1/modeldb/project/getProjectById"
        response = self._conn.make_proto_request("GET", url, params=msg)
        response = self._conn.must_proto_response(response, msg.Response)
        proj_proto = response.project

        if proj_proto.workspace_service_id:
            return self._conn.get_workspace_name_from_id(proj_proto.workspace_service_id)
        else:
            return self._conn._OSS_DEFAULT_WORKSPACE
```

(will require a `from verta._protos.public.modeldb import ProjectService_pb2` at the top of the file)

client/verta/verta/client.py
liuverta: > So as I see some of permission tests in Python Client can't be migrated to v2.

Yes, we will need to end up deleting tests that are no longer applicable to v2 😞

client/verta/verta/dataset/entities/_dataset.py
```diff
@@ -89,7 +89,7 @@ def workspace(self):
         self._refresh_cache()
 
         if self._msg.workspace_id:
```
liuverta:
```suggestion
        if self._msg.workspace_service_id:
```

client/verta/verta/tracking/entities/_project.py
```diff
@@ -82,7 +82,7 @@ def workspace(self):
         self._refresh_cache()
 
         if self._msg.workspace_id:
```
liuverta:
```suggestion
        if self._msg.workspace_service_id:
```

client/verta/verta/_internal_utils/_utils.py
liuverta: Ah, I see entities have `workspace_service_id`. Okay, this sounds good to me 👍

liuverta (changes requested):

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: yes, a PR for that is merged (https://vertaai.atlassian.net/browse/VUMM-808) so reverted this change
update: sorry, wrong ticket. The right one is https://vertaai.atlassian.net/browse/VUMM-809 . Not merged, but expected to be merged soon

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: reverted

client/verta/verta/tracking/entities/_experimentrun.py
evgeniilezhnin: applied your code snippet

client/verta/tests/test_permissions/test_workspace.py
```diff
@@ -61,25 +61,24 @@ def test_model_ver(self, client, organization, created_entities):
...
 
         WEBAPP_WORKSPACE = (
             client._conn.get_default_workspace()
         )  # TODO: first change default workspace
-        CONFIG_WORKSPACE = create_organization().name
-        CLIENT_WORKSPACE = create_organization().name
-        PARAM_WORKSPACE = create_organization().name
+        CONFIG_WORKSPACE = workspace.name
+        CLIENT_WORKSPACE = workspace.name
+        PARAM_WORKSPACE = workspace.name
```
liuverta: If possible, these three workspaces need to be different, not the same one.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -270,6 +270,16 @@ def _get_visible_orgs(self):
         org_names = filter(None, org_names)
```
liuverta, line 270: Just to double check: In permissions V2, `workspace.org_name` is still the accurate workspace name?

client/verta/tests/registry/model_version/test_from_run.py
```diff
@@ -37,29 +37,4 @@ def test_from_run(self, experiment_run, model_for_deployment, registered_model):
...
-
-        with pytest.raises(requests.HTTPError) as excinfo:
-            registered_model.create_version_from_run(
-                run_id=experiment_run.id, name="From Run {}".format(experiment_run.id)
-            )
-
-        exc_msg = str(excinfo.value).strip()
-        assert exc_msg.startswith("404")
-        assert "not found" in exc_msg
+# TODO: test_from_run_diff_workspaces with separate workspace
```
liuverta: Do you know if it is possible to keep these tests for V2?

liuverta (changes requested):
Just a few more questions!

client/verta/tests/test_permissions/test_workspace.py
evgeniilezhnin: yes, it is possible, but we need workspace creation for that, so I added it now to workspace creation PR

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: yes

client/verta/tests/registry/model_version/test_from_run.py
evgeniilezhnin: yes, it is possible, but we need workspace creation for that, so I added it now to workspace creation PR

client/verta/tests/test_permissions/test_workspace.py
liuverta: 👍 Great, I see that it was added in https://github.com/VertaAI/modeldb/pull/3439/commits/33079e293fc9ae1d70ee63bbd8389012e184c6f9

client/verta/tests/registry/model_version/test_from_run.py
liuverta: 👍 Great, I see that it was added in https://github.com/VertaAI/modeldb/pull/3439/commits/29377da62da94712b41db213ebbd1612c12ae478

liuverta (approved):

PR #3408, opened 2022-11-29
[VUMM-766] Replace OR Deprecated collaborator apis

anandjakhaniya:
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


evgeniilezhnin:
is it possible to use config flag permission_V2 to change behavior instead of deleting methods from main (tests should be ok to delete without a flag)? If not we should until everything else is finished before merging this change

anandjakhaniya:
This was just starting optional PR so closing this

PR #3409, opened 2022-11-29
fix: Address a race condition in the migrator if two processes try to…

jkwatson-verta:
… run migrations at the same time.

Addresses the race condition in 2 ways:
1. Lock the database earlier in the process, before figuring out what migrations to apply
2. When updating the `schema_migrations` table, check the current version as a part of the UPDATE SQL to make sure something hasn't updated it underneath us.

## Testing
- [x] Unit test
- [x] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
@@ -34,7 +34,7 @@ public class Migrator {
 
   public void performMigration(Integer desiredVersion) throws SQLException, MigrationException {
     MigrationDatastore migrationDatastore = setupDatastore();
-    migrationDatastore.lock();
+    MigrationTools.lockDatabase(migrationDatastore);
     try {
       MigrationState currentState = findCurrentState();
       if (currentState == null) {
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
```diff
@@ -136,40 +144,45 @@ private MigrationDatastore setupDatastore() throws SQLException {
...
+      throws SQLException {
     int newVersion =
         pendingMigration.isUp() ? pendingMigration.getNumber() : pendingMigration.getNumber() - 1;
-    updateVersion(dirty, newVersion);
+
+    updateVersion(dirty, newVersion, expectedCurrentVersion);
   }
 
-  private void updateVersion(boolean dirty, int newVersion) throws SQLException {
+  private void updateVersion(boolean dirty, int newVersion, int expectedCurrentVersion)
```
conradoverta: Maybe also check the dirty state as extra safety?

conradoverta (approved):
LGTM

backend/common/src/main/java/ai/verta/modeldb/common/db/migration/Migrator.java
jkwatson-verta: good idea. will add!

jkwatson-verta: done

PR #3428, opened 2022-12-06
do not use personal workspaces

evgeniilezhnin:
## Impact and Context
personal workspaces are no more in permission v2
The decision on which workspace to use if workspace not specified should be handled by UAC

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [x] Other (explain): Tested on pmigration1.dev.verta.ai by updating dataset

## Reverting

### DIFF
--- a/backend/server/src/main/java/ai/verta/modeldb/authservice/MDBRoleServiceUtils.java
+++ b/backend/server/src/main/java/ai/verta/modeldb/authservice/MDBRoleServiceUtils.java
@@ -22,6 +22,7 @@ import ai.verta.modeldb.dto.WorkspaceDTO;
 import ai.verta.modeldb.utils.ModelDBUtils;
 import ai.verta.uac.*;
 import ai.verta.uac.ModelDBActionEnum.ModelDBServiceActions;
+import com.google.common.base.Strings;
 import com.google.rpc.Code;
 import io.grpc.Metadata;
 import io.grpc.StatusRuntimeException;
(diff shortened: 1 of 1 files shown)


backend/server/src/main/java/ai/verta/modeldb/authservice/MDBRoleServiceUtils.java
```diff
@@ -318,14 +319,10 @@ public String buildReadOnlyRoleBindingName(
...
       UserInfo currentLoginUserInfo, String workspaceName) {
-    /*from the name for workspace, get the workspace id and type.
-    if no workspace is present assume user's personal workspace*/
-    if (workspaceName == null || workspaceName.isEmpty()) {
-      return authService.workspaceIdByName(
-          true, authService.getUsernameFromUserInfo(currentLoginUserInfo));
-    } else {
-      return authService.workspaceIdByName(true, workspaceName);
+    if (Strings.isNullOrEmpty(workspaceName)) {
+      return Workspace.newBuilder().build();
```
anandjakhaniya, line 323: Have you checked what happens if we are not returning the workspace? I think we have to call UAC to get a workspace for the login user. might be the default workspace?

evgeniilezhnin: The getWorkspaceByWorkspaceName returns a workspace which is used in setResource method. But if we use empty workspace when calling setResource then setResource by itself finds default workspace for the current user. So I think there is no need to search for default workspace during getWorkspaceByWorkspaceName method

anandjakhaniya (approved):

backend/server/src/main/java/ai/verta/modeldb/authservice/MDBRoleServiceUtils.java
eheinlein-verta, line 323: I'm not sure I follow this.  From the description in the PR:

> The decision on which workspace to use if workspace not specified should be handled by UAC

but it looks like here you're returning an empty workspace, not letting UAC handle it. Can you clarify?

evgeniilezhnin: This method is to set request parameters during setResource or getResource call. We should not set in MDB some workspace in request and we can set default workspace by calling Workspace.newBuilder().build().

eheinlein-verta (approved):

PR #3439, opened 2022-12-09
workspace creation

evgeniilezhnin:
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -342,6 +354,21 @@ def get_default_workspace(self):
         else:  # old backend
             return self.get_personal_workspace()
 
+    def get_custom_workspace(self, org_id, workspace_name):
```
liuverta: Could we move this function to `Client` and name it `_create_workspace()`?

evgeniilezhnin: moved, renamed and refactored the method

evgeniilezhnin:
Locally tests work fine except S3 tests because I don't have configs for them. There are still issues in test when I run it using jenkins, but they don't seem to be related to the current changes: https://jenkins.dev.verta.ai/job/client/job/python/job/pytest-3.7/396
https://jenkins.dev.verta.ai/job/client/job/python/job/pytest-3.7/397

client/verta/verta/tracking/_role.py
```diff
@@ -0,0 +1,50 @@
...
+    @classmethod
+    def _create(
+        cls, conn, name, org_id, resource_actions
+    ):
+        Message = _Role.SetRoleV2
+        msg = cls._create_msg(name, org_id, resource_actions)
+
+        response = conn.make_proto_request(
+            "POST",
+            "/api/v2/uac-proxy/organization/{}/roles".format(org_id),
```
ewagner-verta: This is nit-picky, especially in this case, but I would love to do away with `.format()` forever.  Now that we have left Python 2 behind, we can use f-strings which are much more readable (IMHO).
`f"/api/v2/uac-proxy/organization/{org.id}/roles",`

client/verta/verta/tracking/_role.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, conn, msg):
+        self.conn = conn
+        self.msg = msg
+        self.id = msg.id
+        self.org_id = msg.org_id
+        self.name = "{}:{}".format(msg.org_id, msg.name)
+
+    @classmethod
+    def _create(
```
ewagner-verta: There seems to be some inconsistency in the naming conventions we use in these cases.  Elsewhere, we would call this `_create_proto` since it contains that actual proto-request operation.  Perhaps that only exists where it was necessary to nest the request operation inside some additional logic, but it would be good to stay consistent, whatever we choose.

client/verta/verta/client.py
```diff
@@ -1735,3 +1738,23 @@ def find_datasets(
             datasets = datasets.find(predicates)
 
         return datasets
+
+    def _create_workspace(self, org_id, workspace_name, resource_action_groups):
```
ewagner-verta, line 1740: Doc string?

client/verta/verta/client.py
```diff
@@ -1735,3 +1738,23 @@ def find_datasets(
...
+            "GET", "/api/v1/uac-proxy/organization/{}/groups".format(org_id)
+        )
+        response_roles = self._conn.make_proto_request(
+            "GET", "/api/v2/uac-proxy/organization/{}/roles".format(org_id)
+        )
+        groups = self._conn.maybe_proto_response(response_groups, _Group.SearchGroups.Response).groups
+        all_users_group_id = next(iter(set(group.id for group in groups if group.name == "All Users")))
+        admins_group_id = next(iter(set(group.id for group in groups if group.name == "Admins")))
+        roles = self._conn.maybe_proto_response(response_roles, _Role.SearchRolesV2.Response).roles
+        super_user_role_id = next(iter(set(role.id for role in roles if role.name == "Super User")))
```
ewagner-verta: There is a lot of work happening here, namely 4 or 5 different requests that all seem only tangentially related to the operation of creating the workspace itself.  Perhaps `Groups` should have its own class and set of helper functions for returning these various subsets of users.

Something along the lines of 
```
        groups: List[str] = Groups._find_proto(
            self.conn,
            workspace_name,
            org_id,
            )
        custom_role_id: int = Role._create(
            self._conn,
            workspace_name + "role", 
            org_id, 
            resource_action_groups
            ).id
        return Workspace._create(
            self._conn,
            workspace_name,
            org_id,
            [
                _Workspace.Permission(
                    group_id = groups.admins,
                    role_id = groups.supers,
                    ),
                _Workspace.Permission(
                    group_id = groups.all,
                    role_id = custom_role_id,
                    )
            ]
        )
```
Just a thought, as this client file is already a bit unwieldy

ewagner-verta:
I'm curious why the `Role`, `Organization`, and `Workspace` classes are nested under `/tracking/`?  Is there a relationship there I should understand?

client/verta/verta/tracking/_role.py
liuverta: When there is just one or two variables involved, I agree that an f-string is a bit nicer

```suggestion
            f"/api/v2/uac-proxy/organization/{org_id}/roles",
```

client/verta/verta/client.py
liuverta: Agreed—this method definitely needs a docstring to explain what it is doing, and what each of the arguments arts.

For example, I'm not sure what exactly `resource_action_groups` is, or how it will be used.

liuverta (changes requested):
@evgeniilezhnin Could I ask you for a favor? Would you be able to split `_create_workspace()` and the new `class`es into a branch that we can merge into _`main`_? That way we do not have to re-review this code in #3407 which is already quite large.

---

And with regards to @ewagner-verta's comment

>I'm curious why the Role, Organization, and Workspace classes are nested under /tracking/? Is there a relationship there I should understand?

I would recommend putting `_role.py` and `_workspace.py` into a new directory, maybe `modeldb/client/verta/verta/_uac/`?

client/verta/verta/client.py
liuverta: As an alternative, just as a bare minimum to pull this logic out of `client.py`, I think we could have something like a

```python
class Organization:  # has its own _conn and id
    ...

    def create_role(self, name, resource_actions):
        return Role._create(...)

    def get_groups(self):
        ...
        return {group.name: group for group in groups}

    def get_roles(self):
        ...
        return {role.name: role for role in roles}
```

That way we do not have to keep passing around `self._conn` and `org_id`, too.

client/verta/verta/tracking/_role.py
evgeniilezhnin: renamed

client/verta/verta/client.py
evgeniilezhnin: added

client/verta/verta/client.py
evgeniilezhnin: added Organization

evgeniilezhnin:
> @evgeniilezhnin Could I ask you for a favor? Would you be able to split `_create_workspace()` and the new `class`es into a branch that we can merge into _`main`_? That way we do not have to re-review this code in #3407 which is already quite large.
> 
https://github.com/VertaAI/modeldb/pull/3497

> And with regards to @ewagner-verta's comment
> 
> > I'm curious why the Role, Organization, and Workspace classes are nested under /tracking/? Is there a relationship there I should understand?
> 
> I would recommend putting `_role.py` and `_workspace.py` into a new directory, maybe `modeldb/client/verta/verta/_uac/`?

moved

client/verta/verta/client.py
ewagner-verta: Can we consider just expanding the existing `verta.tracking._organization.Organization()` class instead of creating a secondary `v2` class for Organization?  It would be better to have a single class for organization to avoid confusion.  I still think it makes sense to move it into `verta._uac._organization...` since the UAC seems like a better fit than `tracking`.

client/verta/tests/test_endpoint/test_endpoint.py
```diff
@@ -781,7 +780,6 @@ def test_update_from_version_diff_workspace(
         created_entities.append(endpoint)
 
         endpoint.update(model_version, DirectUpdateStrategy(), wait=True)
-        assert endpoint.workspace != model_version.workspace
```
liuverta: Were these two tests failing here? 🤔 I think these checks are important to keep, because the purpose of the test is that the workspaces are different.

client/verta/verta/_internal_utils/_utils.py
```diff
@@ -353,7 +352,6 @@ def get_default_workspace(self):
         else:  # old backend
             return self.get_personal_workspace()
 
-
```
liuverta: Sorry to trouble you, but could we keep the blank lines that were removed in this file? [The convention in Python is](https://pep8.org/#blank-lines)

> Blank Lines
>
> Surround top-level function and class definitions with two blank lines.

client/verta/verta/client.py
```diff
@@ -48,6 +47,8 @@
 from .endpoint import Endpoints
 from .endpoint.update import DirectUpdateStrategy
 from .visibility import _visibility
+from ._protos.public.uac import WorkspaceV2_pb2 as _Workspace
```
liuverta: See my comment in `_organization.py`

```suggestion
from ._protos.public.uac import WorkspaceV2_pb2
```

client/verta/verta/client.py
```diff
@@ -48,6 +47,8 @@
...
+from verta._uac._workspace import Workspace
```
liuverta:
```suggestion
from ._uac._workspace import Workspace
```

client/verta/verta/client.py
```diff
@@ -1735,3 +1736,34 @@ def find_datasets(
...
+        org_id : str
+            ID of organization in which workspace is created.
+        workspace_name : str
+            name of workspace.
+        resource_action_groups : [_Role.RoleResourceActions]
+            role description with allowed actions for resource types.
+
+        Returns
+        -------
+        :class:`~verta.tracking._workspace.Workspace`
```
liuverta:
```suggestion
        :class:`~verta._uac._workspace.Workspace`
```

client/verta/verta/client.py
```diff
@@ -1735,3 +1736,34 @@ def find_datasets(
...
+        """
+        creates a workspace with a custom role for all users.
+
+        Parameters
+        ----------
+        org_id : str
+            ID of organization in which workspace is created.
+        workspace_name : str
+            name of workspace.
+        resource_action_groups : [_Role.RoleResourceActions]
```
liuverta:
```suggestion
        resource_action_groups : list of RoleV2_pb2.RoleResourceActions
```

client/verta/verta/_uac/_organization.py
```diff
@@ -0,0 +1,22 @@
+from ._role import Role
+from verta._protos.public.uac import GroupV2_pb2 as _Group
+from verta._protos.public.uac import RoleV2_pb2 as _Role
```
liuverta: I would actually recommend using `..._pb2` so it's clear it is a protobuf module (older parts of the client use `import ... as ...`, which I know is confusing and I intend to update in the future)

```suggestion
from verta._protos.public.uac import GroupV2_pb2
from verta._protos.public.uac import RoleV2_pb2
```

client/verta/tests/conftest.py
```diff
@@ -36,6 +36,8 @@
     mock_env_jwt_auth,
     mock_env_authn_missing,
 )
+from verta._internal_utils._utils import generate_default_name
+from verta._protos.public.uac import RoleV2_pb2 as _Role
```
liuverta: See my comment in `_organization.py`

```suggestion
from verta._protos.public.uac import RoleV2_pb2
```

client/verta/verta/client.py
```diff
@@ -1735,3 +1736,34 @@ def find_datasets(
...
+            role description with allowed actions for resource types.
```
liuverta: I think this would be more accurate, based on how `custom_role_id` is used.

```suggestion
            Resource actions for non-admins in this workspace.
```

client/verta/verta/client.py
```diff
@@ -1735,3 +1736,34 @@ def find_datasets(
...
+        Returns
+        -------
+        :class:`~verta.tracking._workspace.Workspace`
+
+        """
+        org = OrganizationV2(self._conn, org_id)
+        groups = org.get_groups()
+        all_users_group_id = next(iter(set(group.id for group in groups if group.name == "All Users")))
+        admins_group_id = next(iter(set(group.id for group in groups if group.name == "Admins")))
+        super_user_role_id = next(iter(set(role.id for role in org.get_roles() if role.name == "Super User")))
```
liuverta, line 1762: I'm a bit surprised we don't have a way to get group/role by name from UAC 😮 But this is...alright with me, if it needs to be done this way.

client/verta/verta/_uac/_workspace.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+from verta._protos.public.uac import WorkspaceV2_pb2 as _Workspace
```
liuverta: See my comment in `_organization.py`.

```suggestion
from verta._protos.public.uac import WorkspaceV2_pb2
```

client/verta/verta/_uac/_workspace.py
```diff
@@ -0,0 +1,49 @@
...
+    Object representing a Workspace.
+
+    """
+
+    def __init__(self, conn, msg):
+        self.conn = conn
+        self.msg = msg
+        self.id = msg.id
+        self.org_id = msg.org_id
+        self.name = "{}:{}".format(msg.org_id, msg.name)
```
liuverta: I saw [in Slack](https://vertaai.slack.com/archives/C03PWNZU5CM/p1673376588729409?thread_ts=1673369535.993669&cid=C03PWNZU5CM) that we use `%org_id/%workspace_name`. Should this be

```suggestion
        self.name = "{}/{}".format(msg.org_id, msg.name)
```

instead?

(I don't think it matters too much. We can change later)

client/verta/verta/_uac/_organization.py
```diff
@@ -0,0 +1,22 @@
...
+class OrganizationV2:
+    def __init__(self, conn, org_id):
+        self.conn = conn
+        self.org_id = org_id
+
+    def get_groups(self):
+        response_groups = self.conn.make_proto_request(
+            "GET", "/api/v1/uac-proxy/organization/{}/groups".format(self.org_id)
```
liuverta: I'm not sure about this, but should we use `v2` instead of `v1`? 🤔 

```suggestion
            "GET", "/api/v1/uac-proxy/organization/{}/groups".format(self.org_id)
```

client/verta/verta/_uac/_organization.py
```diff
@@ -0,0 +1,22 @@
...
+        )
+        return self.conn.maybe_proto_response(response_groups, _Group.SearchGroups.Response).groups
+
+    def get_roles(self):
+        response_roles = self.conn.make_proto_request(
+            "GET", "/api/v1/uac-proxy/organization/{}/roles".format(self.org_id)
```
liuverta: Also please use

```suggestion
            "GET", f"/api/v1/uac-proxy/organization/{self.org_id}/roles"
```

! 🙇

client/verta/verta/_uac/_role.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+from verta._protos.public.uac import RoleV2_pb2 as _Role
```
liuverta: See my comment in `_organization.py`

```suggestion
from verta._protos.public.uac import RoleV2_pb2
```

liuverta (changes requested):

client/verta/verta/client.py
evgeniilezhnin: existing Organization in main is used in tests. I think old organization in this PR is not used anymore except 3 CollaboratorType usages which can be easily removed. So I think existing Organization is a candidate for removal and not sure that we need to combine it with the new organization. Also we haven't been adding any backward compatibility for PythonClient that is why I'm not sure that existing Organization needed. OrganizationV2 can be renamed to Organization. I just renamed it to V2 to be able to create a separate change: https://github.com/VertaAI/modeldb/pull/3497

client/verta/verta/_uac/_organization.py
evgeniilezhnin: I think v2 makes more sense, but need to change it in proto first: https://github.com/VertaAI/protos-all/pull/425

client/verta/verta/_uac/_workspace.py
evgeniilezhnin: no, the problem is that "/" doesn't work for grpc proxy. Need to use some other symbol (":" is allowed too currently) or implement your suggestion https://vertaai.slack.com/archives/C03PWNZU5CM/p1673390118488849?thread_ts=1673381601.789169&cid=C03PWNZU5CM

client/verta/tests/test_endpoint/test_endpoint.py
evgeniilezhnin: reverted these changes. Now it is working. Sorry, I think I misunderstood these tests before and I do not know why it was failing before: maybe there was a different MDB version.

client/verta/verta/_internal_utils/_utils.py
evgeniilezhnin: reverted

client/verta/verta/_uac/_organization.py
evgeniilezhnin: applied

client/verta/verta/client.py
evgeniilezhnin: yes, we don't have search by name at the initial version

client/verta/verta/client.py
liuverta: That's a fair question @ewagner-verta. _Technically_ they could be combined. But _pragmatically_, I think what we want is to remove the original `_organization` `Organization` (it was never meant for public use anyway) and only use the new `_uac` `OrganizationV2`.

But like @evgeniilezhnin mentioned, we might still need the old one for compatibility with some existing tests. As soon as we determine it's not needed, `_organization` `Organization` should be removed.

client/verta/verta/_uac/_organization.py
```diff
@@ -0,0 +1,22 @@
...
+from verta._protos.public.uac import GroupV2_pb2
+from verta._protos.public.uac import RoleV2_pb2
+class OrganizationV2:
+    def __init__(self, conn, org_id):
+        self.conn = conn
+        self.org_id = org_id
+
+    def get_groups(self):
+        response_groups = self.conn.make_proto_request(
+            "GET", f"/api/v1/uac-proxy/organization/{self.org_id}/groups"
```
liuverta: Should this also be `v1` instead of `v2`? (I'm not sure what our guidelines are for when to use `v2` 😵‍💫).

client/verta/verta/_uac/_workspace.py
```diff
@@ -0,0 +1,49 @@
...
+    @classmethod
+    def _create_proto(
+        cls, conn, name, org_id, permissions
+    ):
+        Message = WorkspaceV2_pb2.SetWorkspaceV2
+        msg = cls._create_msg(name, org_id, permissions)
+
+        response = conn.make_proto_request(
+            "POST",
+            "/api/v2/uac-proxy/organization/{}/workspaces".format(org_id),
```
liuverta: f-string if we could, please!

```suggestion
            f"/api/v2/uac-proxy/organization/{org_id}/workspaces",
```

client/verta/verta/_uac/_workspace.py
```diff
@@ -0,0 +1,49 @@
...
+                      permissions = permissions)
+        return msg
+
+    def delete(self):
+        """
+        Deletes this workspace.
+
+        """
+        Message = WorkspaceV2_pb2.DeleteWorkspaceV2
+        endpoint = "/api/v2/uac-proxy/organization/{}/workspace/{}".format(self.org_id, self.id)
```
liuverta:
```suggestion
        endpoint = f"/api/v2/uac-proxy/organization/{self.org_id}/workspace/{self.id}"
```

client/verta/verta/_uac/_workspace.py
```diff
@@ -0,0 +1,49 @@
...
+        msg = cls._create_msg(name, org_id, permissions)
+
+        response = conn.make_proto_request(
+            "POST",
+            "/api/v2/uac-proxy/organization/{}/workspaces".format(org_id),
+            body=Message(workspace=msg),
+        )
+        workspace = conn.must_proto_response(response, Message.Response).workspace
+
+        print("created new Workspace : {}".format(workspace.name))
```
liuverta:
```suggestion
        print(f"created new Workspace : {workspace.name}")
```

client/verta/verta/_uac/_role.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, conn, msg):
+        self.conn = conn
+        self.msg = msg
+        self.id = msg.id
+        self.org_id = msg.org_id
+        self.name = msg.name
+
+    @classmethod
+    def _create_proto(
```
liuverta: @ewagner-verta re: https://github.com/VertaAI/modeldb/pull/3439#discussion_r1063627806, these should actually be

```suggestion
    def _create(
```

because they return an instantiated object from our client

```python
return cls(conn, role)
```

which is analogous to, say, [`_ModelDBEntity._create()`](https://github.com/VertaAI/modeldb/blob/1fbe2533c3310f1ee2eeee78348df09056d7cf34/client/verta/verta/tracking/entities/_entity.py#L153-L185) and [`Endpoint._create()`](https://github.com/VertaAI/modeldb/blob/1fbe2533c3310f1ee2eeee78348df09056d7cf34/client/verta/verta/endpoint/_endpoint.py#L132-L157)

[`_create_proto()`](https://github.com/VertaAI/modeldb/blob/1fbe2533c3310f1ee2eeee78348df09056d7cf34/client/verta/verta/tracking/entities/_entity.py#L187-L207) (and the [`_create_proto_internal()`](https://github.com/VertaAI/modeldb/blob/1fbe2533c3310f1ee2eeee78348df09056d7cf34/client/verta/verta/tracking/entities/_experimentrun.py#L178-L209) it calls, implemented by subclasses) returns a _protobuf_ object, instead. And I'm kind of alright with just implementing `_create()` for these, skipping all the protobuf helper methods, because these only really need to be minimally implemented for use in tests.

liuverta: @evgeniilezhnin sorry to trouble you, but would you mind changing these `_create_proto()`s back to `_create()`?

liuverta (changes requested):
I think this should be the last set of requested changes 🙇

client/verta/tests/constants.py
```diff
@@ -29,7 +29,6 @@
 DEV_KEY_2 = os.environ.get(DEV_KEY_2_ENV_VAR)
 EMAIL_3 = os.environ.get(EMAIL_3_ENV_VAR)
 DEV_KEY_3 = os.environ.get(DEV_KEY_3_ENV_VAR)
-WORKSPACE = os.environ.get(WORKSPACE_ENV_VAR, DEFAULT_ENV)
```
liuverta, line 32: Do you think we should remove `WORKSPACE_ENV_VAR` and `DEFAULT_ENV` from this file, too?

liuverta (changes requested):

client/verta/verta/_uac/_organization.py
evgeniilezhnin: yeah, changed it to v2, I think we discussed it already here. It needed proto changes, they are merged now so now can change it to v2

client/verta/verta/_uac/_role.py
evgeniilezhnin: reverted

client/verta/tests/constants.py
evgeniilezhnin: I think it is not used and was introduced in VUMM-760_remove_organization PR. But can revert it if we plan to use it somehow in the future. Reverted

client/verta/tests/conftest.py
```diff
@@ -520,13 +531,43 @@ def predict(self, x):
...
 @pytest.fixture
 def workspace(client, created_entities):
-    return Workspace(client._conn.get_default_workspace())
+    return create_workspace(client, created_entities, [
+    RoleV2_pb2.RoleResourceActions(resource_type=RoleV2_pb2.ResourceTypeV2.ENDPOINT,
+                                   allowed_actions=[RoleV2_pb2.ActionTypeV2.READ]),
+    RoleV2_pb2.RoleResourceActions(resource_type=RoleV2_pb2.ResourceTypeV2.REGISTERED_MODEL,
+                                   allowed_actions=[RoleV2_pb2.ActionTypeV2.READ,
+                                                    RoleV2_pb2.ActionTypeV2.UPDATE]),
+])
```
liuverta: The indentation seems a bit strange to me, here. I would recommend trying this:

```suggestion
    return create_workspace(
        client,
        created_entities,
        [
            RoleV2_pb2.RoleResourceActions(
                resource_type=RoleV2_pb2.ResourceTypeV2.ENDPOINT,
                allowed_actions=[RoleV2_pb2.ActionTypeV2.READ],
            ),
            RoleV2_pb2.RoleResourceActions(
                resource_type=RoleV2_pb2.ResourceTypeV2.REGISTERED_MODEL,
                allowed_actions=[RoleV2_pb2.ActionTypeV2.READ,
                                 RoleV2_pb2.ActionTypeV2.UPDATE],
            ),
        ],
    )
```

(It's a bit more conventional this way, even though it's more lines)

liuverta (approved):

client/verta/tests/constants.py
```diff
@@ -29,6 +29,7 @@
 DEV_KEY_2 = os.environ.get(DEV_KEY_2_ENV_VAR)
 EMAIL_3 = os.environ.get(EMAIL_3_ENV_VAR)
 DEV_KEY_3 = os.environ.get(DEV_KEY_3_ENV_VAR)
+WORKSPACE = os.environ.get(WORKSPACE_ENV_VAR, DEFAULT_ENV)
```
liuverta: I would actually recommend _removing_ `WORKSPACE`, `WORKSPACE_ENV_VAR`, and `DEFAULT_ENV` if they are not being used. Otherwise, I think we might forget, and become confused about why they are there.

evgeniilezhnin: removed

PR #3441, opened 2022-12-12
chore: Add a metric that measures the time a Future waits before being executed

jkwatson-verta:
## Impact and Context
This will have zero impact unless the OpenTelemetry instance has been configured to have an active MeterProvider.

## Testing
- [ ] Unit test
- [x] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
@@ -21,6 +21,7 @@ public class FutureExecutor implements Executor {
   @With private final io.grpc.Context grpcContext;
 
   public FutureExecutor captureContext() {
+    //set the current nano time into the context, so we can measure how long it takes to actually execute the future.
     io.opentelemetry.context.Context otelContext = Context.current();
     otelContext = otelContext.with(EXECUTION_TIMER_KEY, System.nanoTime());
     return withOtelContext(otelContext).withGrpcContext(io.grpc.Context.current());
@@ -60,6 +61,7 @@ public class FutureExecutor implements Executor {
 
   private Runnable wrapWithTimer(Runnable r) {
     return () -> {
+      //pull the start time out of the context, and record how much time has elapsed.
       Long startTime = io.opentelemetry.context.Context.current().get(EXECUTION_TIMER_KEY);
       if (startTime != null) {
         long endTime = System.nanoTime();
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
@@ -3,6 +3,11 @@ package ai.verta.modeldb.common.futures;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.metrics.LongHistogram;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
 import lombok.AccessLevel;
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -10,13 +14,30 @@
...
+
+  private static LongHistogram getSchedulingDelayHistogram() {
+    if (schedulingDelayHistogram != null) {
+      return schedulingDelayHistogram;
+    }
+    schedulingDelayHistogram =
+        GlobalOpenTelemetry.get()
+            .getMeter("verta.future")
+            .histogramBuilder("future_exec_delay")
+            .setDescription("The number of microseconds between creating and executing a Future")
```
liam-verta, line 37: Milliseconds or microseconds?

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -37,8 +58,24 @@ public static FutureExecutor newSingleThreadExecutor() {
...
 
+  private Runnable wrapWithTimer(Runnable r) {
+    return () -> {
+      Long startTime = io.opentelemetry.context.Context.current().get(EXECUTION_TIMER_KEY);
+      if (startTime == null) {
+        r.run();
+        return;
+      }
+      long endTime = System.nanoTime();
+      long millisDelay = (endTime - startTime) / 1_000L;
```
liam-verta, line 69: Milliseconds or microseconds?

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -37,8 +58,24 @@ public static FutureExecutor newSingleThreadExecutor() {
...
+      getSchedulingDelayHistogram().record(millisDelay);
+      r.run();
```
liam-verta: Restructure to have one call to `run()` and name variable with micros instead of millis.
```suggestion
      if (startTime != null) {
        long endTime = System.nanoTime();
        long microsDelay = (endTime - startTime) / 1_000L;
        getSchedulingDelayHistogram().record(microsDelay);
      }
      r.run();
```

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
```diff
@@ -37,8 +58,24 @@ public static FutureExecutor newSingleThreadExecutor() {
     return makeCompatibleExecutor(Executors.newSingleThreadExecutor());
   }
 
+  private Runnable wrapWithTimer(Runnable r) {
+    return () -> {
```
liam-verta, line 64: Is it possible to make this return `r` unchanged if the OpenTelemetry instance hasn't been configured?

liam-verta (approved):
Could use some comments as well, but generally looks good to me.

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
jkwatson-verta: I like it. good cleanup!

backend/common/src/main/java/ai/verta/modeldb/common/futures/FutureExecutor.java
jkwatson-verta: There isn't a way to know if OpenTelemetry has been configured or not... so unfortunately not.

jkwatson-verta: Another option (that I will probably end up with) is to inject the OpenTelemetry MeterProvider into this class, rather than rely on the global. Right now...still experimenting, though!

PR #3450, opened 2022-12-14
chore: [PLAT-67] switch to using the jaeger grpc exporter

jkwatson-verta:
## Testing
- [x] Unit test
- [x] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/pom.xml
+++ b/backend/common/pom.xml
@@ -111,10 +111,6 @@
         </dependency>
         <dependency>
             <groupId>io.opentelemetry</groupId>
-            <artifactId>opentelemetry-exporter-jaeger-thrift</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-exporter-jaeger</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


conradoverta (approved):

backend/common/pom.xml
```diff
@@ -109,6 +113,10 @@
             <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-exporter-jaeger-thrift</artifactId>
         </dependency>
+        <dependency>
```
liam-verta: Should we removed the `opentelemetry-exporter-jaeger-thrift` above?

liam-verta:
You'll also need to update the downstream projects right?

backend/common/pom.xml
jkwatson-verta: yes! good call.

jkwatson-verta:
> You'll also need to update the downstream projects right?

I think now that this code all lives here, that it should be ok? Unless we aren't properly depending on the common module in them?

backend/common/pom.xml
jkwatson-verta: done

jkwatson-verta:
> > You'll also need to update the downstream projects right?
> 
> I think now that this code all lives here, that it should be ok? Unless we aren't properly depending on the common module in them?

unfortunately, it does look like we'll need a couple of updates to the dependencies downstream, yes. I'll jump on them ASAP after this merges.

coryverta (approved):

PR #3470, opened 2022-12-28
chore: Replace usages of InternalFuture with Future in the modeldb-server

jkwatson-verta:
## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


jkwatson-verta:
this is too old. will redo this at some point in the future

PR #3473, opened 2022-12-29
chore: Provide a method on the Future for a flexible retry strategy 

jkwatson-verta:
## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [x] Other (explain) - tried this out in the registry for async http retries. 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/DoublingBackoffStrategy.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/DoublingBackoffStrategy.java
@@ -1,12 +1,11 @@
 package ai.verta.modeldb.common.futures;
 
-import lombok.AccessLevel;
-import lombok.Getter;
-import lombok.Value;
-
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Value;
 
 @Value
 @Getter(AccessLevel.NONE)
--- a/backend/common/src/main/java/ai/verta/modeldb/common/futures/DoublingBackoffStrategy.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/futures/DoublingBackoffStrategy.java
@@ -1,0 +1,27 @@
+package ai.verta.modeldb.common.futures;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Value;
+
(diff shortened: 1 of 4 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/futures/DoublingBackoffStrategy.java
```diff
@@ -0,0 +1,28 @@
...
+@Getter(AccessLevel.NONE)
+class DoublingBackoffStrategy<R> implements RetryStrategy<R> {
+  private static final int INITIAL_DELAY_MILLIS = 10;
+  BiPredicate<R, Throwable> resultChecker;
+  int maxRetries;
+  AtomicInteger numberRetried = new AtomicInteger();
+
+  @Override
+  public Retry shouldRetry(R result, Throwable throwable) {
+    boolean apply = resultChecker.test(result, throwable);
```
eheinlein-verta: I'm not in love with the name `apply`. Maybe something like `success`?

backend/common/src/test/java/ai/verta/modeldb/common/futures/FutureTest.java
```diff
@@ -144,6 +145,57 @@ void retry() throws Exception {
...
 
+  @Test
+  void retry_strategy_failure() {
+    AtomicInteger sequencer = new AtomicInteger();
+
+    assertThatThrownBy(
+            () ->
+                Future.retrying(
+                        () -> {
+                          if (sequencer.getAndIncrement() < 10) {
```
eheinlein-verta: Is the goal of the # 10 here to just be more than 6 (the expected max number)? Or is it more specific? If the goal here is just to be "too many retries" maybe pull the literal into a variable named `tooManyRetries` or something?

eheinlein-verta (approved):
Approved, with just some small naming suggestions if you agree.

backend/common/src/main/java/ai/verta/modeldb/common/futures/DoublingBackoffStrategy.java
jkwatson-verta: yes. this was left over from a refactoring. will definitely clean up. good catch!

backend/common/src/test/java/ai/verta/modeldb/common/futures/FutureTest.java
jkwatson-verta: yeah, it's just intended to be more times than the maximum configured. I'll rename it. good suggestion!

PR #3477, opened 2023-01-02
Remove Global sharing test for permission v2

anandjakhaniya:
## Impact and Context
- I have removed this test because it was checking collaboration sharing for all users when we create resources using the Organization workspace but in the new system by default, we have shared by workspace concept at the organization and we do not have a collaborator concept at the organization so this test will have no sense anymore. The new concept is already been achieved by Other resources test classes so no need to implement it here.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


jkwatson-verta:
The PR description should include some justification for removing this. thanks!

anandjakhaniya:
> The PR description should include some justification for removing this. thanks!

Updated description as per review comment.

backend/server/src/test/java/ai/verta/modeldb/TestSequenceSuite.java
```diff
@@ -30,7 +30,6 @@
   BlobProtoEquality.class,
   DiffAndMerge.class,
   ValidatorBlobTest.class,
-  ValidatorBlobDiffTest.class,
-  GlobalSharingTest.class
+  ValidatorBlobDiffTest.class
```
jkwatson-verta, line 33: curious...is this test suite really still used? Or are we only running the PullRequestSuite now?

jkwatson-verta (approved):

eheinlein-verta (approved):

PR #3497, opened 2023-01-09
create workspace v2, add permission v2 classes

evgeniilezhnin:
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


liuverta (changes requested):
I actually meant "can we close #3439, and have a new PR for `main`?" instead of having two PRs like this 😅 But that is alright.

See my comments in #3439 for requested changes

evgeniilezhnin:
> I actually meant "can we close #3439, and have a new PR for `main`?" instead of having two PRs like this sweat_smile But that is alright.
> 
> See my comments in #3439 for requested changes

I think we can't close it, because besides classes it also contains test changes that is dependent on organization removal PR

PR #3500, opened 2023-01-12
feat: [VRD-620] Implement client function to collect log entries

ewagner-verta:
## Impact and Context
Adds tooling for logging context around predictions.

All that is required of client users is that they add calls like `verta.runtime.log(any_key='any_value')` with any number of arguments, anywhere within the scope of their model's predict method, and that key-value will be automatically added to the logging context for that prediction.  The collection of logs is finalized and uploaded to S3 upon completion of the predict method.

## Sample Script
This simple .py script attempts to illustrate how usage of this code would work.
```python
import sys
from pprint import pprint
from verta import runtime

def print_stats():
    """ Print current content and memory usage of the stored context. """
    print("Logs:")
    [print(f"    {k}: {v}") for k,v in runtime._get_thread_logs().items()]
    print(f"Memory usage: {sys.getsizeof(runtime._get_thread_logs())} \n")

# Imitate a user-defined model predict method.
# Note that no context is required.  Just call runtime.log from anywhere in your predict method.
def predict(data):
    runtime.log('prediction_category', 'category1')
    print_stats()

    # Imagine a transformation here
    transformed_input = data + '_post_transformation'
    runtime.log('transformed_input', transformed_input)
    print_stats()

    # Imagine prediction logic here.
    prediction_output = 'imaginary_output'
    runtime.log('output', prediction_output)
    print_stats()

# Imagine post-processing here
    transformed_output = prediction_output + '_post_processing'
    runtime.log('processed_output', transformed_output)
    print_stats()
    return transformed_output


# Imitate deployment/python-model-service/app/runtime/cherrypy/predict_resource.PredictResource POST
# Uses context manager to ensure an empty logging handler to start with, aggregation of logs
# provided inside the predict() method, and clean-up of logging handler upon exit of context.
def predict_wrapper(data):
    with runtime.context(validate=True) as ctx:
        print(f"Memory usage of context on starting context manager: {sys.getsizeof(runtime._get_thread_logs())}\n")
        runtime.log('outer_key_2', 'another_outer_context_example_val')
        predict(data)
        print("Final log entry:")
        pprint(ctx.logs())

    print(f"\nMemory usage of thread-local var after exiting context manager (empty dict): {sys.getsizeof(runtime._get_thread_logs())}")
    print(f"Value of final logs still stored on context manager object:")
    pprint(ctx.logs())
    print_stats()

if __name__ == '__main__':
    predict_wrapper('prediction_from_requestor')
        
```
Output:
```
Memory usage of context on starting context manager: 64

Logs:
    outer_key_2: another_outer_context_example_val
    prediction_category: category1
Memory usage: 232 

Logs:
    outer_key_2: another_outer_context_example_val
    prediction_category: category1
    transformed_input: prediction_from_requestor_post_transformation
Memory usage: 232 

Logs:
    outer_key_2: another_outer_context_example_val
    prediction_category: category1
    transformed_input: prediction_from_requestor_post_transformation
    output: imaginary_output
Memory usage: 232 

Logs:
    outer_key_2: another_outer_context_example_val
    prediction_category: category1
    transformed_input: prediction_from_requestor_post_transformation
    output: imaginary_output
    processed_output: imaginary_output_post_processing
Memory usage: 232 

Final log entry:
{'outer_key_2': 'another_outer_context_example_val',
 'output': 'imaginary_output',
 'prediction_category': 'category1',
 'processed_output': 'imaginary_output_post_processing',
 'transformed_input': 'prediction_from_requestor_post_transformation'}

Memory usage of thread-local var after exiting context manager (empty dict): 64
Value of final logs still stored on context manager object:
{'outer_key_2': 'another_outer_context_example_val',
 'output': 'imaginary_output',
 'prediction_category': 'category1',
 'processed_output': 'imaginary_output_post_processing',
 'transformed_input': 'prediction_from_requestor_post_transformation'}
Logs:
Memory usage: 64 
```
## Testing
- [X] Unit test (Not yet, still feeling out the implementation)
- [ ] Deployed to dev env
- [ ] Other - (explain)

## Reverting


client/verta/verta/runtime.py
```diff
@@ -0,0 +1,84 @@
...
+    def add(self, **kwargs):
+        # Overwrites existing key-values
+        self.attributes.update(**kwargs)
+
+    def wipe_clean(self):
+        self.attributes = dict()
+
+# Use a single handler object that is wiped clean after each log entry
+# to minimize memory usage
+LOG_HANDLER = LogHandler()
```
conradoverta: we should probably use a thread-local object vs a global? I'd check how other people have done this kind of magic association via context for inspiration

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        # Handler attributes get deleted when context is exited.
+        self.context.wipe_clean()
+
+    @property
+    def as_json(self) -> str:
```
liuverta: Isn't this supposed to be

```suggestion
    def logs(self) -> dict:
```

as per the design doc we agreed on?

https://github.com/VertaAI/docs/blob/main/design-docs/2023-01-model-prediction-logging.md#logging-key-values

I don't see mention of any reason to have deviated from that.

liuverta: Also, as a personal opinion, I don't think this should be a `property` either way. I believe that if a user accesses an object attribute:

```python
ctx.as_json
```

they should almost _never_ expect to encounter an exception (other than a `object has no attribute` `AttributeError`). If an exception can occur due to extra internal processing (or moreso due to data validation), it should be a method call:

```python
ctx.as_json()
```

See https://github.com/psf/requests/issues/1028#issuecomment-11474800

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,84 @@
...
+    Any logged data is deleted upon exiting the context.
+
+    Attributes
+    ----------
+    as_json : str
+        A JSON string representation of all the current saved logging context.
+    """
+    def __init__(self):
+        self.context = LOG_HANDLER
+        if bool(self.context.attributes): # if not empty dict
```
```suggestion
        if self.context.attributes:
```

As far as I know, explicitly casting to `bool` is extremely unidiomatic in Python, and it's almost always clear from context (and better: from type analysis tooling) what's being checked.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,84 @@
...
+
+class LogHandler:
+    """
+    Container class for collecting and dumping log entries.
+    """
+    def __init__(self):
+        self.attributes: Dict[str, Any] = dict()
+
+    def __repr__(self):
+        return str(self.attributes)
```
liuverta: Not sure about this, but maybe

```suggestion
        return repr(self.attributes)
```

?

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,84 @@
...
+    ----------
+    as_json : str
+        A JSON string representation of all the current saved logging context.
+    """
+    def __init__(self):
+        self.context = LOG_HANDLER
+        if bool(self.context.attributes): # if not empty dict
+            self.context.wipe_clean()
+
+    def log(self, **kwargs) -> None:
```
liuverta: Isn't this supposed to be

```suggestion
    def log(self, key, val) -> None:
```

as per the design doc we agreed on?

https://github.com/VertaAI/docs/blob/main/design-docs/2023-01-model-prediction-logging.md#logging-key-values

I don't see mention of any reason to have deviated from that.

liuverta (changes requested):

client/verta/verta/runtime.py
ewagner-verta: This has been refactored out.

client/verta/verta/runtime.py
ewagner-verta: I suppose we need to come to an agreement on exactly how prescriptive the design doc is supposed to be.  Using `**kwargs` here allows the user to add any number of key-value pairs they want to log at that moment.

`runtime.log(key1='val1', key2='val2', key3='val3')`
just makes more sense than
```
runtime.log('key1', 'val1')
runtime.log('key2', 'val2')
runtime.log('key3', 'val3')
```
and forces the key's used to be log-friendly, queryable strings that can be legit column names in the final Athena table

client/verta/verta/runtime.py
ewagner-verta: Refactored out

client/verta/verta/runtime.py
ewagner-verta: Fair point.  I agree with your logic here. ~Fixed~

client/verta/verta/runtime.py
ewagner-verta: Latest iteration is my attempt to use a thread local object for storing the context.  Admittedly, I'm new to working with the `threading` module, so this will likely need some work and I need to write a test to make sure it actually is thread safe.

client/verta/verta/runtime.py
ewagner-verta: Fixed naming convention to `logs()`

client/verta/verta/runtime.py
ewagner-verta: I also updated the sample script in the PR description for the latest changes.

client/verta/verta/runtime.py
liuverta: > I suppose we need to come to an agreement on exactly how prescriptive the design doc is supposed to be.

I'm not opposed in principle to deviating from a design doc, if we have reason to! Buuuuuuuut

> ```python
> runtime.log(key1='val1', key2='val2', key3='val3')
> ```
> just makes more sense

I disagree with this. I think there _is_ merit in an argument that it forces the keys to be strings, but the idea that a user's arbitrary items for logging are _keyword arguments_ to a function is extremely unintuitive to me. Like, I don't think that's what function parameters are meant for, and no other libraries (and certainly no other client functions) come to mind that do this: allow a user to pass their own arbitrary key-value items as named arguments (though if you do have any examples, I'd be interested in considering them).

Calling `log()` multiple times is also a somewhat established practice, for example in TensorBoard (https://www.tensorflow.org/api_docs/python/tf/summary/scalar#arguments). But if we'd like to support

> add any number of key-value pairs they want to log at that moment.

I'd personally recommend just accepting a whole dictionary, like Weights & Biases does (https://docs.wandb.ai/guides/track/log), for which we also have established precedence for in our client via methods like [`log_hyperparameters()`](https://verta.readthedocs.io/en/main/_autogen/verta.tracking.entities.ExperimentRun.html#verta.tracking.entities.ExperimentRun.log_hyperparameters).

liuverta: I think it'd be more prudent to raise this question in #dev-python to hear another opinion on this. If a consensus was reached already during the design doc phase, I'm concerned about the implications of making last-mile changes without a similar breadth of rigor.

client/verta/verta/runtime.py
liuverta: It doesn't look like we've fully addressed my suggestion of

```python
def logs(self) -> dict:
```

The [design doc](https://github.com/VertaAI/docs/blob/main/design-docs/2023-01-model-prediction-logging.md#logging-key-values) outlines that `ctx.logs()` should return a `dict`, not a `str`.

If you are intent on changing this to a `str`, I **highly** recommend raising it #feat-model-data-log to get input from others. As-is, this breaks the API that we had consensus on for everyone consuming this downstream.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,93 @@
...
+
+from collections import defaultdict
+import json
+import threading
+import os
+
+
+_THREAD = threading.local()
+
+def init_thread_context() -> None:
```
liuverta: If you weren't intending to already, I'd recommend marking these `*_thread_context()` functions as private

```suggestion
def _init_thread_context() -> None:
```

as they are magic that we should discourage users from fiddling with or being concerned about, to my understanding.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,93 @@
...
+        Arguments and values to be added as key-value pairs to the current
+        logging context. Values must be serializable to JSON.
+
+    Returns
+    -------
+    None
+    """
+    local_context: defaultdict = get_thread_context().copy()
+    local_context.update(**kwargs)
+    set_thread_context(local_context)
```
liuverta: Out of curiosity, what reason did you have in mind for copy and reassigning `_THREAD.context` instead of

```suggestion
    get_thread_context().update(**kwargs)
```

? I'm unsure at a glance what copying will do for us, and it can be a bit of a costly operation if we don't need it.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,93 @@
...
+import threading
+import os
+
+
+_THREAD = threading.local()
+
+def init_thread_context() -> None:
+    """Initialize our thread-local variable for storing logging context."""
+    if not hasattr(_THREAD, "context"):
+        _THREAD.context = defaultdict()
```
liuverta: I think `.context` might not be the right name. To me, _this_ isn't "the context"; this is just accumulated logs. But no strong feelings, here.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,93 @@
...
+    ----------
+    kwargs : Any
+        Key-value pairs to be added to the current logging context.
+    Attributes
+    ----------
+    as_json : str
+        A JSON string representation of all the current saved logging context.
+    """
+    def __init__(self, **kwargs):
+        self.new_context = kwargs
```
liuverta: I think I'm a bit confused by how this is to be used. Is

```python
with context(key=val) as ctx:
    pass
```

just equivalent to

```python
with context() as ctx:
    ctx.log(key=val)
```

?

Considering that `context` is meant to be instantiated either by our Python Model Service, or by the user trying to emulate the model service, I'm not sure that it makes sense to build logging functionality into `__init__()`

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,93 @@
...
+    Parameters
+    ----------
+    kwargs : Any
+        Key-value pairs to be added to the current logging context.
+    Attributes
+    ----------
+    as_json : str
+        A JSON string representation of all the current saved logging context.
+    """
+    def __init__(self, **kwargs):
```
liuverta: We're missing the `validate` parameter from the [design doc](https://github.com/VertaAI/docs/blob/main/design-docs/2023-01-model-prediction-logging.md#logging-key-values).

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,93 @@
...
+        self.new_context = kwargs
+        self.prior_context = defaultdict()
+
+    def __enter__(self):
+        self.prior_context = get_thread_context()
```
liuverta: I'm tempted to suggest that we raise an error if there's already an active `context`. I don't think our feature is intended to be used with any degree of nesting, and using it that way would probably increase the likelihood of user errors. But I don't have a strong opinion on this.

liuverta (changes requested):

client/verta/verta/runtime.py
ewagner-verta: Got it.  I thought we had discussed in standup whether it should return a dict or a JSON string and I understood JSON to be the answer.  Easy thing to change.

ewagner-verta: Refactored to return `Dict[str, Any]`

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,93 @@
...
+        updated_context: defaultdict = self.prior_context.copy()
+        updated_context.update(self.new_context)
+        set_thread_context(updated_context)
+        return self
+
+    def __exit__(self, *args):
+        """ Return thread-local context to prior state when exiting this context. """
+        set_thread_context(self.prior_context)
```
liuverta: I think we're going to encounter issues if `ctx.logs()` is ever called after the context is exited (which is a bit of a wild case, but isn't totally out of the question in Python).

client/verta/verta/runtime.py
liuverta: Ah, a misunderstanding, then. Sorry about the confusion.

client/verta/verta/runtime.py
ewagner-verta: Refactored to accept separate `key` and `value` parameters and simply update the current context dictionary with that k-v pair.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+
```
liuverta, line 6: fyi if you weren't intending to already, we ought to add a module docstring ([example](https://github.com/VertaAI/modeldb/blob/7d7d578a5f09d40f90b374b369969ecb5610c018/client/verta/verta/credentials.py#L2-L6)) and add `runtime` to [python.rst](https://github.com/VertaAI/modeldb/blob/main/client/verta/docs/python.rst).

client/verta/verta/runtime.py
ewagner-verta: If we find that we need to enforce some constraints on the "key" parameter later, we can always add that in another PR.

liuverta: To be clear, I would've been good with raising this as a discussion point in one of our Slack channels. My personal disagreement and recommendation aren't absolute.

client/verta/verta/runtime.py
ewagner-verta: Updated to private functions as suggested.

client/verta/verta/runtime.py
ewagner-verta: Removed copy operation.

client/verta/verta/runtime.py
ewagner-verta: Refactored to use "logs" instead.  Seems more intuitive.

client/verta/verta/runtime.py
ewagner-verta: I refactored this to simplify it.  I had originally thought nesting these context managers would be a thing, but we do not need that for this particular iteration.  We could add the ability to nest, and even wrap a logging.Logger class to mimic python native logging calls like `warning`, or `error` etc. if customers want a more sophisticated implementation in the future.

client/verta/verta/runtime.py
ewagner-verta: Working on this now.  Just getting some clarification

client/verta/verta/runtime.py
ewagner-verta: This was part of my original design pattern, when I thought we would be allowing nesting.  This now just makes sure the logging context is empty upon entering and exiting the context manager.

client/verta/verta/runtime.py
ewagner-verta: Absolutely.  Will do

client/verta/verta/runtime.py
ewagner-verta: The advantage of writing our own context manager here is that we can store the final logs in a class variable so they could still be accessed after exiting the context.  The thread-local variable holding the logs would still be wiped on exit, but the context manager object would still have the logs if `ctx.logs()` were called outside the context.

I've just updated the PR with that change, we can revert if we change our minds.

client/verta/verta/runtime.py
ewagner-verta: Understood, and thanks. It's not so much that I don't feel comfortable challenging, but that you have a breadth of valuable experience with this product, client code, and certain Python constructs upon which I depend.  Very often we bounce back and forth, and I wind up convinced that you are right :) 

I honestly didn't feel that strongly about it and there is nothing about doing it with `kwargs` that I think would have a significant positive impact on the overall user experience.

client/verta/verta/runtime.py
ewagner-verta: Added `validate` flag, the boolean for which is also stored as a thread local variable.

client/verta/verta/runtime.py
ewagner-verta: Done. Can you sanity check for me that I have added the `runtime` module to `python.rst` correctly?

client/verta/docs/python.rst
```diff
@@ -33,3 +33,4 @@ Verta
         tracking
         utils
         visibility
+        runtime
```
liuverta: This is indeed the correct way to enable documentation for the `verta.runtime` module!

Would you mind slotting it just below `registry` so that it's in alphabetical order?

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+def _set_validate_flag(flag: bool) -> bool:
+    """
+    Sets the thread-local validate flag boolean value.
+    """
+    _THREAD.validate = flag
+    return _THREAD.validate
+
+
+def validate_json(value: Any) -> str:
```
liuverta: If this is just for our internal use, we should mark it as such

```suggestion
def _validate_json(value: Any) -> str:
```

Otherwise we should expand the docstring.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class context:
+    """
+    Context manager for aggregating key-value pairs into a custom log entry.
+    Uses thread-local variables to store context in a thread-safe manner.
+
+    Parameters
+    ----------
+    validate : Optional[bool]
```
liuverta: I'm not sure if `typing` has a convention for defaults, but I think technically (in numpydoc conventions) this should be

```suggestion
    validate : bool, default False
```

rather than `Optional`, to explicitly say "the default is to _not_ validate".

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+    ----------
+    key : str
+        String value to use as data label (key) in logging context.
+    value : Any
+        Any JSON serializable value you wish to include in the context logs.
+
+    Returns
+    -------
+    None
+
```
liuverta, line 114: We should add a `Raises` section here!

https://numpydoc.readthedocs.io/en/latest/format.html#raises

I'd recommend a wording like

```rst
If `validate` was set to ``True`` on the active :class:`context`, and `value` is not JSON-serializable.
```

client/verta/docs/python.rst
liuverta: Also fyi—since apparently the ReadTheDocs PR check is no longer cooperating—we have instructions for locally building docs that can be used to verify things render correctly: https://github.com/VertaAI/modeldb/tree/main/client/verta/docs

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    Attempt to serialize the provided value to JSON.
+    Raise an error on failure.
+    """
+    try:
+        return json.dumps(value)
+    except json.JSONDecodeError as json_err:
+        raise Exception("Unable to convert logging data to JSON") from json_err
+    except TypeError as type_err:
+        raise Exception("Unable to serialize log value to JSON.") from type_err
```
liuverta: Something to consider: To my understanding, `raise Exception` is somewhat bad form. This means that an end-user trying to handle JSON issues while logging (maybe for debugging, maybe for testing) would have to write

```python
try:
    runtime.log("foo", training_dataframe)
except Exception:
    pass  # do something
```

but `log()` does a fair number of things, any of which could raise an exception, all of which would be caught by `except Exception`, and so it becomes difficult to write code to specifically catch the JSON error here. We should try to `raise` as specific of an exception type as possible ([`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError) seems to me the most accurate) or just allow `json.dumps()` to raise its error unhandled.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def log(key: str, value: Any) -> None:
+    """
+    Updates current logging context dict with provided key and value.
+    For use within a model's ``predict()`` method to collect logging
```
liuverta: This might not be exactly accurate, but we could use

```rst
:meth:`~verta.registry.VertaModelBase.predict`
```

instead of

```rst
``predict()``
```

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+    context, which is stored in thread-local variables.
```
liuverta: I'd recommend omitting this part about thread-local variables since [they are a magic](https://vertaai.slack.com/archives/C01RN3BFUMD/p1673978965808889?thread_ts=1673978921.387789&cid=C01RN3BFUMD) implementation detail that an end-user has no need to know about, and is more likely to confuse than inform.

If you feel it may be pertinent, we could simply mention here that `context`s are thread-safe, and further details about thread locality can be placed in [a `Notes` section](https://numpydoc.readthedocs.io/en/latest/format.html#notes) of the docstring.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+        validate_json(value)
+    local_context: Dict[str, Any] = _get_thread_logs()
+    local_context.update({key: value})
+    _set_thread_logs(local_context)
+
+
+class context:
+    """
+    Context manager for aggregating key-value pairs into a custom log entry.
+    Uses thread-local variables to store context in a thread-safe manner.
```
liuverta: See above comment about omitting mention of "thread-local variables".

We perhaps _should_ mention here that when a model is deployed, Verta automatically sets up a `context` to capture logs for each prediction, _and_ that a user can use `context` themselves for testing and validation (and include an `Examples` section with some snippets from our design doc).

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+    ----------
+    validate : Optional[bool]
+        If true, each individual call to ``runtime.log('key', value)`` will
+        verify that the value provided is JSON serializable.
+
+    Attributes
+    ----------
+    logs : Dict[str, Any]
+        Dictionary of the current logging context.  If called after exiting
+        the context manager, the final complete log entry is returned.
```
liuverta: This might be good to remove entirely, now that we're using a function rather than a property (it's no longer accurately an "attribute").

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+                more_logs = ['a', 'list', 'of', 'things']
+                runtime.log('things', more_logs)
+                return prediction_output
+
+    """
+    if _get_validate_flag():
+        validate_json(value)
+    local_context: Dict[str, Any] = _get_thread_logs()
+    local_context.update({key: value})
+    _set_thread_logs(local_context)
```
liuverta: I'm not sure that it's necessary to set thread logs here. `local_context` is already _the actual_ `_THREAD.logs` object, so `local_context.update()` itself has the effect of adding the new item.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+    ----------
+    logs : Dict[str, Any]
+        Dictionary of the current logging context.  If called after exiting
+        the context manager, the final complete log entry is returned.
+    """
+    def __init__(self, validate: Optional[bool] = False):
+        self.validate = validate
+        self.logs_dict = dict()
+
+    def __enter__(self):
```
liuverta, line 209: If (and only if) it's straightforward to do, I think we should raise an exception if there's already an active context, just to cover our bases. Otherwise—as a worst case—`_set_thread_logs()` would wipe user information without warning or recourse.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+        Ensure an empty logging context to start and set validation flag.
+        """
+        _set_thread_logs(dict())
+        if self.validate:
+            _set_validate_flag(True)
+        return self
+
+    def __exit__(self, *args):
+        """
+        Capture the final complete log entry in a class variable, ensure an
```
```suggestion
        Capture the final complete log entry in an instance variable, ensure an
```

Technically [an instance variable rather than a class variable](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables) 😄

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Attributes
+    ----------
+    logs : Dict[str, Any]
+        Dictionary of the current logging context.  If called after exiting
+        the context manager, the final complete log entry is returned.
+    """
+    def __init__(self, validate: Optional[bool] = False):
+        self.validate = validate
+        self.logs_dict = dict()
```
liuverta: I'd recommend marking these as private to discourage enterprising users/developers from tinkering (or at least absolve us of responsibility if things break).

```suggestion
        self._validate = validate
        self._logs_dict = dict()
```

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Return the currently stored, thread-local logs from inside this
+        context manager. JSON serialization is attempted on the whole
+        dictionary and an error is raised on failure.
+
+        If called after exiting the context manager, the final complete
+        log entry is returned.
+        """
+        logs: Dict[str, Any] = self.logs_dict or _get_thread_logs()
+        validate_json(logs)
```
liuverta: Would you mind removing the validation step here? It does make sense on its own, but the Python Model Service will already propagate an error if the final logs can't be JSONified, and having this here as a redundancy can be costly for performance if the logs are large.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+    --------
+    .. code-block:: python
+       :emphasize-lines: 10,13
+
+        from verta import runtime
+
+        # Sample model code:
+        class MyModel(VertaModelBase):
+            def __init__(self, artifacts):
+                "ok"
```
```suggestion
                pass
```

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+
+            def predict(self, data):
```
liuverta: I know this is a bit irrelevant, but I'd like to help ingrain it as a habit for everyone 😬 

```suggestion
            @verify_io
            def predict(self, data):
```

liuverta (changes requested):
Another round of nitpicky stuff. Logic overall makes sense! Were you planning to push unit tests within this PR as well?

client/verta/verta/runtime.py
liuverta: That's a fair (and cool) idea!

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/verta/runtime.py
ewagner-verta: Yeah, this was copy-pasted from an example notebook.  `pass` would be more pythonic.  Done

client/verta/verta/runtime.py
ewagner-verta: Good call.

client/verta/verta/runtime.py
ewagner-verta: Done.  I'm curious what you think about including `JSONDecodeError` as an exception that can be raised.  It seems unlikely that users will ever encounter anything but a `TypeError` if they pass in something invalid, but it is technically possible.

client/verta/verta/runtime.py
ewagner-verta: No objections.

client/verta/verta/runtime.py
ewagner-verta: You are absolutely correct.  It's an inlace update, so the call to `_set..` is superfluous.  Fixed

client/verta/verta/runtime.py
liuverta: I believe `JSONDecodeError`s only occur on `json.load()`, so we wouldn't need to worry about it!

client/verta/verta/runtime.py
ewagner-verta: Fixed

client/verta/verta/runtime.py
ewagner-verta: Done.  Let me know if you think it is too wordy.  We could always link to an outside knowledge doc later if we need more words.

client/verta/verta/runtime.py
ewagner-verta: Agreed.  Done.

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/verta/runtime.py
ewagner-verta: Let's think about this.  Logging context is cleaned up when entering and when exiting the context manager, so as long as calls to `runtime.log()` are made within the user's prediction function, there should be no issue.  If a user calls `runtime.log()` in the model code, but outside of the `predict()` function, then those logs would indeed disappear as soon as the predict function runs.  It should be pretty straightforward to throw an exception with a custom message if the logs are not empty when entering the context.

client/verta/docs/python.rst
ewagner-verta: Done, and thank you.

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/verta/runtime.py
ewagner-verta: I thought the `from` clause there would keep the exception intact, but it may not have actually been working that way.  In any case, I already refactored this to raise the original exception for the reasons you stated.

client/verta/verta/runtime.py
ewagner-verta: I share your concerns about performance, especially now that we are adding key validation too.  Perhaps we could just make it contingent on the `validate` argument?  That way it would still be useful as a check if users open their own context managers (for sending logs to a different service, for example).

client/verta/verta/runtime.py
ewagner-verta: Fixed

client/verta/verta/runtime.py
liuverta: That's a fair proposal, but after thinking about this some more I've realized that it might not be a good idea to have a check here at all.

Here's one way to look at this: If the user set `validate=True`, then we are checking each value for JSON serializability in `runtime.log()`, right? _Therefore_ the items here in `logs` would already be guaranteed to be JSON-serializable: We've already checked the values, and the keys already have to be simple strings. So, if `validate=True`, this check here is a performance hit with no additional value: Everything has already been validated.

And furthermore—commenting on performance—this check actually ends up being a marginally _worse_ equivalent to simply checking in `runtime.log()` (it has to process the full `logs` in memory, plus it has to check the keys). Which is another reason not to do it by default: It's a slower version of the thing we already decided against for being too slow.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,211 @@
...
+    None
+
+    Raises
+    -------
+    TypeError
+        If `validate` was set to ``True`` on the active :class:`context`, and
+        `value` is not a JSON-serializable type.
+    JSONDecodeError
+        If `validate` was set to ``True`` on the active :class:`context`, and
+        `value` is not JSON-serializable.
```
liuverta: I actually wouldn't expect to encounter `JSONDecodeError`s (https://github.com/VertaAI/modeldb/pull/3500#discussion_r1081979078) 🤔 Did you see any while testing/iterating?

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,211 @@
...
+    Updates current logging context dict with provided key and value.
+    For use within a model's :meth:`~verta.registry.VertaModelBase.predict`
+    method to collect logging context.
+
+    Parameters
+    ----------
+    key : str
+        String value to use as data label (key) in logging context.
+    value : Any
+        Any JSON serializable value you wish to include in the context logs.
```
liuverta: Is this too paranoid of me? 😆 

```suggestion
        Any `JSON serializable <https://docs.python.org/3/library/json.html#json.JSONEncoder>`__
        value you wish to include in the context logs.
```

client/verta/verta/runtime.py
ewagner-verta: Changed my mind and just dropped it altogether.  If `validate` is set to `True`, every key and value has been validated already when added.  If `validate` is set to `False`, the PyMS will through the `TypeError` back down in the 400 response if bad values were given.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,227 @@
...
+"""
+
+import json
+import re
+import threading
+from typing import Any, Dict, Optional
+
+
+_THREAD = threading.local()
+_S3_REGEX = re.compile("^[a-zA-Z0-9!_.*'()-]+(/[a-zA-Z0-9!_.*'()-]+)*$")
```
ewagner-verta: [Explanation of regex pattern](https://regex101.com/r/EFTruv/1) which conforms to the "Safe characters" described by AWS in their [object key naming doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html)

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/verta/runtime.py
ewagner-verta: I've been considering dropping this after trying to write a test for it.  It turns out it's really hard to trigger that error from `json.dumps()`.  I think you would have to read in json data from a file with malformed JSON or corrupted somehow.  It's technically possible, but I don't think we need to catch it.  It they manage to pull it off somehow, we can let the error bubble up naturally and feel just fine about it :)

liuverta: Well, that is sort of what I'm asking about. You mention that "it's technically possible", but I don't actually think that it is. `dumps()` _encodes_ JSON, so I'm having trouble imaging how we'd encounter a decode error, and `json`'s documentation only mentions `JSONDecodeError` in its load methods.

So I was curious what motivated its inclusion here in the first place. Did you truly encounter any `JSONDecodeError` while working on this PR?

ewagner-verta: I think I just did some digging on what other sorts of exceptions could be encountered and stumbled across examples like [this](https://stackoverflow.com/questions/68000458/json-dump-lead-to-json-decoder-jsondecodeerror-when-json-loadf) and [this](https://stackoverflow.com/questions/55093573/json-loadsstring-throws-json-decoder-jsondecodeerror-when-trying-to-convert-st).  In any case, I don't think catching such a fringe exception makes sense.  I've already dropped all the references to it.

client/verta/verta/runtime.py
ewagner-verta: nope, I like it.  Done

client/verta/verta/runtime.py
ewagner-verta: Simplified already.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,220 @@
...
+def _validate_s3(
+        value: str,
+        pattern: re.Pattern = _S3_REGEX
+    ) -> None:
+    """
+    Check the provided value for any special characters that would
+    violate Amazon S3 object key naming rules. Raise ValueError if any
+    are found.
+    """
+    if len(value) > 100:
```
ewagner-verta, line 86: I used `timeit` to measure the execution time of Python's native `len()` function versus adding the length check to the regex like `[0-9a-zA-Z_-]{1,100}$`.  The difference was very negligible and both are pretty fast, so I went with the Python length check so I could throw a more informative error message back to the user in that case.

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,172 @@
...
+    100 threads are completed by 5 workers, which each thread verifying
+    that it only holds it own thread-local value for logs.
+    """
+    def log_in_thread(value):
+        runtime.log('test_key', value)
+        assert runtime._get_thread_logs() == {'test_key': value}
+
+    with futures.ThreadPoolExecutor(max_workers=5) as executor:
+        with runtime.context():
+            executor.map(log_in_thread, [{f'thread_{x}_key': f'thread_{x}_val'} for x in range(101)])
```
liuverta: `log_in_thread()` isn't getting called here. `executor.map()` is lazy: Nothing actually gets evaluated until we actually iterate through the return value from `executor.map()`.

This can be verified by putting an `assert False` in `log_in_thread` and seeing whether `test_thread_safety` passes.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,172 @@
...
+"""
+
+from concurrent import futures
+import unittest
+import pytest
+from typing import Any, Dict
+from verta import runtime
+
+
+def test_thread_safety() -> None:
```
liuverta: I think the true test we need here is whether `ctx.logs()` returns the logs after everything is done, and I'm not sure that it will with our current implementation because it can't read the logs from child threads 🤔

liuverta (changes requested):

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Ok, now I understand why many of the examples I found were wrapping the `executor.map()` statement in a `list()` or `print()`.  I couldn't understand why, so I stripped that out on one of the final passes through.  I just wrapped it with list and now changing to `assert False` fails as expected.

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: I think perhaps we still want this to make sure the thread local variable for storing the logs is actually thread local and functions as expected.  However, we could write another test and push the test method called by each thread back one layer of scope, so to speak, so that each thread creates it's own context manager and verifies the content in `ctx.logs()` independently.  Is that kind of what you're thinking?

ewagner-verta: Ok, check out the new `test_thread_safe_context` function below.  I believe it does what you're thinking.  I tried adding a bunch of random sleeps throughout the inner `log_in_context` function to ensure randomized timing of each thread and even though it makes the test really slow, it still passes.

ewagner-verta: Ok, I was able to get the tests to work reliably.

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Fixed (wrapped with `list()`)

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    def __init__(self, validate: Optional[bool] = False):
+        self._validate = validate
+        self._logs_dict = dict()
+
+    def __enter__(self):
+        """
+        Ensure an empty logging context to start and set validation flag.
+        """
+        if bool(_get_thread_logs()):  # If not an empty dict.
```
liuverta: I thought we'd be getting rid of the `bool()` (https://github.com/VertaAI/modeldb/pull/3500#discussion_r1073823319) 🤔

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+            raise RuntimeError(
+                " cannot overwrite prior logging context. Please ensure all calls"
+                " to runtime.log() are made within the predict() method of your"
+                " model, or inside the scope of an instance of verta.runtime.context()."
```
liuverta: This error message seems strange to me. It's advising the user to put `runtime.log()` in the correct place, but this is more about the location of `with runtime.context():`, isn't it?

I think a more relevant message to provide here would be about not being allowed to create a new logging context when one is already active (e.g. nesting contexts within each other).

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Validate that the logging context being logged is thread local.
+    100 threads are completed by 5 workers, which each thread verifying
+    that it only holds it own thread-local value for logs.
+    """
+    def log_in_thread(value):
+        runtime.log('test_key', value)
+        time.sleep(random.uniform(0, 1))
```
liuverta: Does this actually do anything for the test? It would surprise me if a sleep affects logging/fetching logs at all.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+        assert runtime._get_thread_logs() == {'test_key': value}
+
+    with futures.ThreadPoolExecutor(max_workers=10) as executor:
+        list(
+            executor.map(
+                log_in_thread,
+                [{f'thread_{x}_key': f'thread_{x}_val'} for x in range(101)]
+            )
+        )
+    runtime._set_thread_logs({})  # clean-up logs left by test.
```
liuverta: We should ensure this line gets executed even if this test encounters an error/failure.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+class TestThreadLocalFunctions(unittest.TestCase):
+    """ Unit tests for the functions that handle thread local variables """
+
+    def test_init_thread_logs(self):
+        """
+        The logs stored in the thread local variable are blank on init.
+        """
+        local_dict: Dict[str, Any] = runtime._THREAD.__dict__
```
liuverta: What does this line do?

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+        assert runtime._get_thread_logs() == {}
+
+    def test_set_thread_logs(self):
+        runtime._set_thread_logs({'test_set': 'test_set'})
+        assert runtime._THREAD.logs ==  {'test_set': 'test_set'}
```
liuverta: Should this be cleaned up?

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def test_init_thread_validate(self):
+        """
+        The logs stored in the thread local variable are blank on init.
+        """
+        assert runtime._get_validate_flag() == False
+
+    def test_set_thread_logs(self):
+        runtime._set_validate_flag(True)
+        assert runtime._THREAD.validate == True
```
liuverta: Should this be cleaned up?

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+            )
+        )
+
+
+class TestThreadLocalFunctions(unittest.TestCase):
+    """ Unit tests for the functions that handle thread local variables """
+
+    def test_init_thread_logs(self):
+        """
+        The logs stored in the thread local variable are blank on init.
```
liuverta: This test (and `test_init_thread_validate`) both depend on prior code not touching `logs` & `validate`, which is not a good practice.

Relying on external state can introduce difficult-to-debug flakiness, and having the success/failure of a test depend on where it is written in its `.py` file (not to mention the possibility of _other_ test files elsewhere) makes the suite harder to maintain.

We could have these tests spawn new threads and perform the checks within them.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+    def test_final_log_entry_instance_variable(self):
+        """ The instance of the context manager maintains the final log entry in logs(). """
+        with runtime.context() as ctx:
+            runtime.log('fake_log', 'fake_value')
+        assert ctx.logs() == {'fake_log': 'fake_value'}  # outside the context
+
+    def test_validate_flag_true(self):
+        """ The validate argument triggers json validation as expected. """
+        with pytest.raises(TypeError):
+            with runtime.context(validate=True):
```
liuverta: `pytest.raises()` should be as specific as possible in its body. I don't think this test is intended to capture the possibility of `runtime.context(validate=True)` raising a `TypeError`.

```suggestion
        with runtime.context(validate=True):
            with pytest.raises(TypeError):
```

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+    def test_validate_flag_false(self):
+        """ The validate argument defaults to False and does not trigger json validation. """
+        with runtime.context():
+            runtime.log('obviously_not_jsonable', unittest.TestCase)
+
+    def test_validate_reset_on_exit(self):
+        """ Thread local var for validate is reset to default (False) after exiting context. """
+        with runtime.context(validate=True):
+            runtime.log('test', {'test_key': 'test_val'})
+        assert runtime._get_validate_flag() == False  # outside the context
```
liuverta: We're missing a test to check `runtime._get_validate_flag() == True` inside the context.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+class TestLog(unittest.TestCase):
+    """ Unit tests for the verta.runtime.log() function for logging prediction context """
+
+    def setUp(self) -> None:
+        """ Some useful logging bits """
+        self.log1 = dict(test_1_key=['list', 'of', 'things'])
+        self.log2 = dict(test_2_key=dict(this='that'))
+        self.log3 = dict(test1=self.log1, test2=self.log2)
```
liuverta: Out of curiosity, why are we using `dict(key=val)` here when we've been using `{"key": val}` everywhere else?

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,201 @@
...
+
+def test_exception_on_prior_logging_context() -> None:
+    """
+    A Runtime error is thrown if logging context is added outside the context
+    manager's scope that would otherwise be overwritten.
+    """
+    with pytest.raises(RuntimeError):
+        runtime.log('outside_of_scope', {'this': 'that'})
+        with runtime.context():
+            pass
```
liuverta: Which line, exactly, is meant to raise the `RuntimeError`?

`RuntimeError` is very broad so we should be as specific as possible.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        self._logs_dict = _get_thread_logs()
+        _set_thread_logs(dict())
+        _set_validate_flag(False)
+
+    def logs(self) -> Dict[str, Any]:
+        """
+        Return the currently stored, thread-local logs from inside this
+        context manager. If called after exiting the context manager, the
+        final complete log entry is returned.
```
liuverta: The "entry" in "final complete log entry" makes it sound like only the last call to `runtime.log()` is returned (each item being a single entry).

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+       :emphasize-lines: 4,8,11
+
+        import json
+        from verta import runtime
+
+        with runtime.context() as ctx:
+            runtime.log('labels', {'model_type': 'scikit_learn'})
+            output = 'this_is_output'
+            runtime.log('output', output)
+            print json.dumps(ctx.logs())
```
```suggestion
            print(json.dumps(ctx.logs()))
```

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+                more_logs = ['a', 'list', 'of', 'things']
+                runtime.log('things', more_logs)
+                return prediction_output
+
+    """
+    if _get_validate_flag():
+        _validate_json(value)
+    _validate_s3(key)
+    local_logs: Dict[str, Any] = _get_thread_logs()
+    local_logs.update({key: value})
```
liuverta, line 159: I mentioned this in another comment, but we should raise an error if `log()` is called outside of a `context`.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+        class MyModel(VertaModelBase):
+            def __init__(self, artifacts):
+                pass
+
+            @verify.io
+            def predict(self, data):
+                logs = {'some_stuff': 'I_care_about'}
+                runtime.log('my_logging_key', logs)
+                prediction_output = 'some_output'
+                more_logs = ['a', 'list', 'of', 'things']
```
liuverta: We should use values that are actually vaguely realistic. This looks unprofessional, or internal test suite-y.

Our design doc already has sample snippets we can use here.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+
+def _validate_json(value: Any) -> str:
+    """
+    Attempt to serialize the provided value to JSON.
+    Raise an error on failure.
+    """
+    try:
+        return json.dumps(value)
+    except TypeError as type_err:
+        print("provided data type cannot be converted to JSON for logging.")
```
liuverta: We shouldn't be `print`ing what appears to be an error message. It either goes in an exception object or it gets logged.

client/verta/verta/runtime.py
liuverta: To put it another way, if we wait until `with runtime.context():` to tell the user that they called `runtime.log()` in the wrong place, it's already too late! There are no public functions for them to access that log they made, nor to fix their thread state. We should raise that error in `runtime.log()` to prevent the problem from happening in the first place.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,237 @@
...
+    _validate_s3(key)
+    local_logs: Dict[str, Any] = _get_thread_logs()
+    local_logs.update({key: value})
+
+
+class context:
+    """
+    Context manager for aggregating key-value pairs into a custom log entry.
+    It should not be necessary to instantiate this class directly unless you
+    wish to export log entries manually to a resource other than Amazon S3.
```
liuverta: I wouldn't advise mentioning that at this time. It implicitly puts us on the hook to support users who wish to export to other object stores, which is not in-scope for this feature.

```suggestion
    It should not be necessary to instantiate this class directly.
```

liuverta: Our design doc also lists "user debugging their own model logging calls" as a use case.

liuverta (changes requested):

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: It does actually.  The test is somewhat flaky without it.  If each thread has exactly the same execution time (super fast in this case), they complete so quickly that they do not reliably collide and create a conflict, even though they _could_.  Adding a randomized sub-second delay here makes each thread's execution time slightly different, and thus guarantees a collision if the variable being tested is not thread safe.

client/verta/verta/runtime.py
ewagner-verta: Copy pasted the examples from the doc.

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Debugging line I forgot to remove.  -> Fixed

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Good call.  I converted this to a `unitest.TestCase` so we can rely on the `setUp` and `tearDown` functions to make sure the logs are cleaned up before and after the test.

client/verta/verta/runtime.py
ewagner-verta: Whoops.  Fixed.

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Done

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Done

client/verta/tests/unit_tests/test_runtime.py
liuverta: Hmm...that...doesn't make too much sense to me: That introducing randomness would help ensure collisions.

But I'll take your word for it. Would you mind adding a comment to the code about that?

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: They are essentially the same thing, and thus a matter of personal preference.  The rest of the repo just happens to contain yours.  In any case, I don't see a problem with standardizing on one vs the other so I just flipped it.

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: `with runtime.context():` opens the context manager and thus runs its `__enter__` method immediately, which is where the Runtime is raised, currently.  The idea there was to raise the error at the moment that logs would be overwritten.  I'll move the broader conversation about where to raise the Runtime (context() vs log()) in Slack.

client/verta/verta/runtime.py
ewagner-verta: Moving to a Slack thread for further discussion.

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Solved this with a pytest fixture to ensure clean state before and after tests

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Implemented

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Added.

client/verta/verta/runtime.py
ewagner-verta: Decided to drop the custom error message.  The TypeError is already informative enough on its own, I think.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,205 @@
...
+def test_thread_safe_context_manager() -> None:
+    """
+    Validate that multiple threads running a context manager simultaneously
+    will stay thread-safe. 50 threads are completed by 10 workers, with
+    each thread verifying that it only holds it own thread-local value for logs.
+    """
+    def log_in_context(value):
+        with runtime.context() as test_ctx:
+            runtime.log('test_key', value)
+            # time.sleep(random.uniform(0, 1))
```
liuverta: As I mentioned in [my previous comment](https://github.com/VertaAI/modeldb/pull/3500#discussion_r1086983418), would you please add a comment about what the intention of `time.sleep()` was, so that future readers know what to be aware of? Otherwise, it should be removed.

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,240 @@
...
+    """
+    _THREAD.logs = logs
+    return _THREAD.logs
+
+
+def _delete_thread_logs() -> None:
+    """
+    Drop the `log` attribute from the thread local variable.
+    """
+    _THREAD.__delattr__('logs')
```
liuverta: Accessing a dunder directly like this isn't...great? As I proposed [in Slack](https://vertaai.slack.com/archives/C01RN3BFUMD/p1674685341003589?thread_ts=1674684085.583239&cid=C01RN3BFUMD): `del _THREAD.logs` would be perfectly idiomatic.

Alternatively if we're using `hasattr(_THREAD, 'logs')` everywhere, we might as well use `delattr(_THREAD, 'logs')` for consistency.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,205 @@
...
+
+def test_set_thread_validate():
+    runtime._set_validate_flag(True)
+    assert runtime._THREAD.validate == True
+
+
+def test_get_thread_validate():
+    runtime._set_validate_flag(True)
+    assert runtime._get_validate_flag() == True
+    runtime._set_validate_flag(False)
```
liuverta: Why does one test reset


```python
runtime._set_validate_flag(False)
```

at the end but not the other? Maybe it should be in the cleanup fixture? And maybe it should be a `delattr()` instead?

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -0,0 +1,205 @@
...
+    with runtime.context():
+        runtime.log('obviously_not_jsonable', unittest.TestCase)
+
+
+def test_validate_reset_on_exit():
+    """ Thread local var for validate is reset to default (False) after exiting context. """
+    with runtime.context(validate=True):
+        runtime.log('test', {'test_key': 'test_val'})
+        assert runtime._get_validate_flag() == True  # inside the context
+    assert runtime._get_validate_flag() == False  # outside the context
```
liuverta: `context.__exit__()` deletes `_THREAD.validate`, but here `runtime._get_validate_flag()` sets `_THREAD.validate = False`. It seems like the implementation and this assertion are at odds with each other.

What is the state that we want when a context is exited? That `_THREAD.validate` doesn't exist, or that `_THREAD.validate` is `False`?

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,240 @@
...
+                pass
+
+            @verify.io
+            def predict(self, x):
+                embeddings = self.get_embeddings(x)
+                return self.nn(embeddings)
+
+            def get_embeddings(x):
+                embedding = self.embedding[x]
+                client.runtime.log("embedding", embedding)
```
liuverta: To match the `from verta import runtime` above:

```suggestion
                runtime.log("embedding", embedding)
```

client/verta/verta/runtime.py
```diff
@@ -0,0 +1,240 @@
...
+    validate : bool, default False
+        If true, each individual call to :meth:`~verta.runtime.log`  will
+        verify that the value provided is JSON serializable.
+
+    Raises
+    ------
+    RuntimeError
+        If logs has been added via :meth:`~verta.runtime.log()` outside the
+        scope of the model's :meth:`~verta.registry.VertaModelBase.predict`
+        method, or outside the scope of this context manager.
```
liuverta: Something like

```suggestion
        If this context manager is being nested inside an existing instance.
```

liuverta (changes requested):

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -22,6 +22,8 @@ def clean_thread():
     yield
     if hasattr(runtime._THREAD, 'logs'):
         runtime._THREAD.__delattr__('logs')  # clean-up any existing logs.
+    if hasattr(runtime._THREAD, 'validate'):
+        runtime._THREAD.__delattr__('validate')  # clean-up any existing flag.
```
liuverta: Like I mentioned in [my other comment](https://github.com/VertaAI/modeldb/pull/3500#discussion_r1089220329), I'd recommend sticking with `delattr(runtime._THREAD, 'logs')` rather than directly calling a dunder method.

client/verta/verta/runtime.py
liuverta: It doesn't appear that this was resolved.

liuverta (approved):

PR #3505, opened 2023-01-13
refactor: Deprecate ModelDBException constructors in favor of strong types

conradoverta:
## Impact and Context
Instead of passing `Code.INTERNAL` to an argument, we should use the strong type `InternalErrorException`. This makes it easier to catch just some specific scenarios.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/exceptions/ModelDBException.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/exceptions/ModelDBException.java
@@ -12,11 +12,13 @@ public class ModelDBException extends RuntimeException {
 
   private final Code code;
 
+  @Deprecated(forRemoval = true)
   public ModelDBException() {
     super();
     code = Code.INTERNAL;
   }
 
+  @Deprecated(forRemoval = true)
   public ModelDBException(String message) {
     super(message);
     code = Code.INTERNAL;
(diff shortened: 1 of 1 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/exceptions/ModelDBException.java
```diff
@@ -27,21 +29,25 @@ public ModelDBException(String message, Throwable cause) {
...
+  @Deprecated(forRemoval = true)
   public ModelDBException(String message, com.google.rpc.Code code, Throwable cause) {
     super(message, cause);
     this.code = Code.valueOf(code.name());
   }
 
+  @Deprecated(forRemoval = true)
   public ModelDBException(String message, com.google.rpc.Code code) {
     super(message);
     this.code = Code.valueOf(code.name());
```
anandjakhaniya, line 53: @conradoverta I think this will be needed because we are throwing some GRPC code like Code.ALREADY_EXISTS, Code.NOT_FOUND etc. and we are managing it on global exception. So if we deprecate it then what is the alternative of the same?

conradoverta: I'm not sure I follow. We should be able to create the type-specific exceptions, and have a static method to convert a grpc error to the correct exception type.

jkwatson-verta:
TBH, I didn't know the subtypes even existed! Yeah, I don't know if we need all of the constructors if we have the subtypes. 

If it had been me, I might have had static methods to create the various codes (like `ModelDbException.invalidArgument(String message)`) but Java likes it's exception hierarchies. These seem fine, and redundant constructors are redundant.

jkwatson-verta:
We'll want to make at least one of the constructors `protected` for the subtypes access, but this is a good first step to that.

backend/common/src/main/java/ai/verta/modeldb/common/exceptions/ModelDBException.java
anandjakhaniya: Ok then it makes sense, So you mean Instead of a common ModelDB exception, we have multiple exceptions for each type? right?

jkwatson-verta: We already have them, yes. :)

conradoverta:
One of the things I've learned is that you should use strong types for exceptions so that you can catch just want you want (e.g. handle not found and transform into optional) to avoid catch-check-rethrow which is harder to follow.

jkwatson-verta (approved):

PR #3510, opened 2023-01-13
chore: capture the callsite on ListenableFuture wrapping 

jkwatson-verta:
## Impact and Context
This adds the callsite as an suppressed Exception to whatever causes the ListenableFuture to fail.

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
@@ -163,7 +163,7 @@ public class CommonUtils {
                 .build();
       } else {
         LOGGER.error(
-            "Stacktrace with {} elements for {} {}", stack.length, e.getClass(), e.getMessage());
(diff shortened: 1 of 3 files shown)


conradoverta (commented):
Should we collect the stack traces as we create futures or just the last one? Bc we won't have the full path with the last one only 🤔

jkwatson-verta:
> Should we collect the stack traces as we create futures or just the last one? Bc we won't have the full path with the last one only 🤔

This particular change was just for ListenableFuture based futures. I'm not opposed to adding more stack capture (IIRC you can turn this one via an environment variable `FUTURE_TESTING_STACKS_ENABLED` already), but wanted to address a small thing first before diving in whole-hog. Perhaps this should also be behind the same env var?

jkwatson-verta:
closing. can re-open later if we decide we need it

PR #3520, opened 2023-01-19
refactor: Expand getInvalidFieldException from registry

conradoverta:
## Impact and Context
The equivalent method in the registry (which is being removed in favor of this) had more logic to handle other types of exceptions. So we just move the logic here.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


jkwatson-verta (approved):

jkwatson-verta:
!rebuild

PR #3550, opened 2023-01-31
feature: [VRD-662] Throw ValueError on attempts to overwrite existing logs

ewagner-verta:
## Impact and Context
Adding this exception will protect the user from unknowingly overwriting previous logs when calling `runtime.log(key, value)` with a key that already exists on the logs for that same prediction.  This explicitly prevents confusion around the difference between our context manager and other context-loggers that write immediately on each call to "log".

## Risks and Area of Effect
Low risk, as this only affects the new runtime functionality (not yet released, officially)

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


client/verta/verta/runtime.py
```diff
@@ -156,6 +160,8 @@ def get_embeddings(x):
         _validate_json(value)
     _validate_s3(key)
     local_logs: Dict[str, Any] = _get_thread_logs()
+    if key in local_logs.keys():
+        raise KeyError(f" cannot overwrite existing value for \"{key}\"")
```
liuverta: This might be a hot take, but I think this should be a `ValueError`.

Python's documentation describes `KeyError` exclusively as a lookup error (in fact, its base class is `LookupError`), rather than an assignment error:

> _exception_ **KeyError**
> Raised when a mapping (dictionary) key is not found in the set of existing keys.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -202,4 +202,14 @@ def test_s3_validation_good_keys(key: str):
...
-
+def test_overwrite_existing_key_throws_error() -> None:
+    """
+    If an attempt is made to write to the same key twice within a single
+    context, a KeyError exception is raised.
+    """
+    with pytest.raises(KeyError)as err:
+        with runtime.context() as ctx:
+            runtime.log('key_123', 'value_1')
+            runtime.log('key_123', 'value_2')
```
liuverta: As I've mentioned in a few comments before already, I'd encourage making the `with pytest.raises(...):` body _as specific as possible_ in what it contains.

```suggestion
    with runtime.context() as ctx:
        runtime.log('key_123', 'value_1')
        with pytest.raises(KeyError)as err:
            runtime.log('key_123', 'value_2')
```

liuverta (changes requested):

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Done.

client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -202,4 +202,14 @@ def test_s3_validation_good_keys(key: str):
...
-
+def test_overwrite_existing_key_throws_error() -> None:
+    """
+    If an attempt is made to write to the same key twice within a single
+    context, a KeyError exception is raised.
+    """
+    with runtime.context() as ctx:
+        with pytest.raises(ValueError)as err:
+                runtime.log('key_123', 'value_1')
+                runtime.log('key_123', 'value_2')
```
liuverta: As I mentioned in [my previous comment](https://github.com/VertaAI/modeldb/pull/3550#discussion_r1092569335), this should be

> _as specific as possible_

This isn't the case in your change here, because line `212` isn't expected to throw an error. Could we just use the suggestion from my comment?

Also there's extra unnecessary indentation here.

client/verta/verta/runtime.py
```diff
@@ -94,11 +94,13 @@ def log(key: str, value: Any) -> None:
...
-    method to collect logs.
+    method to collect logs.  All logs passed to this function are aggregated into a
+    single dictionary within the context of one prediction, which is processed into
+    storage after the prediction result is returned, to minimize added latency.
 
     .. note::
-        Multithreading of calls to log() within a model's predict() method is not
-        currently supported.
+        Existing keys cannot be overwritten.  Multithreading of calls to log() within a
+        model's predict() method is notcurrently supported.
```
```suggestion
        Existing keys cannot be overwritten. Multithreading of calls to log() within a
        model's predict() method is not currently supported.
```

client/verta/verta/runtime.py
```diff
@@ -118,8 +120,9 @@ def log(key: str, value: Any) -> None:
         If `validate` was set to ``True`` on the active :class:`context`, and
         `value` is not a JSON-serializable type.
     ValueError
-        If `key` provided contains non-alphanumeric characters.  Dashes `-`
-        and underscores `_` are permitted.
+        If `key` provided contains non-alphanumeric characters (Dashes `-`
+        and underscores `_` are permitted) or the key already exists and the
```
liuverta: A comma here will help a bit with clarity.

```suggestion
        and underscores `_` are permitted), or the key already exists and the
```

liuverta (changes requested):

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Done.

liuverta (approved):

PR #3552, opened 2023-02-01
[FE-641] Fix fuzzy search using 'user_id' which doesn't exists

anandjakhaniya:
## Impact and Context
There was a bug when the user try to search resources based on owners which don't exist then the backend was throwing an `Internal Server` error instead of returning an empty resource list and it was a bug that I have fixed but that logic and code in a common module so I have created common module changes PR here with test cases.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/authservice/UACApisUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/authservice/UACApisUtil.java
@@ -387,7 +387,7 @@ public class UACApisUtil {
                             executor)
                         .thenCompose(
                             resourceIds -> {
-                              if (resourceIds.isEmpty()){
+                              if (resourceIds.isEmpty()) {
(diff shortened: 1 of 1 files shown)


AndreyKuznetsovQS (approved):

anandjakhaniya:
@eheinlein-verta

eheinlein-verta (approved):
It would be great to have a test for this, but it's not a blocker. You can create it as part of this PR or in a follow-up.

anandjakhaniya:
> It would be great to have a test for this, but it's not a blocker. You can create it as part of this PR or in a follow-up.

Some tests are already merged in the main for this and others are in separate PR at https://github.com/VertaAI/registry/pull/1883 which has actual requirements so might be all tests are already covered.

PR #3558, opened 2023-02-01
test: Fix h5py constraint

liuverta:
## Impact and Context

Fix a dependency issue blocking our test suite. See **Testing** **Before**.

## Risks and Area of Effect

—

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

I ran `deployable_entity/test_artifacts.py::TestModels::test_keras` in our CI `pytest` job.

### Before

It actually fails during setup at

```sh
python -m spacy download en_core_web_sm
```

trying to import `h5py`, encountering

```
AttributeError: module 'numpy' has no attribute 'typeDict'
```

### After

Jobs succeed (build number 15, in our CI) all the way through in every Python version (except 3.11, because we don't support that yet).

## Reverting

Revert this PR.


liuverta:
The linting error is unrelated to this PR, and will be fixed in #3557

ewagner-verta (approved):

PR #3569, opened 2023-02-03
fix: [VRD-620] Fix False Runtime Error on Nested Contexts

ewagner-verta:
## Impact and Context
Calling predict rapidly within a loop (as in many of our example notebooks) can cause a RuntimeError related to the `runtime.context()` class.  This minor refactor fixes that problem.

## Risks and Area of Effect
This is fairly low risk, and it fixes a bug currently in `main`. 

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/client/verta/tests/unit_tests/test_runtime.py
+++ b/client/verta/tests/unit_tests/test_runtime.py
@@ -77,8 +77,9 @@ def test_logs_are_clean_on_entry():
 
 def test_logs_attribute_removed_on_exit():
     """ Thread local var for logs is removed after exiting the context. """
-    with runtime.context():
+    with runtime.context() as ctx:
         runtime.log('fake_log', 'fake_value')
+        assert ctx.logs()
     assert not hasattr(runtime._THREAD, 'logs')  # outside the context
 
 
--- a/client/verta/tests/unit_tests/test_runtime.py
+++ b/client/verta/tests/unit_tests/test_runtime.py
@@ -49,13 +49,6 @@ def test_thread_safe_context_manager() -> None:
         )
 
 
-def test_init_thread_logs():
-    """
-    An empty dict is initialized by default.
-    """
-    assert runtime._get_thread_logs() == {}
-
-
 def test_set_thread_logs():
     runtime._set_thread_logs({'test_set': 'test_set'})
     assert runtime._THREAD.logs ==  {'test_set': 'test_set'}
@@ -64,13 +57,6 @@ def test_set_thread_logs():
 def test_get_thread_logs():
     runtime._set_thread_logs({'test_get': 'test_get'})
     assert runtime._get_thread_logs() ==  {'test_get': 'test_get'}
-
-
-def test_init_thread_validate():
-    """
-    A boolean defaulting to False is initialized by default.
-    """
-    assert runtime._get_validate_flag() == False
 
 
 def test_set_thread_validate():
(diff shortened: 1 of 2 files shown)


client/verta/tests/unit_tests/test_runtime.py
```diff
@@ -89,15 +75,16 @@ def test_logs_are_clean_on_entry():
...
 
-def test_logs_are_clean_on_exit():
-    """ Thread local var for logs is a blank dict after exiting context. """
-    with runtime.context() as ctx:
+def test_logs_attribute_removed_on_exit():
+    """ Thread local var for logs is removed after exiting the context. """
+    with runtime.context():
         runtime.log('fake_log', 'fake_value')
-    assert runtime._get_thread_logs() == {}  # outside the context
+    assert not hasattr(runtime._THREAD, 'logs')  # outside the context
```
liuverta, line 83: Optionally, we might want to add a

```python
ctx.logs()
```

before the `hasattr` check, since that was the exact issue we were seeing. But probably not necessary, now that we've removed the behavior altogether.

client/verta/verta/runtime.py
```diff
@@ -16,10 +16,8 @@
 
 def _get_thread_logs() -> Dict[str, Any]:
     """
-    Return the current thread-local logs or initialize an empty dict.
+    Return the 'logs' attribute of the thread-local variable.
     """
-    if not hasattr(_THREAD, 'logs'):
-        _set_thread_logs(dict())
```
liuverta, line 22: We still need this `if`, maybe just to `return {}` or `return None`.

Otherwise PyMS will encounter an error if nothing was logged

```python
from verta import runtime

with runtime.context() as ctx:
    pass
ctx.logs()
```

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [1], line 6
      3 with runtime.context() as ctx:
      4     pass
----> 6 ctx.logs()

File ~/Documents/modeldb/client/verta/verta/runtime.py:242, in context.logs(self)
    231 def logs(self) -> Dict[str, Any]:
    232     """
    233     Return the currently stored, thread-local logs from inside this
    234     context manager. If called after exiting the context manager, the
   (...)
    240         Dictionary of logs collected within this context manager.
    241     """
--> 242     return  self._logs_dict or _get_thread_logs()

File ~/Documents/modeldb/client/verta/verta/runtime.py:21, in _get_thread_logs()
     17 def _get_thread_logs() -> Dict[str, Any]:
     18     """
     19     Return the 'logs' attribute of the thread-local variable.
     20     """
---> 21     return _THREAD.logs

AttributeError: '_thread._local' object has no attribute 'logs'
```

liuverta: ☝️ That might also be a good test to add.

client/verta/verta/runtime.py
```diff
@@ -40,11 +38,8 @@ def _delete_thread_logs() -> None:
 
 def _get_validate_flag() -> bool:
     """
-    Return the current thread-local variable for validate or initialize a
-    new boolean.
+    Return the 'validate' attribute of the thread local variable.
     """
-    if not hasattr(_THREAD, 'validate'):
-        _set_validate_flag(False)
```
liuverta, line 47: Similar to my comment above: let's keep the check for `_THREAD.validate` as a safeguard.

client/verta/verta/runtime.py
```diff
@@ -209,7 +204,7 @@ class context:
     """
     def __init__(self, validate: Optional[bool] = False):
         self._validate = validate
-        self._logs_dict = dict()
+        self._logs_dict = {}
```
liuverta: Since you mentioned preferring `dict()` elsewhere before, I think:

I actually prefer using `dict()`, myself, if we're only creating an empty dictionary. To me personally, `{}` looks confusing because I forget whether it's an empty `dict` or an empty `set`.

liuverta (changes requested):

client/verta/verta/runtime.py
ewagner-verta: That's funny.  I only changed it because I noticed that pylint has started complaining about it, but I prefer `dict()` for exactly that reason.  Swapped it back

client/verta/verta/runtime.py
ewagner-verta: Done

client/verta/verta/runtime.py
ewagner-verta: Added this test.

client/verta/tests/unit_tests/test_runtime.py
ewagner-verta: Good call.  Verifying that logs are there before we verify that they get removed makes the test a bit more complete.  >> Done

liuverta: I actually meant a `ctx.logs()` _outside_ the context, since the original bug was that calling `ctx.logs()` after `context.__exit__()`! But it's still probably fine either way 😆

liuverta (approved):

PR #3604, opened 2023-02-15
fix: [VUMM-877] Make sure a failed migration results in the VM exiting

jkwatson-verta:
## Impact and Context
We were seeing cases where the Migration work was failing, because the database server was not initialized yet. In those cases, the migration container would not exit properly due to non-daemon threads being started for "cronjob" type tasks in the application.

In addition to fixing the underlying issue, this PR also adds logging to the database connection initialization to make it easier to debug when things are going wrong.

## Testing
- [x] Unit test
- [x] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
@@ -77,7 +77,7 @@ public abstract class CommonDBUtil {
   protected static boolean checkDBConnection(RdbConfig rdb, Integer timeout) {
     LOGGER.info("Checking DB connection with timeout [" + timeout + "]...");
     try (var con = getDBConnection(rdb)) {
-      LOGGER.info("Got connection: " + con);
+      LOGGER.debug("Got connection: " + con);
       boolean valid = con.isValid(timeout);
       LOGGER.info("DB Connection valid? " + valid);
       return valid;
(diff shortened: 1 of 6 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
```diff
@@ -75,8 +75,12 @@ protected static Connection getDBConnection(RdbConfig rdb) throws SQLException {
   }
 
   protected static boolean checkDBConnection(RdbConfig rdb, Integer timeout) {
+    LOGGER.info("Checking DB connection with timeout [" + timeout + "]...");
     try (var con = getDBConnection(rdb)) {
-      return con.isValid(timeout);
+      LOGGER.info("Got connection: " + con);
```
eheinlein-verta: This seems at a glance like it would be better as DEBUG level, but I don't feel super strongly about it.

eheinlein-verta (approved):
A few of the log lines seem like they might be better at DEBUG level, but that's my only nit pick.

backend/common/src/main/java/ai/verta/modeldb/common/CommonDBUtil.java
jkwatson-verta: yeah, probably a good idea. I think this happens only at startup time, so I don't know that it matters all that much, but I'm happy to downgrade it to debug. 👍🏼

jkwatson-verta:
```suggestion
      LOGGER.debug("Got connection: " + con);
```

PR #3611, opened 2023-02-17
feat: [VRD-692] Validate model_test() in create_standard_model()

liuverta:
## Impact and Context

In `RegisteredModel.create_standard_model()`, validate that the user-defined `model_test()` method doesn't have any additional parameters.

Also slightly changes the exception message for the `__init__()` validation.

## Risks and Area of Effect

Changing the `__init__()` validation's exception message is _technically_ an API breakage, but I'd consider it so unlikely for any user to be relying on this that it's fine.

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

I successfully ran `registry/test_standard_model.py::TestModelValidator` in our CI in all supported versions of Python.

## Reverting

Revert this PR.


client/verta/verta/_internal_utils/model_validator.py
```diff
@@ -81,11 +81,15 @@ def must_verta(model):
...
+    )
+    init_params = list(
+        inspect.signature(model.__init__).parameters.keys()
+    )
     if init_params != expected_init_params:
         raise TypeError(
-            "model __init__() parameters must be named {},"
+            "model __init__() parameters must be {},"
             " not {}".format(expected_init_params, init_params)
         )
```
liuverta, line 94: Changing this exception message is _technically_ an API breakage, but I'd consider it so unlikely for any user to be relying on this that it's fine.

client/verta/verta/_internal_utils/model_validator.py
```diff
@@ -95,4 +99,16 @@ def must_verta(model):
             " argument and return types may change unintuitively when deployed"
         )
 
+    expected_test_params = list(
+        inspect.signature(VertaModelBase.model_test).parameters.keys()
```
ewagner-verta, line 103: This would always return `['self']` right?  If the idea here to ensure that no arguments are passed to `model_test()` should we be more explicit in the error?  i.e. `model_test cannot accept arguments` type of message.

liuverta: This does return `['self']`, yes! And the implied idea _is_ to ensure no arguments are passed to `model_test()`.

But technically the core idea is to ensure that the user's implementation of `model_test()` has the same signature as the interface `VertaModelBase.model_test()` since Python itself doesn't have a way to enforce that, I don't think

liuverta: This does return `['self']`, yes! And the implied idea _is_ to ensure no arguments are passed to `model_test()`.

But technically the core idea is to ensure that the user's implementation of `model_test()` has the same signature as the interface `VertaModelBase.model_test()` since Python itself doesn't have a way to enforce that, I don't think.

liuverta: Hmm I see your point, but for me the more important thing is "implementation-wise, the function signature's parameters match" rather than "conceptually, `model_test()` doesn't accept arguments". And even that wording is confusing to me, since `self` is an argument.

To me, this error message is actually already as explicit as possible, in that it says exactly what parameters the method signature should have. 🤔

ewagner-verta: Granted, the user likely arrives at the same conclusion whether it is via `['self'] != ['self', 'thing_you_added']` or `don't use args with this function`.  It's the precise language of "parameters must be" within a message intended to convey "don't use parameters here" that arguably could lead to confusion.  Point taken about `self` _technically_ being a parameter.  Honestly, I don't feel strongly enough to fight for a change here.  I just wanted to point out how another reasonable person might interpret it.

ewagner-verta (approved):

client/verta/verta/_internal_utils/model_validator.py
liuverta: haha this was helpful to point out—I wouldn't have considered it myself. If users start bringing it up, I'll make the change.

PR #3625, opened 2023-02-24
test: [VRD-680] add test class for model dependency validation.

ewagner-verta:
## Impact and Context
Adds a new test class for use in model env/dependency validation.  The idea is to import a bunch of 3rd party dependencies and use them in a variety of ways within the class and this will give us a resource we can use to test the forthcoming logic to extract dependencies from model classes written by customers.

This class attempts to cover the following ways in which dependencies _could_ be introduced:
* Direct usage of a module within the body of a function
* Type hints in function arguments
* Type hints in return type annotation
* Type hints for function-scoped variables
* All of the above wrapped in `verify_io` and not wrapped in `@verify_io`
* Class functions that are wrapped as `@staticmethod` and `@propery`
* Type hints that are nested within type constructs like `Type[things.Thingy]` or `Union[things.Thingy, stuff.Whatever]`

_All dependencies used were pulled from existing client dependencies, and this PR does not introduce net new ones._

## Risks and Area of Effect
Very low.  This only introduces a new test class and it doesn't actually use it.  Subsequent PRs will add the tooling and tests.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [X] Other (explain) - This IS a testing component.

## Reverting

### DIFF
--- a/client/verta/tests/registry/conftest.py
+++ b/client/verta/tests/registry/conftest.py
@@ -1,8 +1,28 @@
 # -*- coding: utf-8 -*-
 
-import pytest
+import calendar
+import collections as collecs
+import datetime as dt
+import json
+from typing import Type, Union
 
-from verta.registry import DockerImage
+import boto3
+import click
+import cloudpickle as cp
+import numpy
+import pandas as pd
+import pytest
+import requests
+import sklearn
+import spacy
+import torch
+import urllib3
+import yaml
+from google.protobuf import message as msg
+from PIL import UnidentifiedImageError
+
+from verta import runtime
+from verta.registry import DockerImage, verify_io, VertaModelBase
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


client/verta/tests/registry/conftest.py
```diff
@@ -1,8 +1,28 @@
...
+import pandas as pd
 import pytest
+import requests
+import sklearn
+import spacy
+import torch
+import urllib3
+import yaml
+from google.protobuf import message as msg
+from PIL import UnidentifiedImageError
```
liuverta: I mentioned this in Slack, but just to codify it in a code review: We should use `pytest.importorskip()` for these rather than `import`. Otherwise `pytest` will fail at collection time due to non-installed libraries.

Something that might look a bit hacky but is acceptable imo for tests is to create a `dependency_testing_model()` fixture that calls `importorskip()`s, and defines (and returns) `class DependencyTestingModel` in its body.

liuverta (changes requested):

client/verta/tests/registry/conftest.py
ewagner-verta: Working on that right now actually.

ewagner-verta: Ok, let me know what you think of the latest implementation.

liuverta (approved):
Looks comprehensive! Should be good to start with.

PR #3632, opened 2023-02-27
feature: [VRD-681] Client - Model Dependency Extraction Utils

ewagner-verta:
## Impact and Context
Adds utility functions for extracting dependencies from a model class.

## Risks and Area of Effect
This should not affect any existing features, as this is all new code and it will be tested more thoroughly as the broader feature develops.

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,31 @@
...
+        pass
+        # unwrap here
+    return [
+        value for key, value in inspect.getclosurevars(func).globals.items()
+        if isinstance(value, ModuleType)
+    ]
+
+def is_wrapped(func: Callable, **kwargs) -> bool:
+    try:
+        return func.__wrapped__(**kwargs) is not None
```
liuverta: I _think_ we'd want to do a `hasattr(func, "__wrapped__")` instead. This looks like it'd actual call the underlying function, which might have undesired side effects.

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,54 @@
...
+        '__init__',
+        'predict',
+        'unwrapped_predict',
+        'make_dataframe',
+        'make_timeout',
+        'make_spacy_error',
+        'make_boto_session',
+        'nested_multiple_returns_hint',
+        'nested_type_hint'
+    ]
```
liuverta: There's probably pros and cons either way, but we might want to have this as an attribute on `DependencyTestingModel`. That way if we ever change the model code, we can update this list right there, instead of having to remember that it's defined in the test code.

Just as a thought—I'm not totally convinced on it, myself.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+import ast
+import inspect
+import re
+from importlib.metadata import packages_distributions
```
liuverta: Don't forget we'll need to add [`importlib-metadata`](https://pypi.org/project/importlib-metadata/) as [a client dependency](https://github.com/VertaAI/modeldb/blob/main/client/verta/requirements.txt). Looks like the version we need for `packages_distributions` is [`>=3.7.0`](https://github.com/python/importlib_metadata/blob/main/CHANGES.rst#v370)

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,26 @@
...
+def list_class_functions(model_class: Type[VertaModelBase]) -> List[Tuple[str, Callable]]:
+    """List all the functions present in the provided class object."""
+    return inspect.getmembers(model_class, predicate=inspect.isfunction)
+
+
+def list_modules_in_function_definition(func: Callable) -> List[ModuleType]:
+    """List all modules called within the body of the provided function."""
+    try:
+        _func = inspect.unwrap(func)  # returns last object in chain
+    except ValueError:  # if a cycle is encountered
```
liuverta, line 22: Just out of curiosity, did you encounter any particular cases with cycles while experimenting? I'm curious what libraries might have them‚ or if that even works in Python at all.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,26 @@
...
+
+def list_class_functions(model_class: Type[VertaModelBase]) -> List[Tuple[str, Callable]]:
+    """List all the functions present in the provided class object."""
+    return inspect.getmembers(model_class, predicate=inspect.isfunction)
+
+
+def list_modules_in_function_definition(func: Callable) -> List[ModuleType]:
+    """List all modules called within the body of the provided function."""
+    try:
+        _func = inspect.unwrap(func)  # returns last object in chain
```
liuverta: I know we've discussed not worrying too much about wrapper for now, but just to call it out explicitly: We _might_ want to inspect the wrapper `func` itself as well for modules used, rather than jumping straight to the innermost wrapped `_func`. Could be a follow-up for next sprint.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Agreed.  I'm actually dropping this out of this PR, since it is more related to the next batch of work anyway.  I put a comment on that Jira issue so we won't forget.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: We could theoretically iterate through each layer with `inspect.unwrap(stop=?)`, but I agree it would be best left to future improvements.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: I did not.  I'm basically just recycling [documentation](https://docs.python.org/3/library/inspect.html#inspect.unwrap) here.

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    extracted_func_names = [f[0] for f in md.list_class_functions(dependency_testing_model)]
+    assert extracted_func_names == dependency_testing_model.expected_function_names()
+
+
+def test_list_modules_in_function_definition_wrapped(dependency_testing_model) -> None:
+    """ Verify that modules used within a function definition are extracted
+    as expected, for a function that is not wrapped in verify_io """
+    func: Callable = dependency_testing_model.predict
+    expected_module_names = ['verta', 'yaml']
```
liuverta: Well...if we're still going to define `expected_module_names` in _some_ test functions—rather than defining all of them in the `DependencyTestingModel`—then we should move `DependencyTestingModel.expected_function_names` and `DependencyTestingModel.expected_modules` back from conftest.py into their respective test functions here.

We need to have all these "expected name lists" defined in a consistent way: either in the `class`, or in the test functions. Otherwise it's just confusing.

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,148 @@
...
+    """ Verify that all the functions in the test class are recognized and
+    returned.
+    """
+    extracted_func_names = [f[0] for f in md.list_class_functions(dependency_testing_model)]
+    assert extracted_func_names == dependency_testing_model.expected_function_names()
+
+
+def test_list_modules_in_function_definition_wrapped(dependency_testing_model) -> None:
+    """ Verify that modules used within a function definition are extracted
+    as expected, for a function that is not wrapped in verify_io """
```
liuverta:
```suggestion
    as expected, for a function that is wrapped in verify_io """
```

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,148 @@
...
+    func: Callable = dependency_testing_model.predict
+    expected_module_names = ['verta', 'yaml']
+    extracted_modules: List[ModuleType] = [
+        f.__package__ for f in md.list_modules_in_function_definition(func)
+    ]
```
liuverta: Note that `__package__` seems to only go one level up the module hierarchy, i.e. it doesn't return the root module name. For example, `verta._internal_utils.model_dependences.__package__` returns `verta._internal_utils` rather than `verta`. Is that the intended behavior?

liuverta: I'm also wondering if there's a reason why some tests are using `__package__` while others use `__name__`.

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,148 @@
...
+
+def test_list_modules_in_function_return_type_hint_unwrapped(dependency_testing_model) -> None:
+    """ Verify that modules used in function return type hints are extracted
+    as expected with no function wrappers.
+    """
+    func: Callable = dependency_testing_model.unwrapped_predict
+    extracted_modules: List[str] = [
+        f.__name__ for f in md.list_modules_in_function_signature(func)
+    ]
+    assert 'requests.exceptions' in extracted_modules
```
liuverta: I think we should add new functions to `DependencyTestingModel` specifically for these two tests that have the return type hint and nothing else. `in` is a pretty weak assertion to use in tests in general, and glancing at this code I'd have very low confidence that it wouldn't encounter false positives if someone in the future shuffled around the signatures.

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,148 @@
...
+    """ Verify that modules used in function return type hints are extracted
+    as expected when multiple return types are specified.
+    """
+    func: Callable = dependency_testing_model.nested_multiple_returns_hint
+    expected_module_names = ['urllib3.util.retry', 'PIL']
+    extracted_modules: List[str] = [
+        f.__name__ for f in md.list_modules_in_function_signature(func)
+    ]
+    for name in expected_module_names:
+        assert name in extracted_modules
```
liuverta: Same thing about the `in`s here as my above comment. I think `assert extracted_modules == expected_module_names` would make these tests much stronger.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,98 @@
...
+
+def unwrap(func: Callable) -> Callable:
+    """Unwraps a function to get the original function object in a non-blocking manner."""
+    try:
+        return inspect.unwrap(func)  # returns last object in chain
+    except ValueError:  # if a cycle is encountered
+        return func
+
+
+def list_modules_in_function_definition(func: Callable) -> List[ModuleType]:
```
liuverta: I think it might be more accurate to call this

```suggestion
def list_modules_in_function_body(func: Callable) -> List[ModuleType]:
```

since a signature is technically also part of a function definition.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,98 @@
...
+    ]
+
+
+def list_modules_in_function_signature(func: Callable) -> List[ModuleType]:
+    """List all modules used in type hints in the provided function's arguments
+    and return type hint."""
+    _func = unwrap(func)
+    hints = get_type_hints(_func)
+    arg_hints = { k: v for k, v in hints.items() if k != 'return' }
+    return_hint = { k: v for k, v in hints.items() if k == 'return' }
```
liuverta: Bit of a loose suggestion here, but if `return_hint` is always going to be one item, I think

```suggestion
    return_hint = hints.get('return')
```

would be preferable to having a one-item dictionary (and having to fetch the `'return'` key multiple times below).

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,98 @@
...
+
+    modules = [ inspect.getmodule(value) for key, value in arg_hints.items() ]
+
+    if return_hint:
+        mod = inspect.getmodule(return_hint['return'])
+        if mod.__name__ == 'typing':
```
liuverta, line 54: fyi in I think 3.10, Python starts allowing

```python
def banana() -> list[str]:
```

in addition to

```python
def banana() -> typing.List[str]:
```

We can handle this as a follow-up further down the line.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,98 @@
...
+            ret_ann = inspect.signature(_func).return_annotation
+            nested_args = ret_ann.__args__
+            for a in nested_args:
```
liuverta, line 57: This probably doesn't handle deeper nesting, right? Like `Dict[str, List[np.array]]`

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,98 @@
...
+
+        def visit_AnnAssign(self, node) -> None:
+            if node.simple:
+                # 'simple' == a single name, not an attribute or subscription, thus
+                # `node.target.id` should exist.
+                self.annotations[node.target.id] = node.annotation
+
+    source = textwrap.dedent(inspect.getsource(func))
+    mod = ast.parse(source)
+    assert mod.body and isinstance(mod.body[0], (ast.FunctionDef, ast.AsyncFunctionDef))
```
liuverta: Was this line for debugging?

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,98 @@
...
+    collector = AnnotationsCollector()
+    collector.visit(mod.body[0])
+    return [
+        ast.get_source_segment(source, node)
+        for node in collector.annotations.values()
+    ]
```
liuverta: This function appears to be returning the verbatim type annotation (e.g. `"spacy.Errors"`—an object) rather than the module name.

liuverta: And consequently `get_source_segment()` means that if—for whatever bizarre reason—they're using some other identifier

```python
sp_err = spacy.Errors
spacy_error: sp_err = self.make_spacy_error()
```

it'll return `sp_err`.

liuverta (changes requested):

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: In this case we are testing logic that extracts modules at the _function_ level.  It is necessarily a tiny subset of the modules found throughout the class.  If we simply verify that the module extracted form this individual function exists in the list of all modules expected to be extracted for the class, the test is less specific.

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: Latest version strips the submodules since it was a fairly trivial change.

Good catch. I started using `__package__` but some of the object I was dealing with didn't have that attribute. I have updated the tests to use only __name__ for consistency.

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: All `assert in` statements have been refactored into more explicit `==` assertions.

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: Fixed

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Updated `definition` >> `body`

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Likely not.  We should be able to do this with a recursive function of some kind though right?  I'll add it to the "would be nice to fix if we have time to spare at the end of next sprint" pile.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Yup.  removed

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Because of the way Python handles annotations on function local variables, they can't be accessed easily via `inspect` the way annotations in signatures can for example, hence this additional hackery to TRY and get them.    This is a trade-off we might just have to accept and document as a weakness for now.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: I'm going to write up a list of these little things and we should weigh them into the plans for next sprint.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Same here. Adding to the list

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Captured as a potential future improvement

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: Resolved in [slack thread](https://vertaai.slack.com/archives/C04NVN48LAV/p1677772742079909)

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Also captured as future improvement

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,106 @@
...
+                # 'simple' == a single name, not an attribute or subscription, thus
+                # `node.target.id` should exist.
+                self.annotations[node.target.id] = node.annotation
+
+    source = textwrap.dedent(inspect.getsource(func))
+    mod = ast.parse(source)
+    collector = AnnotationsCollector()
+    collector.visit(mod.body[0])
+    return [
+        ast.get_source_segment(source, node)
```
liuverta: Linter says `ast.get_source_segment()` isn't available until Python 3.8

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,194 @@
...
+        'collections',
+        'sklearn.base',
+        'cloudpickle.cloudpickle_fast',
+        'requests.exceptions'
+    ]
+    extracted_modules: List[str] = [
+        f.__name__ for f in md.list_modules_in_function_signature(func)
+    ]
+    assert extracted_modules == expected_modules
+
```
liuverta: Aren't these exactly the same as `test_list_modules_in_function_signature_wrapped` and `test_list_modules_in_function_signature_unwrapped`, now?

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,106 @@
...
+            if node.simple:
+                # 'simple' == a single name, not an attribute or subscription, thus
+                # `node.target.id` should exist.
+                self.annotations[node.target.id] = node.annotation
+
+    source = textwrap.dedent(inspect.getsource(func))
+    mod = ast.parse(source)
+    collector = AnnotationsCollector()
+    collector.visit(mod.body[0])
+    return [
```
liuverta: I realized I bundled two comments in one, in my last review 😅 To clarify

> [This function appears to be returning the verbatim type annotation (e.g. "spacy.Errors"—an object) rather than the module name.](https://github.com/VertaAI/modeldb/pull/3632#discussion_r1122376822)

Based on its name and docstring, it sounds like this function is supposed to return _module_ names. But the way it's written, it is returning _class_ names. It returns the type annotation `spacy.Errors` rather than the module `spacy`.

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,194 @@
...
+    func: Callable = dependency_testing_model.unwrapped_predict
+    expected_modules = ['boto3.Session']
+    extracted_modules =  md.list_function_local_annotation_module_names(func)
+    assert extracted_modules == expected_modules
+
+
+def test_module_names_in_class(dependency_testing_model) -> None:
+    """ Verify that all expected module names are extracted as expected.
+    """
+    expected_module_names = [
```
liuverta: Shall we call this

```suggestion
    expected_module = [
```

to match the other test functions?

client/verta/tests/registry/conftest.py
```diff
@@ -98,12 +98,15 @@ def make_boto_session():                 # No modules in function signature
...
         # 3rd-party modules nested inside type constructs should still be extracted
-        @property
         def nested_type_hint(self) -> Type[torch.NoneType]:
             return torch.NoneType
 
     return DependencyTestingModel
+
+
+def make_spacy_error():  # No modules in function signature
+    return spacy.Errors
```
liuverta: Was this pasted here by mistake?

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,106 @@
...
+        ast.get_source_segment(source, node)
+        for node in collector.annotations.values()
+    ]
+
+
+def module_names_in_class(model_class: Type[VertaModelBase]) -> List[str]:
```
liuverta: What would you think about calling this

```suggestion
def list_class_module_names(model_class: Type[VertaModelBase]) -> List[str]:
```

to match the other functions?

(and don't forget a docstring!)

liuverta (changes requested):

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: Yeah, these are duplicates now that we dropped the `in` assertions.  If we want to test the return type hint specifically, we'll have to separate out that logic into another function.  If we revisit this to deal with nested type hints, we should do that.  For now, I'll drop these tests.

ewagner-verta: Done.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Aw man.  Does that mean we'll need to find of way of doing this that is compatible with 3.6+?

ewagner-verta: Or do we just have to walk away from type hints in the function body for now?....

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Done

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: Done

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Fixed so that all the extraction functions return a List[str] with base module names only.  This actually simplifies things a bit.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Dropped type hints in function body altogether, so this is moot now.

client/verta/tests/registry/conftest.py
ewagner-verta: My thinking here was to have 3rd party modules called "indirectly" (not sure if that is the correct terminology) from within another function to see if we would catch those in the body (like line 65 of contest.py for example).  I think perhaps we should just rip those out for now, since we are not supporting type hints in the body.

liuverta: I don't quite understand. This function isn't even inside the `class`. There's already a `DependencyTestingModel.make_spacy_error()` defined on line 93. What is this one here on line 111 for? I don't see it being used anywhere.

liuverta: Also as a side note, I don't think we'd catch the use of `spacy.Errors` from the body of `predict()`, but our utilities _would_ catch it when we directly inspect the body of `make_spacy_error()`.

ewagner-verta: Done

client/verta/tests/registry/conftest.py
```diff
@@ -90,19 +84,11 @@ def make_dataframe(self, input):         # No modules in function signature
         def make_timeout(self):                  # No modules in function signature
             return requests.Timeout()            # 3rd party module in function body
 
-        def make_spacy_error(self):              # No modules in function signature
-            return spacy.Errors                  # 3rd party module in function body
-
-        @staticmethod
-        def make_boto_session():                 # No modules in function signature
-            return boto3.DEFAULT_SESSION         # 3rd-party module in function body
```
liuverta, line 98: Maybe we should keep this? Having coverage of `@staticmethod`-decorated functions seems useful.

client/verta/verta/_internal_utils/model_dependencies.py
liuverta: fyi we support >=3.7, not 3.6

client/verta/tests/registry/conftest.py
ewagner-verta: It actually isn't catching those right now, because the module only exists in the `return` line, and thus does not get returned by `inspect.getclosurevars()`.   I think I may have a solution pending

ewagner-verta: OK, I think I've fixed this.  When a module is aliased (as `pd` in the `make_dataframe()` function of the test class for example), it gets stored inside a different variable when returned by `inspect.getclosurevars()`.  Basically its in`nonlocals` instead of `globals` which is why it wasn't returning in some cases.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Ah, good to know!  Thanks

client/verta/tests/registry/conftest.py
ewagner-verta: I just made another function static, so we should have that covered now.  Good thinking

client/verta/tests/registry/conftest.py
```diff
@@ -73,36 +69,27 @@ def unwrapped_predict(
...
-        def make_timeout(self):                  # No modules in function signature
-            return requests.Timeout()            # 3rd party module in function body
-
-        def make_spacy_error(self):              # No modules in function signature
-            return spacy.Errors                  # 3rd party module in function body
-
         @staticmethod
-        def make_boto_session():                 # No modules in function signature
-            return boto3.DEFAULT_SESSION         # 3rd-party module in function body
+        def make_timeout():                  # No modules in function signature
```
liuverta:
```suggestion
        def make_timeout():                      # No modules in function signature
```

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,129 @@
...
+    expected_func_names = [
+                '__init__',
+                'make_dataframe',
+                'make_timeout',
+                'model_test',
+                'nested_multiple_returns_hint',
+                'nested_type_hint',
+                'predict',
+                'unwrapped_predict',
+            ]
```
liuverta:
```suggestion
    expected_func_names = [
        '__init__',
        'make_dataframe',
        'make_timeout',
        'model_test',
        'nested_multiple_returns_hint',
        'nested_type_hint',
        'predict',
        'unwrapped_predict',
    ]
```

client/verta/tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,129 @@
...
+def test_list_modules_in_function_body_wrapped(dependency_testing_model) -> None:
+    """ Verify that modules used within a function body are extracted
+    as expected, for a function that is wrapped in verify_io """
+    func: Callable = dependency_testing_model.predict
+    expected_modules = ['verta', 'yaml']
+    extracted_modules: List[str] = md.list_modules_in_function_body(func)
+    assert extracted_modules == expected_modules
+
+
+def test_list_modules_in_function_body_unwrapped(dependency_testing_model) ->None:
```
liuverta:
```suggestion
def test_list_modules_in_function_body_unwrapped(dependency_testing_model) -> None:
```

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,73 @@
...
+import ast
+import inspect
+import textwrap
+from types import ModuleType
+from typing import Callable, get_type_hints, List, Tuple, Type
+
+from verta.registry import VertaModelBase
+
+
+def list_class_functions(model_class: Type[VertaModelBase]) -> List[Tuple[str, Callable]]:
```
liuverta: Not a big deal, but we might want to make all our `list_*()` functions return the same format; this returns both function names and objects, but the others return just the module names.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,73 @@
...
+
+def list_modules_in_function_signature(func: Callable) -> List[str]:
+    """List all base modules used in type hints in the provided function's arguments
+    and return type hint."""
+    _func = unwrap(func)
+    hints = get_type_hints(_func)
+    arg_hints = { k: v for k, v in hints.items() if k != 'return' }
+    return_hint = hints.get('return')
+
+    modules = [ inspect.getmodule(value) for key, value in arg_hints.items() ]
```
liuverta:
```suggestion
    modules = [ inspect.getmodule(value) for value in arg_hints.values() ]
```

liuverta: Or for extra fun

```suggestion
    modules = list(map(inspect.getmodule, arg_hints.values()))
```

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -0,0 +1,73 @@
...
+def list_class_module_names(model_class: Type[VertaModelBase]) -> List[str]:
+    """Attempt to list all base modules used in the provided class object."""
+    funcs = list_class_functions(model_class)
+    modules_found = list()
+    for function in funcs:
+        _func = function[1]
+        mods_in_body = list_modules_in_function_body(_func)
+        mods_in_signature = list_modules_in_function_signature(_func)
+        modules_found += mods_in_body + mods_in_signature
+    return list(set(modules_found))
```
liuverta: Any particular reason for not returning the set, rather than a list?

liuverta: Heck, I wonder if all of these should return sets 🤔

liuverta (approved):

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: The more I though on this, the more I agree.  I switched these to return sets throughout, and I simplified the return type of `list_class_functions`  to just `Set[Callable]`.  Seems more readable / simpler

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Changed return type to be more consistent (now just a set of callables)

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Went with the list comprehension just for consistency

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: Fixed

client/verta/tests/registry/test_model_dependencies.py
ewagner-verta: Yikes.  How did I miss that <facepalm>

client/verta/tests/registry/conftest.py
ewagner-verta: Fixed

liuverta (changes requested):
Just marking this as needing changes while we figure out how to remove `check_model_dependencies.py`, so that we don't accidentally merge this.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -58,16 +58,15 @@ def list_modules_in_function_signature(func: Callable) -> List[str]:
...
-    for function in funcs:
-        _func = function[1]
+    modules_found = set()
+    for function in list_class_functions(model_class):
+        _func = function
         mods_in_body = list_modules_in_function_body(_func)
         mods_in_signature = list_modules_in_function_signature(_func)
-        modules_found += mods_in_body + mods_in_signature
-    return list(set(modules_found))
+        modules_found.update(mods_in_body | mods_in_signature)
```
liuverta: fyi Python has `|=`, analogous to `+=`.

client/verta/tests/unit_tests/registry/test_model_dependencies.py
```diff
@@ -0,0 +1,130 @@
...
+    expected_func_names = [
+                '__init__',
+                'make_dataframe',
+                'make_timeout',
+                'model_test',
+                'nested_multiple_returns_hint',
+                'nested_type_hint',
+                'predict',
+                'unwrapped_predict',
+            ]
```
liuverta:
```suggestion
    expected_func_names = [
        '__init__',
        'make_dataframe',
        'make_timeout',
        'model_test',
        'nested_multiple_returns_hint',
        'nested_type_hint',
        'predict',
        'unwrapped_predict',
    ]
```

liuverta (approved):

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: updated

client/verta/tests/unit_tests/registry/test_model_dependencies.py
ewagner-verta: fixed

PR #3641, opened 2023-03-03
feature: [VRD-683] Client function to check model dependencies

ewagner-verta:
## Impact and Context
Adds a new customer facing client function to allow users to pass a model class and a Python environment object for comparison to identify any missing packages from the environment.

Also covers [VRD-684] (testing ticket)

## Risks and Area of Effect
Net new code that does not touch any existing functionality (other than the utility functions recently merged for this purpose). Will be thoroughly tested before the next client release.

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

[VRD-684]: https://vertaai.atlassian.net/browse/VRD-684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,76 @@
...
+def check_model_dependencies(
+        model: Type[VertaModelBase],
+        environment: Python,
+        raise_for_missing: bool = False,
+        ) -> bool:
+    """
+    Attempt to scan the provided model class for dependencies and verify that
+    the provided environment includes the necessary packages to support them.
+
+    .. version_added:: 0.22.1
```
liuverta: `0.22.1` is going out first thing next week alongside our platform's `2023_03` release, so this probably shouldn't be considered user-ready until the following version.

```suggestion
    .. versionadded:: 0.22.2
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Parameters
+    ----------
+    model: Type[VertaModelBase]
+        Model class, inherited from :class:`~verta.registry.VertaModelBase` to be scanned.
+    environment:
+        Instance of `~verta.environment.Python` to validate against.
+    raise_for_missing: bool, default False
```
liuverta: To be a bit more consistent with [`create_standard_model()`](https://github.com/VertaAI/modeldb/blob/5cf73b6f9d6cd648a7fc180c13540029040f8b76/client/verta/verta/registry/entities/_model.py#L363-L366)

```suggestion
    model_cls : subclass of :class:`~verta.registry.VertaModelBase`
        Model class to be scanned.
    environment : :class:`~verta.environment.Python`
        pip dependencies to validate against.
    raise_for_missing : bool, default False
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,76 @@
...
+        If True, raises an exception if any dependencies detected in the model class
+        are missing from the environment. Defaults to printing a warning.
+
+    Returns
+    -------
+    True if all dependencies detected are found in the environment, False otherwise.
```
liuverta:
```suggestion
    bool
        True if all dependencies detected are found in the environment, False otherwise.
```

client/verta/verta/registry/check_model_dependencies.py
liuverta: On that note, if we're looking to merge this today/Monday, we should hide this function at least from our API documentation by either

1. making it private by renaming the function to `_check_model_dependencies()`
2. commenting out the docstring

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -71,3 +71,14 @@ def class_module_names(model_class: Type[VertaModelBase]) -> Set[str]:
...
+
+# TODO VRD-682 convert module names to pkg names here
+def package_names(module_names: Set[str]) -> Set[str]:
+    """Return a set of all packages detected in the provided set of base module names."""
+    return module_names
+
+
+def strip_version(pkg_str: str) -> str:
+    """Return the base package name from a package string."""
+    return pkg_str.split('==')[0]
```
liuverta: I recommend using ~`verta._internal_utils._pip_requirements_utils.parse_version(pkg_str)[0]`~  `verta._internal_utils._pip_requirements_utils.parse_req_spec(pkg_str)[0]`instead.

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,76 @@
...
+    ------
+    RuntimeError
+        If `raise_for_missing` is True and any dependencies are missing.
+
+    """
+    detected_modules: Set[str] = md.class_module_names(model)
+    detected_packages: Set[str] = md.package_names(detected_modules)
+    env_packages: Set[str] = { md.strip_version(r) for r in environment.requirements }
+
+    if detected_packages !=  env_packages:
```
liuverta:
```suggestion
    if detected_packages != env_packages:
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,76 @@
...
+                    f"the following packages are not required by the model but are present "
+                    f"in the environment: {extra_packages}"
+                )
+            else:
+                warnings.warn(
+                    f"the following packages are not required by the model but are present "
+                    f"in the environment: {extra_packages}",
+                    category=RuntimeWarning,
+                )
+                return False
```
liuverta: I wouldn't consider extra packages an issue. I suppose we can show a warning if we'd really like, but I don't think this is error-worthy.

liuverta: There could be all sorts of cases in Python where a library is required to be installed in the environment, but isn't directly `import`ed or invoked by the model. For example, `urllib3` is required by `requests`, and I might want a specific version of `urllib3` to ensure some underlying behavior even if my model doesn't use it directly.

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,76 @@
...
+    env_packages: Set[str] = { md.strip_version(r) for r in environment.requirements }
+
+    if detected_packages !=  env_packages:
+        missing_packages = detected_packages - env_packages
+        extra_packages = env_packages - detected_packages
+        if missing_packages:
+            if raise_for_missing:
+                raise RuntimeError(
+                    f"the following packages are required by the model but missing "
+                    f"from the environment: {missing_packages}"
```
liuverta: I'd recommend defining this error message string once, rather than twice. In general, we should avoid duplicating things—both for maintainability in the future, and so code reviewers don't have to carefully check things word-by-word to make sure they're not missing a subtle deliberate difference

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -0,0 +1,117 @@
...
+
+# TODO VRD-682 convert module names to pkg names and update this
+@pytest.fixture(scope='session')
+def complete_env() -> Python:
+    """ Environment with all expected packages from dependency_testing_model
+    fixture.  For completeness, some are versioned and some are not.
+    """
+    return Python([
+        'calendar',
+        'click',
```
liuverta: I'm not sure if this was intentional, but it looks like you're exploiting a fallback in `Python` 🤔 

So, `Python()` is supposed to insta-magically identify and assign version numbers to all libraries. For example,

```python
env = Python(["click"])
assert env.requirements == ["click==8.1.3", "cloudpickle==2.2.0", "verta==0.22.0"]
```

but it will skip this behavior if it encounters an error while checking the environment for version numbers. In this case, because `calendar` comes from the std lib and doesn't have a `__version__` attribute, `Python()` encounters an error and just rolls with `["calendar", "click", ...]`. If you set your logging level to `INFO`, you'll see

```
INFO - failed to manually parse requirements; falling back to capturing raw contents
Traceback (most recent call last):
  File "/Users/miliu/Documents/modeldb/client/verta/verta/environment/_python.py", line 266, in _capture_requirements
    _pip_requirements_utils.set_version_pins(requirements_copy)
  File "/Users/miliu/Documents/modeldb/client/verta/verta/_internal_utils/_pip_requirements_utils.py", line 267, in set_version_pins
    six.raise_from(error, None)
  File "<string>", line 3, in raise_from
ValueError: unable to determine a version number for requirement 'calendar'; it might not be installed; please manually specify it as 'calendar==x.y.z'
```

liuverta: All that is to say: If you're intentionally doing this so that `complete_env` stores libraries without version strings, I'd recommend adding a comment

```suggestion
        'calendar',  # from std lib; forces Python() to skip capturing version number
        'click',
```

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -0,0 +1,117 @@
...
+            environment=incomplete_env,
+            raise_for_missing=False,
+        )
+    warn_msg = caught_warnings[0].message.args[0]
+    assert warn_msg == "the following packages are required by the model but " \
+                       "missing from the environment: {'pandas'}"
+
+
+def test_check_model_dependencies_extra_raise(dependency_testing_model, complete_env) -> None:
+    """ Verify that check_model_dependencies does not raise an exception for
```
liuverta:
```suggestion
    """ Verify that check_model_dependencies raises an exception for
```

liuverta (changes requested):

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Fixed, and I'm assuming we will not merge this before EOD, so we can take the necessary steps to hide it next week.

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Fixed

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,103 @@
...
+    extraneous packages in the provided environment.
+
+    .. note::
+        This function is not guaranteed to detect all dependencies in all cases.
+
+    .. versionadded:: 0.22.2
+
+    Parameters
+    ----------
+    model: subclass of :class:`~verta.registry.VertaModelBase`
```
liuverta: To match [`create_standard_model()`](https://github.com/VertaAI/modeldb/blob/5cf73b6f9d6cd648a7fc180c13540029040f8b76/client/verta/verta/registry/entities/_model.py#L363-L366)

```suggestion
    model_cls: subclass of :class:`~verta.registry.VertaModelBase`
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,103 @@
...
+        Model class object (not an instance) to be scanned.
```
liuverta: I think the word "object" makes it a bit confusing; in most languages, "object" directly implies "instance" and I'm not sure how common the knowledge is that Python `class`es are objects.

```suggestion
        Model class (not an instance) to be scanned.
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,103 @@
...
+    environment: subclass of :class:`~verta.environment.Python`
```
liuverta: Not a subclass here.

```suggestion
    environment: :class:`~verta.environment.Python`
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,103 @@
...
+        Instance of a :class:`~verta.environment.Python` environment against
+        which to validate pip dependencies.
```
liuverta: We already document the type of this argument in two other places, so we can skip having it in the description.

```suggestion
        Environment against which to validate pip dependencies.
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Returns
+    -------
+        True
+            if all 3rd-party dependencies detected in the model class have
+            corresponding packages in the environment, and no extraneous packages
+            exist in the environment.
+        False
+            if any 3rd-party dependencies detected in the model class are missing
+            or extraneous packages are detected in the environment.
```
liuverta:
```suggestion
    bool
        True
            if all 3rd-party dependencies detected in the model class have
            corresponding packages in the environment, and no extraneous packages
            exist in the environment.
        False
            if any 3rd-party dependencies detected in the model class are missing
            or extraneous packages are detected in the environment.
```

liuverta (changes requested):

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: I've looked through the functions in `_pip_requirements_utils` and none of them seems to do exactly what we need here.  Given the string `pandas==1.0.6`, we need back the package name.  `parse_version` parses the version bit (`1.0.6`) and `remove_pinned_requirements` isn't it either.  Would it be better to add a new utility here, rather than with the `model_dependenies` utilities?

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: I was thinking more about preventing the loading of unused packages for performance reasons, bit I hadn't considered that.  I'm leaning toward just ripping this out then

client/verta/verta/_internal_utils/model_dependencies.py
liuverta: Oops a bad copy and paste on my part. I meant `parse_req_spec()`

ewagner-verta: Updated to use `parse_req_spec()` to get the package name.

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Fixed

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: fixed

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Removed all functions and tests for extraneous packages in the env, and updated docs strings accordingly.

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
ewagner-verta: Dropped tests

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: fixed

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: fixed

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: fixed

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: fixed

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Good call.  Updated

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
ewagner-verta: I just gave all package strings proper versions for now under the assumption we'll handle all this package related stuff all at once in the next block of work. (VRD-682)

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Updated.

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,80 @@
...
+        Environment against which to validate pip dependencies.
+    raise_for_missing: bool, default False
+        If `True`, raises an exception if any dependencies detected in the model class
+        are missing from the environment.  Defaults to printing a warning.
+
+    Returns
+    -------
+        bool
+            `True` if all 3rd-party dependencies detected in the model class have
+            corresponding packages in the environment. `False` if any are missing.
```
liuverta:
```suggestion
    bool
        `True` if all 3rd-party dependencies detected in the model class have
        corresponding packages in the environment. `False` if any are missing.
```

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import Set, Type
+import warnings
+
+from verta.registry import VertaModelBase
+from verta.environment import Python
+from verta._internal_utils import model_dependencies as md
+from verta._internal_utils._pip_requirements_utils import parse_req_spec
+
+
+def check_model_dependencies(
```
liuverta: Like I [mentioned last week](https://github.com/VertaAI/modeldb/pull/3641#discussion_r1125093066), we should hide this from our public API before merging since it's not yet in an end-user usable state.

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,80 @@
...
+
+        env = Python(["numpy", "pandas"])
+        check_model_dependencies(model=MyModelClass, environment=env)
+
+    """
+    detected_modules: Set[str] = md.class_module_names(model_cls)
+    detected_packages: Set[str] = md.package_names(detected_modules)
+    env_packages: Set[str] = { parse_req_spec(e)[0] for e in environment.requirements }
+    error_msg = "the following packages are required by the model but missing " \
+                "from the environment: "
```
liuverta: I realized I was a bit unclear in [my previous comment](https://github.com/VertaAI/modeldb/pull/3641#discussion_r1125103748); I'd recommend

```suggestion
    error_msg = "the following packages are required by the model but missing " \
                "from the environment: " + str(missing_packages)
```

We should really just define the whole thing at once, if it's used exactly the same way multiple times below.

client/verta/verta/registry/check_model_dependencies.py
```diff
@@ -0,0 +1,80 @@
...
+
+    if detected_packages != env_packages:
+        missing_packages = detected_packages - env_packages
+        if missing_packages:
```
liuverta: Could this just be

```suggestion
    missing_packages = detected_packages - env_packages
    if missing_packages:
```

?

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
liuverta: I don't think VRD-682 has anything to do with whether `Python()` captures library version numbers. Using my suggestion of adding a comment really would've been fine.

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
+""" Unit tests for verta.registry.check_model_dependencies """
+
+import pytest
+from typing import Set
+import warnings
+
+from verta.registry.check_model_dependencies import check_model_dependencies
```
liuverta: This doesn't seem to match the design doc, nor the example in `check_model_dependencies`'s docstring

> ```python
> from verta.registry import check_model_dependencies
> ```

liuverta (changes requested):

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Fixed.

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Yes it should.  I neglected to change this when I ripped out the checks for extra packages. Fixed

client/verta/verta/registry/check_model_dependencies.py
ewagner-verta: Made the function hidden.  Thanks for the reminder.

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -0,0 +1,82 @@
...
+        'torch==0.0.1',
+        'urllib3==0.0.1',
+        'verta==0.0.1',
+        'yaml==0.0.1',
+    ])
+
+
+def test_check_model_dependencies_complete(dependency_testing_model, complete_env) -> None:
+    """ Verify that check_model_dependencies returns all the expected packages from
+    the test model class (dependency_testing_model fixture).
```
liuverta: This docstring isn't quite accurate (says the function returns packages, but it doesn't!)

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -0,0 +1,82 @@
...
+    assert _check_model_dependencies(
+        model_cls=dependency_testing_model,
+        environment=complete_env,
+        raise_for_missing=False,
+    )
+
+
+def test_check_model_dependencies_missing_raise(dependency_testing_model, complete_env) -> None:
+    """ Verify that check_model_dependencies raises an exception for missing
+    packages when `raise_for_missing`` is True.
```
liuverta:
```suggestion
    packages when `raise_for_missing` is True.
```

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -4,7 +4,7 @@
 from types import ModuleType
 from typing import Callable, get_type_hints, Set, Type
 
-from verta.registry import VertaModelBase
+from ..registry._verta_model_base import VertaModelBase
```
liuverta, line 7: Out of curiosity, was there a particular reason for this change?

```python
from verta.registry import VertaModelBase
```

would've been the more canonical way to do this import in my opinion, though we don't really have a strict guideline for it.

client/verta/verta/registry/_check_model_dependencies.py
```diff
@@ -0,0 +1,78 @@
...
+def _check_model_dependencies(
+        model_cls: Type[VertaModelBase],
+        environment: Python,
+        raise_for_missing: bool = False,
+        ) -> bool:
+    """Scan for missing dependencies in a model's environment.
+
+    This function attempts to scan the provided model class for 3rd-party (not
+    python standard library) dependencies and identify any packages in the
+    provided environment.
```
liuverta: Should this be something like

```suggestion
    python standard library) dependencies and identify any missing packages in the
    provided environment.
```

since it's specifically about checking for missing packages? Unsure though

client/verta/verta/registry/_check_model_dependencies.py
```diff
@@ -0,0 +1,78 @@
...
+
+    .. note::
+        This function is not guaranteed to detect all dependencies in all cases.
```
liuverta: I'd like to assure the user that this function has a value/use-case rather than simply "hey this may not work" haha

```suggestion
        This function is not guaranteed to detect all dependencies in all cases, and should be considered a fast sanity check rather than a test.
```

liuverta (approved):
A few minor comments. Thanks for the experimentation and the follow-ups! Looks good

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -37,8 +37,9 @@ def complete_env() -> Python:
 
 
 def test_check_model_dependencies_complete(dependency_testing_model, complete_env) -> None:
-    """ Verify that check_model_dependencies returns all the expected packages from
-    the test model class (dependency_testing_model fixture).
+    """ Verify that check_model_dependencies extracts all the expected packages from
+    the test model class (dependency_testing_model fixture) and correctly reconciles
```
liuverta, line 41: "reconcile" isn't quite the right term imo. In software and more generally too I think, it usually means "to change something's state to match another's", and the function doesn't change the model class.

PR #3645, opened 2023-03-06
feat: [VRD-711] Add batch prediction method to client

ghost:
## Impact and Context
Create a `batch_predict` method that accepts and returns `pandas.DataFrame`s. This method takes a dataframe, splits it into smaller dataframes of the provided `batch_size` to make predictions against the model, then reassembles the output to return to the user as one dataframe.

## Risks and Area of Effect
Doesn't affect existing features.

## Testing
- [X] Unit test
- [X] Deployed to dev env
- [ ] Other (explain) 

Added unit tests to cover the basic functionality. Planning to add more complex property tests shortly, in a separate PR. Also deployed to a dev env to ensure the pipeline works from end to end.

Also, python model tests look good with the updated json request with nans thing: https://jenkins.dev.verta.ai/job/test/job/python-models/job/python-models/34/

## Reverting

### DIFF
--- a/client/verta/tests/unit_tests/test_deployed_model.py
+++ b/client/verta/tests/unit_tests/test_deployed_model.py
@@ -400,7 +400,7 @@ def test_batch_predict_with_one_batch_with_no_index(mocked_responses) -> None:
         creds=creds,
         token=TOKEN,
         )
-    # the input below is entirely irrelevant since it"s smaller than the batch size
+    # the input below is entirely irrelevant since it's smaller than the batch size
     prediction_df = dm.batch_predict(pd.DataFrame({"hi": "bye"}, index=[1]), 10)
     pd.testing.assert_frame_equal(expected_df, prediction_df)
 
--- a/client/verta/tests/unit_tests/test_deployed_model.py
+++ b/client/verta/tests/unit_tests/test_deployed_model.py
@@ -449,7 +449,7 @@ def test_batch_predict_with_five_batches_of_one_with_no_indexes(mocked_responses
         )
     input_df = pd.DataFrame({"a": [1, 2, 3, 4, 5], "b": [11, 12, 13, 14, 15]})
     prediction_df = dm.batch_predict(input_df, 1)
-    expected_final_df = pd.concat(expected_df_list)
+    expected_df = pd.concat(expected_df_list)
     # Since no index was provided, we can"t guarantee the index type for assertions
     pd.testing.assert_frame_equal(expected_df, prediction_df, check_index_type=False)
 
--- a/client/verta/tests/unit_tests/test_deployed_model.py
+++ b/client/verta/tests/unit_tests/test_deployed_model.py
@@ -428,8 +428,8 @@ def test_batch_predict_with_one_batch_with_output_index(mocked_responses) -> Non
     pd.testing.assert_frame_equal(expected_df, prediction_df)
 
 
-def test_batch_predict_with_five_batches_of_one_with_no_indexes(mocked_responses) -> None:
-    """ Call batch_predict with five batches. """
+def test_batch_predict_with_batches_with_no_indexes(mocked_responses) -> None:
+    """ Call batch_predict with batches. """
     expected_d_list = [pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}),
                        pd.DataFrame({"B": [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]}),
                        pd.DataFrame({"C": [21, 22, 23, 24, 25, 26, 27, 28, 29, 30]}),
--- a/client/verta/tests/unit_tests/test_deployed_model.py
+++ b/client/verta/tests/unit_tests/test_deployed_model.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """ Unit tests for the verta.deployment.DeployedModel class. """
+import json
 import os
 from typing import Any, Dict
 
@@ -384,13 +385,13 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
     )
 
 
-def test_batch_predict_with_one_batch_with_no_output_index(mocked_responses) -> None:
+def test_batch_predict_with_one_batch_with_no_index(mocked_responses) -> None:
     """ Call batch_predict with a single batch. """
     expected_df = pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], "B": [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]})
-    expected_df_json = expected_df.to_dict(orient="index")
+    expected_df_body = json.dumps(expected_df.to_dict(orient="split")).encode("utf-8")
     mocked_responses.post(
         BATCH_PREDICTION_URL,
-        json=expected_df_json,
+        body=expected_df_body,
         status=200,
         )
     creds = EmailCredentials.load_from_os_env()
@@ -401,17 +402,16 @@ def test_batch_predict_with_one_batch_with_no_output_index(mocked_responses) -> 
         )
     # the input below is entirely irrelevant since it"s smaller than the batch size
     prediction_df = dm.batch_predict(pd.DataFrame({"hi": "bye"}, index=[1]), 10)
-    # Since no index was provided, we can"t guarantee the index type for assertions
-    pd.testing.assert_frame_equal(expected_df, prediction_df, check_index_type=False)
+    pd.testing.assert_frame_equal(expected_df, prediction_df)
 
 
-def test_batch_predict_with_one_batch_with_output_index(mocked_responses) -> None:
+def test_batch_predict_with_one_batch_with_index(mocked_responses) -> None:
     """ Call batch_predict with a single batch, where the output has an index. """
     expected_df = pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], "B": [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]}, index=["a", "b", "c", "d", "e", "f", "g", "h", "i", "j"])
-    expected_df_json = expected_df.to_dict(orient="index")
+    expected_df_body = json.dumps(expected_df.to_dict(orient="split")).encode("utf-8")
     mocked_responses.post(
         BATCH_PREDICTION_URL,
-        json=expected_df_json,
+        body=expected_df_body,
         status=200,
         )
     creds = EmailCredentials.load_from_os_env()
@@ -420,13 +420,13 @@ def test_batch_predict_with_one_batch_with_output_index(mocked_responses) -> Non
         creds=creds,
         token=TOKEN,
         )
-    # the input below is entirely irrelevant since it"s smaller than the batch size
+    # the input below is entirely irrelevant since it's smaller than the batch size
     prediction_df = dm.batch_predict(pd.DataFrame({"hi": "bye"}, index=[1]), 10)
     # Since an index WAS provided, we should be able to assert with indexes included
     pd.testing.assert_frame_equal(expected_df, prediction_df)
 
 
-def test_batch_predict_with_five_batches_of_one_with_no_indexes(mocked_responses) -> None:
+def test_batch_predict_with_five_batches_with_no_indexes(mocked_responses) -> None:
     """ Call batch_predict with five batches. """
     expected_df_list = [pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}),
                        pd.DataFrame({"B": [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]}),
@@ -434,11 +434,11 @@ def test_batch_predict_with_five_batches_of_one_with_no_indexes(mocked_responses
                        pd.DataFrame({"D": [31, 32, 33, 34, 35, 36, 37, 38, 39, 40]}),
                        pd.DataFrame({"E": [41, 42, 43, 44, 45, 46, 47, 48, 49, 50]}),
                        ]
-    for expected_d in expected_df_list:
+    for expected_df in expected_df_list:
         mocked_responses.add(
             responses.POST,
             BATCH_PREDICTION_URL,
-            json=expected_d.to_dict(orient="index"),
+            body=json.dumps(expected_df.to_dict(orient="split")).encode("utf-8"),
             status=200,
             )
     creds = EmailCredentials.load_from_os_env()
@@ -450,8 +450,7 @@ def test_batch_predict_with_five_batches_of_one_with_no_indexes(mocked_responses
     input_df = pd.DataFrame({"a": [1, 2, 3, 4, 5], "b": [11, 12, 13, 14, 15]})
     prediction_df = dm.batch_predict(input_df, 1)
     expected_df = pd.concat(expected_df_list)
-    # Since no index was provided, we can"t guarantee the index type for assertions
-    pd.testing.assert_frame_equal(expected_df, prediction_df, check_index_type=False)
+    pd.testing.assert_frame_equal(expected_df, prediction_df)
 
 
 def test_batch_predict_with_batches_and_indexes(mocked_responses) -> None:
@@ -462,11 +461,11 @@ def test_batch_predict_with_batches_and_indexes(mocked_responses) -> None:
                        pd.DataFrame({"D": [31, 32, 33, 34, 35, 36, 37, 38, 39, 40]}, index=["a", "b", "c", "d", "e", "f", "g", "h", "i", "j"]),
                        pd.DataFrame({"E": [41, 42, 43, 44, 45, 46, 47, 48, 49, 50]}, index=["a", "b", "c", "d", "e", "f", "g", "h", "i", "j"]),
                        ]
-    for expected_d in expected_df_list:
+    for expected_df in expected_df_list:
         mocked_responses.add(
             responses.POST,
             BATCH_PREDICTION_URL,
-            json=expected_d.to_dict(orient="index"),
+            body=json.dumps(expected_df.to_dict(orient="split")).encode("utf-8"),
             status=200,
             )
     creds = EmailCredentials.load_from_os_env()
--- a/client/verta/tests/unit_tests/test_deployed_model.py
+++ b/client/verta/tests/unit_tests/test_deployed_model.py
@@ -385,8 +385,7 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
 
 def test_batch_predict_with_one_batch_with_no_output_index(mocked_responses) -> None:
     """ Calling batch_predict with a small dataset (fits within one batch) and getting a 200 response returns the response as expected. """
-    d = pd.DataFrame({'A': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 'B': [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]})
-    expected_df = pd.DataFrame(d)
+    expected_df = pd.DataFrame({'A': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 'B': [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]})
     expected_df_json = expected_df.to_dict(orient="index")
     mocked_responses.post(
         BATCH_PREDICTION_URL,
@@ -408,8 +407,7 @@ def test_batch_predict_with_one_batch_with_no_output_index(mocked_responses) -> 
 
 def test_batch_predict_with_one_batch_with_output_index(mocked_responses) -> None:
     """ Calling batch_predict with a small dataset (fits within one batch) and getting a 200 response returns the response as expected. """
-    d = pd.DataFrame({'A': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 'B': [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]}, index=['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'])
-    expected_df = pd.DataFrame(d)
+    expected_df = pd.DataFrame({'A': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 'B': [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]}, index=['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'])
     expected_df_json = expected_df.to_dict(orient="index")
     mocked_responses.post(
         BATCH_PREDICTION_URL,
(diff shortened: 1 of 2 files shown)


client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +360,81 @@ def predict_with_id(
...
+            status_forcelist=retry_status,
+            backoff_factor=backoff_factor,
+        )
+
+        # TODO: add compression
+        # Split into batches
+        out_df_list = []
+        batches = np.array_split(x, batch_size)
+        for batch in batches:
+            # response = self._single_batch_predict(batch, batch_size, prediction_id)
```
ghost: will uncomment this at some point; currently contriving a response for testing

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -157,12 +163,44 @@ def _predict(
...
+            except ValueError:  # not JSON response; 502 not from model back end
+                pass
+            else:  # from model back end; contains message (maybe)
+                # try to directly print message, otherwise line breaks appear as '\n'
+                msg = data.get("message") or json.dumps(data)
+                raise RuntimeError(f"deployed model encountered an error: {msg}")
+
+        if not response.ok:
+            _utils.raise_for_http_error(response=response)
+        return response
```
ewagner-verta: We may want to consider putting this into a separate function like `response_handler(response: Response) -> Response` or similar and calling it from both places since it is ~16 lines of identical code.  Not a big deal functionally, so I'm not inclined to hold anything up.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -181,16 +219,16 @@ def get_curl(self):
...
-        ) -> Dict[str, Any]:
+            self,
+            x: List[Any],
+            compress=False,
+            max_retries: int = http_session.DEFAULT_MAX_RETRIES,
+            always_retry_404=False,
+            always_retry_429=False,
+            retry_status: Set[int] = http_session.DEFAULT_STATUS_FORCELIST,
+            backoff_factor: float = http_session.DEFAULT_BACKOFF_FACTOR,
+            prediction_id: str = None,
```
ewagner-verta, line 200: Did you do this manually, or do you use an auto-formatter of some kind?  (Just curious, and thanks btw)

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +360,81 @@ def predict_with_id(
...
+        out_df_list = []
+        batches = np.array_split(x, batch_size)
+        for batch in batches:
+            # response = self._single_batch_predict(batch, batch_size, prediction_id)
+            # out_batches.extend(_utils.body_to_json(response))
+
+            # TODO: obviously undo before merging
+            temp = pd.DataFrame.from_dict({'name': ['Alice', 'Bob', 'Charlie'],
+                                           'age': [25, 30, 35],
+                                           'city': ['New York', 'London', 'Paris']})
```
ewagner-verta: In my last PR, @liuverta mentioned that we should prefix incomplete functions under development with an underscore such that they don't show up in the docs before they are actually usable.  Can I assume there is some means by which the publicly available docs get updated independently of a client release?  (e.g. pulled from `main` periodically)

liuverta: The only reason we really needed to do that was because we we're right about to ship a new client release.

Docs currently are updated every time we merge to `main`.

client/verta/verta/deployment/_deployedmodel.py
liuverta: Could we also—instead of having this `_single_batch_predict()` at all—do the splitting and JSONification in `batch_predict()` and simply use `_predict()`?

client/verta/verta/deployment/_deployedmodel.py
ghost: it happened automagically without my consent 😅

client/verta/verta/deployment/_deployedmodel.py
ghost: good idea! Now that I've got the base functionality working, time to clean up(:

client/verta/verta/deployment/_deployedmodel.py
ghost: (I assume it was PyCharm)

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -7,13 +7,17 @@
 from requests import Session
 from typing import Any, Dict, List, Optional, Set, Tuple
 from urllib.parse import urlparse
+
+from verta._internal_utils import importer
+
 from verta import credentials
 
 from .._internal_utils import _utils, http_session
```
liuverta:
```suggestion
from verta import credentials

from .._internal_utils import _utils, http_session, importer
```

client/verta/verta/deployment/_deployedmodel.py
liuverta: Our indentation throughout this file is wildly inconsistent.

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -107,14 +111,18 @@ def prediction_url(self, value):
             raise ValueError("not a valid prediction_url")
         self._prediction_url = parsed.geturl()
 
+    def batch_prediction_url(self):
```
liuverta:
```suggestion
    @property
    def _batch_prediction_url(self):
```

Recommending

1. making it a `@property` to match `prediction_url` (`self.batch_prediction_url` instead of `self.batch_prediction_url()`)
2. marking it as internal with an initial underscore because a user shouldn't be interested in using this URL directly (as opposed to `prediction_url`, which we support cURLing)

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -107,14 +111,18 @@ def prediction_url(self, value):
...
+        # Adding the v1/ to make sure we don't accidentally replace the wrong part of the URL
+        return self.prediction_url.replace("v1/predict", "v1/batch-predict")
```
liuverta: This seems hacky 🤔 

```suggestion
        return self.prediction_url.replace("/predict/", "/batch-predict/")
```

might be a bit less hacky (so we don't rely on the `/v1` routing)

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -107,14 +111,18 @@ def prediction_url(self, value):
...
 
     # TODO: Implement dynamic compression via separate utility and call it from here
     def _predict(
             self,
             x: Any,
-            compress: bool=False,
-            prediction_id: Optional[str]=None,
-            ):
+            prediction_url: bytes,
```
liuverta: Is this not a `str`? 🤔

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +328,81 @@ def predict_with_id(
...
+
+        # Split into batches
+        out_df_list = []
+        for i in range(0, len(df), batch_size):
+            batch = df[i:i+batch_size]
+            serialized_batch = batch.to_dict(orient="records")
+            response = self._predict(serialized_batch, self.batch_prediction_url(), compress, prediction_id)
+            out_df = pd.DataFrame.from_dict(response.json())
+            out_df_list.append(out_df)
+        out_df = pd.concat(out_df_list, ignore_index=True)
```
liuverta: Wouldn't we want to try keeping the index intact? Or was it causing issues?

```suggestion
        out_df = pd.concat(out_df_list)
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +328,81 @@ def predict_with_id(
...
+            self._session,
+            max_retries=max_retries,
+            status_forcelist=retry_status,
+            backoff_factor=backoff_factor,
+        )
+
+        # Split into batches
+        out_df_list = []
+        for i in range(0, len(df), batch_size):
+            batch = df[i:i+batch_size]
```
liuverta: I think this needs to be

```suggestion
            batch = df.iloc[i:i+batch_size]
```

to index rows. Otherwise it's indexing columns 🤔

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +328,81 @@ def predict_with_id(
...
+            serialized_batch = batch.to_dict(orient="records")
+            response = self._predict(serialized_batch, self.batch_prediction_url(), compress, prediction_id)
+            out_df = pd.DataFrame.from_dict(response.json())
```
liuverta: Like we do in `predict_with_id()`:

```suggestion
            out_df = pd.DataFrame.from_dict(_utils.body_to_json(response))
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +328,81 @@ def predict_with_id(
...
+            Set of status codes, as integers, for which retry attempts should be made.  Overwrites default value.
+            Expand the set to include more. For example, to add status code 409 to the existing set, use:
+            ``retry_status={404, 429, 500, 503, 504, 409}``
+        backoff_factor : float, default 0.3
+            A backoff factor to apply between retry attempts.  Uses standard urllib3 sleep pattern:
+            ``{backoff factor} * (2 ** ({number of total retries} - 1))`` with a maximum sleep time between requests of
+            120 seconds.
+        prediction_id: str, default None
+            A custom string to use as the ID for the prediction request.  Defaults to a randomly generated UUID.
+        Returns
```
liuverta, line 374:
```suggestion

        Returns
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +328,81 @@ def predict_with_id(
...
+        )
 
-        response = self._predict(x, compress, prediction_id)
+        response = self._predict(x, self.prediction_url, compress, prediction_id)
         id = response.headers.get('verta-request-id', '')
         return (id, _utils.body_to_json(response))
 
+    def batch_predict(
+            self,
+            df,
```
liuverta, line 341: Ahhhhh we can't use `pandas` in type annotations here, huh 😞

liuverta (changes requested):

client/verta/verta/deployment/_deployedmodel.py
ghost: it's... not

client/verta/verta/deployment/_deployedmodel.py
ghost: oopsie

client/verta/verta/deployment/_deployedmodel.py
ghost: I think chatgpt suggested it and I didn't follow up 😅

client/verta/verta/deployment/_deployedmodel.py
ghost: sadly not :(

client/verta/verta/deployment/_deployedmodel.py
ghost: idk WHY, but pycharm is very insistent that it's not a str 😒

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +325,82 @@ def predict_with_id(
...
+
+        # Split into batches
+        out_df_list = []
+        for i in range(0, len(df), batch_size):
+            batch = df.iloc[i:i+batch_size]
+            serialized_batch = batch.to_dict(orient="records")
+            response = self._predict(serialized_batch, self._batch_prediction_url, compress, prediction_id)
+            out_df = pd.DataFrame.from_dict(_utils.body_to_json(response))
+            out_df_list.append(out_df)
+        out_df = pd.concat(out_df_list)
```
liuverta: Oh wait I'm sorry, we can't keep the index because the returned DataFrame doesn't necessarily use the same index as the input (i.e. the `out_df`s might have conflicting indices). ChatGPT wins this time.

liuverta (changes requested):
🙇

client/verta/verta/deployment/_deployedmodel.py
liuverta: Hrm...Could we leave out the type hint altogether, then? There's no way it's supposed to be `bytes`.

liuverta (changes requested):

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +326,86 @@ def predict_with_id(
...
+
+        # Split into batches
+        out_df_list = []
+        for i in range(0, len(df), batch_size):
+            batch = df.iloc[i:i+batch_size]
+            serialized_batch = batch.to_dict(orient="index")
+            # Predict with one batch at a time
+            response = self._predict(serialized_batch, self._batch_prediction_url, compress, prediction_id)
+            re = _utils.body_to_json(response)
+            out_df = pd.DataFrame.from_dict(re, orient="index")
```
liuverta: I don't have any great ideas for what variable name to use, but we should avoid `re` because that's [Python's regex library](https://docs.python.org/3/library/re.html).

```suggestion
            data = _utils.body_to_json(response)
            out_df = pd.DataFrame.from_dict(data, orient="index")
```

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -323,12 +326,86 @@ def predict_with_id(
...
+            out_df_list.append(out_df)
+        # Reassemble output before returning to user
+        out_df = pd.concat(out_df_list)
```
liuverta: As a nit, I might recommend not reusing the `out_df` variable name as both the final concatenated DataFrame _and_ the batch output DataFrame inside the `for` loop. But if it takes more than 3 seconds to think of a better name, then it's fine to just leave this 🤷

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+        status=200,
+        headers={"verta-request-id": "hereISaTESTidFROMtheUSER"},
+        )
+    creds = EmailCredentials.load_from_os_env()
+    dm = DeployedModel(
+        prediction_url=PREDICTION_URL,
+        creds=creds,
+        token=TOKEN,
+        )
+    # the input below is entirely irrelevant since it"s smaller than the batch size
```
liuverta: did you cmd-F replace `'` with `"`

```suggestion
    # the input below is entirely irrelevant since it's smaller than the batch size
```

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+    prediction_df = dm.batch_predict(pd.DataFrame({"hi": "bye"}, index=[1]), 10)
+    # Since no index was provided, we can"t guarantee the index type for assertions
+    pd.testing.assert_frame_equal(expected_df.reset_index(drop=True), prediction_df.reset_index(drop=True))
```
liuverta: Hmmm I feel like resetting the index conceptually weakens the test (now we can't check index values if we ever wanted to).

`assert_frame_equal()` [appears](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html) to have a `check_index_type` parameter. Would that have worked?

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+        token=TOKEN,
+        )
+    # the input below is entirely irrelevant since it"s smaller than the batch size
+    prediction_df = dm.batch_predict(pd.DataFrame({"hi": "bye"}, index=[1]), 10)
+    # Since an index WAS provided, we should be able to assert with indexes included
+    pd.testing.assert_frame_equal(expected_df, prediction_df)
+
+
+def test_batch_predict_with_five_batches_of_one_with_no_indexes(mocked_responses) -> None:
+    """ Call batch_predict with five batches. """
```
liuverta: imo the test using five batches of one is more detail than is necessary for the test name + docstring. The key point would be that we have coverage of batch

```suggestion
def test_batch_predict_with_batches_with_no_indexes(mocked_responses) -> None:
    """ Call batch_predict with batches. """
```

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+                       pd.DataFrame({"D": [31, 32, 33, 34, 35, 36, 37, 38, 39, 40]}),
+                       pd.DataFrame({"E": [41, 42, 43, 44, 45, 46, 47, 48, 49, 50]}),
+                       ]
+    for expected_d in expected_d_list:
+        mocked_responses.add(
+            responses.POST,
+            BATCH_PREDICTION_URL,
+            json=expected_d.to_dict(orient="index"),
+            status=200,
+            headers={"verta-request-id": "hereISaTESTidFROMtheUSER"},
```
liuverta: I think we should remove the header from these tests because it's unrelated haha (unless we do need it??)

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+        )
+    # the input below is entirely irrelevant since it"s smaller than the batch size
+    prediction_df = dm.batch_predict(pd.DataFrame({"hi": "bye"}, index=[1]), 10)
+    # Since an index WAS provided, we should be able to assert with indexes included
+    pd.testing.assert_frame_equal(expected_df, prediction_df)
+
+
+def test_batch_predict_with_five_batches_of_one_with_no_indexes(mocked_responses) -> None:
+    """ Call batch_predict with five batches. """
+    expected_d_list = [pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}),
```
liuverta: I assume it should be `df` because the tests above use that in variable names

```suggestion
    expected_df_list = [pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}),
```

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+                       pd.DataFrame({"B": [11, 12, 13, 14, 15, 16, 17, 18, 19, 20]}),
+                       pd.DataFrame({"C": [21, 22, 23, 24, 25, 26, 27, 28, 29, 30]}),
+                       pd.DataFrame({"D": [31, 32, 33, 34, 35, 36, 37, 38, 39, 40]}),
+                       pd.DataFrame({"E": [41, 42, 43, 44, 45, 46, 47, 48, 49, 50]}),
+                       ]
+    for expected_d in expected_d_list:
```
liuverta:
```suggestion
    for expected_df in expected_df_list:
```

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+    creds = EmailCredentials.load_from_os_env()
+    dm = DeployedModel(
+        prediction_url=PREDICTION_URL,
+        creds=creds,
+        token=TOKEN,
+        )
+    input_df = pd.DataFrame({"a": [1, 2, 3, 4, 5], "b": [11, 12, 13, 14, 15]})
+    prediction_df = dm.batch_predict(input_df, 1)
+    expected_final_df = pd.concat(expected_d_list)
+    # Since no index was provided, we can"t guarantee the index type for assertions
```
liuverta:
```suggestion
    # Since no index was provided, we can't guarantee the index type for assertions
```

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+    pd.testing.assert_frame_equal(expected_final_df.reset_index(drop=True), prediction_df.reset_index(drop=True))
```
liuverta: Ditto my above comment about `assert_frame_equal(check_index_type)`

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+
+
+def test_batch_predict_with_five_batches_of_one_with_indexes(mocked_responses) -> None:
+    """ CCall batch_predict with five batches, where each dataframe has an explicitly defined index. """
```
liuverta:
```suggestion
    """ Call batch_predict with five batches, where each dataframe has an explicitly defined index. """
```

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +382,134 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+    expected_d_list = [pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}, index=["a", "b", "c", "d", "e", "f", "g", "h", "i", "j"]),
```
liuverta:
```suggestion
    expected_df_list = [pd.DataFrame({"A": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}, index=["a", "b", "c", "d", "e", "f", "g", "h", "i", "j"]),
```

liuverta (changes requested):
nitpicknitpick

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -3,6 +3,8 @@
 import os
 from typing import Any, Dict
 
+import numpy as np
+import pandas as pd
```
liuverta: We should make these

```python
pd = pytest.importorskip("pandas")
```

after `import pytest`

liuverta (changes requested):

client/verta/tests/unit_tests/test_deployed_model.py
ghost: I have a different tests with indexes, does that change your opinion?

client/verta/tests/unit_tests/test_deployed_model.py
ghost: can you tell I find-replaced all the single quotes? 😅

client/verta/tests/unit_tests/test_deployed_model.py
ghost: I tried the `check_index_type` and it didn't do enough 😭

ghost: well, somewhere along the line, number indexes are converting to string indexes and that's sketchy. I should look into that.

client/verta/tests/unit_tests/test_deployed_model.py
ghost: ya

client/verta/verta/deployment/_deployedmodel.py
```diff
@@ -130,20 +137,25 @@ def _predict(
...
                 headers=request_headers,
                 data=gzstream.read(),
-                )
+            )
         else:
+            # when passing json=x, requests sets `allow_nan=False` by default (as of 2.26.0), which we don't want
+            # so we're going to dump ourselves
+            body = json.dumps(x, allow_nan=True)
+            if not isinstance(body, bytes):
+                body = body.encode("utf-8")
```
liuverta: We can omit this since we're firmly in Python 3, now. `json.dumps()` will give us a `utf-8` `str`

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -379,3 +383,97 @@ def test_predict_400_error_message_missing(mocked_responses) -> None:
...
+            status=200,
+            )
+    creds = EmailCredentials.load_from_os_env()
+    dm = DeployedModel(
+        prediction_url=PREDICTION_URL,
+        creds=creds,
+        token=TOKEN,
+        )
+    input_df = pd.DataFrame({"a": [1, 2, 3, 4, 5], "b": [11, 12, 13, 14, 15]})
+    prediction_df = dm.batch_predict(input_df, 1)
```
liuverta: This test is a bit hard to read (and therefore probably hard to main/debug); it's unclear that the `batch_size=1` here forces us to have five batches because `input_df` has five rows, and that it's related to `expected_df_list` having five DataFrames.

But I don't know how I'd write it any better 🤷

liuverta (approved):
🚀

PR #3649, opened 2023-03-07
fix: [VUMM-629] Update client's XGBoost callback (and test_integrations.py::TestXGBoost::test_callback)

baasitsh-verta:
## Impact and Context

(https://verta.readthedocs.io/en/main/_autogen/verta.integrations.xgboost.verta_callback.html#verta.integrations.xgboost.verta_callback)

```
bst = xgb.train(
            params, dtrain,
            num_boost_round=num_rounds,
            evals=[(dtrain, train_dataset_name)],
bq.           callbacks=[verta_callback(experiment_run)],
        )
E           ValueError: Old style callback was removed in version 1.6.  See: [https://xgboost.readthedocs.io/en/latest/python/callbacks.html].
```

## Risks and Area of Effect

XGBoost Integration with ModelDB

## Testing
- [x] Unit test - test_integrations.py
```
=================================================================== test session starts ====================================================================
platform darwin -- Python 3.9.13, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/baasitsharief/code/VertaAI/modeldb/client/verta/tests, configfile: pytest.ini
plugins: anyio-3.5.0, hypothesis-6.68.2
collected 10 items                                                                                                                                         

tests/test_integrations.py ..........                                                                                                                [100%]

===================================================================== warnings summary =====================================================================
test_integrations.py: 10 warnings
  /Users/baasitsharief/code/VertaAI/modeldb/client/verta/verta/_internal_utils/_utils.py:89: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
    self.retry = Retry(

test_integrations.py::TestScikitLearn::test_patch_log
  /Users/baasitsharief/.pyenv/versions/anaconda3-2022.10/lib/python3.9/site-packages/sklearn/linear_model/_ridge.py:157: DeprecationWarning: The 'sym_pos' keyword is deprecated and should be replaced by using 'assume_a = "pos"'. 'sym_pos' will be removed in SciPy 1.11.0.
    return linalg.solve(A, Xy, sym_pos=True, overwrite_a=True).T

test_integrations.py::TestScikitLearn::test_patch_log
  /Users/baasitsharief/.pyenv/versions/anaconda3-2022.10/lib/python3.9/site-packages/sklearn/neural_network/_multilayer_perceptron.py:692: ConvergenceWarning: Stochastic Optimizer: Maximum iterations (200) reached and the optimization hasn't converged yet.
[...]

PR #3650, opened 2023-03-07
docs: Fix docstring typo

liuverta:
## Impact and Context

#3649 introduced an `s`, but

```
``eval_metric``\ s
```

already renders as

> `eval_metric`s

so the extra `s` is unnecessary.

## Risks and Area of Effect

—

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

—

## Reverting

Revert this PR.


baasitsh-verta (approved):
Looks good. Thanks a lot.

PR #3652, opened 2023-03-08
feature: [VRD-682] Client utility function to translate import module names into distribution package names

ewagner-verta:
## Impact and Context
This is the missing piece of functionality for the client method `check_model_dependencies` to work properly.

The [importlib.metadata module](https://importlib-metadata.readthedocs.io/en/stable/using.html#mapping-import-to-distribution-packages) is used to capture info about the distribution packages for each of the import modules extracted from a model class.

## Risks and Area of Effect
Fairly low risk, as this is net new functionality on the first iteration and it's all hidden from the docs for now.
The Python ecosystem does not enforce package naming conventions, and the `importlib` docs warn:
> Some editable installs, [do not supply top-level names](https://github.com/pypa/packaging-problems/issues/609), and thus this function is not reliable with such installs.

This seems to be related specifically to editable installs.  However I experimented with `verta` as an `editable` installation, and the `name` attribute we need is there.  

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
+++ b/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
@@ -41,8 +41,8 @@ def test_check_model_dependencies_complete(dependency_testing_model, complete_en
 
 
 def test_check_model_dependencies_missing_raise(dependency_testing_model, complete_env) -> None:
-    """ Verify that check_model_dependencies raises an exception for missing
-    packages when `raise_for_missing` is True.
+    """ Verify that check_model_dependencies raises an exception, with the
+    correct message, for missing packages when `raise_for_missing` is True.
     """
     incomplete_env = Python(
         [r for r in complete_env.requirements if r != 'click==0.0.1']
@@ -53,16 +53,18 @@ def test_check_model_dependencies_missing_raise(dependency_testing_model, comple
             environment=incomplete_env,
             raise_for_missing=True,
         )
-    assert err.value.args[0] == "the following packages are required by the model " \
-                                "but missing from the environment: {'click'}"
+    assert err.value.args[0] == "the following import modules are required by the model " \
+                                "but missing any corresponding distribution package in the" \
+                                " environment:\n{'import_module': 'click', " \
+                                "'distribution_packages': ['click']}"
 
 
 def test_check_model_dependencies_missing_warning(dependency_testing_model, complete_env) -> None:
-    """ Verify that check_model_dependencies defaults to raising a warning for
-    missing packages when `raise_for_missing` is False.
+    """ Verify that check_model_dependencies defaults to raising a warning, with
+    the correct message, for missing packages when `raise_for_missing` is False.
     """
     incomplete_env = Python(
-        [r for r in complete_env.requirements if r != 'pandas==0.0.1']
+        [r for r in complete_env.requirements if r not in ['pandas==0.0.1', 'PyYAML==0.0.1']]
     )  # drop a single dependency to be caught
     with warnings.catch_warnings(record=True) as caught_warnings:
         assert not _check_model_dependencies(
@@ -71,5 +73,8 @@ def test_check_model_dependencies_missing_warning(dependency_testing_model, comp
             raise_for_missing=False,
         )
     warn_msg = caught_warnings[0].message.args[0]
-    assert warn_msg == "the following packages are required by the model but " \
-                       "missing from the environment: {'pandas'}"
+    assert warn_msg == "the following import modules are required by the model " \
+                       "but missing any corresponding distribution package in the" \
+                       " environment:\n{'import_module': 'pandas', 'distribution_packages':" \
+                       " ['pandas']}\n{'import_module': 'yaml', 'distribution_packages': " \
+                       "['PyYAML']}"
--- a/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
+++ b/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
@@ -16,7 +16,7 @@ def complete_env() -> Python:
     """
     return Python([
         'click==0.0.1',
-        'google==0.0.1',
+        'googleapis-common-protos==0.0.1',
         'numpy==0.0.1',
         'pandas==0.0.1',
         'Pillow==0.0.1',
--- a/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
+++ b/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
@@ -53,10 +53,8 @@ def test_check_model_dependencies_missing_raise(dependency_testing_model, comple
             environment=incomplete_env,
             raise_for_missing=True,
         )
-    assert err.value.args[0] == "the following import modules are required by the model " \
-                                "but missing any corresponding distribution package in the" \
-                                " environment:\n{'import_module': 'click', " \
-                                "'distribution_packages': ['click']}"
+    assert err.value.args[0] == "the following packages are required by the model but missing " \
+                                "from the environment:\nclick (installed via ['click'])"
 
 
 def test_check_model_dependencies_missing_warning(dependency_testing_model, complete_env) -> None:
@@ -64,7 +62,7 @@ def test_check_model_dependencies_missing_warning(dependency_testing_model, comp
     the correct message, for missing packages when `raise_for_missing` is False.
     """
     incomplete_env = Python(
-        [r for r in complete_env.requirements if r not in ['pandas==0.0.1', 'PyYAML==0.0.1']]
+        [r for r in complete_env.requirements if r not in ['pandas==0.0.1']]
     )  # drop a single dependency to be caught
     with warnings.catch_warnings(record=True) as caught_warnings:
         assert not _check_model_dependencies(
@@ -73,8 +71,5 @@ def test_check_model_dependencies_missing_warning(dependency_testing_model, comp
             raise_for_missing=False,
         )
     warn_msg = caught_warnings[0].message.args[0]
-    assert warn_msg == "the following import modules are required by the model " \
-                       "but missing any corresponding distribution package in the" \
-                       " environment:\n{'import_module': 'pandas', 'distribution_packages':" \
-                       " ['pandas']}\n{'import_module': 'yaml', 'distribution_packages': " \
-                       "['PyYAML']}"
+    assert warn_msg == "the following packages are required by the model but missing " \
+                       "from the environment:\npandas (installed via ['pandas'])"
--- a/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
+++ b/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
@@ -62,7 +62,7 @@ def test_check_model_dependencies_missing_warning(dependency_testing_model, comp
     the correct message, for missing packages when `raise_for_missing` is False.
     """
     incomplete_env = Python(
-        [r for r in complete_env.requirements if r not in ['PyYAML==0.0.1']]
+        [r for r in complete_env.requirements if r not in ['PyYAML==0.0.1', 'pandas==0.0.1']]
     )  # drop a single dependency to be caught
     with warnings.catch_warnings(record=True) as caught_warnings:
         assert not _check_model_dependencies(
@@ -72,4 +72,5 @@ def test_check_model_dependencies_missing_warning(dependency_testing_model, comp
         )
     warn_msg = caught_warnings[0].message.args[0]
     assert warn_msg == "the following packages are required by the model but missing " \
-                       "from the environment:\nyaml (installed via ['PyYAML'])"
+                       "from the environment:\npandas (installed via ['pandas'])" \
+                       "\nyaml (installed via ['PyYAML'])"
--- a/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
+++ b/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
@@ -62,7 +62,7 @@ def test_check_model_dependencies_missing_warning(dependency_testing_model, comp
     the correct message, for missing packages when `raise_for_missing` is False.
     """
     incomplete_env = Python(
-        [r for r in complete_env.requirements if r not in ['pandas==0.0.1']]
+        [r for r in complete_env.requirements if r not in ['PyYAML==0.0.1']]
     )  # drop a single dependency to be caught
     with warnings.catch_warnings(record=True) as caught_warnings:
         assert not _check_model_dependencies(
@@ -72,4 +72,4 @@ def test_check_model_dependencies_missing_warning(dependency_testing_model, comp
         )
     warn_msg = caught_warnings[0].message.args[0]
     assert warn_msg == "the following packages are required by the model but missing " \
-                       "from the environment:\npandas (installed via ['pandas'])"
+                       "from the environment:\nyaml (installed via ['PyYAML'])"
--- a/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
+++ b/client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
@@ -45,7 +45,7 @@ def test_check_model_dependencies_missing_raise(dependency_testing_model, comple
     packages when `raise_for_missing` is True.
     """
     incomplete_env = Python(
-        [ r for r in complete_env.requirements if r != 'click==0.0.1' ]
+        [r for r in complete_env.requirements if r != 'click==0.0.1']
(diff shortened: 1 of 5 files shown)


client/verta/verta/registry/_check_model_dependencies.py
```diff
@@ -23,7 +23,8 @@ def _check_model_dependencies(
     provided environment.
 
     .. note::
-        This function is not guaranteed to detect all dependencies in all cases, and should be considered a fast sanity check rather than a test.
+        This function is not guaranteed to detect all dependencies in all cases,
+        and should be considered a fast sanity check rather than a test.
```
ewagner-verta, line 27: The "commit suggestion" button edited this on a previous PR.  This change just adds a newline to make it readable in the file.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -73,7 +74,24 @@ def class_module_names(model_class: Type[VertaModelBase]) -> Set[str]:
...
+    """
+    pkg_dist = packages_distributions()
+    pkgs = set()
+    for mod in module_names:
+        dist_pkgs: Union[List[str], None] = pkg_dist.get(mod)
+        if dist_pkgs:
+            if len(dist_pkgs) == 1:
+                pkgs.add(dist_pkgs[0])
+            if len(dist_pkgs) > 1:
+                pkgs.add(distribution(mod).name)
```
ewagner-verta: A "namespace distribution package" will have multiple options for packages that _could_ install the given module.

```python
from importlib.metadata import packages_distributions
pkgs = packages_distributions()
print(pkgs.get('google'))
```
Output:
`['googleapis-common-protos', 'google-cloud-bigquery', 'protobuf', 'google-cloud-core', 'google-api-core', 'google-resumable-media', 'google-auth']`
Hence why the length is checked and the two flavors are handled differently

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -53,7 +45,7 @@ def test_check_model_dependencies_missing_raise(dependency_testing_model, comple
     packages when `raise_for_missing` is True.
     """
     incomplete_env = Python(
-        [ r for r in complete_env.requirements if r != 'click==0.0.1' ]
+        [r for r in complete_env.requirements if r != 'click==0.0.1']
```
ewagner-verta, line 48: Figured out how to run the auto-formatter in PyCharm.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -32,9 +33,9 @@ def modules_in_function_body(func: Callable) -> Set[str]:
         if isinstance(value, ModuleType)
     ]
     _non_locals = [
-       value.__name__.split('.')[0]  # strip off submodules and classes
-       for key, value in inspect.getclosurevars(_func).nonlocals.items()
-       if isinstance(value, ModuleType)
+        value.__name__.split('.')[0]  # strip off submodules and classes
+        for key, value in inspect.getclosurevars(_func).nonlocals.items()
+        if isinstance(value, ModuleType)
```
ewagner-verta, line 38: auto formatted

client/verta/verta/_internal_utils/model_dependencies.py
liuverta: It doesn't seem that this works 🤔 

```python
distribution("google").name
```

raises

```
PackageNotFoundError: No package metadata was found for google
```

liuverta: I'm actually not sure what is meant to happen here. I think the logic we might want is "`pkgs.get('google')` returns a handful of potential packages for this module; the user's environment` should supply at least one of them to hopefully provide that module."

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -73,7 +74,24 @@ def class_module_names(model_class: Type[VertaModelBase]) -> Set[str]:
...
+    This fact is used to capture only 3rd-party packages.  Modules associated
+    with a single package name are easily captured.  Modules associated with a
+    namespace package (e.g.`google`), may come from a variety of distribution
+    packages.  To avoid making an arbitrary choice in those cases, the main
+    namespace package name is fetched from the distribution metadata.
+    """
+    pkg_dist = packages_distributions()
+    pkgs = set()
+    for mod in module_names:
+        dist_pkgs: Union[List[str], None] = pkg_dist.get(mod)
```
liuverta: The docstring makes it sounds like `dist_pkgs` is only `None` if `mod` comes from the standard library, but this doesn't seem to be the case:

Screen Shot 2023-03-08 at 1 42 00 PM

We might want to at least log those cases in an `else` clause below.

liuverta (changes requested):

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: `google` is a [namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) , so you could run `pip install google` and get all the distribution packages under that umbrella.  In that case `packages_distributions().get('google')` returns the list of them all.  The challenge here is handling both cases:  User env has a single package installed that contains the required module OR user env has the whole namespace package.  I _think_ I've found a way to do that "gracefully" but it required tweaking return types and using dictionaries to track modules with their list of possible packages.  Check out the latest commits.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: This seems like something we'll need to investigate, but I propose we do it in a separate PR.  I have a new branch with the fix for modules introduced as Class instances (problem you raised in this [slack thread](https://vertaai.slack.com/archives/C04NVN48LAV/p1678310608042619)). We could wrap the two things together in that

liuverta: A follow-up PR is alright, but please don't bundle it together with the other issue—they're unrelated things, as far as I can tell.

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -49,28 +41,30 @@ def test_check_model_dependencies_complete(dependency_testing_model, complete_en
...
             model_cls=dependency_testing_model,
             environment=incomplete_env,
             raise_for_missing=True,
         )
-    assert err.value.args[0] == "the following packages are required by the model " \
-                                "but missing from the environment: {'click'}"
+    assert err.value.args[0] == "the following import modules are required by the model " \
+                                "but missing any corresponding distribution package in the" \
+                                " environment:\n{'import_module': 'click', " \
+                                "'distribution_packages': ['click']}"
```
liuverta: I don't think users (especially data scientists) are going to know what a "distribution package" or "import module" is.

I'd recommend just saying "the following packages are required by the model but missing from the environment" and list just the distribution package names (concatenating all the namespace packages, because that's the best we can do for now).

Or if we have to get in-depth, something like "yaml (installed via ['PyYAML']), pandas (installed via ['pandas'])".

client/verta/tests/unit_tests/registry/test_model_dependencies.py
```diff
@@ -126,5 +126,26 @@ def test_class_module_names(dependency_testing_model) -> None:
...
+
+def test_package_names(dependency_testing_model) -> None:
+    """Verify that distribution package names are returned correctly for the
+    given module names when they differ from the package name.  Test fixture
+    ensures packages required for the test are installed or this test will be
+    skipped.
+    """
+    expected_packages = {
+        'sklearn': ['scikit-learn'],
+        'PIL': ['Pillow', ],
```
liuverta: Maybe black would catch this in a future PR, but

```suggestion
        'PIL': ['Pillow'],
```

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -73,7 +74,11 @@ def class_module_names(model_class: Type[VertaModelBase]) -> Set[str]:
...
-    """Return a set of all packages detected in the provided set of base module names."""
-    return module_names
+def package_names(module_names: Set[str]) -> Dict[str, List[str]]:
+    """Return a dictionary where the key is the name of the import module
+    and the value is the list of possible distribution packages for each
+    3rd-party import module in the provided set that can be found in the
+    locally installed package_distributions.
+    """
+    pkg_dist = packages_distributions()
+    return {m: pkg_dist.get(m) for m in module_names if pkg_dist.get(m)}
```
liuverta, line 84: I know you mentioned covering this in a follow-up PR, but I'd strongly recommend putting at least a `# TODO` here about cases where `pkg_dist.get(m)` returns `None`. It's a sneaky case that's hard to catch when just looking at this code, and I don't want it to be missed forever (if, say, we end up not getting around to the follow-up).

client/verta/verta/registry/_check_model_dependencies.py
```diff
@@ -60,14 +61,26 @@ def _check_model_dependencies(
...
+    env_packages: Set[str] = {parse_req_spec(e)[0] for e in environment.requirements}
     detected_modules: Set[str] = md.class_module_names(model_cls)
-    detected_packages: Set[str] = md.package_names(detected_modules)
-    env_packages: Set[str] = { parse_req_spec(e)[0] for e in environment.requirements }
+    detected_packages: Dict[str, List[str]] = md.package_names(detected_modules)
+
+    missing_packages: Dict[str, List[str]] = {
+        mod: pkgs
+        for mod, pkgs in detected_packages.items()
+        if not env_packages.intersection(set(pkgs))
```
liuverta: Trying to remove potentially unnecessary nesting because this comprehension is already quite complex.

```suggestion
        if not env_packages.intersection(pkgs)
```

client/verta/verta/registry/_check_model_dependencies.py
```diff
@@ -60,14 +61,26 @@ def _check_model_dependencies(
...
+           and mod not in env_packages  # namespace package
```
liuverta: It's a bit surprising to me that we'd need to check for `mod not in env_packages` if we're already checking `pkg`s 🤔 Are you sure we need this?

liuverta (changes requested):

client/verta/tests/unit_tests/registry/test_model_dependencies.py
ewagner-verta: fixed

client/verta/verta/registry/_check_model_dependencies.py
ewagner-verta: Ah, I did not realize you could just use `intersection` with a list.  Thanks

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Done

client/verta/verta/registry/_check_model_dependencies.py
ewagner-verta: Yeah, once again `google` is the example problem child.  A user could just do `Python(['google'])`.  The result of `pip install google` would be the installation of ~5 different packages, none of which is named "google".  So when we compare packages, we have to add the plain "google" option to the list of acceptable package options for supporting the given module.  We could move that logic to `package_names()` instead of here...

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: Closing, since there are multiple threads on this issue

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
ewagner-verta: Updated

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
```diff
@@ -49,28 +41,28 @@ def test_check_model_dependencies_complete(dependency_testing_model, complete_en
...
 
 def test_check_model_dependencies_missing_warning(dependency_testing_model, complete_env) -> None:
-    """ Verify that check_model_dependencies defaults to raising a warning for
-    missing packages when `raise_for_missing` is False.
+    """ Verify that check_model_dependencies defaults to raising a warning, with
+    the correct message, for missing packages when `raise_for_missing` is False.
     """
     incomplete_env = Python(
-        [ r for r in complete_env.requirements if r != 'pandas==0.0.1' ]
+        [r for r in complete_env.requirements if r not in ['pandas==0.0.1']]
```
liuverta: Out of curiosity, why was `PyYAML` [removed from this test](https://github.com/VertaAI/modeldb/pull/3652/commits/e47bab11310125cf6cb4feec68b8f9bf1d69c853)? Seems like it would have been useful to have coverage of a module whose name differed from its package.

client/verta/verta/registry/_check_model_dependencies.py
liuverta: Hmmm I'm still not sure about this. If `md.package_names()` says "The root module `'google'` is supplied by the pip-installable packages `['googleapis-common-protos', 'google-cloud-bigquery', 'protobuf', 'google-cloud-core', 'google-api-core', 'google-resumable-media', 'google-auth']`", and `env_packages` contains pip-installable package names, I don't quite see the purpose of also checking for the root module name.

> The result of `pip install google` would be the installation of ~5 different packages

This isn't true.

liuverta (changes requested):

client/verta/verta/registry/_check_model_dependencies.py
ewagner-verta: Just installed `google` on a clean virtual env and you are absolutely right.  `pip install google` just installs beautifulsou4 and a google search module, and even then the pkg_dist name is "googlesearch".  I don't know how my main modeldb virtual env has a package named "google" in `packages_distributions()`, but it does and it's causing the unit tests to fail when I don't catch it.

ewagner-verta: In any case, passing "google" as a dependency to the env would not work to support the modules we're using in the test case.  I'll pull this last line of the comprehension out and adjust the test env to use any of the appropriate google packages.

ewagner-verta: Done.

client/verta/tests/unit_tests/registry/test_check_model_dependencies.py
ewagner-verta: Having 2 or more missing packages in this test would have required more refactoring to deal with the ordering of the dictionary returned so that the error message can be tested.  (Ordering of the missing packages in the error message is random because each is pulled from a dict).  We already have explicit coverage of "module name != pkg name" in `test_model_dependencies.test_package_names()`, but there's no harm in swapping in `yaml` for `pandas` in this test.

ewagner-verta: Just added that change.  Even though the util function is tested, it's nice to see the final step tested here as well.

liuverta: It can be a good practice to sort items before inserting them in the error message, to make things more deterministic.

We should want to cover cases with multiple missing packages.

liuverta: And as an aside,

> Ordering of the missing packages in the error message is random because each is pulled from a dict

It's not quite accurate to say it's random; Python dictionaries maintain order.

liuverta (changes requested):
I re-opened a prior comment thread: https://github.com/VertaAI/modeldb/pull/3652#discussion_r1132829201. We should have some test coverage of multiple missing modules.

That should be the last thing! Everything else looks good.

client/verta/verta/registry/_check_model_dependencies.py
```diff
@@ -60,14 +61,21 @@ def _check_model_dependencies(
...
+        for mod, pkgs in detected_packages.items()
+        if not env_packages.intersection(pkgs)
+    }
 
-    missing_packages = detected_packages - env_packages
     if missing_packages:
         error_msg = f"the following packages are required by the model but missing " \
-                    f"from the environment: {missing_packages}"
+                    f"from the environment:"
+        for m, p in missing_packages.items():
```
liuverta:
```suggestion
        for m, p in sorted(missing_packages.items(), key=lambda item: item[0]):
```

I think?

ewagner-verta: Accepted changes and updated the test

liuverta (approved):
🚀

PR #3656, opened 2023-03-09
feat: [VRD-698] Verify model against code_dependencies

liuverta:
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain)

I used `pandas` here just because I was too lazy to create another custom module of my own.

Screen Shot 2023-03-09 at 3 07 09 PM

## Reverting


liuverta:
Cut from scope 😢

PR #3659, opened 2023-03-10
fix: [VRD-682] Fix module extraction for class instances in function body

ewagner-verta:
## Impact and Context
Prior to this fix, any 3rd-party module introduced via a class instance inside the function body will not be identified and extracted.

Example:
```python
from pandas import DataFrame

class MyModel(VertaModelBase):
    def __init__(self, artifacts):
        pass
    @verify_io
    def predict(self, x):
        df = DataFrame(x)
        return df
```
`pandas` would not be detected as a dependency of the model.  This change fixes that.

## Risks and Area of Effect
Low.  Unreleased client functionality that is well tested.

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


client/verta/tests/unit_tests/registry/test_model_dependencies.py
```diff
@@ -108,6 +119,7 @@ def test_class_module_names(dependency_testing_model) -> None:
     """ Verify that all expected module names are extracted as expected.
     """
     expected_modules = {
+        'builtins',
```
ewagner-verta, line 131: Another std-lib package that gets caught by the new logic.  Ignored like the others.

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -27,17 +27,22 @@ def unwrap(func: Callable) -> Callable:
...
         if isinstance(value, ModuleType)
-    ]
-    _non_locals = [
-        value.__name__.split('.')[0]  # strip off submodules and classes
-        for key, value in inspect.getclosurevars(_func).nonlocals.items()
-        if isinstance(value, ModuleType)
-    ]
-    return set(_globals + _non_locals)
+    }
+    _global_classes = {
```
liuverta: This variable name makes it sound like this set contains classes or class names, but it actually contains modules right?

client/verta/tests/unit_tests/registry/test_model_dependencies.py
```diff
@@ -52,6 +53,16 @@ def test_modules_in_function_body_return_line(dependency_testing_model) -> None:
     assert extracted_modules == expected_modules
 
 
+def test_modules_in_function_body_as_class_instance(dependency_testing_model) -> None:
+    """ Verify that modules introduced only via class instance are extracted
+    as expected.
```
liuverta, line 59: What does the "instance" in "class instance" mean exactly? Isn't it just a "class", not an instance of one?

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -27,17 +27,22 @@ def unwrap(func: Callable) -> Callable:
...
+        inspect.getmodule(value).__name__.split('.')[0]  # strip off submodules
+        for key, value in inspect.getclosurevars(_func).globals.items()
+        if inspect.isclass(value)
```
liuverta: How about functions?

```python
from verta._internal_utils import model_dependencies as md

from numpy import array
from pandas import DataFrame

class Model():
    def predict(self):
        arr = array([1, 2, 3])
        df = DataFrame(arr)
        return df

assert md.class_module_names(Model) == {"pandas"}  # no numpy
```

Could we try doing something more general, if possible? Data scientists could import anything from libraries.

liuverta (dismissed):

client/verta/tests/unit_tests/registry/test_model_dependencies.py
ewagner-verta: The means by which this module is introduced is the declaration of an instance of a class provided by a 3rd-party module.  Meaning `DataFrame(some_data)` declares an instance of the DataFrame class, vs. a type annotation that may simply use the class itself `df: DataFrame = make_a_df()`.  The distinction matters because each case is handled in a slightly different manner.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: image
 Apparently, when a function is used from a 3rd party module like this, the value of `type()` is "class 'builtin_function_or_method'".  However,  the inspect functions `isclass()`, `isfunction()` and `is method()`. all return `False`.  Only `inspect.isbuiltin()` returns `True`.

ewagner-verta: Ok, I think I've found a way to make this simpler and more generic by excluding only `unbound` findings from [inspect.getclosurevars](https://docs.python.org/3/library/inspect.html#inspect.getclosurevars) which allows everything else to be caught.  That function sorts out things that could not be resolved (into the 'unbound` category), then [inspect.getmodule](https://docs.python.org/3/library/inspect.html#inspect.getmodule) is already set up to handle the various types.

client/verta/verta/_internal_utils/model_dependencies.py
ewagner-verta: obsolete

client/verta/verta/_internal_utils/model_dependencies.py
```diff
@@ -25,19 +25,21 @@ def unwrap(func: Callable) -> Callable:
...
-        for key, value in inspect.getclosurevars(_func).nonlocals.items()
-        if isinstance(value, ModuleType)
+    closure_vars: List[Dict[str, Any]] = [
+        val
+        for key, val in inspect.getclosurevars(_func)._asdict().items()
+        if key != "unbound"
     ]
-    return set(_globals + _non_locals)
+    return {
+        inspect.getmodule(val).__name__.split(".")[0]
```
ghost: Could you rewrite this section to be clearer? I have truly no idea what's happening here(: maybe taking these lines out of the return and adding some variable names could help?

ewagner-verta: Any time I start nesting inside of a comprehension, I ask myself - "Should I really be doing this?"  The answer is usually no.  I dropped that in favor of a clearer, more readable process.  Let me know what you think.

ghost (approved):

PR #3664, opened 2023-03-14
docs: Mention importlib-metadata in client changelog

liuverta:
## Impact and Context

I totally missed this! It's here now.

## Risks and Area of Effect

I'm not sure whether to consider a dependency addition a "backwards incompatibility". It's what I've been doing, and it's true that the client isn't compatible with environments that _don't_ have the dependency.

Still, I think `importlib-metadata` is lightweight enough that it doesn't warrant any corrective action, nor a minor version release.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

Built and rendered locally:

Screen Shot 2023-03-14 at 2 10 58 AM

## Reverting

Revert this PR.

### DIFF
--- a/client/verta/requirements.txt
+++ b/client/verta/requirements.txt
@@ -1,7 +1,7 @@
 click >= 7.0, < 9.0
 cloudpickle >= 1.0, < 3.0
 googleapis-common-protos >= 1.5, < 2.0
-importlib_metadata >= 3.7.0
+importlib-metadata >= 3.7.0
 protobuf >= 3.8, < 4.0
 pytimeparse >= 1.1.8, < 2.0
 pyyaml >= 5.1, < 7.0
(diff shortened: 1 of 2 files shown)


client/verta/requirements.txt
```diff
@@ -1,7 +1,7 @@
 click >= 7.0, < 9.0
 cloudpickle >= 1.0, < 3.0
 googleapis-common-protos >= 1.5, < 2.0
-importlib_metadata >= 3.7.0
+importlib-metadata >= 3.7.0
```
liuverta, line 4: Switched to a hyphen to be consistent with how the package is listed in PyPI: https://pypi.org/project/importlib-metadata/

Doesn't seem to make a difference for `pip`, though.

ewagner-verta: Oh Python

ewagner-verta (approved):

PR #3666, opened 2023-03-15
Fix sonarqube suggested bugs

anandjakhaniya:
## Impact and Context
This PR only contains all `try-with-resource` changes.

**Before:**
This is from my local sonar setup.
image (3)

**After:**
image

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/config/Config.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/config/Config.java
@@ -36,7 +36,7 @@ public abstract class Config {
   @JsonProperty private boolean event_system_enabled = false;
   @JsonProperty private ServerInterceptor tracingServerInterceptor = null;
   @JsonProperty private ClientInterceptor tracingClientInterceptor = null;
-  @JsonProperty private volatile OpenTelemetry openTelemetry;
+  @JsonProperty private OpenTelemetry openTelemetry;
   @JsonProperty private ArtifactStoreConfig artifactStoreConfig;
 
   @JsonProperty private boolean permissionV2Enabled = false;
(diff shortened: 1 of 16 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/config/Config.java
```diff
@@ -36,7 +36,7 @@ public abstract class Config {
   @JsonProperty private boolean event_system_enabled = false;
   @JsonProperty private ServerInterceptor tracingServerInterceptor = null;
   @JsonProperty private ClientInterceptor tracingClientInterceptor = null;
-  @JsonProperty private volatile OpenTelemetry openTelemetry;
+  @JsonProperty private OpenTelemetry openTelemetry;
```
jkwatson-verta, line 39: This field isn't actually used by anything, anywhere, as far as I can tell. We should remove it.

backend/server/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
```diff
@@ -139,9 +142,6 @@ public static boolean needToRetry(Exception ex) {
   }
 
   public static Throwable findCommunicationsFailedCause(Throwable throwable) {
-    if (throwable == null) {
-      return null;
-    }
     var rootCause = throwable;
     while (rootCause.getCause() != null
         && !(rootCause.getCause() instanceof SocketException
```
jkwatson-verta, line 147: are we sure this null check isn't necessary? It looks like if we skip it, we'll get an NPE on line 146 if a null `throwable` is passed in.

jkwatson-verta (approved):
Just one question about a deleted null check.

backend/server/src/main/java/ai/verta/modeldb/utils/ModelDBUtils.java
anandjakhaniya: We have that check already but at the one step ahead where this function is called.

PR #3673, opened 2023-03-17
docs: [VRD-703] Add batch_predict to verta model base

ghost:
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/client/verta/verta/registry/_verta_model_base.py
+++ b/client/verta/verta/registry/_verta_model_base.py
@@ -119,7 +119,7 @@ class VertaModelBase(object):
         .. [1] https://pandas.pydata.org/docs/reference/frame.html
 
         """
-        pass
+        raise NotImplementedError
 
     def model_test(self):
         """Test a model's behavior for correctness.
--- a/client/verta/verta/registry/_verta_model_base.py
+++ b/client/verta/verta/registry/_verta_model_base.py
@@ -109,11 +109,11 @@ class VertaModelBase(object):
 
         Parameters
         ----------
-        df : a pandas.DataFrame
+        df : pandas.DataFrame
 
         Returns
         -------
-        a pandas.DataFrame
+        pandas.DataFrame
 
         References
         ----------
--- a/client/verta/verta/registry/_verta_model_base.py
+++ b/client/verta/verta/registry/_verta_model_base.py
@@ -102,7 +102,7 @@ class VertaModelBase(object):
         .. note::
 
             :meth:`batch_predict` must be written to both receive and return
-            pandas.DataFrames [1]_.
+            ``pandas.DataFrame``\ s [1]_.
 
             At this time, your subclass must still implement :meth:`predict`, regardless
             of whether it contains anything meaningful.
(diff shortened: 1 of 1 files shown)


client/verta/verta/registry/_verta_model_base.py
```diff
@@ -93,6 +93,35 @@ def predict(self, input):
         """
         raise NotImplementedError
 
+    @abc.abstractmethod
```
liuverta:
```suggestion
```

`@abc.abstractmethod` makes it required, but this method is actually optional (um I think, right?) so we should omit the decorator.

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -93,6 +93,35 @@ def predict(self, input):
...
+        Returns
+        -------
+        a pandas.DataFrame
+
+        References
+        ----------
+        .. [1] https://pandas.pydata.org/docs/reference/frame.html
+
+        """
+        pass
```
liuverta: And since it's optional, and wouldn't make sense to be called unless the user implements it, we should raise an exception if this method isn't overridden

```suggestion
        raise NotImplementedError
```

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -93,6 +93,35 @@ def predict(self, input):
...
+            At this time, your subclass must still implement :meth:`predict`, regardless
+            of whether it contains anything meaningful.
+
+        Parameters
+        ----------
+        df : a pandas.DataFrame
+
+        Returns
+        -------
+        a pandas.DataFrame
```
liuverta:
```suggestion
        Parameters
        ----------
        df : pandas.DataFrame

        Returns
        -------
        pandas.DataFrame
```

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -67,7 +67,7 @@ def predict(self, input):
             will be fully compatible with the Verta platform as you iterate
             locally.
 
-            :meth:`predict` must be written to both recieve [1]_ and return
+            :meth:`predict` must be written to both receive [1]_ and return
```
liuverta, line 70: UGH gets me every time

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -93,6 +93,35 @@ def predict(self, input):
...
+    @abc.abstractmethod
+    def batch_predict(self, df):
+        """Produce an output from `df`.
+
+        This method is called when batch predictions are made against a Verta endpoint.
+
+        .. note::
+
+            :meth:`batch_predict` must be written to both receive and return
+            pandas.DataFrames [1]_.
```
liuverta:
```suggestion
            ``pandas.DataFrame``\ s [1]_.
```

liuverta (changes requested):

client/verta/verta/registry/_verta_model_base.py
ghost: I see! I was thinking of it the opposite way but what you're saying makes sense

client/verta/verta/registry/_verta_model_base.py
ghost: I of course can't remember it now but I have a similar nemesis

liuverta (approved):

PR #3677, opened 2023-03-21
Improve code with sonarqube suggestions

anandjakhaniya:
## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
@@ -31,7 +31,6 @@ import org.springframework.web.server.ResponseStatusException;
 public class CommonUtils {
   private static final Logger LOGGER = LogManager.getLogger(CommonUtils.class);
   private static final int STACKTRACE_LENGTH = 4;
-  private static final String LOGGER_PLACE_HOLDER = "{}: {}";
 
   public static String appendOptionalTelepresencePath(String filePath) {
     String telepresenceRoot = System.getenv("TELEPRESENCE_ROOT");
@@ -208,18 +207,18 @@ public class CommonUtils {
       if (isLongStack) {
         for (; n < STACKTRACE_LENGTH + 1; ++n) {
           if (isClientError) {
-            LOGGER.debug(LOGGER_PLACE_HOLDER, n, stack[n]);
+            LOGGER.debug("{}: {}", n, stack[n]);
           } else {
-            LOGGER.warn(LOGGER_PLACE_HOLDER, n, stack[n]);
+            LOGGER.warn("{}: {}", n, stack[n]);
           }
         }
       }
       for (; n < stack.length; ++n) {
         if (stack[n].getClassName().startsWith("ai.verta") || !isLongStack) {
           if (isClientError) {
-            LOGGER.debug(LOGGER_PLACE_HOLDER, n, stack[n]);
+            LOGGER.debug("{}: {}", n, stack[n]);
           } else {
-            LOGGER.warn(LOGGER_PLACE_HOLDER, n, stack[n]);
+            LOGGER.warn("{}: {}", n, stack[n]);
           }
         }
       }
@@ -348,12 +347,12 @@ public class CommonUtils {
     boolean isLongStack = stack.length > STACKTRACE_LENGTH;
     if (isLongStack) {
       for (; n < STACKTRACE_LENGTH + 1; ++n) {
-        logger.warn(LOGGER_PLACE_HOLDER, n, stack[n].toString());
+        logger.warn("{}: {}", n, stack[n].toString());
       }
     }
     for (; n < stack.length; ++n) {
       if (stack[n].getClassName().startsWith("ai.verta") || !isLongStack) {
-        logger.warn(LOGGER_PLACE_HOLDER, n, stack[n].toString());
+        logger.warn("{}: {}", n, stack[n].toString());
       }
     }
   }
--- a/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
+++ b/backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
@@ -18,7 +18,6 @@ import com.google.rpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.grpc.protobuf.StatusProto;
 import io.grpc.stub.StreamObserver;
-import java.io.File;
 import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.List;
@@ -32,6 +31,7 @@ import org.springframework.web.server.ResponseStatusException;
 public class CommonUtils {
   private static final Logger LOGGER = LogManager.getLogger(CommonUtils.class);
   private static final int STACKTRACE_LENGTH = 4;
+  private static final String LOGGER_PLACE_HOLDER = "{}: {}";
 
   public static String appendOptionalTelepresencePath(String filePath) {
     String telepresenceRoot = System.getenv("TELEPRESENCE_ROOT");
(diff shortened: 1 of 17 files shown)


backend/common/src/main/java/ai/verta/modeldb/common/artifactStore/storageservice/s3/S3Client.java
```diff
@@ -28,9 +28,9 @@ public class S3Client {
 
   private final String bucketName;
 
-  private volatile AmazonS3 s3Client;
-  private volatile AtomicInteger referenceCounter;
-  private volatile AWSCredentials awsCredentials;
```
jkwatson-verta: I don't think it's safe to remove these. We're swapping references, and to guarantee visibility across threads, we need to have them marked as `volatile` or to have some other thread-safe way to make sure there is a memory barrier around the swapping.

jkwatson-verta (changes requested):

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
```diff
@@ -34,6 +34,10 @@
   private static boolean captureStacksAtCreation;
 
   public static final AttributeKey<String> STACK_ATTRIBUTE_KEY = AttributeKey.stringKey("stack");
+  public static final String FUTURE_EXECUTOR_REQUIRED_ERROR =
+      "A FutureExecutor is required to create a new Future.";
+  public static final String CALL_WITH_EXECUTOR_FUTURE_EXECUTOR_REQUIRED_ERROR =
+      "A FutureExecutor is required to be present for this method. Please call withExecutor before calling this.";
```
jkwatson-verta: we should fix this message while we're here. Just use the FUTURE_EXECUTOR_REQUIRED_ERROR. This message is outdated and refers to a method that doesn't exist any more. :)

backend/server/src/test/java/ai/verta/modeldb/ArtifactStore/NFSArtifactStoreTest.java
```diff
@@ -324,7 +324,7 @@ public void getUrlForArtifactTest() throws IOException {
 
       int responseCode = httpClient.getResponseCode();
       LOGGER.info("POST Response Code :: {}", responseCode);
-      assumeTrue(responseCode == HttpURLConnection.HTTP_OK);
+      Assertions.assertEquals(responseCode, HttpURLConnection.HTTP_OK);
```
jkwatson-verta: would generally prefer statically importing these methods

backend/common/src/main/java/ai/verta/modeldb/common/artifactStore/storageservice/s3/S3Client.java
anandjakhaniya: As per the SonarQube, it is not required. it said that if there is an object then no need to add volatile so I have removed it. but Yes, I agree with you, we have to keep it for the thread safe so reverting this.

anandjakhaniya: Reverted it.

backend/common/src/main/java/ai/verta/modeldb/common/futures/Future.java
anandjakhaniya: Used `FUTURE_EXECUTOR_REQUIRED_ERROR` instead of `CALL_WITH_EXECUTOR_FUTURE_EXECUTOR_REQUIRED_ERROR`

backend/server/src/test/java/ai/verta/modeldb/ArtifactStore/NFSArtifactStoreTest.java
anandjakhaniya: Used assert with static

backend/common/src/main/java/ai/verta/modeldb/common/CommonUtils.java
```diff
@@ -32,6 +31,7 @@
 public class CommonUtils {
   private static final Logger LOGGER = LogManager.getLogger(CommonUtils.class);
   private static final int STACKTRACE_LENGTH = 4;
+  private static final String LOGGER_PLACE_HOLDER = "{}: {}";
```
jkwatson-verta: this name is meaningless. Please give it a name that makes it obvious what it's being used for, or just inline the string. It seems to provide negative value to obscure what the logging message format is from the reader at the call site.

anandjakhaniya: reverted this

backend/server/src/main/java/ai/verta/modeldb/experiment/FutureExperimentDAO.java
```diff
@@ -228,20 +235,16 @@ public InternalFuture<FindExperiments.Response> findExperiments(FindExperiments
...
-                                                      .setVersionNumber(
-                                                          rs.getLong("version_number"));
-                                              return runBuilder;
-                                            })
+                                            (rs, ctx) ->
+                                                Experiment.newBuilder()
+                                                    .setId(rs.getString("id"))
+                                                    .setProjectId(rs.getString("project_id"))
+                                                    .setName(rs.getString("name"))
+                                                    .setDescription(rs.getString(DESCRIPTION_FIELD))
```
jkwatson-verta: this doesn't seem like a good change. Makes it harder when there's a mix of constants and plain Strings to see what's going on. Recommend reverting.

backend/server/src/main/java/ai/verta/modeldb/experiment/FutureExperimentDAO.java
```diff
@@ -471,7 +474,7 @@ private InternalFuture<Map<String, String>> getProjectIdByExperimentId(
                   "select id, project_id from experiment where id IN (<ids>) AND deleted = :deleted")) {
             List<Map<String, String>> experimentEntitiesMap =
                 query
-                    .bind("deleted", false)
+                    .bind(ModelDBConstants.DELETED, false)
```
jkwatson-verta: I don't think this is a good change. It just obfuscates what's going on.

backend/server/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
```diff
@@ -283,7 +283,7 @@ private void insertVersioningInput(
                 + " VALUES (:repository_id, :commit, :versioning_key, :versioning_location, :entity_type, :versioning_blob_type, :blob_hash, :entityId)",
             App.getInstance().mdbConfig.getDatabase().getRdbConfiguration().isMssql()
                 ? "\"commit\""
-                : "commit",
+                : COMMIT_QUERY_PARAM,
```
jkwatson-verta: would revert this. Just makes things more confusing

backend/server/src/main/java/ai/verta/modeldb/experiment/FutureExperimentDAO.java
anandjakhaniya: Reverted

backend/server/src/main/java/ai/verta/modeldb/experiment/FutureExperimentDAO.java
anandjakhaniya: Reverted

backend/server/src/main/java/ai/verta/modeldb/experimentRun/subtypes/VersionInputHandler.java
anandjakhaniya: Reverted

jkwatson-verta (approved):

PR #3709, opened 2023-03-30
Change package name based on SonarQube suggestions

anandjakhaniya:
## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


jkwatson-verta:
this will break UAC and the registry, correct?

anandjakhaniya:
> this will break UAC and the registry, correct?

Yes correct, Registry: https://github.com/VertaAI/registry/pull/1993 and UAC will not break.

PR #3714, opened 2023-03-31
feature: [VRD-768] Add client utility functions for Kafka

ewagner-verta:
## Impact and Context
Adds new utility functions to allow the fetching of Kafka configurations and valid Kafka topics to assist users in configuration of the Kafka integration and endpoints.  This PR is the 

## Risks and Area of Effect
Based on testing in my dev env, it seems that the following error is returned BOTH when no Kafka config exists and when a config exists but the user performing the search is not a system admin.  We'll need to sort out the permissions issue there.
```
requests.exceptions.HTTPError: 400 Client Error: Kafka validation failed (check broker addresses?): : unable to dial: dial tcp :9092: connect: connection refused for url: https://ewagner.dev.verta.ai/api/v1/deployment/list/kafka-topics at <TIME>
```

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

Screen Shot 2023-03-30 at 7 44 24 PM

## Reverting

### DIFF
--- a/client/verta/tests/unit_tests/conftest.py
+++ b/client/verta/tests/unit_tests/conftest.py
@@ -6,6 +6,7 @@ import os
 from unittest.mock import patch
 
 import pytest
+import responses
 
 from verta._internal_utils._utils import Connection, Configuration
 from verta.credentials import EmailCredentials
@@ -27,3 +28,9 @@ def mock_conn() -> Connection:
 def mock_config() -> Configuration:
     """ Return a mocked object of the _internal_utils._utils.Configuration class for use in tests """
     return Configuration(use_git=False, debug=False)
+
+
+@pytest.fixture
+def mocked_responses():
+    with responses.RequestsMock() as rsps:
+        yield rsps
--- a/client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
+++ b/client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
@@ -1,0 +1,101 @@
+# -*- coding: utf-8 -*-
+
+"""Unit tests for Kafka utilities."""
+
+from typing import Any, Dict, List
+
+import pytest
+
+from verta._internal_utils import kafka
+
+
+@pytest.fixture
+def mock_kafka_configs_response() -> Dict[str, Any]:
+    """
+    Provide mocked API result from `api/v1/uac-proxy/system_admin/listKafkaConfiguration`
+    with a single Kafka configuration.
+    """
+    return {
+        "configurations": [
+            {
+                "id": "1232abcdefg4-567h-891i-0jkl-1m1n1o2131pq",
+                "kerberos": {
+                    "enabled": True,
+                    "conf": (
+                        "[libdefaults]\ndefault_realm ="
+                        " EXAMPLE.COM\n\n[realms]\nEXAMPLE.COM"
+                        " = {\n    kdc_ports = 12,345\n    kadmind_port = 123\n    kdc ="
+                        " testing--kafka-server-X.integrations--kafka-server-headless.ci.svc.cluster.local\n"
+                        "    admin_server ="
+                        " testing--kafka-server-X.integrations--kafka-server-headless.ci.svc.cluster.local\n}"
+                    ),
+                    "keytab": "SLG45KJlksajdfglakjWlskadfj098lkjakdfjl1k3kjflakdj-lakf+alkqfvmLPWZNVTdfS84UHads3dlkfj97alfkj!lakdfjlkaj8fg=",
+                    "client_name": "test-kafka-client_name",
+                    "service_name": "test-kafka-utils",
+                },
+                "brokerAddresses": "integrations--kafka-X.integrations--kafka-headless.ci.svc.cluster.local:00001",
+                "enabled": True,
+                "name": "Test Kafka Name",
+            }
+        ]
+    }
+
+
+@pytest.fixture
+def mock_kafka_topics() -> List[str]:
+    """Provide a mock API response from `api/v1/deployment/list/kafka-topics`."""
+    return ["test-topic-1", "test-topic-2", "test-topic-3"]
+
+
+def test_list_kafka_configurations(
+    mocked_responses, mock_conn, mock_kafka_configs_response
+):
+    """Verify that list_kafka_configurations function makes the expected calls
+    and handles results correctly.
+    """
+    url = "https://test_socket/api/v1/uac-proxy/system_admin/listKafkaConfiguration"
+    mocked_responses.get(url=url, status=200, json=mock_kafka_configs_response)
+    config = kafka.list_kafka_configurations(mock_conn)
+    mocked_responses.assert_call_count(url, 1)
+    assert config == mock_kafka_configs_response["configurations"]
+
+
+def test_format_kafka_config_for_topic_search(mock_kafka_configs_response) -> None:
+    """Verify that format_kafka_config_for_topic_search function formats the
+    mocked kafka config into the expected results.
+    """
+    mock_config = mock_kafka_configs_response["configurations"][0]
+    config = kafka.format_kafka_config_for_topic_search(mock_config)
+    assert config == {
+        "broker_addresses": [
+            "integrations--kafka-X.integrations--kafka-headless.ci.svc.cluster.local:00001"
+        ],
+        "kerberos": {
+            "enabled": True,
+            "conf": (
+                "[libdefaults]\ndefault_realm = EXAMPLE.COM\n\n[realms]\nEXAMPLE.COM"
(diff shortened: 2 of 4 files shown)


client/verta/tests/unit_tests/conftest.py
```diff
@@ -27,3 +28,9 @@ def mock_conn() -> Connection:
 def mock_config() -> Configuration:
     """ Return a mocked object of the _internal_utils._utils.Configuration class for use in tests """
     return Configuration(use_git=False, debug=False)
+
+
+@pytest.fixture
+def mocked_responses():
+    with responses.RequestsMock() as rsps:
+        yield rsps
```
ewagner-verta, line 36: Just moved this here to `contest.py` since it can now be reused but the new tests.

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+"""Unit tests for Kafka utilities."""
+
+from typing import Any, Dict, List
+
+import pytest
+
+from verta._internal_utils import kafka
+
+
```
ewagner-verta, line 11: This fixture has been sanitized.  IDs, keytab, addresses etc. are fake.

client/verta/tests/unit_tests/test_deployed_model.py
```diff
@@ -35,12 +35,6 @@
 VERTA_CLASS = 'verta.deployment._deployedmodel'
 
 
-@pytest.fixture
-def mocked_responses():
-    with responses.RequestsMock() as rsps:
-        yield rsps
```
ewagner-verta, line 41: Moved to `conftest.py`

client/verta/verta/_internal_utils/kafka.py
```diff
@@ -0,0 +1,50 @@
...
+"""Utilities for working with Kafka."""
+
+from typing import Any, Dict, List
+
+from verta._internal_utils import _utils
+
+
+def list_kafka_configurations(conn: _utils.Connection) -> List[Dict[str, Any]]:
+    """Make an HTTP call to fetch a list of Kafka configurations. If no
+    configurations exist, an empty dict is returned by the API.  In that
```
liuverta: Is this true for the actual `/listKafkaConfiguration`? That seems like a backend bug. If there are no configurations, I'd expect a list with zero elements, not a list with one empty item. We should mention this in the Slack channel

ewagner-verta: Moving conversation to `#feat-kafka-deployment`

ewagner-verta: Also, I've updated the doc string to "fetch a dict of Kafka configurations" to be more clear about the API return type

ewagner-verta: [Slack Thread](https://vertaai.slack.com/archives/C024FCNP404/p1680278268843749)

client/verta/verta/_internal_utils/kafka.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+
+"""Utilities for working with Kafka."""
```
liuverta, line 3: I'll pass along some advice I had been given before, which is that it's not a _great_ practice to have a bunch of standalone util functions (even though the client has a lot of them—my fault 😛). It creates a risk of turning util modules into catch-all buckets of unorganized, vaguely-related functions that are hard to sift through and maintain.

Something more extensible would be to start organizing things into an structure like

```python
class Kafka:  # can be abstracted into a `MessageQueue` interface in the future?
    def __init__(self, conn):
        self._conn = conn

    @staticmethod
    def _format_kafka_config_for_topic_search(config):
        pass

    def list_configurations(self):
        pass

    def list_topics(self, config):
        pass

File: client/verta/verta/_internal_utils/kafka.py [hunk 6]
```diff
@@ -0,0 +1,50 @@
...
+        f"{conn.scheme}://{conn.socket}/api/v1/uac-proxy/system_admin/listKafkaConfiguration",
+        conn,
+    )
+    _utils.raise_for_http_error(response)
+    return response.json().get("configurations", [])
+
+
+def format_kafka_config_for_topic_search(config: Dict[str, Any]) -> Dict[str, Any]:
+    """Extract and format relevant data from a Kafka configuration for use
+    with the `deployment/list//kafka-topics` API.
```
liuverta:
```suggestion
    with the `deployment/list/kafka-topics` API.
```

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -0,0 +1,101 @@
...
+        "configurations": [
+            {
+                "id": "1232abcdefg4-567h-891i-0jkl-1m1n1o2131pq",
+                "kerberos": {
+                    "enabled": True,
+                    "conf": (
+                        "[libdefaults]\ndefault_realm ="
+                        " EXAMPLE.COM\n\n[realms]\nEXAMPLE.COM"
+                        " = {\n    kdc_ports = 12,345\n    kadmind_port = 123\n    kdc ="
+                        " testing--kafka-server-X.integrations--kafka-server-headless.ci.svc.cluster.local\n"
```
liuverta: I think this kind of reveals how we're running Kafka in our cluster. I'd recommend replacing these (and `"brokerAddresses"` below) with maybe `kerberos.example.com`?

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -0,0 +1,101 @@
...
+                " = {\n    kdc_ports = 12,345\n    kadmind_port = 123\n    kdc ="
+                " testing--kafka-server-X.integrations--kafka-server-headless.ci.svc.cluster.local\n"
+                "    admin_server ="
+                " testing--kafka-server-X.integrations--kafka-server-headless.ci.svc.cluster.local\n}"
+            ),
+            "keytab": "SLG45KJlksajdfglakjWlskadfj098lkjakdfjl1k3kjflakdj-lakf+alkqfvmLPWZNVTdfS84UHads3dlkfj97alfkj!lakdfjlkaj8fg=",
+            "client_name": "test-kafka-client_name",
+            "service_name": "test-kafka-utils",
+        },
+    }
```
liuverta: I think having these values hard-coded into the test makes a suite harder to maintain if we make changes to it in the future. Would you be opposed to re-using values directly from the original object?

```suggestion
    assert config == {
        "broker_addresses": [mock_config["brokerAddresses"]],
        "kerberos": mock_config["kerberos"],
    }
```

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -0,0 +1,101 @@
...
+                "name": "Test Kafka Name",
+            }
+        ]
+    }
+
+
+@pytest.fixture
+def mock_kafka_topics() -> List[str]:
+    """Provide a mock API response from `api/v1/deployment/list/kafka-topics`."""
+    return ["test-topic-1", "test-topic-2", "test-topic-3"]
```
liuverta, line 42: If you'd like (and I'd mildly recommend it), these two fixtures would be straightforward to use `hypothesis` for.

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_list_kafka_configurations(
+    mocked_responses, mock_conn, mock_kafka_configs_response
+):
+    """Verify that list_kafka_configurations function makes the expected calls
+    and handles results correctly.
+    """
+    url = "https://test_socket/api/v1/uac-proxy/system_admin/listKafkaConfiguration"
```
liuverta: Could we grab `"https"` and `"test_socket"` from the `mock_conn` object? It's not obvious to me that they're intrinsically tied to that fixture value, and could be a pain to debug if the test started failing.

liuverta (approved):

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
ewagner-verta: Done

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
ewagner-verta: Done

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
ewagner-verta: I'm going to try to tackle this in another PR.

PR #3716, opened 2023-03-31
test: [VRD-768] Convert kafka utils unit tests to property tests

ewagner-verta:
## Impact and Contex
Converts the recently added unit tests for Kafka utilities into property tests.

## Risks and Area of Effect
Change to testing only.

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

Screen Shot 2023-03-31 at 3 06 33 PM

## Reverting

### DIFF
--- a/client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
+++ b/client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
@@ -2,64 +2,57 @@
 
 """Unit tests for Kafka utilities."""
 
-from typing import Any, Dict, List
+from typing import Any, Dict
 
-import pytest
+from hypothesis import given, HealthCheck, settings, strategies as st
 
 from verta._internal_utils import kafka
 
 
-@pytest.fixture
-def mock_kafka_configs_response() -> Dict[str, Any]:
+@st.composite
+def mock_kafka_configs_response(draw) -> Dict[str, Any]:
     """
     Provide mocked API result from `api/v1/uac-proxy/system_admin/listKafkaConfiguration`
     with a single Kafka configuration.
     """
+    recursive_json = st.recursive(
+        st.none() | st.booleans() | st.integers() | st.text(),
+        lambda children: st.lists(children) | st.dictionaries(st.text(), children),
+        max_leaves=5,
+    )
     return {
         "configurations": [
             {
-                "id": "1232abcdefg4-567h-891i-0jkl-1m1n1o2131pq",
-                "kerberos": {
-                    "enabled": True,
-                    "conf": (
-                        "kerberos.example.com"
-                    ),
-                    "keytab": "test-key-tab=",
-                    "client_name": "test-kafka-client_name",
-                    "service_name": "test-kafka-utils",
-                },
-                "brokerAddresses": "kerberos.example.com",
-                "enabled": True,
-                "name": "Test Kafka Name",
+                "id": draw(st.integers()),
+                "kerberos": draw(recursive_json),
+                "brokerAddresses": draw(st.text()),
+                "enabled": draw(st.booleans()),
+                "name": draw(st.text()),
             }
         ]
     }
 
 
-@pytest.fixture
-def mock_kafka_topics() -> List[str]:
-    """Provide a mock API response from `api/v1/deployment/list/kafka-topics`."""
-    return ["test-topic-1", "test-topic-2", "test-topic-3"]
-
-
-def test_list_kafka_configurations(
-    mocked_responses, mock_conn, mock_kafka_configs_response
-):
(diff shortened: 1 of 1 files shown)


client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -2,81 +2,80 @@
...
     url = f"{mock_conn.scheme}://{mock_conn.socket}/api/v1/uac-proxy/system_admin/listKafkaConfiguration"
-    mocked_responses.get(url=url, status=200, json=mock_kafka_configs_response)
-    config = kafka.list_kafka_configurations(mock_conn)
-    mocked_responses.assert_call_count(url, 1)
-    assert config == mock_kafka_configs_response["configurations"]
+    with mocked_responses as _responses:
+        _responses.get(url=url, status=200, json=kafka_configs_response)
+        config = kafka.list_kafka_configurations(mock_conn)
+        _responses.assert_call_count(url, 1)
+        assert config == kafka_configs_response["configurations"]
```
ewagner-verta, line 48: This fixture has to be used as a context manager in order to be compatible with hypothesis.

[This guide from hypothesis](https://hypothesis.works/articles/hypothesis-pytest-fixtures/) says:
>In pytest you can declare fixtures which do set up and tear down per function. These will “work” with Hypothesis, but they will run once for the entire test function rather than once for each time given calls your test function

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -2,81 +2,80 @@
...
-    mocked_responses.assert_call_count(url, 1)
-    assert topics == mock_kafka_topics
+    url = f"{mock_conn.scheme}://{mock_conn.socket}/api/v1/deployment/list/kafka-topics"
+    with mocked_responses as _responses:
+        _responses.post(url=url, status=200, json=kafka_topics)
+        topics = kafka.list_kafka_topics(
+            conn=mock_conn, kafka_config=kafka_configs_response["configurations"][0]
+        )
+        _responses.assert_call_count(url, 1)
+        assert topics == kafka_topics
```
ewagner-verta, line 82: Same here: mocked_responses fixture ---> context manager

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
liuverta: Oh, interesting 🤔 Is it because otherwise the same `mocked_responses` instance would get used between tests?

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -2,81 +2,80 @@
...
+def mock_kafka_configs_response(draw) -> Dict[str, Any]:
     """
     Provide mocked API result from `api/v1/uac-proxy/system_admin/listKafkaConfiguration`
     with a single Kafka configuration.
     """
+    recursive_json = st.recursive(
+        st.none() | st.booleans() | st.integers() | st.text(),
+        lambda children: st.lists(children) | st.dictionaries(st.text(), children),
+        max_leaves=5,
+    )
```
liuverta: We have something almost identical to this as [`tests.strategies.json_strategy`](https://github.com/VertaAI/modeldb/blob/08681e3541f15a67000e21e05aa053a754fa26e7/client/verta/tests/strategies.py#L25-L30). Could we possibly reuse that?

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -2,81 +2,80 @@
...
-                    ),
-                    "keytab": "test-key-tab=",
-                    "client_name": "test-kafka-client_name",
-                    "service_name": "test-kafka-utils",
-                },
-                "brokerAddresses": "kerberos.example.com",
-                "enabled": True,
-                "name": "Test Kafka Name",
+                "id": draw(st.integers()),
+                "kerberos": draw(recursive_json),
```
liuverta: This doesn't seem right. `"kerberos"` isn't an arbitrarily nested unstructured JSON; it has specific keys and value types, right?

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
ewagner-verta: Exactly.  The fixture is only generated once, then repeating runs of the function by hypothesis means that the count of calls is cumulative, so the assertion fails

liuverta: Ugh dang. I'll have to keep that in mind when looking at our mocked tests 😕

liuverta (approved):

PR #3727, opened 2023-04-05
feature: [VRD-679] New client function `get_kafka_topics()`

ewagner-verta:
## Impact and Context
* Adds a new function to `verta.client.Client` that will allow users to fetch a list of available Kafka topics.
* Updates `KafkaSettings` class to make the `cluster_config_id` argument optional.
* Refactors `verta._endpoint.Endpoint()` to fetch the Kafka config ID when it is not included.
* Adds property/unit tests to verify behavior with and without `cluster_config_id`.
* Adds property/unit tests for new client function.
* Includes some minor `black` formatting changes to bring changed files up to standard.

## Testing
- [X] Unit test
- **test_endpoint.py**
Screen Shot 2023-04-05 at 12 07 16 PM

- **test_client_get_kafka_topics.py**
Screen Shot 2023-04-05 at 12 09 16 PM

- [ ] Deployed to dev env
- [X] Other (explain) 
Validation script for running against a real env with a Kafka config:
```python
from verta import Client
from verta.endpoint import KafkaSettings
from verta.environment import Python
from verta.registry import verify_io, VertaModelBase


class Model(VertaModelBase):
    def __init__(self, artifacts=None):
        pass

    @verify_io
    def predict(self, input):
        return input


def main():
    client = Client()

    model_ver = client.get_or_create_registered_model(
        f"Kafka Model Testing"
    ).create_standard_model(
        Model,
        code_dependencies=[],
        environment=Python([]),
    )

    kafka_settings = KafkaSettings(
        input_topic="inputTopic",
        output_topic="outputTopic",
        error_topic="errorTopic",
    )
    assert kafka_settings._cluster_config_id is None
    print(f"\nKafka Settings Provided\n----------------------\n{kafka_settings}\n")

    client.create_endpoint("kafka-model-testing", kafka_settings=kafka_settings)
    endpoint = client.get_endpoint("kafka-model-testing")
    assert endpoint.kafka_settings.cluster_config_id is not None

    print(f"\nFinal Kafka Settings\n--------------------\n{endpoint.kafka_settings}\n")

if __name__ == "__main__":
    main()
```
Output:
```plaintext
got VERTA_HOST from environment
got VERTA_EMAIL from environment
got VERTA_DEV_KEY from environment
created new RegisteredModel: Kafka Model Testing in workspace: ewagner
created new ModelVersion: ModelVersion 421771680719193545579
uploading model to Registry
upload complete
uploading model_api.json to Registry
upload complete
uploading custom_modules to Registry
upload complete

Kafka Settings Provided
----------------------
KafkaSettings('inputTopic', 'outputTopic', 'errorTopic', None)

Final Kafka Settings
--------------------
KafkaSettings('inputTopic', 'outputTopic', 'errorTopic', 'd5a8b0b8-c4ab-4d75-97e7-ecddc494402f')
```
Note the `cluster_config_id` (4th arg) has been fetched.

## Reverting


client/verta/tests/unit_tests/conftest.py
```diff
@@ -8,29 +8,50 @@
...
-    """ Return a mocked object of the _internal_utils._utils.Connection class for use in tests """
-    with patch.dict(os.environ, {'VERTA_EMAIL': 'test_email@verta.ai', 'VERTA_DEV_KEY':'123test1232dev1232key123'}):
+    """Return a mocked object of the _internal_utils._utils.Connection class for use in tests"""
+    with patch.dict(
+        os.environ,
+        {
+            "VERTA_EMAIL": "test_email@verta.ai",
+            "VERTA_DEV_KEY": "123test1232dev1232key123",
+        },
+    ):
```
ewagner-verta, line 36: `black` formatting change only

client/verta/tests/unit_tests/internal_utils/test_kafka_utils.py
```diff
@@ -6,34 +6,10 @@
...
-                    "conf": draw(st.text()),
-                    "keytab": draw(st.text()),
-                    "service_name": draw(st.text()),
-                },
-                "brokerAddresses": draw(st.text()),
-                "enabled": draw(st.booleans()),
-                "name": draw(st.text()),
-            }
-        ]
-    }
```
ewagner-verta, line 34: moved to `strategies.py`

client/verta/tests/unit_tests/strategies.py
```diff
@@ -21,7 +22,7 @@ def uint64(draw) -> int:
 @st.composite
```
ewagner-verta, line 22: Any changes to the existing strategies are `black` formatting only

client/verta/tests/unit_tests/strategies.py
```diff
@@ -95,3 +98,49 @@ def code_blob_proto(draw) -> Code_pb2.CodeBlob:
...
+                "name": draw(st.text()),
+            }
+        ]
+    }
+
+
+@st.composite
+def mock_kafka_settings(draw) -> KafkaSettings:
+    """Generate a mocked KafkaSettings object with no value for cluster config ID."""
+    topics = draw(st.lists(st.text(min_size=1), min_size=3, max_size=3, unique=True))
```
ewagner-verta, line 131: has to be unique to avoid errors thrown when the input topic matches output or error topics.

client/verta/tests/unit_tests/test_endpoint.py
```diff
@@ -2,109 +2,207 @@
...
 
 import pytest
+from hypothesis import given, HealthCheck, settings
 
+from tests.unit_tests.strategies import (
+    mock_kafka_configs_response,
+    mock_kafka_settings,
+    mock_kafka_settings_with_config_id,
+)
 from verta.endpoint import Endpoint
```
ewagner-verta, line 18: Changes from lines 20 - 123 are `black` formatting only

client/verta/tests/unit_tests/test_endpoint.py
```diff
@@ -2,109 +2,207 @@
...
+) -> None:
+    """Verify that the get_deployed_model method returns the correct value for the full_url when it is returned
+    by the get_json_by_id method.  The 'full_url' key was added to the schema of the response 10/2022.
     """
     deployed_model = mock_endpoint.get_deployed_model()
-    assert deployed_model.prediction_url == 'https://test_socket/api/v1/predict/test_path'
+    assert (
+        deployed_model.prediction_url == "https://test_socket/api/v1/predict/test_path"
+    )
+
```
ewagner-verta, line 128: new tests start here >>

client/verta/verta/client.py
```diff
@@ -1782,3 +1784,26 @@ def _create_workspace(self, org_id, workspace_name, resource_action_groups):
...
+
+        Raises
+        ------
+        HTTPError
+            If no valid Kafka configuration can be found.
+        """
+        kafka_configs: List[Dict[str, Any]] = kafka.list_kafka_configurations(self._conn)
+        if kafka_configs:
+            return kafka.list_kafka_topics(self._conn, kafka_configs[0])
+        return []
```
ewagner-verta, line 1809: Rendered docs:
Screen Shot 2023-04-05 at 12 42 09 PM

client/verta/verta/endpoint/_kafka_settings.py
```diff
@@ -33,15 +40,14 @@ class KafkaSettings(object):
             input_topic="my_input_data",
             output_topic="my_predictions",
             error_topic="my_endpoint_errors",
-            cluster_config_id="kafka_config_id",
         )
 
         endpoint = client.create_endpoint(path="/my-endpoint", kafka_settings=kafka_settings)
         print(endpoint.kafka_settings)
```
ewagner-verta, line 45: Rendered docs:
Screen Shot 2023-04-05 at 12 52 02 PM

client/verta/tests/unit_tests/conftest.py
```diff
@@ -8,29 +8,50 @@
...
 import responses
 
-from verta._internal_utils._utils import Connection, Configuration
+from verta._internal_utils._utils import Configuration, Connection
+from verta._protos.public.registry import RegistryService_pb2 as _RegistryService
+from verta.client import Client
 from verta.credentials import EmailCredentials
+from verta.registry.entities import RegisteredModelVersion
+
+
```
ewagner-verta, line 17: new fixture >>

client/verta/tests/unit_tests/conftest.py
```diff
@@ -8,29 +8,50 @@
...
+@pytest.fixture(scope="session")
+def mock_client(mock_conn) -> Client:
+    """Return a mocked object of the Client class for use in tests"""
+    # with patch.dict(os.environ, {'VERTA_EMAIL': 'test_email@verta.ai', 'VERTA_DEV_KEY':'123test1232dev1232key123'}):
+    client = Client(ignore_conn_err=True, _connect=False)
+    client._conn = mock_conn
+    return client
 
 
 @pytest.fixture(scope="session")
```
ewagner-verta, line 27: `black` formatting only >>

client/verta/tests/unit_tests/strategies.py
```diff
@@ -95,3 +98,49 @@ def code_blob_proto(draw) -> Code_pb2.CodeBlob:
         proto.notebook.CopyFrom(draw(notebook_code_blob_proto()))
 
     return proto
+
+
+@st.composite
```
ewagner-verta, line 103: moved from `test_kafka_utils.py` >>

client/verta/tests/unit_tests/strategies.py
liuverta: Minor point, but I think the more idiomatic way to write this would be something like

```python
input_topic = draw(st.text(min_size=1))
output_topic = draw(st.text(min_size=1))
error_topic = draw(st.text(min_size=1))
hypothesis.assume(input_topic != output_topic != error_topic)
```

because using the `st.lists(min_size=3, max_size=3)` generator purely as a mechanism to ensure uniqueness feels like...co-opting it for something it's not meant to be.

client/verta/verta/endpoint/_kafka_settings.py
```diff
@@ -10,6 +10,12 @@ class KafkaSettings(object):
     For use during :class:`~verta.endpoint.Endpoint` creation or with
     :meth:`Endpoint.update() <verta.endpoint.Endpoint.update>`.
 
+    Search for available Kafka topics with :meth:`Client.get_kafka_topics() <verta.Client.get_kafka_topics>`.
+
+    .. versionadded:: 0.22.3
```
liuverta: This feels more like a `versionchanged` to me, personally

client/verta/verta/endpoint/_kafka_settings.py
```diff
@@ -10,6 +10,12 @@ class KafkaSettings(object):
...
+        ``cluster_config_id`` is now fetched from the active Kafka
+        configuration by default.  It remains optional for backwards
+        compatibility.
```
liuverta: I think we can leave this out.

```suggestion
        configuration if not provided, if there is only one configured.
```

It's not optional for backwards compatibility; like we mentioned in Slack, it's for the future possibility of multiple Kafka clusters. Either way, I don't think we need to justify it in the docstring 😬

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -508,6 +508,9 @@ def _set_kafka_settings(self, stage_id, kafka_settings):
                 "`kafka_settings` must be of type verta.endpoint.KafkaSettings,"
                 " not {}".format(type(kafka_settings))
             )
+        if kafka_settings._cluster_config_id is None:
+            config = kafka.list_kafka_configurations(self._conn)[0]
+            kafka_settings._cluster_config_id = config.get("id", None) if config else None
```
liuverta: What error does the backend return if we pass `None` as the config ID?

liuverta: I'm asking because I feel like if `config` doesn't contain `"id"` (`config.get("id", None)`), or if `config` is somehow empty (`if config`), tyhen that should be an error condition rather than just defaulting to `None`.

But if the backend will return an error a couple lines below, then that's probably fine.

client/verta/tests/unit_tests/conftest.py
```diff
@@ -8,29 +8,50 @@
...
+from verta.client import Client
 from verta.credentials import EmailCredentials
+from verta.registry.entities import RegisteredModelVersion
+
+
+@pytest.fixture(scope="session")
+def mock_client(mock_conn) -> Client:
+    """Return a mocked object of the Client class for use in tests"""
+    # with patch.dict(os.environ, {'VERTA_EMAIL': 'test_email@verta.ai', 'VERTA_DEV_KEY':'123test1232dev1232key123'}):
+    client = Client(ignore_conn_err=True, _connect=False)
```
liuverta: `ignore_conn_err` might not be what we want. I think it does some wild thing where it converts `5XX`s and `4XX`s into `200`s, which might mess with some tests. But maybe not?

client/verta/tests/unit_tests/test_endpoint.py
```diff
@@ -2,109 +2,207 @@
...
+    )
+    put_config_url = f"{deployment_url}/stages/654/kafka"
+    builds_url = f"{deployment_url}/workspace/{WORKSPACE_ID}/builds"
+
+    get_configs_url = f"{mock_conn.scheme}://{mock_conn.socket}/api/v1/uac-proxy/system_admin/listKafkaConfiguration"
+
+    with mocked_responses as _responses:
+        # Mock all HTTP requests involved
+        _responses.get(url=stages_url, status=200, json={"id": 654})
+        _responses.get(url=stages_url + "/654", status=200, json={"id": 654})
```
liuverta: I'd really recommend accessing `mock_registered_model_version.id` instead of hard-coding the value into tests. It's not clear in this test which values are kind of arbitrary (like `"cluster_config_id": 987`) versus which are actually tied to fixture values and shouldn't be changed without checking other bits of code.

liuverta (changes requested):

client/verta/tests/unit_tests/conftest.py
ewagner-verta: Dropping it does not seem to affect the outcome of the tests, so I think it's best to rip it out and avoid any unnecessary confusion when writing future tests that may use the fixture.  >> Done

client/verta/tests/unit_tests/strategies.py
ewagner-verta: The use of `min_size=3, max_size=3` is just meant to ensure it has the general shape of a Kafka config (input, output, error topics) which must be a list of 3 text items.  The `unique=True` arg is what ensures the values in the list are not the same in order to avoid the topic name check (input cannot = output or error)

ewagner-verta: I don't have any objection to using the suggestion though, for the record.  It is more readable.

ewagner-verta: Huh.  Maybe I'm not doing this right but this change:
```python
@st.composite
def mock_kafka_settings(draw) -> KafkaSettings:
    """Generate a mocked KafkaSettings object with no value for cluster config ID."""
    input_topic = draw(st.text(min_size=1))
    output_topic = draw(st.text(min_size=1))
    error_topic = draw(st.text(min_size=1))
    assume(input_topic != output_topic != error_topic)
    return KafkaSettings(
        input_topic=input_topic,
        output_topic=output_topic,
        error_topic=error_topic
    )
```
is failing with 
`ValueError: input_topic must not be equal to either the output or error topics`

client/verta/tests/unit_tests/test_endpoint.py
ewagner-verta: I have stripped out all the arbitrary ID numbers in the test to make things a bit clearer.  All of these global variables should be converted into property tests at some point, but that seems like it should be a separate PR.

client/verta/verta/endpoint/_endpoint.py
ewagner-verta: `list_kafka_configurations` itself will raise an HTTPError if there is no valid config, but I don't know what happens if we actually manage to pass `None` as the ID to back-end.  I'll test that out real quick

ewagner-verta: It looks like passing `None` along to the backend throws
`requests.exceptions.HTTPError: 500 Server Error: Kafka config ID does not correlate with valid kafka configuration for url: https://ewagner.dev.verta.ai/api/v1/deployment/stages/237/kafka` in the next step.

A 500 seems appropriate, since this would have to be caused by us breaking the "contract" in the API response.

client/verta/verta/endpoint/_kafka_settings.py
ewagner-verta: Good call.  Fixed

client/verta/verta/endpoint/_kafka_settings.py
ewagner-verta: Fixed

ewagner-verta:
!rebuild

liuverta (commented):
There are past comments that weren't addressed. Are they fine?

- https://github.com/VertaAI/modeldb/pull/3727#discussion_r1158922361
- https://github.com/VertaAI/modeldb/pull/3727#discussion_r1158936274

ewagner-verta:
@liuverta Did you happen to see my prior responses in each of the two threads you referenced?

In the [first case](https://github.com/VertaAI/modeldb/pull/3727#discussion_r1158922361), using `assume` to ensure uniqueness in a list as suggested is failing for me, so I'm asking if you can make a specific code suggestion there or can we leave it working as is?

In the [second case](https://github.com/VertaAI/modeldb/pull/3727#discussion_r1158936274), sending a `None` value to the back-end would result in a 500 error that I think is ok to pass along as is.  Again, if you have a specific suggestion or vision for that process, can you share?

client/verta/tests/unit_tests/strategies.py
liuverta: Alright, let's just go with the original.

client/verta/verta/endpoint/_endpoint.py
liuverta: How do you feel about this:

I think it's a better practice for the client to explicitly raise an error here saying "We expected a non-empty config with the backend" or "The config should've contained an ID, but didn't". Rather than silently set it to `None` and rely on a subsequent call to the backend to raise a `500`.

We already know our program's state here is going to result in an issue; we shouldn't have it continue running.

And I don't think a client should rely on the backend to raise an error for an issue that itself originated from the backend.

liuverta (changes requested):

client/verta/verta/endpoint/_endpoint.py
ewagner-verta: Ok, cool.  No objections to that and it gives us an opportunity to be a bit more specific with the error message.  I'll ping you when that update is in place shortly.

ewagner-verta: Changes and related tests are now included

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -509,8 +509,15 @@ def _set_kafka_settings(self, stage_id, kafka_settings):
                 " not {}".format(type(kafka_settings))
             )
         if kafka_settings._cluster_config_id is None:
-            config = kafka.list_kafka_configurations(self._conn)[0]
-            kafka_settings._cluster_config_id = config.get("id", None) if config else None
+            configs = kafka.list_kafka_configurations(self._conn)
+            if not configs:
+                raise RuntimeError("no valid Kafka configuration found.")
```
liuverta: Throwing in a little actionable suggestion in the error message:

```suggestion
                raise RuntimeError("no Kafka configuration found; please ensure that Kafka is configured in Verta")
```

liuverta: Also, I'd recommend against having a trailing period in our error messages. I don't think the client does that anywhere else.

client/verta/verta/endpoint/_endpoint.py
```diff
@@ -509,8 +509,15 @@ def _set_kafka_settings(self, stage_id, kafka_settings):
...
+            config_id = configs[0].get("id", None)
+            if config_id is None:
+                raise RuntimeError(
+                    "Kafka configuration is missing the required cluster_configuration_id."
```
liuverta: This error message kind of sounds like the client user neglected to provide something, but it's actually bad state in our backend that they can't do anything about. Also not sure where "cluster_configuration_id" came from, because I don't think we have that field name anywhere 🤔 

Maybe

```suggestion
                    "active Kafka configuration is missing its ID; please notify the Verta development team"
```

liuverta (changes requested):
Thanks for tackling the tests 😵‍💫 Some feedback on the error messages, then I think we'll be good

ewagner-verta:
@liuverta - No problem.  Thanks for the review.  I accepted the changes to the error messages and I just pushed a commit to update the unit tests accordingly.  Cheers

liuverta (approved):

PR #3778, opened 2023-04-26
feature: [VRD-840] Client function to trigger build scan

ewagner-verta:
## Impact and Context
Expands the existing `Build` class with the ability to trigger a scan of the build.

Only external scans are currently supported and a `NotImplemented` exception will be returned if any attempt is made to trigger an internal scan.  However, this logic assumes that internal scans will be the default option in the future, and thus users are required to pass the `external=True` argument when working with external scanning services.  This also ensures backwards compatibility if/when we do add internal scans.  Any users working with external scans will already have the parameter in place.

## Risks and Area of Effect
Relatively low.  This is a new API and client function wrapper that should not affect any other components.

## Testing
- [X] Unit test
test_build.py
Screen Shot 2023-04-26 at 2 52 33 PM
test_build_scan.py
Screen Shot 2023-04-26 at 2 53 27 PM

- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


client/verta/verta/endpoint/build/_build.py
```diff
@@ -102,3 +102,30 @@ def get_scan(self) -> _build_scan.BuildScan:
 
         """
         return _build_scan.BuildScan._get(self._conn, self.id)
+
```
ewagner-verta, line 105: Screen Shot 2023-04-26 at 4 10 08 PM

client/verta/verta/endpoint/build/_build.py
```diff
@@ -102,3 +102,30 @@ def get_scan(self) -> _build_scan.BuildScan:
...
+        Returns
+        -------
+        :class:`~verta.endpoint.build.BuildScan`
+            Build scan.
+
+        """
+        if not external:
+            raise NotImplementedError(
+                "internal scans are not yet supported.  Please use `external=True`"
+                " parameter."
```
liuverta: Don't use trailing periods in exception messages.

```suggestion
                "internal scans are not yet supported; please use `external=True`"
                " parameter"
```

client/verta/tests/unit_tests/deployment/test_build_scan.py
```diff
@@ -50,3 +50,23 @@ def test_get_scan(mock_conn, mocked_responses, build_dict, build_scan_dict):
         build_scan = build.get_scan()
 
     assert_build_scan_fields(build_scan, build_scan_dict)
+
+@settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+@given(build_dict=build_dict(), build_scan_dict=build_scan_dict())
+def test_start_scan(mock_conn, mocked_responses, build_dict, build_scan_dict):
+    """Verify we can construct a BuildScan object from start_scan."""
+    build_dict["creator_request"]["scan_external"] = True
+    build_scan_dict["creator_request"]["scan_external"] = True
```
liuverta: We should add a parameter to the strategies to optionally force `scan_external` instead of overriding the item here. This is going to throw off `hypothesis`'s example shrinking behavior, and also breaks the `build_scan_dict` strategy because it has a check for `scan_external`.

client/verta/tests/unit_tests/deployment/test_build_scan.py
```diff
@@ -50,3 +50,23 @@ def test_get_scan(mock_conn, mocked_responses, build_dict, build_scan_dict):
         build_scan = build.get_scan()
 
     assert_build_scan_fields(build_scan, build_scan_dict)
+
+@settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+@given(build_dict=build_dict(), build_scan_dict=build_scan_dict())
+def test_start_scan(mock_conn, mocked_responses, build_dict, build_scan_dict):
```
liuverta: Since this test seems to explicitly be for external scans, we might want to call that out.

```suggestion
def test_start_external_scan(mock_conn, mocked_responses, build_dict, build_scan_dict):
```

client/verta/verta/endpoint/build/_build_scan.py
```diff
@@ -87,6 +87,18 @@ def _get(cls, conn: _utils.Connection, build_id: int):
 
         return cls(response.json())
 
+    @classmethod
+    def _start(cls, conn: _utils.Connection, build_id: int, external: bool) -> "BuildScan":
+        if external:
+            data = {"scan_external": True}
+        else:
+            data = {"scan_external": False}
```
liuverta: Can we do

```suggestion
        data = {"scan_external": external}
```

?

client/verta/verta/endpoint/build/_build_scan.py
```diff
@@ -87,6 +87,18 @@ def _get(cls, conn: _utils.Connection, build_id: int):
 
         return cls(response.json())
 
+    @classmethod
+    def _start(cls, conn: _utils.Connection, build_id: int, external: bool) -> "BuildScan":
```
liuverta: Hmmmmm I think we ought to call this `_create` to match our usual get/create terminology.

client/verta/tests/unit_tests/deployment/test_build.py
```diff
@@ -2,12 +2,12 @@
 
 from typing import Any, Dict
 
-from hypothesis import given, HealthCheck, settings
 import hypothesis.strategies as st
+import pytest
+from hypothesis import given, HealthCheck, settings
```
liuverta: Would you mind keeping the two `hypothesis` lines grouped together?

client/verta/tests/unit_tests/deployment/test_build.py
```diff
@@ -92,3 +92,20 @@ def test_model_version_list_builds(
...
+    build = Build(mock_conn, build_dict)
+    with pytest.raises(
+        NotImplementedError,
+        match=(
+            "internal scans are not yet supported.  Please use `external=True`"
+            " parameter."
+        ),
+    ):
+        build.start_scan()
+        build.start_scan(external=False)
```
liuverta: This isn't going to work. `build.start_scan()` is going to raise an error and the `with` will exit, so `build.start_scan(external=False)` will never get run.

client/verta/verta/endpoint/build/_build.py
```diff
@@ -102,3 +102,30 @@ def get_scan(self) -> _build_scan.BuildScan:
 
         """
         return _build_scan.BuildScan._get(self._conn, self.id)
+
+    def start_scan(self, external: bool = False) -> _build_scan.BuildScan:
+        """Start a new scan for this build. Internal scans are not yet supported.
+        Use ``external=True`` parameter.
```
liuverta: I'd find it very unpleasant for the default behavior of a function be to encounter an error, but that might just be me 🤔

client/verta/verta/endpoint/build/_build.py
```diff
@@ -102,3 +102,30 @@ def get_scan(self) -> _build_scan.BuildScan:
...
+
+        This function only awaits a success response for the request, not any result
+        from the scan itself. Use
```
liuverta: It's a bit vague what "the request" is or does since it's an implementation detail of our code (and we don't technically wait for a _success_, either). Maybe something like

```suggestion
        This function only starts a scan; it does not wait for the finished scan to
        return results. Use
```

liuverta (changes requested):

client/verta/tests/unit_tests/deployment/test_build.py
ewagner-verta: I ran PyCharm's `Optimize Imports` feature and this seems to be what it came up with.  Not sure why that would be the outcome, but I'll happily rearrange to keep like elements together.

liuverta: Ah, I didn't realize it was from a tool.

Well...I guess that's fine? I have no idea why a tool would want to split up a package like this but it's not a big deal 🤷

client/verta/verta/endpoint/build/_build_scan.py
ewagner-verta: Done

client/verta/verta/endpoint/build/_build_scan.py
ewagner-verta: good call.  Much cleaner. Done

client/verta/verta/endpoint/build/_build.py
ewagner-verta: Fixed

client/verta/verta/endpoint/build/_build.py
ewagner-verta: Done.

client/verta/verta/endpoint/build/_build.py
ewagner-verta: I agree, but at least it's an easy solution handed to you at the moment.  The alternative could be that a working script breaks because a new version of the client comes out and your formerly external-by-default scans are now treated as internal because of a missing parameter.  No great options, but at least this could be temporary and force users into a future-proof state right from the introduction of the feature.

client/verta/tests/unit_tests/deployment/test_build_scan.py
ewagner-verta: and Done

client/verta/tests/unit_tests/deployment/test_build_scan.py
ewagner-verta: Done

client/verta/tests/unit_tests/deployment/test_build.py
ewagner-verta: Doh.  Fixed it with separate context managers for each condition.

client/verta/tests/unit_tests/deployment/test_build.py
ewagner-verta: Done

client/verta/verta/endpoint/build/_build.py
ewagner-verta: Updated for most recent changes ^^

client/verta/verta/endpoint/build/_build.py
liuverta: Couldn't we just remove the default for now?

```suggestion
    def start_scan(self, external: bool) -> _build_scan.BuildScan:
        """Start a new scan for this build. Internal scans are not yet supported.
        Use ``external=True`` parameter.
```

So a user has to explicitly understand what the parameter does and enter `False`, rather than someone autocompleting thinking it'll work then encountering the `NotImplementedError`.

client/verta/tests/unit_tests/deployment/test_build.py
ewagner-verta: @liuverta I'll leave this to you to resolve if you are good with the solution.

client/verta/tests/unit_tests/deployment/test_build.py
```diff
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 
+import pytest
```
liuverta: `pytest` shouldn't be here; standard library imports come first. https://pep8.org/#imports

client/verta/verta/endpoint/build/_build.py
liuverta: Couldn't we just remove the default for now?

```suggestion
    def start_scan(self, external: bool) -> _build_scan.BuildScan:
        """Start a new scan for this build. Internal scans are not yet supported.
        Use ``external=True`` parameter.
```

So a user has to explicitly understand what the parameter does and enter `False`, rather than someone autocompleting or leaving it empty thinking it'll work then encountering the `NotImplementedError`.

ewagner-verta: Yeah, I like that argument.  committing suggestion now.

client/verta/tests/unit_tests/deployment/test_build.py
```diff
@@ -92,3 +92,27 @@ def test_model_version_list_builds(
...
+        ),
+    ):
+        build.start_scan()
+    with pytest.raises(
+            NotImplementedError,
+            match=(
+                    "internal scans are not yet supported.  Please use `external=True`"
+                    " parameter"
+            ),
+    ):
```
liuverta:
```suggestion
    with pytest.raises(
        NotImplementedError,
        match=(
            "internal scans are not yet supported.  Please use `external=True`"
            " parameter"
        ),
    ):
```

client/verta/tests/unit_tests/deployment/test_build.py
```diff
@@ -92,3 +92,27 @@ def test_model_version_list_builds(
...
+    """Test that start_scan() raises the expected error when the `external` parameter is not
+    included or False.  This can be removed when we support internal scans.
+    """
+    build = Build(mock_conn, build_dict)
+    with pytest.raises(
+        NotImplementedError,
+        match=(
+            "internal scans are not yet supported.  Please use `external=True`"
+            " parameter"
+        ),
```
liuverta: Did you re-run tests before re-requesting review? This doesn't match the exception message, and I'd expect this test to fail.

client/verta/tests/unit_tests/strategies.py
```diff
@@ -194,11 +196,13 @@ def _build_scan_detail(draw) -> Dict[str, Any]:
 
 
 @st.composite
-def build_scan_dict(draw) -> Dict[str, Any]:
-    """Generate a Verta build scan, as returned by /api/v1/deployment/builds/{build_id}/scan."""
+def build_scan_dict(draw, external_scan: bool = False) -> Dict[str, Any]:
```
liuverta: I know this is a bit of extra work, but I'd recommend

```suggestion
def build_scan_dict(draw, external_scan: Optional[bool] = None) -> Dict[str, Any]:
```

What if we wanted to force `external_scan=False`?

liuverta: Or we can file a follow-up ticket and get this PR in first.

liuverta (changes requested):

client/verta/tests/unit_tests/deployment/test_build.py
liuverta: Oh wow it looks like it passes 🤔

client/verta/tests/unit_tests/deployment/test_build.py
ewagner-verta: Forgot to re-run black after the fix-up.  ~facepalm~

client/verta/tests/unit_tests/deployment/test_build.py
liuverta: Am...am I losing it? 😵‍💫

ewagner-verta: I fixed up the trailing period, but missed the semi-colon change in the middle.  Just reran tests on my end and it did fail on that until I fixed it.  Sorry for the mixup.  Should be fixed now.

client/verta/tests/unit_tests/deployment/test_build.py
ewagner-verta: fixed

liuverta: We can just use the PyCharm thing haha

client/verta/tests/unit_tests/deployment/test_build.py
liuverta: Hmmmmm black does have other options for working with PyCharm 🤔 I wonder if one of them would work https://black.readthedocs.io/en/stable/integrations/editors.html

client/verta/tests/unit_tests/deployment/test_build.py
ewagner-verta: This is moot now that `external` is a required param.  I dropped check for `start_scan()` with no params

client/verta/verta/endpoint/build/_build.py
```diff
@@ -102,3 +102,30 @@ def get_scan(self) -> _build_scan.BuildScan:
...
+        Use ``external=True`` parameter.
+
+        This function only starts a scan; it does not wait for the finished scan
+        to return results. Use
+        :meth:`Build.get_scan() <verta.endpoint.build.Build.get_scan()>` to view
+        progress and results of the scan.
+
+        Parameters
+        ----------
+        external : bool, default False
```
liuverta:
```suggestion
        external : bool
```

liuverta (approved):
One last comment.

Thank you for bearing with me 🙇

client/verta/tests/unit_tests/strategies.py
liuverta: @ewagner-verta I might have been unclear on this feedback. This one line change alone doesn't address my core comment of

> What if we wanted to force external_scan=False?

We also need to change the body of this function, because

```
"scan_external": True if external_scan else draw(st.booleans()),
```

doesn't facilitate forcing `False` at all. We should've also changed the `build_dict` strategy as well

PR #3785, opened 2023-05-02
fix: Correct Build Scan Endpoint

ewagner-verta:
## Impact and Context
Fixes the `Build.start_scan()` function to use the correct URL for the endpoint (needs to include `workspace`)

## Risks and Area of Effect
Low risk, fixes a bug.

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


client/verta/verta/endpoint/build/_build.py
```diff
@@ -110,7 +110,7 @@ def get_scan(self) -> _build_scan.BuildScan:
         """
         return _build_scan.BuildScan._get(self._conn, self.id)
 
-    def start_scan(self, external: bool) -> _build_scan.BuildScan:
+    def start_scan(self, external: bool, workspace: str) -> _build_scan.BuildScan:
```
liuverta: A user's not going to know off the top of their head which workspace this build exists in. I'd argue that a `Build` itself should be aware of what workspace it's in.

I'd recommend changing `Build.__init__()` to take and store `workspace`

```python
def __init__(self, conn, workspace, json):
```

similar to what we do for `Endpoint.__init__()` and other classes.

liuverta (changes requested):

client/verta/verta/endpoint/build/_build.py
ewagner-verta: I have this mostly implemented but I'm now discovering that this breaks the `_list_model_version_builds()`  function because we now need to pass `workspace` to `cls(conn, build_json)` for each build in the list.  (`_build.py#78`) Unfortunately, the API response does not contain the workspace in the build info, so I'm not sure where we'll get it.

liuverta: We can add `workspace` as a parameter to `_list_model_version_builds()` too, or even move that method to `RegisteredModelVersion` to just instantiate `Build`s using `self.workspace` directly.

ewagner-verta: Just pushed several commits to address this.

client/verta/tests/unit_tests/deployment/test_build.py
```diff
@@ -80,6 +82,11 @@ def test_model_version_list_builds(
             ],
             json={"builds": build_dicts},
         )
+        rsps.get(
+            url=model_version_url,
+            status=200,
+            json={"workspace_id": "123"},
+        )
```
ewagner-verta, line 89: Added this because the usage of `self.workspace` means that RMV [makes another call out to registry](https://github.com/VertaAI/modeldb/blob/7b48f0570d56f0f6c842a47e812ea25debc796b2/client/verta/verta/registry/entities/_modelversion.py#L176) to get it's workspace, so that call had to be mocked.

client/verta/tests/unit_tests/conftest.py
```diff
@@ -74,5 +74,5 @@ def __repr__(self):  # avoid network calls when displaying test results
     return MockRegisteredModelVersion(
         mock_conn,
         mock_config,
-        _RegistryService.ModelVersion(id=555),
+        _RegistryService.ModelVersion(id=555, registered_model_id=123)
```
ewagner-verta, line 77: This is used in the call to registry that now occurs (see [comment below](https://github.com/VertaAI/modeldb/pull/3785/files#r1183118508))

client/verta/tests/unit_tests/strategies.py
```diff
@@ -221,3 +221,14 @@ def build_scan_dict(draw, external_scan: Optional[bool] = None) -> Dict[str, Any
             d["details"] = draw(st.lists(_build_scan_detail()))
 
     return d
+
+
+@st.composite
+def mock_workspace(draw):
+    workspace = draw(
+        st.text(
+            alphabet=st.characters(whitelist_categories=("Ll", "Lu", "Nl", "Pd")),
```
liuverta: Would you mind adding a brief comment about what these categories mean?

client/verta/tests/unit_tests/deployment/test_build_scan.py
```diff
@@ -51,17 +55,21 @@ def test_get_scan(mock_conn, mocked_responses, build_dict, build_scan_dict):
...
+def test_start_external_scan(
+    mock_conn, mocked_responses, build_dict, build_scan_dict, workspace
+):
     """Verify we can construct a BuildScan object from start_scan(external=True)."""
-    build = Build(mock_conn, build_dict)
+    build = Build(mock_conn, workspace, build_dict)
 
     deployment_url = f"{mock_conn.scheme}://{mock_conn.socket}/api/v1/deployment"
-    scan_url = f"{deployment_url}/builds/{build.id}/scan"
+    scan_url = f"{deployment_url}/workspace/{workspace}/builds/{build.id}/scan"
```
ewagner-verta, line 72: This is the real fix ^^

client/verta/tests/unit_tests/deployment/test_build_scan.py
```diff
@@ -2,10 +2,10 @@
 
 from typing import Any, Dict
 
-from hypothesis import given, HealthCheck, settings
+from hypothesis import given, HealthCheck, settings, strategies as st
```
liuverta: Is this being used?

client/verta/verta/endpoint/build/_build.py
```diff
@@ -59,13 +60,11 @@ def _get(cls, conn: _utils.Connection, workspace: str, id: int) -> "Build":
...
 
-        return cls(conn, response.json())
+        return cls(conn, workspace, response.json())
 
     @classmethod
     def _list_model_version_builds(
-        cls,
-        conn: _utils.Connection,
-        id: int,
+        cls, conn: _utils.Connection, id: int, workspace: str
```
liuverta: I'd prefer if we had `workspace` come right after `conn` for consistency with all our of other functions (plus it's an internal and yet-unreleased method, so it's fine), but not a huge deal.

client/verta/verta/endpoint/build/_build_scan.py
```diff
@@ -101,9 +101,11 @@ def _get(cls, conn: _utils.Connection, build_id: int):
         return cls(response.json())
 
     @classmethod
-    def _create(cls, conn: _utils.Connection, build_id: int, external: bool) -> "BuildScan":
+    def _create(
+        cls, conn: _utils.Connection, build_id: int, external: bool, workspace: str
```
liuverta: Here I'd actually recommend having `workspace` come right after `conn` to match all of our other creation methods.

liuverta (changes requested):
Looks good overall. Some minor things.

client/verta/tests/unit_tests/strategies.py
ewagner-verta: Added a doc string with details.

client/verta/tests/unit_tests/deployment/test_build_scan.py
ewagner-verta: Ah, shoot.  I haven't been running the import optimizer on this file to avoid dealing with the 'pytest' placement oddity.  Fixed now.  Thanks

client/verta/verta/endpoint/build/_build.py
ewagner-verta: Easy fix.  Done

client/verta/verta/endpoint/build/_build_scan.py
ewagner-verta: Also done.

client/verta/tests/unit_tests/strategies.py
```diff
@@ -221,3 +221,19 @@ def build_scan_dict(draw, external_scan: Optional[bool] = None) -> Dict[str, Any
             d["details"] = draw(st.lists(_build_scan_detail()))
 
     return d
+
+
+@st.composite
+def mock_workspace(draw):
+    """ Return a valid workspace name.
+    Unicode categories allowed: Ll (lowercase letter), Lu (Uppercase letters),
+    Nl (Decimal number), Pd (Dash punctuation).  `%` disallowed to prevent
```
liuverta:
```suggestion
    Nd (Decimal number), Pd (Dash punctuation).  `%` disallowed to prevent
```

liuverta (approved):
One small docstring comment. 🚀

PR #3835, opened 2023-05-30
feat: [VRD-909] Add pii to client

ghost:
## Impact and Context
Add new registered model PII field to client.

## Risks and Area of Effect
It's a new field so not bad.

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [X] Other (explain) 

Also added the client changes to a pyenv and used the client like normal to make sure I could change the PII value.
## Reverting


client/verta/tests/registry/test_model.py
```diff
@@ -132,6 +132,24 @@ def test_description(self, client, created_entities):
...
+        # created_entities.append(registered_model)
+        # assert registered_model.get_pii()
+
+        registered_model = client.get_or_create_registered_model(pii=True)
+        # created_entities.append(registered_model)
+        # registered_model.set_pii(True)
+        # assert registered_model.get_pii()
+        print(f"setting false for {registered_model.id}")
+        registered_model.set_pii(False)
+        assert registered_model.get_pii() is False
```
ghost, line 150: this is failing. And looking at the db, the change isn't going through. Why??

client/verta/verta/registry/entities/_model.py
```diff
@@ -87,6 +90,11 @@ def workspace(self):
         else:
             return self._conn._OSS_DEFAULT_WORKSPACE
 
+    @property
+    def pii(self):
+        self._refresh_cache()
+        return self._msg.pii
```
liuverta: I'd say we should omit this; `get_pii()` should suffice, and it's not greeeaat to have multiple ways to do something simple.

I don't have a clear guideline for when something should or shouldn't be a `property`, but it's something along the lines of "`property` is for fundamental things about the entity that the user can't directly change"

client/verta/tests/registry/test_model.py
```diff
@@ -132,6 +132,24 @@ def test_description(self, client, created_entities):
         registered_model.set_description(desc)
         assert desc == registered_model.get_description()
 
+    def test_pii(self, client, created_entities):
+        # Test setting on creation
+        registered_model = client.get_or_create_registered_model(pii=True)
+        created_entities.append(registered_model)
+        assert registered_model.get_pii()
```
liuverta: For maximum correctness because in Python, lots of non-`bool` objects can be `True`y and therefore pass an `assert`. We should explicitly check that it's `True`.

```suggestion
        assert registered_model.get_pii() is True
```

client/verta/tests/registry/test_model.py
```diff
@@ -132,6 +132,24 @@ def test_description(self, client, created_entities):
...
+        registered_model = client.create_registered_model(pii=True)
+        created_entities.append(registered_model)
+        assert registered_model.get_pii()
```
liuverta:
```suggestion
        assert registered_model.get_pii() is True
```

client/verta/tests/registry/test_model.py
```diff
@@ -132,6 +132,24 @@ def test_description(self, client, created_entities):
...
+
+        # Test updating
+        registered_model = client.get_or_create_registered_model()
+        created_entities.append(registered_model)
+        registered_model.set_pii(True)
+        assert registered_model.get_pii()
```
liuverta:
```suggestion
        assert registered_model.get_pii() is True
```

client/verta/verta/client.py
```diff
@@ -780,6 +780,7 @@ def get_or_create_registered_model(
         id=None,
         task_type=None,
         data_type=None,
+        pii=None,
```
liuverta:
```suggestion
        pii=False,
```

See my other comment below, on the docstring

client/verta/verta/client.py
```diff
@@ -815,6 +816,8 @@ def get_or_create_registered_model(
             Task type of the registered_model.
         data_type : :mod:`~verta.registry.data_type`, optional
             Data type of the registered_model.
+        pii : :mod:`~verta.registry.pii`, optional
```
liuverta: This is a simple `bool`, so we don't want/need to reference another Python `mod`ule.

Also, I think we can say this is `False` by default—rather than optional—since that's the result in the backend if the user doesn't explicitly pass a value.

```suggestion
        pii : bool, default False
```

liuverta (changes requested):

client/verta/tests/registry/test_model.py
ghost: ah, caught this for the False case but didn't think of it for True

client/verta/verta/client.py
```diff
@@ -1250,6 +1254,7 @@ def create_registered_model(
         visibility=None,
         task_type=None,
         data_type=None,
+        pii=None,
```
liuverta:
```suggestion
        pii=False,
```

liuverta (approved):

PR #3867, opened 2023-06-07
chore: build protos as a separate module

liam-verta:
## Impact and Context

This results in significant build time improvements due to not rebuild protos multiple times.

Does not change build behavior for downstream projects that are building common or protos on their own.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


jkwatson-verta (approved):
Thanks!

PR #3878, opened 2023-06-12
feat: Allow specifying org id in client

conradoverta:
## Impact and Context
This will include the provided value to the `organization-id` metadata header, which will be processed by the backend. The value is optional.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/client/verta/verta/credentials.py
+++ b/client/verta/verta/credentials.py
@@ -46,14 +46,17 @@ class EmailCredentials(Credentials):
         A user email address.
     dev_key : str
         A dev key to use for authentication.
+    organization_id : str, optional
+        An organization ID to use for authentication.
     """
 
     EMAIL_ENV = "VERTA_EMAIL"
     DEV_KEY_ENV = "VERTA_DEV_KEY"
 
-    def __init__(self, email, dev_key):
+    def __init__(self, email, dev_key, organization_id=None):
         self.email = email
         self.dev_key = dev_key
+        self.organization_id = organization_id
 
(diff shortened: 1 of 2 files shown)


client/verta/verta/credentials.py
```diff
@@ -67,11 +70,12 @@ def headers(self):
             # without underscore, for NGINX support
             # https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls#missing-disappearing-http-headers
             "developer-key": self.dev_key,
+            "organization-id": self.organization_id or "",
```
liuverta: Would it be preferable to do a

```python
if self.organization_id:
    headers["organization-id"] = self.organization_id
```

instead of passing an empty string?

liuverta (approved):

client/verta/verta/credentials.py
conradoverta: Sounds good.

client/verta/verta/credentials.py
```diff
@@ -113,16 +122,19 @@ def headers(self):
             # without underscore, for NGINX support
             # https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls#missing-disappearing-http-headers
             "bearer-access-token": self.jwt_token,
+            "organization-id": self.organization_id or "",
```
ewagner-verta, line 125: Since we are checking for and adding the org id below, we could maybe just drop this line.

PR #3879, opened 2023-06-13
feat: Create a JetstreamConnector for managing Jetstream connectivity

jkwatson-verta:
## Impact and Context

Combines proven working code from both UAC and the registry into one common class that both can use.

## Testing
- [x] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


backend/common/src/test/java/ai/verta/modeldb/common/nats/JetstreamConnectorTest.java
```diff
@@ -0,0 +1,158 @@
...
+import lombok.extern.log4j.Log4j2;
+import np.com.madanpokharel.embed.nats.EmbeddedNatsConfig;
+import np.com.madanpokharel.embed.nats.EmbeddedNatsServer;
+import np.com.madanpokharel.embed.nats.NatsServerConfig;
+import np.com.madanpokharel.embed.nats.NatsVersion;
+import org.junit.jupiter.api.*;
+
+@Log4j2
+class JetstreamConnectorTest {
+  public static EmbeddedNatsServer embeddedNatsServer;
```
gustavoteixeirah, line 24: This is awesome!

gustavoteixeirah: Wow I just saw that you are a contributor of this project! That's even more awesome!

backend/common/src/test/java/ai/verta/modeldb/common/nats/JetstreamConnectorTest.java
```diff
@@ -0,0 +1,158 @@
...
+          msg.ack();
+          message.set(new String(msg.getData()));
+        },
+        "testDurable",
+        "testGroup");
+
+    JetStream cheeseStream = connector.getJetStream("cheese");
+    cheeseStream.publish("cheese.create", "Swiss".getBytes(StandardCharsets.UTF_8));
+
+    await().untilAsserted(() -> assertThat(message.get()).isEqualTo("Swiss"));
```
gustavoteixeirah, line 91: I was wondering if we could capture a exception that occurs within the event handler. I tried to do a test in the registry for this but I couldn't find a way to capture the exception and do some assertion on it. I was trying with the "messageHandler" object which was the only one available I saw. Do you know if there's a way to do it with this EmbeddedNatsServer project?

marcelo-verta-ai (approved):

backend/common/src/test/java/ai/verta/modeldb/common/nats/JetstreamConnectorTest.java
jkwatson-verta: what? I don't think I'm a contributor unless I missed something!

jkwatson-verta: oh! haha. https://github.com/madansp/nats-embedded/pull/6 That's my only contribution.

backend/common/src/test/java/ai/verta/modeldb/common/nats/JetstreamConnectorTest.java
jkwatson-verta: What specific case are you thinking of? When our handler throws an exception, or some other exception?

gustavoteixeirah: When the handler throws exception. For example that parse message method that can throw a exception of not being able to parse, I couldn't catch it. I think it is because it is running async...

gustavoteixeirah (approved):

backend/common/src/test/java/ai/verta/modeldb/common/nats/JetstreamConnectorTest.java
jkwatson-verta: hm. that's probably a test of the handler, rather than a test of this code, I believe. All we can do here is provide a handler that is implemented however we need to test that it's being called appropriately.

PR #3884, opened 2023-06-15
feat: [VRD-960] Implement LocalOrchestrator

liuverta:
## Impact and Context

Implements `LocalOrchestrator` and `ModelObjectStepHandler` (and `ModelContainerStepHandler`) to prepare and run a pipeline that uses locally-instantiated model objects retrieved from registered model versions.

## Risks and Area of Effect

None—all new functionality

I don't yet have tests, but I have a separate ticket for that.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [x] Other (explain)

Tested manually for now because things are still a bit in flux 😭 

Using a pared-down version of the pipeline spec we have exemplified in [our design doc](https://github.com/VertaAI/docs/blob/main/design-docs/2023-05-pipelines.md):

```python
spec = {
  "steps": [
    {
      "rmvId": model1.id,
      "attributes": {"name": "multiply_by_1"}
    },
    {
      "rmvId": model2.id,
      "attributes": {"name": "multiply_by_2"}
    },
    {
      "rmvId": model3.id,
      "attributes": {"name": "multiply_by_3"}
    }
  ],
  "graph": {
    "inputs": {
      "multiply_by_2": ["multiply_by_1"],
      "multiply_by_3": ["multiply_by_2"]
    }
  }
}
```

we have a silly pipeline that multiplies an input number by 1, then by 2, then by 3:

```python
from verta._pipeline_orchestrator import LocalOrchestrator

orchestrator = LocalOrchestrator(client._conn, spec)
orchestrator.run(1)
# 6
```

and outputs are stored as state for that run of the pipeline:

```python
orchestrator._outputs
# {'multiply_by_1': 1, 'multiply_by_2': 2, 'multiply_by_3': 6}
```

## Reverting

Revert this PR.

### DIFF
--- a/client/verta/verta/_pipeline_orchestrator/_orchestrator.py
+++ b/client/verta/verta/_pipeline_orchestrator/_orchestrator.py
@@ -18,7 +18,7 @@ class _OrchestratorBase(abc.ABC):
     def __init__(
         self,
         pipeline_spec: Dict[str, Any],
-        step_handlers: Dict[int, _StepHandlerBase],
+        step_handlers: Dict[str, _StepHandlerBase],
     ):
         self._pipeline_spec = pipeline_spec
         self._step_handlers = step_handlers
@@ -26,15 +26,15 @@ class _OrchestratorBase(abc.ABC):
 
     def _run_step(
         self,
-        model_version_id: int,
+        name: str,
         input: Any,
     ) -> Any:
         """Run a pipeline step.
 
         Parameters
         ----------
-        model_version_id : int
-            ID of the step's model version.
+        name : str
+            Step name.
         input : object
             Step input.
 
@@ -47,13 +47,11 @@ class _OrchestratorBase(abc.ABC):
         if self._dag is None:
             raise RuntimeError("DAG not initialized")
 
-        step_handler = self._step_handlers[model_version_id]
-
+        step_handler = self._step_handlers[name]
 
         output = step_handler.run(input)
 
-
-        self._dag.done(model_version_id)
+        self._dag.done(name)
         return output
 
     @staticmethod
@@ -70,7 +68,7 @@ class _OrchestratorBase(abc.ABC):
         Returns
         -------
         :class:`~graphlib.TopologicalSorter`
-            Pipeline graph. Each node is a model version ID.
+            Pipeline graph. Each node is a step name.
 
         Raises
         ------
@@ -105,21 +103,19 @@ class _OrchestratorBase(abc.ABC):
 
         # run input node
         # NOTE: assumes only one input node
-        model_version_id: int = self._dag.get_ready()[0]
+        step_name: str = self._dag.get_ready()[0]
         outputs = {
-            model_version_id: self._run_step(model_version_id, input),
+            step_name: self._run_step(step_name, input),
         }
 
         while self._dag.is_active():
-            for model_version_id in self._dag.get_ready():
-                input =
+            for step_name in self._dag.get_ready():
+                input = None
 
-                self._run_step(
-                    model_version_id,
-                )
+                self._run_step(step_name, input)
 
-                step_handler = self._step_handlers[model_version_id]
-                outputs[model_version_id] = step_handler.run(  # TODO: async?
+                step_handler = self._step_handlers[step_name]
+                outputs[step_name] = step_handler.run(  # TODO: async?
                     outputs[step_handler.predecessors[0]]  # just to test
                     # {
                     #     predecessor_id: outputs[predecessor_id]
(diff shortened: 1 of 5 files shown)


client/verta/verta/_pipeline_orchestrator/_orchestrator.py
```diff
@@ -0,0 +1,191 @@
...
+from verta._internal_utils import _utils
+from verta.external.cpython.graphlib import TopologicalSorter
+
+from ._step_handler import (
+    _StepHandlerBase,
+    ModelObjectStepHandler,
+)
+
+
+class _OrchestratorBase(abc.ABC):
```
liuverta, line 15: Basically, there's

- `DeployedOrchestrator` to orchestrate deployed model containers (stub for now)
- `LocalOrchestrator` to orchestrate locally-instantiated model containers

These subclasses implement logic to prepare step handlers. This base class (`_OrchestratorBase`) has the logic to prepare and run the pipeline.

client/verta/verta/_pipeline_orchestrator/_step_handler.py
```diff
@@ -0,0 +1,165 @@
...
+import abc
+import json
+from typing import Any, Dict, Iterable, List, Optional, Set, Type
+
+from verta._internal_utils import _utils, http_session
+from verta.registry.entities import RegisteredModelVersion
+from verta.tracking.entities._entity import _MODEL_ARTIFACTS_ATTR_KEY
+
+
+class _StepHandlerBase(abc.ABC):
```
liuverta, line 12: Basically, there's

- `ModelContainerStepHandler` to route predictions to a deployed model container
- `ModelObjectStepHandler` to route predictions to a locally-instantiated model object

These subclasses implement that logic to route predictions. This base class (`_StepHandlerBase`) doesn't do much.

client/verta/verta/external/cpython/LICENSE
```diff
@@ -0,0 +1,279 @@
+A. HISTORY OF THE SOFTWARE
```
liuverta, line 1: Everything in `client/verta/verta/external/cpython/` is vendored from Python 3.11. We specifically want the `graphlib` standard library.

client/verta/verta/external/cpython/graphlib.py
```diff
@@ -0,0 +1,252 @@
...
+        self._npassedout = 0
+        self._nfinished = 0
+
+        if graph is not None:
+            for node, predecessors in graph.items():
+                self.add(node, *predecessors)
+
+    def _get_nodeinfo(self, node):
+        result = self._node2info.get(node)
+        if result is None:
```
liuverta, line 58: I had to change this from

```python
if (result := self._node2info.get(node)) is None:
```

because assignment expressions aren't supported in Python 3.7.

client/verta/verta/external/cpython/graphlib.py
```diff
@@ -0,0 +1,253 @@
...
+
+        if self._ready_nodes is None:
+            raise ValueError("prepare() must be called first")
+
+        n2i = self._node2info
+
+        for node in nodes:
+            # Check if we know about this node (it was added previously using add()
+            nodeinfo = n2i.get(node)
+            if nodeinfo is None:
```
liuverta, line 174: I had to change this from

```python
if (nodeinfo := n2i.get(node)) is None:
```

because assignment expressions aren't supported in Python 3.7.

client/verta/verta/_pipeline_orchestrator/_orchestrator.py
```diff
@@ -0,0 +1,200 @@
...
+            )
+
+        step_handler = self._step_handlers[name]
+        if input is None:
+            predecessors = step_handler.predecessors
+            if not predecessors:
+                raise ValueError(
+                    f"unexpected error: step {name} has no predecessors, but no input was provided",
+                )
+            if not predecessors <= self._outputs.keys():
```
ewagner-verta, line 85: Can you help me understand exactly what this is checking?  (not disputing the logic of the check, just not immediately able to understand the purpose).  Thanks

ewagner-verta (approved):

client/verta/verta/_pipeline_orchestrator/_orchestrator.py
liuverta: `predecessors` is a set (I just added a type hint for it); `set <= other` checks whether all the members of `set` are present in `other`.

So `if not predecessors <= self._outputs.keys()` is "if there are members of `predecessors` _not_ in `self._outputs.keys()`", i.e. if this step has input steps that we don't output values for.

PR #3910, opened 2023-06-30
feat: [VRD-991] Create input validation decorator

ghost:
## Impact and Context
Note: this won't be usable until `set_schema` is implemented in a follow-up PR.

Create a decorator to validate prediction input and output on json schemas.

## Risks and Area of Effect
This is a new feature, so no risk to existing functionality.

## Testing
- [X] Unit test
- [X] Deployed to dev env
- [ ] Other (explain) 

Deployed to dev env and tested the feature e2e (with a client hack to set schemas since that isn't implemented yet).

## Reverting


client/verta/requirements.txt
```diff
@@ -7,3 +7,4 @@ pytimeparse >= 1.1.8, < 2.0
 pyyaml >= 5.1, < 7.0
 requests >= 2.25, < 3.0
 urllib3 >= 1.26.0
+jsonschema>=4.17.3
```
liuverta: Also, please insert this in alphabetical order!

```suggestion
jsonschema >= 4.17.3
```

client/verta/verta/registry/_validate_input.py
```diff
@@ -0,0 +1,34 @@
...
+                schema = json.load(file)
+                input_schema = schema['input']
+                jsonschema.validate(instance=prediction_input, schema=input_schema)
+
+            output = f(self, *args, **kwargs)
+            return output
+        except FileNotFoundError:
+            print("No schema found for model. Did you remember to `set_schema`?")
+
+    setattr(wrapper, _DECORATED_FLAG, True)
```
liuverta: I'm not sure if we addressed this already, but is `validate_input` ever meant to be used together with `verify_io`? I'm curious whether the `setattr()`s carry over.

client/verta/verta/registry/_validate_input.py
```diff
@@ -0,0 +1,34 @@
...
+            with open(_INPUT_SCHEMA_PATH, 'r') as file:
+                # Load the JSON data into a variable
+                schema = json.load(file)
+                input_schema = schema['input']
+                jsonschema.validate(instance=prediction_input, schema=input_schema)
+
+            output = f(self, *args, **kwargs)
+            return output
+        except FileNotFoundError:
+            print("No schema found for model. Did you remember to `set_schema`?")
```
liuverta: This seems like it should be an exception, rather than a print, since it flat-out won't work.

client/verta/verta/registry/_validate_input.py
```diff
@@ -0,0 +1,34 @@
...
+    """Decorator to validate prediction input against previously provided schema.
+
+    """
+
+    @functools.wraps(f)
+    def wrapper(self, *args, **kwargs):
+        # accepts dict input
+        # TODO: validate length of args and type
+        prediction_input = args[0]
+        try:
```
liuverta: A `try` block in a `try-except` should contain the minimal amount of code that's relevant to the error, as a general practice to avoid accidentally capturing errors from other code (like what if `jsonschema.validate()` or `f()` raises `FileNotFoundError`s? but it's not what we're interested in here).

So something like

```python
try:
    with open(_INPUT_SCHEMA_PATH, 'r') as file:
        schema = json.load(file)
except FileNotFoundError:
    ...
```

client/verta/verta/registry/_validate_input.py
```diff
@@ -0,0 +1,34 @@
...
+            with open(_INPUT_SCHEMA_PATH, 'r') as file:
+                # Load the JSON data into a variable
+                schema = json.load(file)
+                input_schema = schema['input']
+                jsonschema.validate(instance=prediction_input, schema=input_schema)
```
liuverta: In general: I'd recommend having only the `json.load()` in the `with`—to close the file as soon as we're actually done using it, in case another process needs it.

```suggestion
            with open(_INPUT_SCHEMA_PATH, 'r') as file:
                # Load the JSON data into a variable
                schema = json.load(file)
            input_schema = schema['input']
            jsonschema.validate(instance=prediction_input, schema=input_schema)
```

client/verta/verta/registry/_validate_input.py
ghost: I ran it with `verify_io` (both orientations) and nothing weird happened, so that's promising. But I suppose idk about the `setattr()`s

client/verta/verta/registry/_validate_input.py
ghost: 🤦🏼‍♀️

client/verta/verta/registry/_validate_input.py
ghost: I forgot this is allowed in python "scoping"!

client/verta/verta/registry/_validate_input.py
ghost: once again, python's scopes 😅

client/verta/requirements.txt
```diff
@@ -2,6 +2,7 @@ click >= 7.0, < 9.0
 cloudpickle >= 1.0, < 3.0
 googleapis-common-protos >= 1.5, < 2.0
 importlib-metadata >= 3.7.0
+jsonschema>=4.17.3
```
liuverta: We should give it some space (not actually required, but makes it slightly slightly easier to read)

```suggestion
jsonschema >= 4.17.3
```

client/verta/tests/registry/pydantic_models.py
```diff
@@ -0,0 +1,33 @@
...
+
+class InnerInputClass(BaseModel):
+    h_dict: dict[str, str]
+    i_list: list[str]
+
+    def __init__(self, h, i):
+        super().__init__(
+            h_dict=h,
+            i_list=i,
+        )
```
liuverta, line 12: Are these `__init__()`s actually necessary? 🤔 I would've guessed that `pydantic` only requires the above

```python
    h_dict: dict[str, str]
    i_list: list[str]
```

client/verta/verta/registry/_validate_input.py
liuverta: @hmacdonald-verta If you haven't yet, the thing to verify would be `create_standard_model()`, because that's where the `getattr()` check is performed.

client/verta/verta/registry/_validate_input.py
```diff
@@ -0,0 +1,35 @@
...
+        # TODO: validate length of args and type
+        prediction_input = args[0]
+        try:
+            with open(_MODEL_SCHEMA_PATH, "r") as file:
+                # Load the JSON data into a variable
+                schema = json.load(file)
+        except FileNotFoundError:
+            raise Exception(
+                "No schema found for model. Did you remember to `set_schema`?"
+            )
```
liuverta: - we should raise `FileNotFoundError` instead of `Exception` to be as specific as possible
- Python's standard library has this weird practice of exception messages _not_ being capitalized, so it's dumb but I'm nudging the client to use lowercase and semicolons 🤷 
- we should raise `from e`; this tells Python "I'm deliberately handling the underlying error as a new error" instead of "oh no an unexpected new error ? ? ?"

```suggestion
        except FileNotFoundError as e:
            raise FileNotFoundError(
                "no schema found for model; did you remember to `set_schema`?"
            ) from e
```

client/verta/tests/registry/conftest.py
```diff
@@ -17,4 +20,9 @@ def docker_image():
     )
 
 
-
+@pytest.fixture
+def make_model_schema_file(tmp_path, monkeypatch):
+    path = tmp_path / "model_schema.json"
+    schema = {"input": InputClass.schema()}
+    path.write_text(json.dumps(schema))
+    monkeypatch.setenv(_MODEL_SCHEMA_PATH_ENV_VAR, str(path))
```
ghost, line 36: this is obviously not propagating but idk WHY (I've thrown a print in and verified the actual code is in fact using the default value of the env var):
```        except FileNotFoundError as e:
>           raise FileNotFoundError(
                "no schema found for model. Did you remember to `set_schema`?"
            ) from e
E           FileNotFoundError: no schema found for model. Did you remember to `set_schema`?
E           Falsifying example: test_validate_input_allow(
```

client/verta/tests/registry/pydantic_models.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass
+
+from pydantic import BaseModel
```
liuverta, line 3: We should add `pydantic` to https://github.com/VertaAI/modeldb/blob/main/client/verta/requirements-unit-tests.txt

client/verta/verta/registry/_validate_input.py
ghost: done! looks good. The `predict()` method's `__dict__` includes both attributes.

client/verta/requirements-unit-tests.txt
```diff
@@ -3,4 +3,5 @@ hypothesis
 pytest >= 4.3
 pytest-xdist >= 1.32
 six >= 1.11
-responses >= 0.22
\ No newline at end of file
+responses >= 0.22
+pydantic
```
liuverta: I realize [I literally asked for exactly this](https://github.com/VertaAI/modeldb/pull/3910/files#r1248305601), but it occurs to me this should be in [`requirements-dev.txt`](https://github.com/VertaAI/modeldb/blob/60c03fefb30bd8f64dd11211ade0fec3ebb7893c/client/verta/requirements-dev.txt) instead 😬

client/verta/tests/registry/conftest.py
```diff
@@ -17,4 +20,17 @@ def docker_image():
...
+def make_model_schema_file(tmp_path, monkeypatch):
+    path = tmp_path / "model_schema.json"
+    schema = {"input": AClass.schema(), "output": AnotherClass.schema()}
+    path.write_text(json.dumps(schema))
+    monkeypatch.setenv(_MODEL_SCHEMA_PATH_ENV_VAR, str(path))
 
+
+@pytest.fixture
+def make_model_schema_file_no_output(tmp_path, monkeypatch):
+    path = tmp_path / "other_model_schema.json"
```
liuverta: Just to make it a bit more descriptive

```suggestion
    path = tmp_path / "model_schema_no_output.json"
```

client/verta/tests/registry/pydantic_models.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, h, i):
+        super().__init__(
+            h_dict=h,
+            i_list_str=i,
+        )
+
+
+@dataclass
+class AClass(BaseModel):
```
liuverta: Based on the way we're using these, I think we should give them more descriptive names like

```suggestion
class InputModel(BaseModel):
```

and `OutputModel`. Same with the `generate_*()` composite strategies.

Or at least, I'm doubtful these would get re-used for anything else, so might as well name them according to what they're doing for now

client/verta/tests/registry/pydantic_models.py
```diff
@@ -0,0 +1,51 @@
...
+@dataclass
+class AnInnerClass(BaseModel):
+    h_dict: dict
+    i_list_str: list[str]
+
+    def __init__(self, h, i):
+        super().__init__(
+            h_dict=h,
+            i_list_str=i,
+        )
```
liuverta: Are these `__init__()`s actually necessary? 🤔 I would've guessed that `pydantic` only requires the above

```python
    h_dict: dict[str, str]
    i_list: list[str]
```

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+    ----------
+    .. [#] https://python-jsonschema.readthedocs.io/en/stable/
+    """
+
+    @functools.wraps(f)
+    def wrapper(self, *args, **kwargs):
+        # accepts dict input
+        # TODO: validate length of args and type
+        # Fetch schema
+        prediction_input = args[0]
```
liuverta: It looks like this wrapper is very specifically designed to expect one argument (unlike `_verify_io()` which chose to be more general for no particular reason).

We could just match the signature of `VertaModelBase.predict()`:

```suggestion
    def wrapper(self, input: Dict):
```

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+                "no schema found for model. Did you remember to `set_schema`?"
+            ) from e
+        input_schema = schema["input"]
+        output_schema = None
+        if "output" in schema:
+            output_schema = schema["output"]
+
+        # Validate input
+        if not isinstance(prediction_input, dict):
+            raise TypeError("input must be a dict. Did you remember to call `.dict()`?")
```
liuverta: We should do this validation at the very beginning of the function

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+                # Load the JSON data into a variable
+                schema = json.load(file)
+        except FileNotFoundError as e:
+            raise FileNotFoundError(
+                "no schema found for model. Did you remember to `set_schema`?"
+            ) from e
+        input_schema = schema["input"]
+        output_schema = None
+        if "output" in schema:
+            output_schema = schema["output"]
```
liuverta:
```suggestion
        output_schema = schema.get("output")
```

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+        model_schema_path = os.environ.get(
+            _MODEL_SCHEMA_PATH_ENV_VAR, "/app/model_schema.json"
+        )
+        try:
+            with open(model_schema_path, "r") as file:
+                # Load the JSON data into a variable
+                schema = json.load(file)
+        except FileNotFoundError as e:
+            raise FileNotFoundError(
+                "no schema found for model. Did you remember to `set_schema`?"
```
liuverta:
```suggestion
                "no schema found for model. Did you remember to call `model_ver.set_schema()`?"
```

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+
+def validate_schema(f):
+    """Decorator to validate prediction input and output against previously provided schema.
+
+    Validation is done with the ``jsonschema`` library [#]_. If no schema has been provided, an exception is raised.
+
+    Note that an input schema is required but output is not. If the output schema was not set, then the output will
+    not be validated.
+
+    Can safely be used with `@verify_io`.
```
liuverta:
```suggestion
    Can safely be used with :func:`~verta.registry.verify_io`.
```

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+
+# for use like: `if getattr(model.predict, _DECORATED_FLAG, False)`
+_VALIDATE_DECORATED_FLAG = "_verta_validate_schema"
+_MODEL_SCHEMA_PATH_ENV_VAR = "VERTA_MODEL_SCHEMA_PATH"
+
+
+def validate_schema(f):
+    """Decorator to validate prediction input and output against previously provided schema.
+
+    Validation is done with the ``jsonschema`` library [#]_. If no schema has been provided, an exception is raised.
```
liuverta:
```suggestion
    Validation is done with the ``jsonschema`` library [#]_. If no schema has been provided via :meth:`RegisteredModelVersion.set_schema() <verta.registry.entities.RegisteredModelVersion.set_schema>`, an exception is raised.
```

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+                pass
+
+            @validate_schema
+            def predict(self, input):
+                return {'c': 3, 'd': 4}
+
+        def main():
+            client = Client()
+
+            # Register
```
liuverta: We should be consistent in whether comments start with lowercase or uppercase (the client codebase has been using lowercase)

client/verta/verta/registry/_validate_schema.py
liuverta: Also, less important but the client codebase has been using semicolons

```
... for model; did you remember ...
```

rather than periods

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,125 @@
...
+            # Deploy
+            endpoint = client.get_or_create_endpoint("my-model")
+            endpoint.update(model_ver, wait=True)
+            deployed_model = endpoint.get_deployed_model()
+
+            # succeeds; input matches previously provided schema
+            input = Input(a=1, b=2)
+            output = deployed_model.predict(input.dict())
+
+            # fails; a list will be given to the deployed model as-is
```
liuverta: I'm not sure what this means by

> a list will be given to the deployed model as-is

I thought this would raise an exception 🤔

client/verta/tests/registry/pydantic_models.py
hmacdonald-verta: I think autofill got overly aggressive

client/verta/verta/registry/_validate_schema.py
ghost: I have no idea what I meant 😅

client/verta/tests/registry/pydantic_models.py
```diff
@@ -0,0 +1,22 @@
+from pydantic import BaseModel
+
+
+class AnInnerClass(BaseModel):
+    h_dict: dict
+    i_list_str: list[str]
```
liuverta: We actually need to `from typing import List` and do

```suggestion
    i_list_str: List[str]
```

across this file.

`list[...]` isn't allowed until Python 3.9, and we still have to support 3.8 (including in our tests)

client/verta/verta/registry/_validate_schema.py
```diff
@@ -0,0 +1,124 @@
...
+            bad_input = {"something": "random"}
+            deployed_model.predict(bad_input)
+
+    References
+    ----------
+    .. [#] https://python-jsonschema.readthedocs.io/en/stable/
+    """
+
+    @functools.wraps(f)
+    def wrapper(self, prediction_input: Dict):
```
liuverta: It's probably better to call this `input` instead of `prediction_input` (unless there's a reason not to) so that users can still pass the argument by name `my_model.predict(input=...)`

client/verta/tests/registry/test_validate_schema.py
```diff
@@ -0,0 +1,122 @@
...
+        recwarn,
+        make_model_schema_file,
+        matching_input_value,
+        matching_output_value,
+    ):
+        @validate_schema
+        def predict(self, input):
+            return matching_output_value.dict()
+
+        predict(None, matching_input_value.dict())
```
liuverta, line 31: We should either just define a simple class here, or get rid of `self`; `predict(None, ...)` feels kinda janky

client/verta/tests/registry/test_validate_schema.py
```diff
@@ -0,0 +1,122 @@
...
+    def test_validate_schema_deny_verify_io_first(self, make_model_schema_file):
+        array = pytest.importorskip("numpy").array([1, 2, 3])
+
+        @verify_io
+        @validate_schema
+        def predict(self, input):
+            return input
+
+        # when verify_io is first, it will raise a TypeError before validate_schema is called
+        with pytest.raises(TypeError):
```
liuverta: Is there a way to put a `match=` on this, to make sure we're catching the `TypeError` that we expect?

liuverta (commented):
A few more minor things

liuverta (approved):
Preemptively approved! 🚀

client/verta/tests/registry/test_validate_schema.py
ghost: It looks like a bunch of our other tests do this and I don't reaaaaaally want to deal with it right now. You okay letting this jankiness slide?

ghost: I tried getting rid of the self and nothing worked 😭

liuverta: Oh. Do we? Welp. I guess it's fine haha

PR #3928, opened 2023-07-07
feat: [VRD-975] set_ and get_schema for open api schemas

ghost:
## Impact and Context
Write the code to set and get the model schema file. Also tests. I'll release a client alpha version for e2e testing sooooon.

## Risks and Area of Effect
It's a new feature!

## Testing
- [X] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting


client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
 
+    def set_schema(self, input: dict, output: dict = None) -> None:
+        """
+        Sets the input and output schemas of this Deployable Entity. Schemas are stored as
+        model artifacts.
+
+        The output schema is optional.
+
+        To validate a prediction's input and output against these schemas, use the
+        :func:`~verta.registry.validate_schema` decorator.
```
ewagner-verta: Might be helpful to say _where_ to use the decorator ("... on your model's `predict()` function").  I suppose they could just click the link and read the docs for `validate_schema` but I think we share a sense of pessimism about the willingness of engineers to take any extra steps.

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
     def _get_artifact(self, key):
         raise NotImplementedError
 
+    def set_schema(self, input: dict, output: dict = None) -> None:
```
ewagner-verta:
```suggestion
    def set_schema(
        self, input: Dict[Any, Any], output: Optional[Dict[Any, Any]] = None
        ) -> None:
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+        The output schema is optional.
+
+        To validate a prediction's input and output against these schemas, use the
+        :func:`~verta.registry.validate_schema` decorator.
+
+        Parameters
+        ----------
+        input : dict
+            Input schema as an OpenAPI-compatible json dict. Easiest to create using pydantic.BaseModel.schema() [#]_.
+        output : dict
```
ewagner-verta:
```suggestion
        output : dict, optional
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -5,13 +5,15 @@
...
 import importlib
+import json
 import os
 import shutil
 import sys
 import tarfile
 import tempfile
 import warnings
 import zipfile
+from typing import Dict
```
ewagner-verta:
```suggestion
from typing import Any, Dict, Optional
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+
+
+        References
+        ----------
+        .. [#] https://docs.pydantic.dev/1.10/usage/schema/
+
+        """
+        if not isinstance(input, dict):
+            raise TypeError(
+                f"`input` must be of type dict, not {type(input)}; did you remember to call `.schema()`?"
```
ewagner-verta: `.schema()` would only applicable if using `Pydantic.BaseModel`, which we are NOT requiring? (Asking since I thought we were).  Not opposed to the current setup, just curious about the reasons for not explicitly requiring Pydantic as the input type.

client/verta/requirements-unit-tests.txt
```diff
@@ -3,4 +3,5 @@ hypothesis
 pytest >= 4.3
 pytest-xdist >= 1.32
 six >= 1.11
-responses >= 0.22
\ No newline at end of file
+responses >= 0.22
+pydantic
```
liuverta: I don't think we need this, since we added `pydantic` to `requirements-dev.txt` in #3910.

client/verta/verta/tracking/entities/_deployable_entity.py
liuverta:
```suggestion
        output : dict, optional
```

client/verta/verta/tracking/entities/_deployable_entity.py
liuverta:
```suggestion
    def set_schema(self, input: dict, output: Optional[dict] = None) -> None:
```

liuverta: Is it too late to propose naming this `log_schema()` to be consistent with all our other methods that create artifacts?

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+            )
+        if output is not None and not isinstance(output, dict):
+            raise TypeError(
+                f"`output` must be of type dict, not {type(output)}; did you remember to call `.schema()`?"
```
ewagner-verta:
```suggestion
                f"`output` must be of type dict, not {type(output)}; did you remember to"
                " call `.schema()`?"
```

client/verta/verta/tracking/entities/_deployable_entity.py
ewagner-verta:
```suggestion
                f"`input` must be of type dict, not {type(input)}; did you remember to" 
                " call `.schema()`?"
```
Nit-picky, I know, but a `black` formatted for length

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+            "input": input,
+        }
+        if output is not None:
+            schema["output"] = output
+
+        # write a temp file because otherwise `log_artifact` will think the artifact contents are the file path
+        with tempfile.NamedTemporaryFile(mode="w") as temp_file:
+            temp_filename = temp_file.name
+            with open(temp_filename, "w") as file:
+                json.dump(schema, file)
```
liuverta: Hmmmmm I might be wrong, but I feel like having a file double-opened in write mode like this risks encountering a deadlock or something. Could we just use the one file handle object?

```suggestion
        with tempfile.NamedTemporaryFile(mode="w+") as temp_file:
            json.dump(schema, temp_file)
            temp_file.flush()  # flush object buffer
            os.fsync(temp_file.fileno())  # flush OS buffer
            temp_file.seek(0)

```

The client uses this approach in a few other places like here: https://github.com/VertaAI/modeldb/blob/7ef3f21e1a6e163a0375e690d0552811bc1927f1/client/verta/tests/versioning/environment/conftest.py#L14-L18

liuverta: And not that it matters today, but opening the file a second time like that wouldn't work on Windows 😆 

> [Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).](https://docs.python.org/3.8/library/tempfile.html#tempfile.NamedTemporaryFile)

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+
+    def get_schema(self) -> Dict[str, dict]:
+        """
+        Gets the input and output json schemas of this Deployable Entity, in the format:
+
+        .. code-block:: python
+        {
+            "input": <input schema>,
+            "output": <output schema>
+        }
```
ewagner-verta:
```suggestion
        .. code-block:: python

            {
                "input": <input schema>,
                "output": <output schema>
            }
```
This needs a blank line and slight change to indentation to render correctly.  Otherwise it looks like 
image

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+        if output is not None:
+            schema["output"] = output
+
+        # write a temp file because otherwise `log_artifact` will think the artifact contents are the file path
+        with tempfile.NamedTemporaryFile(mode="w") as temp_file:
+            temp_filename = temp_file.name
+            with open(temp_filename, "w") as file:
+                json.dump(schema, file)
+            self.log_artifact(
+                key="model_schema_json", artifact=temp_filename, overwrite=True
```
liuverta: I'd recommend

1. Above, using `tempfile.NamedTemporaryFile(..., suffix=".json")` which will have the client log the artifact as a JSON file
   ```
   filename_extension: "json"
   ```
   and allow the web UI to recognize and download the artifact as JSON
   Screenshot 2023-07-13 at 10 00 07 AM
2. Making the artifact key `model_schema` because it's redundant, but also in general I don't think we need `json` in the key anyway 😬 
   ```suggestion
                   key="model_schema", artifact=temp_file.name, overwrite=True
   ```
3. Maybe `overwrite` should be a parameter on this method 🤔

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+        {
+            "input": <input schema>,
+            "output": <output schema>
+        }
+
+        If no output schema was provided, output will not be included in the returned dict.
+
+        Returns
+        -------
+        Dict[str, dict]
```
liuverta:
```suggestion
        dict of str to dict
```

client/verta/verta/tracking/entities/_deployable_entity.py
liuverta:
```suggestion
        .. code-block:: python
        
            {
                "input": <input schema>,
                "output": <output schema>,
            }
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
     def _get_artifact(self, key):
         raise NotImplementedError
 
+    def set_schema(self, input: dict, output: dict = None) -> None:
+        """
+        Sets the input and output schemas of this Deployable Entity. Schemas are stored as
```
ewagner-verta: Since "Deployable Entity" is somewhat of an internal construct customers won't recognize, perhaps we could just say "Sets the input and output schemas, which are stored ..." and leave it at that?

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+            temp_filename = temp_file.name
+            with open(temp_filename, "w") as file:
+                json.dump(schema, file)
+            self.log_artifact(
+                key="model_schema_json", artifact=temp_filename, overwrite=True
+            )
+
+    def get_schema(self) -> Dict[str, dict]:
+        """
+        Gets the input and output json schemas of this Deployable Entity, in the format:
```
liuverta:
```suggestion
        Gets the input and output JSON schemas of this deployable entity, in the format:
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+            "input": <input schema>,
+            "output": <output schema>
+        }
+
+        If no output schema was provided, output will not be included in the returned dict.
+
+        Returns
+        -------
+        Dict[str, dict]
+            Input and output json schemas.
```
liuverta:
```suggestion
            Input and output JSON schemas.
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,74 @@ def _get_artifact_msg(self, key):
...
+        }
+        if output is not None:
+            schema["output"] = output
+
+        # write a temp file because otherwise `log_artifact` will think the artifact contents are the file path
+        with tempfile.NamedTemporaryFile(mode="w") as temp_file:
+            temp_filename = temp_file.name
+            with open(temp_filename, "w") as file:
+                json.dump(schema, file)
+            self.log_artifact(
```
liuverta:
```suggestion
            self.log_artifact(  # pylint: disable=no-member
```

client/verta/tests/registry/model_version/test_set_schema.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
```
liuverta: Is this test module redundant since we already have tests covering `deployable_entity`?

liuverta (changes requested):

ewagner-verta (dismissed):
Only additional comment would be that some unit tests would be nice.  i.e. testing the underlying calls to set and get work as expected.  I'm happy to help with that if you'd like.

client/verta/tests/deployable_entity/test_artifacts.py
```diff
@@ -679,3 +681,33 @@ def test_setup_script(self, deployable_entity):
...
+
+    def test_set_schema_no_output(self, deployable_entity):
+        input_schema = InputClass.schema()
+        deployable_entity.set_schema(input=input_schema)
+
+        assert deployable_entity.get_schema() == {
+            "input": input_schema,
+        }
+
+    def test_set_schema_no_input_which_is_bad(self, deployable_entity):
```
liuverta: This is how I've been naming other tests that check for an error to be raised.

```suggestion
    def test_set_schema_no_input_error(self, deployable_entity):
```

liuverta (changes requested):

client/verta/tests/registry/model_version/test_set_schema.py
ghost: oop, yeah, forgot to clean up

client/verta/verta/tracking/entities/_deployable_entity.py
ewagner-verta: Not that this applies to our customer base necessarily, but Windows is still the most widely used OS globally, [by a long shot](https://www.statista.com/statistics/268237/global-market-share-held-by-operating-systems-since-2009/)

client/verta/verta/tracking/entities/_deployable_entity.py
liuverta: I was kind of thinking that too

client/verta/verta/tracking/entities/_deployable_entity.py
ghost: We were and had to abort mission. It's a product requirement (see prd) that we support Open API. So we would have had to suppose BOTH pydantic and an open api schema and I didn't want to do that. I know `.schema()` is specific to pydantic buuuuuut the most likely time for a user to get this error message would be if they forgot to call `.schema()` and passed in their pydantic class. If they're using open API out-of-the-box (sans pydantic), they wouldn't get this error message. Probably. 🤷🏼‍♀️

client/verta/verta/tracking/entities/_deployable_entity.py
liuverta: It doesn't matter for us today because our client doesn't work in Windows anyway, due to issues elsewhere in the codebase 😅

client/verta/verta/tracking/entities/_deployable_entity.py
ghost: your `black` is more aggressive than mine apparently!

client/verta/verta/tracking/entities/_deployable_entity.py
ghost: I literally copied this code from the internet and didn't know it was opening a file twice 😅

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,77 @@ def _get_artifact_msg(self, key):
...
 
+    def log_schema(self, input: dict, output: Optional[dict] = None) -> None:
+        """
+        Sets the input and output schemas of this deployable entity. Schemas are stored as
+        model artifacts.
+
+        The output schema is optional.
+
+        To validate a prediction's input and output against these schemas, use the
+        :func:`~verta.registry.validate_schema` decorator on your model's predict function.
```
ghost: Is there a way to link to the model's predict function docs? Not sure what's best to link to/reference, here.

liuverta: We could link to

```suggestion
        :func:`~verta.registry.validate_schema` decorator on your model's :meth:`~verta.registry.VertaModelBase.predict` function.
```

client/verta/verta/tracking/entities/_deployable_entity.py
ghost: 3. if this function is supposed to update a model_schema, I don't see why we'd ever not want to overwrite it?

client/verta/verta/tracking/entities/_deployable_entity.py
ewagner-verta: Sounds good.  The type error is descriptive enough itself.  Thanks for catching me up.

ghost:
Checked the docs and they look good!

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,82 @@ def _get_artifact_msg(self, key):
...
+
+        To validate a prediction's input and output against these schemas, use the
+        :func:`~verta.registry.validate_schema` decorator on your model's
+        :meth:`~verta.registry.VertaModelBase.predict` function.
+
+        Parameters
+        ----------
+        input : dict
+            Input schema as an OpenAPI-compatible JSON dict. Easiest to create using
+            pydantic.BaseModel.schema() [#]_.
```
liuverta, line 107: This changes it from 

> pydantic.BaseModel.schema()

to

> `pydantic.BaseModel.schema()`

```suggestion
            ``pydantic.BaseModel.schema()`` [#]_.
```

client/verta/verta/tracking/entities/_deployable_entity.py
```diff
@@ -88,6 +90,82 @@ def _get_artifact_msg(self, key):
...
+        output : dict, optional
+            Output schema as an OpenAPI-compatible JSON dict. Easiest to create using
+            pydantic.BaseModel.schema().
```
liuverta, line 110:
```suggestion
            ``pydantic.BaseModel.schema()``.
```

liuverta (approved):

PR #3940, opened 2023-07-13
feat: Allow specifying the organization by name

conradoverta:
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/client/verta/verta/client.py
+++ b/client/verta/verta/client.py
@@ -151,6 +151,8 @@ class Client(object):
         organization_name=None,
         _connect=True,
     ):
+        if organization_id is not None and organization_name is not None:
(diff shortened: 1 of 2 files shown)


client/verta/verta/client.py
```diff
@@ -145,6 +148,7 @@ def __init__(
         jwt_token=None,
         jwt_token_sig=None,
         organization_id=None,
+        organization_name=None,
         _connect=True,
     ):
```
liuverta, line 153:
```suggestion
    ):
        if organization_id is not None and organization_name is not None:
            raise ValueError("cannot provide both `organization_id` and `organization_name`")
```

ewagner-verta (approved):

PR #3957, opened 2023-07-19
feat: [VRD-1031] Add access to request headers to model predict methods

ghost:
## Impact and Context
Add access to headers on the model's prediction methods. As of now, these headers will be provided on curl requests.

## Risks and Area of Effect
High area of effect because every model's predict method is affected. We'll need to make sure PyMS can safely call models that don't have the `headers` param on their predict methods.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

Tested locally as much as I could. Now I need to cut a client alpha version to continue testing.

## Reverting


client/verta/verta/runtime.py
```diff
@@ -249,3 +276,16 @@ def logs(self) -> Dict[str, Any]:
...
+
+    def headers(self) -> Dict[str, Any]:
+        """
+        Return the currently stored, thread-local request headers from inside this
+        context manager. If called after exiting the context manager, the
+        final collection of headers captured at the time of exit is returned.
+
+        Returns
+        -------
+        headers : Dict[str, str]
```
```suggestion
        headers : dict of str to str
```

client/verta/verta/runtime.py
```diff
@@ -249,3 +276,16 @@ def logs(self) -> Dict[str, Any]:
             Dictionary of logs collected within this context manager.
         """
         return self._logs_dict or _get_thread_logs()
+
+    def headers(self) -> Dict[str, Any]:
```
```suggestion
    def headers(self) -> Dict[str, str]:
```

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -55,7 +56,7 @@ def __init__(self, artifacts):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def predict(self, input):
+    def predict(self, input, headers=Dict[str, str]):
```
liuverta: That's a type annotation! It should be after a `:`, not a `=`

```suggestion
    def predict(self, input, headers: Dict[str, str]):
```

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -93,7 +94,7 @@ def predict(self, input):
         """
         raise NotImplementedError
 
-    def batch_predict(self, df):
+    def batch_predict(self, df, headers=Dict[str, str]):
```
liuverta:
```suggestion
    def batch_predict(self, df, headers: Dict[str, str]):
```

client/verta/verta/registry/_validate_schema.py
```diff
@@ -104,7 +104,7 @@ def wrapper(self, input: Dict):
             ) from e
 
         # run function
-        output = f(self, input)
+        output = f(self, input, headers)
```
liuverta: This is going to break if the user didn't opt into headers for their model predict.

I think we could instead do

```python
def wrapper(self, input: Dict, *args, **kwargs):
    ...
    output = f(self, input, *args, **kwargs)
```
and that should work fine?

ghost: I was wondering about this.

client/verta/verta/registry/_verta_model_base.py
ghost: oops it used to be None that's why. Why does this work???

liuverta (approved):

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -80,6 +81,9 @@ def predict(self, input):
         input : any JSON-compatible Python type
             Model input.
```
liuverta: No blank lines between parameters

```suggestion
```

client/verta/verta/registry/_verta_model_base.py
```diff
@@ -114,6 +118,9 @@ def batch_predict(self, df):
         ----------
         df : pandas.DataFrame
```
liuverta: No blank lines between parameters

```suggestion
```

PR #3977, opened 2023-07-28
docs: Note that `headers` is optional

ghost:
## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/client/verta/verta/registry/_verta_model_base.py
+++ b/client/verta/verta/registry/_verta_model_base.py
@@ -65,7 +65,7 @@ class VertaModelBase(object):
 
             The `headers` parameter is optional: overriding `predict` methods
             do not need to include this parameter. For usage details, see
-            https://docs.verta.ai/verta/deployment/guides/accessing-headers-from-predict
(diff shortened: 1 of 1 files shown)


client/verta/verta/registry/_verta_model_base.py
```diff
@@ -61,6 +61,11 @@ def predict(self, input, headers: Dict[str, str]):
 
         This method is called when requests are made against a Verta endpoint.
 
+        .. note::
+
+            The `headers` parameter is optional: subclassed `predict` methods
+            do not need to include this parameter.
```
liuverta: I beliieeeve "override" is the more apt term than "subclass".

I also feel like we should link to the docs (either on this note, or in the `Parameters` section below), otherwise someone seeing this might be like, "It's optional? But how do I know if it's right for me?? What it do???"

```suggestion
            The `headers` parameter is optional: overriding `predict` methods
            do not need to include this parameter. For usage details, see
            https://docs.verta.ai/verta/deployment/guides/accessing-headers-from-predict
```

liuverta (approved):

PR #3998, opened 2023-08-07
feat: remove explicit role binding manipulations for experiment and runs

jkwatson-verta:
## Testing
- [ ] Unit test
- [x] Deployed to dev env
- [ ] Other (explain) 

## Reverting


marcelo-verta-ai (approved):

backend/server/src/main/java/ai/verta/modeldb/authservice/PublicMDBRoleServiceUtils.java
```diff
@@ -83,15 +51,7 @@ public void validateEntityUserWithUserInfo(
...
-      if (modelDBServiceResourceTypes.equals(ModelDBServiceResourceTypes.PROJECT)) {
-        try {
-          if (futureProjectDAO.getProjectById(resourceId).blockAndGet() != null) {
-            throw new NotFoundException(ModelDBMessages.PROJECT_NOT_FOUND_FOR_ID);
-          }
-        } catch (Exception e) {
-          throw new ModelDBException(e);
-        }
-      } else if (modelDBServiceResourceTypes.equals(ModelDBServiceResourceTypes.DATASET)) {
+      if (modelDBServiceResourceTypes.equals(ModelDBServiceResourceTypes.DATASET)) {
```
eheinlein-verta, line 54: Why don't we need to worry about the project case anymore?

eheinlein-verta (commented):
Just one question

backend/server/src/main/java/ai/verta/modeldb/authservice/PublicMDBRoleServiceUtils.java
jkwatson-verta: ok...this is super weird, but IDEA told me... the futureProjectDAO instance variable _was never initialized_ in this class, so the call to `getProjectById` would _always_ fail with an NPE.

eheinlein-verta (approved):

PR #4003, opened 2023-08-09
feat: [VRD-1043] Add gpu settings to `Resources`

ghost:
## Impact and Context
Allow users to pass GPU settings when updating an endpoint. This includes creating _builds_ with gpu hardware compatibility settings.

## Risks and Area of Effect
New stuff! Verified that it didn't break the existing resources code.

## Testing
- [X] Unit test - updated and ran the endpoint tests with resources in em
- [X] Deployed to dev env
- [ ] Other (explain) 

Successfully deployed endpoints with these settings. And verified that the builds are created with gpu settings (and checked the build logs to verify they actually were built with gpu base images).

## Reverting


client/verta/verta/endpoint/nvidia_gpu.py
```diff
@@ -0,0 +1,40 @@
...
+            d["model"] = self.model.name
+
+        return d
+
+    @classmethod
+    def _from_dict(cls, rule_dict):
+        return cls(**rule_dict)
+
+
+class NvidiaGPUModel(Enum):
```
liuverta:
```suggestion
class NvidiaGPUModel(str, Enum):
```

https://docs.python.org/3.7/library/enum.html#others

client/verta/verta/endpoint/nvidia_gpu.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def _validate_model(self, model):
+        if not isinstance(model, NvidiaGPUModel):
+            raise TypeError(self.MODEL_ERR_MSG)
+
+    def _as_dict(self):
+        d = dict()
+        d["number"] = self.number
+        if self.model is not None:
+            d["model"] = self.model.name
```
liuverta: Since we're using strings as the enum values, we may as well use them

```suggestion
            d["model"] = self.model.value
```

client/verta/verta/endpoint/nvidia_gpu.py
```diff
@@ -0,0 +1,40 @@
...
+    def _validate_number_of_gpus(self, number_of_gpus):
+        if not isinstance(number_of_gpus, int):
+            raise TypeError(self.NUM_GPUS_ERR_MSG)
+        if number_of_gpus <= 0:
+            raise ValueError(self.NUM_GPUS_ERR_MSG)
+
+    def _validate_model(self, model):
+        if not isinstance(model, NvidiaGPUModel):
+            raise TypeError(self.MODEL_ERR_MSG)
+
```
liuverta: idk where this check should go, but something like

```suggestion
        if isinstance(model, str):
            NvidiaGPUModel(model)  # raises exception if `model` isn't a NvidiaGPUModel value

```

client/verta/verta/endpoint/resources.py
```diff
@@ -73,12 +78,23 @@ def _validate_memory(self, memory):
         if not re.match(r"^[0-9]+[e]?[0-9]*[E|P|T|G|M|K]?[i]?$", memory):
             raise ValueError(self.MEMORY_ERR_MSG)
 
+    def _validate_nvidia_gpu(self, nvidia_gpu):
+        if isinstance(nvidia_gpu, dict):
+            nvidia_gpu = NvidiaGPU._from_dict(nvidia_gpu)
```
liuverta: I don't think we should accept a `dict` here 🤔 

I'd recommend putting the dict conversion in `_from_dict()`, probably

client/verta/verta/endpoint/nvidia_gpu.py
```diff
@@ -0,0 +1,70 @@
...
+    NUM_GPUS_ERR_MSG = "`number` must be a number greater than 0"
+    MODEL_ERR_MSG = "`model` must be an instance of `verta.endpoint.NvidiaGPUModel`"
+
+    def __init__(self, number, model=None):
+        self._validate_number_of_gpus(number)
+        if model is not None:
+            # convert to NvidiaGPUModel instance if passed in as str
+            model = self._validate_model(model)
+        self.number = number
+        self.model = model
```
liuverta: I'd discourage having `_validate_model()` return because we've had a general pattern of `validate()` methods not returning anything 😬 

I think we could just do

```suggestion
            self._validate_model(model)
        self.number = number
        self.model = NvidiaGPUModel(model)
```

which works if `model` is either an `int` or already an `NvidiaGPUModel`

I think.

client/verta/verta/endpoint/nvidia_gpu.py
```diff
@@ -0,0 +1,85 @@
...
+        Number of GPUs requested to run this endpoint's model.
+    model: :class:`NvidiaGPUModel`, optional
+        Model of GPU requested to run this endpoint's model. Currently available models are
+        specified in the `NvidiaGPUModel` enum.
+
+    Examples
+    --------
+    .. code-block:: python
+
+        from verta.endpoint.resources import Resources, NvidiaGPU, NvidiaGPUModel
```
liuverta: I like this `import` structure, but it looks like this isn't what we've actually implemented.

We should restructure our files from

```
verta/endpoint/
    nvidia_gpu.py
    resources.py
```

to

```
verta/endpoint/resources/
    __init__.py  # for example see https://github.com/VertaAI/modeldb/blob/main/client/verta/verta/endpoint/__init__.py
    _nvidia_gpu.py
    _resources.py
```

client/verta/verta/endpoint/nvidia_gpu.py
```diff
@@ -0,0 +1,85 @@
...
+            raise TypeError(self.NUM_GPUS_ERR_MSG)
+        if number_of_gpus <= 0:
+            raise ValueError(self.NUM_GPUS_ERR_MSG)
+
+    def _validate_model(self, model):
+        if isinstance(model, str):
+            model = NvidiaGPUModel(model)
+        if not isinstance(model, NvidiaGPUModel):
+            raise TypeError(self.MODEL_ERR_MSG)
+        return model
```
liuverta: I'd recommend against `return`ing from a validation method. Validation should just do validation, not conversion.

client/verta/verta/endpoint/resources.py
```diff
@@ -73,15 +80,26 @@ def _validate_memory(self, memory):
         if not re.match(r"^[0-9]+[e]?[0-9]*[E|P|T|G|M|K]?[i]?$", memory):
             raise ValueError(self.MEMORY_ERR_MSG)
 
+    def _validate_nvidia_gpu(self, nvidia_gpu):
+        if not isinstance(nvidia_gpu, NvidiaGPU):
+            raise TypeError(
+                "`nvidia_gpu` must be an instance of `verta.endpoint.NvidiaGpu`"
+            )
+        return nvidia_gpu
```
liuverta: I'd recommend against `return`ing from a validation method.

client/verta/verta/endpoint/resources.py
```diff
@@ -73,15 +80,26 @@ def _validate_memory(self, memory):
...
+        if self.nvidia_gpu is not None:
+            d["nvidia_gpu"] = self.nvidia_gpu._as_dict()
 
         return d
 
     @classmethod
-    def _from_dict(cls, rule_dict):
-        return cls(**rule_dict)
+    def _from_dict(cls, resources_dict):
+        if "nvidia_gpu" in resources_dict:
```
liuverta, line 102: We have to copy the dict if we're modifying its items. Otherwise it may affect the calling code in surprising ways.

```suggestion
        resources_dict = resources_dict.copy()
        if "nvidia_gpu" in resources_dict:
```

client/verta/verta/endpoint/resources.py
```diff
@@ -52,14 +55,18 @@ class Resources(object):
...
     )
 
-    def __init__(self, cpu=None, memory=None):
+    def __init__(self, cpu=None, memory=None, nvidia_gpu=None):
         if cpu is not None:
             self._validate_cpu(cpu)
         if memory is not None:
             self._validate_memory(memory)
+        if nvidia_gpu is not None:
+            # convert to NvidiaGPU instance if passed in as dict
```
liuverta:
```suggestion
```

client/verta/verta/endpoint/resources.py
```diff
@@ -32,13 +33,15 @@ class Resources(object):
...
             power-of-two equivalents: **Ei, Pi, Ti, Gi, Mi, Ki**. For example, the
             following represent roughly the same value: 128974848, 129e6, 129M, 123Mi.
+    nvidia_gpu: :class:`NvidiaGPU`, optional
+        Nvidia GPU resources allowed for an endpoint's model.
 
     Examples
     --------
     .. code-block:: python
 
         from verta.endpoint.resources import Resources
```
liuverta:
```suggestion
        from verta.endpoint.resources import Resources, NvidiaGPU, NvidiaGPUModel
```

client/verta/verta/endpoint/resources.py
```diff
@@ -32,13 +33,15 @@ class Resources(object):
             using one of these suffixes: **E, P, T, G, M, K**. You can also use the
             power-of-two equivalents: **Ei, Pi, Ti, Gi, Mi, Ki**. For example, the
             following represent roughly the same value: 128974848, 129e6, 129M, 123Mi.
+    nvidia_gpu: :class:`NvidiaGPU`, optional
```
liuverta:
```suggestion
    nvidia_gpu: :class:`~verta.endpoint.resources.NvidiaGPU`, optional
```

client/verta/verta/endpoint/build/_build_hardware_compatibility.py
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self, T4=None, V100=None, all=None):
+        self.T4 = T4
+        self.V100 = V100
+        self.all = all
+
+
+class BuildHardwareCompatibility:
+    """The hardware compatibility of a Verta model build.
+
+    .. versionadded:: 0.25.0
```
liuverta: The next version is prooobably going to be `0.24.1`?

client/verta/verta/endpoint/build/_build.py
```diff
@@ -110,6 +111,19 @@ def get_scan(self) -> _build_scan.BuildScan:
...
+        returns None.
+
+        .. versionadded:: 0.25.0
+
+        Returns
+        -------
+        :class:`~verta.endpoint.build.BuildHardwareCompatibility` or None
+
+        """
+        return BuildHardwareCompatibility._get(self._conn, self.id)
```
liuverta: I think we should implement this as

```python
@property
def hardware_compatibility(self):
```

and construct `BuildHardwareCompatibility` using just `self._json`. A build's hardware compatibility won't change after it's created, right? If so we can just base it on the data we already fetched

client/verta/verta/endpoint/build/_build_hardware_compatibility.py
```diff
@@ -0,0 +1,62 @@
...
+        if hardware_compatibility is None:
+            return None
+
+        return cls(hardware_compatibility)
+
+    def _as_dict(self):
+        return self._json
+
+    @property
+    def nvidia_gpu(self):
```
liuverta: I don't think it's good for consistency to have `Resources.nvidia_gpu` and `BuildHardwareCompatibility.nvidia_gpu` return different types.

client/verta/verta/endpoint/build/_build_hardware_compatibility.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+
+from verta._internal_utils import _utils
+
+
+class BuildHardwareCompatibilityNvidiaGPU:
+    def __init__(self, T4=None, V100=None, all=None):
+        self.T4 = T4
+        self.V100 = V100
+        self.all = all
```
liuverta: This class looks...strange to me. How is it being used? I can't seem to find it

ghost: here: https://github.com/VertaAI/modeldb/pull/4003/files#diff-c02f63168a48821dc0b68a146b95ee0f8a2b1b631c1e8a8deca2b6ecfa9f4f41R58 it's literally just trying to match the spec here: dev.verta.ai/api/v1/deployment/swagger-ui/#/builds/get_builds__buildId_

liuverta: I see it's `return`ed by `BuildHardwareCompatibility.nvidia_gpu`, but where do we use `BuildHardwareCompatibility.nvidia_gpu` itself?

client/verta/verta/endpoint/resources/nvidia_gpu.py
```diff
@@ -0,0 +1,75 @@
...
+        Number of GPUs requested to run this endpoint's model.
+    model: :class:`NvidiaGPUModel`, optional
+        Model of GPU requested to run this endpoint's model. Currently available models are
+        specified in the `NvidiaGPUModel` enum.
+
+    Examples
+    --------
+    .. code-block:: python
+
+        from verta.endpoint.resources import Resources, NvidiaGPU, NvidiaGPUModel
```
liuverta, line 28: It looks like we're doing 

```suggestion
        from verta.endpoint.resources import Resources
        from verta.endpoint.resources.nvidia_gpu import NvidiaGPU, NvidiaGPUModel
```

I [was deliberately recommending](https://github.com/VertaAI/modeldb/pull/4003#discussion_r1290728666) sticking with having everything `import`able `from verta.endpoint.resources` because I don't think the extra `.nvidia_gpu` is necessary, but I don't mind too much either way (as long as this docstring is correct).

client/verta/verta/endpoint/resources/resources.py
```diff
@@ -32,13 +33,15 @@ class Resources(object):
...
+    nvidia_gpu: :class:`~verta.endpoint.resources.NvidiaGPU`, optional
+        Nvidia GPU resources allowed for an endpoint's model.
 
     Examples
     --------
     .. code-block:: python
 
-        from verta.endpoint.resources import Resources
-        resources = Resources(cpu=.25, memory="512Mi")
+        from verta.endpoint.resources import Resources, NvidiaGPU, NvidiaGPUModel
```
liuverta: See my comment in [nvidia_gpu.py](https://github.com/VertaAI/modeldb/pull/4003/files#diff-195496e7caec4198a03787ef6e7ac92298c3eeae609124738b7ff5c591132246) about the `import `path

client/verta/verta/endpoint/resources/resources.py
```diff
@@ -4,7 +4,8 @@
 import abc
```
liuverta: I [was recommending](https://github.com/VertaAI/modeldb/pull/4003#discussion_r1290728666) naming this file `_resources.py`, with the leading underscore. It's to help declutter `import` paths (otherwise `Resources` can be technically imported from two public submodules) and I've also just discovered that documentation breaks otherwise.

client/verta/verta/endpoint/resources/nvidia_gpu.py
```diff
@@ -0,0 +1,75 @@
...
+            return None
+        if self.model is None:
+            return {"nvidia_gpu": {"all": True}}
+        else:
+            return {"nvidia_gpu": {self.model.value: True}}
+
+
+class NvidiaGPUModel(str, Enum):
+    T4 = "T4"
+    V100 = "V100"
```
liuverta: https://github.com/VertaAI/modeldb/pull/4006

client/verta/verta/endpoint/build/_build.py
```diff
@@ -97,6 +99,35 @@ def message(self) -> str:
     def is_complete(self) -> bool:
         return self.status in ("finished", "error")
 
+    @property
+    def nvidia_gpu_compatible_hardware(self):
```
liuverta: Ughh...maybe let's just remove this? We've tucked away "hardware compatibility" as a concept deep inside `endpoint.update()` so we're not really obligated to expose this (like you mentioned before).

And Eric is likely to implement an `endpoint.get_resources()` as a part of pipelines, so we can expose "here is the GPU you ordered" there.

client/verta/verta/endpoint/build/_build.py
```diff
@@ -97,6 +99,35 @@ def message(self) -> str:
...
+        hardware_compatibility = self._json["creator_request"].get("hardware_compatibility")
+        if hardware_compatibility is None:
+            return None
+        nvidia_gpu = hardware_compatibility.get("nvidia_gpu")
+        if nvidia_gpu is None:
+            return None
+
+        allows_all_models = nvidia_gpu.get("all")
+        allowed_models = set()
+        for model, allowed in nvidia_gpu.items():
```
liuverta: Is it guaranteed that all models will be present in the `nvidia_gpu` dict, even when they're `False?`

[Like](https://vertaai.slack.com/archives/CJ2MM54Q3/p1691708171921869?thread_ts=1691704020.401389&cid=CJ2MM54Q3)

```json
  "creator_request": {
    "hardware_compatibility": {
      "nvidia_gpu": {
        "T4": true
      }
    },
```

client/verta/verta/endpoint/resources/nvidia_gpu.py
ghost: I fixeded it

liuverta (approved):

PR #4050, opened 2023-08-31
Feature: Add OpenTelemetry AutoConfigured SDK initialisation to TracingExtension

gustavoteixeirah:
The TracingExtension.java class was extended by introducing a new method to initialize an OpenTelemetrySdk instance using
AutoConfiguredOpenTelemetrySdk builder. This instance sets the global OpenTelemetry SDK, and configures various properties including tracing exporter, propagators, and disabled resource providers.

The purpose of this change is to streamline the handling of tracing in testing environment. Developers can now easily initialise the Open Telemetry SDK, which may be useful for enhanced tracking of testing processes.

Also, the pom.xml file in the common-testing directory was updated to include a dependency on 'opentelemetry-sdk-extension-autoconfigure' for enabling the AutoConfiguredOpenTelemetrySdk usage.

## Testing
- [ ] Unit test
- [ ] Deployed to dev env
- [ ] Other (explain) 

## Reverting

### DIFF
--- a/backend/common-testing/src/main/java/ai/verta/common/testing/utils/TracingExtension.java
+++ b/backend/common-testing/src/main/java/ai/verta/common/testing/utils/TracingExtension.java
@@ -193,8 +193,8 @@ public class TracingExtension
 
   /**
    * This method initiates an OpenTelemetrySdk instance using AutoConfiguredOpenTelemetrySdk
-   * builder. It also configures various properties including
-   * tracing exporter, propagators, and disabled resource providers.
+   * builder. It also configures various properties including tracing exporter, propagators, and
+   * disabled resource providers.
    *
    * <p>For tracing in tests, consider the following steps:
    *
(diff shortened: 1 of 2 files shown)


backend/common-testing/src/main/java/ai/verta/common/testing/utils/TracingExtension.java
```diff
@@ -188,6 +191,55 @@ public void postProcessTestInstance(Object o, ExtensionContext extensionContext)
...
+                      "none",
+                      "otel.logs.exporter",
+                      "none",
+                      "otel.service.name",
+                      "registry-local-its");
+                })
+            .build();
+    var openTelemetrySdk = autoConfiguredOpenTelemetrySdk.getOpenTelemetrySdk();
+    log.debug("openTelemetrySdk initialized = " + openTelemetrySdk);
+    return openTelemetrySdk;
```
jkwatson-verta, line 241: would it make sense to set the static OpenTelemetry instance here, before returning?

jkwatson-verta (approved):

backend/common-testing/src/main/java/ai/verta/common/testing/utils/TracingExtension.java
gustavoteixeirah: I tought I would set the static instance in the bean declaration in each test config beans file, no?

jkwatson-verta: Won't all of the usages just be like:
```
var otel = TracingExtension.initializeOpenTelemetrySdk();
TracingExtension.setOpenTelemetry(otel);
```

And we could just do that in one step here, I would think.

PR #4209, opened 2023-11-19
Fix loading yaml config issue.

am0o0:
## Impact and Context
when we are loading config.yaml file for the backend, we have an error like the following:
```
Unable to find property 'test' on class: ai.verta.modeldb.config.MDBConfig
```
this is because of that we are not skipping the parts of the config that we don't need to load by snakeyaml.
I fixed the issue by simply adding a `Representer`.


jkwatson-verta (approved):
This is great! Thanks!

PR #4630, opened 2026-05-14
Add timeout to notebook session lookup

massy-o:
## Summary
- add a timeout to the Jupyter sessions request used while resolving the current notebook path

## Why
If a listed notebook server accepts a connection but does not respond, path discovery can hang indefinitely. A bounded timeout lets the helper continue to the existing failure path instead of waiting on the socket forever.

## Testing
- python3 -m py_compile client/verta/verta/_internal_utils/_utils.py


client/verta/verta/_internal_utils/_utils.py
```diff
@@ -1297,6 +1297,7 @@ def get_notebook_filepath():
             response = requests.get(
                 urljoin(server["url"], "api/sessions"),
                 params={"token": server.get("token", "")},
+                timeout=5,
```
massy-o, line 1300: Self-review: This lookup runs against locally listed notebook servers, so a short timeout is enough to avoid blocking forever while preserving the existing behavior of trying the remaining servers and eventually raising OSError.

